PR #13, opened 2025-11-06
Fix/proper pr workflow

robertDouglass:


robertDouglass:
Rebased onto current main after the history rewrite. Diff now only bumps actions/configure-pages, upload-pages-artifact, and deploy-pages to their v5/v3/v4 releases.

PR #43, opened 2025-11-26
add activation of .venv if any

sdementen:
will look if there is a .venv folder in the repo and activate the related python virtual environment. without this, whenever spec-kitty run python code (for tests, etc) it runs the default python and not the one for the project

### DIFF
--- a/scripts/powershell/common.ps1
+++ b/scripts/powershell/common.ps1
@@ -227,21 +227,25 @@ function Test-DirHasFiles {
     }
 }
 
-
 function Activate-RepoVenv {
     $repoRoot = Get-RepoRoot
-    $venvPath = Join-Path $repoRoot ".venv"
+    # If inside a .worktrees directory, use its root to detect a VIRTUAL_ENV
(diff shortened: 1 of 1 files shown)


robertDouglass:
Thank you for this PR, @sdementen! However, this PR is obsolete due to the Python rewrite completed in v0.10.0 (commit b2d2059, December 18, 2025).

**What changed:**
- All PowerShell scripts were deleted, including `scripts/powershell/common.ps1`
- The entire bash and PowerShell wrapper layer was removed
- Commands now run directly as Python CLI: `spec-kitty agent ...`

**About automatic .venv activation:**
The Python CLI (`spec-kitty`) should be run within an activated virtual environment by the user, not auto-activated by scripts. This is standard Python practice and gives users control over their environment.

If you need to use a specific virtual environment:
```bash
source .venv/bin/activate  # Unix
# or
.venv\Scripts\activate   # Windows
spec-kitty agent feature create-feature
```

Thank you for your contribution!

PR #344, opened 2026-03-28
fix(implement): preserve actionable --json errors

LoneWolf36:
## Summary
- keep `spec-kitty implement --json` machine-readable without collapsing every failure into the same opaque error string
- capture the recent console output for `typer.Exit` failures and return that text in the JSON envelope
- add a regression test proving lane-validation failures surface actionable text

## Why
External orchestrators and wrappers need an actionable failure message from `--json` mode. Returning only `implement command failed` forces callers to scrape human output or rerun the command interactively.

## Scope
- `src/specify_cli/cli/commands/implement.py`
- `tests/agent/test_implement_command.py`

## Validation
- `python3 -m pytest tests/agent/test_implement_command.py -q`

## Notes
This restacks the original PR after `#342` landed and intentionally drops the stale orchestrator docs/API claims that no longer matched current `main`.

### DIFF
--- a/src/specify_cli/cli/commands/implement.py
+++ b/src/specify_cli/cli/commands/implement.py
@@ -7,6 +7,7 @@ import json
 import re
 import subprocess
 from datetime import datetime, timezone
+from io import StringIO
 from pathlib import Path
 
 import typer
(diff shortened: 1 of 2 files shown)


robertDouglass:
Great contribution, thanks.

PR #621, opened 2026-04-14
doctrine: add black-box integration testing directive and tactic

### LINKED ISSUE #203: Feature proposition: mandatory black-box integration tests?
Hello,
I know that specs are meant to be very flexible, but case in point: 

1. I have implemented an app that took 1 week to spec out iterate over with spec-kitty;
2. All features were accepted and unit tests passed (or mostly passed, 370/400 unit tests),
3. I've added (around 170) integration tests which were actually reflecting my functionality via black-box testing,
4. It took 2.5 more weeks to debug integration tests, and still going. In process, missing functionality was found, critical API mismatches were found, everything as you don't want it. I had to rewrite basically every line of code in the app.

The particular app is, in localhost, 6 microservices: 3 services, Postgress, Minio, and another container; also frontend.

Now, obviously, not every app needs integration tests. But when we do HTTP transfer, we almost definitely need them.

Steer the prompts to lean heavily on black-box integration tests? By black box, I mean: We can't import anything from the app except models, we can't mock anything; we can only assert a response.
Ideally, in the process we would test multiple containers as possible to verify their transfer between each other.

Maybe this is just a lesson for others to use with spec-driven dev. 

My integration test document (you probably want to read it!):
<details><summary>Details</summary>

```markdown

## **CRITICAL NOTE TO IMPLEMENTATION AGENTS**

These workflows are integration workflows. Do **not** satisfy them with unit tests disguised as integration tests.
Use mocks only when they are strictly unavoidable, and document each mock with a justification in the test docstring.

**CONTROL THE TEST EXECUTION *ONLY* BY HTTP REQUESTS IN EVERY TEST. DO NOT DO TESTS THAT ARE EFFECTIVELY UNIT TESTS**
**FILE CREATION SHOULD BE MADE VIA TOOL CALLS. EXECUTION LOGIC SHOULD BE MADE VIA TOOL CALLS. EVERYTHING SHOULD BE MADE VIA HTTP REQUESTS, NOT VIA MOCKS AND UNIT TESTS!**

**ASSERT AGAINST *CONTAINERS* NOT AGAINST PYTHON**

*If you cannot assert something via HTTP calls, assert it via the `events.jsonl` event stream.*

Additionally:

1. `kitty-specs/desired_architecture.md` is the central source of truth for the application; all integration tests should be relevant to it and reviewed against it before implementation. Every added test must map to an INT-xxx ID from `kitty-specs/integration-tests.md`.
2. Prefer black-box integration (compose services + real boundaries); avoid heavy mocks except where unavoidable.
3. Include both happy-path and required fail-path assertions (especially gating/validation/refusal/event emission).
4. Do not silently drop existing integration coverage; replacements must preserve or improve mapped coverage.
5. Mark unimplemented IDs explicitly as deferred (P1/P2) instead of omitting them.

## Non-negotiable Integration Execution Contract (applies to every `INT-xxx`)

1. Test target is a running compose stack (`controller`, `worker`, `controller-worker`, infra services), not imported Python functions.
2. Test traffic goes through HTTP APIs only.
3. Files/artifacts are created by API/tool-call pathways, not by direct local writes to app internals.
4. No `patch`, `monkeypatch`, or fake clients for controller/worker/temporal/s3 paths in integration tests.
5. Assertions are on observable boundaries: HTTP responses, container logs, DB rows, object storage objects, emitted events.
6. Every test must include at least one expected-fail assertion where architecture specifies fail behavior.
7. Any unavoidable mock must be isolated to external third-party instability only, and must not mock project modules.
8. A test that imports app internals (`controller.*`, `worker.*`) to invoke business logic directly is **not** integration coverage.
9. Every `build123d` script used in tests must ensure every part has a `.metadata` attribute initialized with a `PartMetadata` or `CompoundMetadata` instance (imported from `shared.models.schemas`), following strict typing rules.
10. Every JSON, YAML, XML is converted to models e.g. Pydantic models and only then assertions are happening against them (for test maintainability and explainability). If there is no model for a JSON or similar schema, add it.

*Exception to the importing rules*: you can import python models and enums to use appropriate schema to avoid using pure json which will need to be manually updated later.
Commonly, these models and enums would be in `shared/` folder.


[... skipped some text ...]

## Required integration test suite

Priorities:

- `P0`: Must pass before merge to `main`.
- `P1`: Must pass in nightly/pre-release.
- `P2`: Extended production-quality/evaluation suite.

### P0: Architecture parity baseline

| ID | Test | Required assertions |
|---|---|---|
| INT-001 | Compose boot + health contract | `controller`, `worker`, `controller-worker`, `postgres`, `minio`, `temporal` become healthy/started; health endpoints return expected status payload. |
| INT-002 | Controller-worker execution boundary | Agent-generated execution happens on worker only; controller never runs LLM-generated code. |
| INT-003 | Session filesystem isolation | Two concurrent sessions cannot read each other's files. |
| INT-004 | Simulation serialization | Multiple agents may run, but only one simulation runs at a time (queue/lock behavior enforced). |

[ ... 170 tests differed by p0, p1, p2; also frontend_p0, frontend_p1, frontend_p2 ... ]

```
</details>

robertDouglass:
## Summary
- Adds `DIRECTIVE_036` (Black-Box Integration Testing) and companion tactic `black-box-integration-testing` directly to `shipped/`, using the new direct-to-approved flow.
- Scope and intent language is drawn from standard testing literature ([Wikipedia](https://en.wikipedia.org/wiki/Black-box_testing), [GeeksforGeeks](https://www.geeksforgeeks.org/software-testing/software-engineering-black-box-testing/), [Imperva](https://www.imperva.com/learn/application-security/black-box-testing/)) so the directive stands on the established paradigm rather than on any single project's phrasing.
- Addresses #203 — steers agents toward testing through external interfaces with observable-output assertions instead of unit-style coverage dressed up as integration.

## Test plan
- [x] New YAML files validate against `directive.schema.yaml` and `tactic.schema.yaml`
- [ ] CI doctrine validation passes

Closes #203

### DIFF
--- a/src/doctrine/tactics/shipped/black-box-integration-testing.tactic.yaml
+++ b/src/doctrine/tactics/shipped/black-box-integration-testing.tactic.yaml
@@ -3,20 +3,35 @@ id: black-box-integration-testing
 name: Black-Box Integration Testing
 purpose: >
   Validate the functionality of an application under test without peering
-  into its internal structures or workings, by driving the application
-  through its external interfaces and asserting on observable outputs.
-  Tests are derived from external descriptions of the software —
-  specifications, requirements, and design parameters — rather than from
-  its code.
+  into its internal structures or workings, by driving it through an external
+  interface and asserting on observable outputs. Tests are derived from
+  external descriptions of the software — specifications, requirements, and
+  design parameters — rather than from its code. "Black-box" describes the
+  testing technique (no structural knowledge of the system under test),
+  not the size of the slice under test. Applied here at the integration
+  boundary — exercising more than one collaborator together through a
+  single external interface — but the same discipline also applies at the
+  module, acceptance, and system levels.
 steps:
+  - title: Select the responsibility boundary
+    description: >
+      Before choosing what to run for real and what to substitute, decide
+      which components are directly responsible for the behavior under test
+      and which are outside that responsibility. Use the
+      test-boundaries-by-responsibility tactic to make this selection
+      explicit. The responsibility boundary — not the deployment boundary —
+      defines the "system under test" for this tactic.
+    examples:
+      - "Responsibility: job intake + validation + persistence. Inside: JobService, Validator, JobRepository. Outside: object storage, email delivery, upstream identity provider."
   - title: Anchor the test in a specification
     description: Identify the specification, requirement, or design parameter the test covers. Record it in the test name or docstring so the test can be read without reference to the implementation.
     examples:
       - "Covers REQ-user-registration-3: a POST /users with a duplicate email returns 409."
   - title: Identify the external interface
-    description: Determine the published interface through which the application under test will be driven — HTTP endpoint, CLI command, message topic, or equivalent. All test traffic goes through this interface.
+    description: Determine the published interface through which the system under test will be driven — HTTP endpoint, CLI command, message topic, public module API, or equivalent. All test traffic goes through this interface. The interface is "external" relative to the responsibility boundary selected above; it does not have to be a network boundary.
     examples:
       - "HTTP: POST /api/jobs. Assertions: response status, response body, GET /api/jobs/{id} read-back, emitted event on the job-events topic."
+      - "Module public API: JobService.submit(...) with a stubbed object-storage port. Assertions: returned Job aggregate, persisted row, emitted domain event."
   - title: Select inputs using black-box techniques
     description: >
       Choose valid and invalid inputs using standard black-box design techniques —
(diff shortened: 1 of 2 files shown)


src/doctrine/tactics/shipped/black-box-integration-testing.tactic.yaml
```diff
@@ -0,0 +1,65 @@
...
+references:
+  - name: Test and Typecheck Quality Gate
+    type: directive
+    id: DIRECTIVE_030
+    when: Black-box integration tests are part of the integration-level gate that must pass before handoff.
+  - name: Test-First Development
+    type: directive
+    id: DIRECTIVE_034
+    when: Black-box integration tests can be written first to capture the external contract before implementation.
+  - name: Test Boundaries by Responsibility
```
stijn-dejongh: This one is important for behavioural testing. Defining the test boundary / scope of codepath/system under test is crucial to successful use of this tactic: 

A black-box test refers to having no explicit structural knowledge of HOW the code is constructed, but it does not equate to "E2E / integration test" automatically. If we are testing a full functional slice, we can do so while still mocking out filesystem / DB / external systems.

stijn-dejongh (approved):
Minor comment in tactics and applicability: I am concerned about an over-reliance on e2e / integration tests, rather than acceptance/module tests.

src/doctrine/tactics/shipped/black-box-integration-testing.tactic.yaml
robertDouglass: Good catch — this landed as a real gap. I read your [define_test_boundaries](https://patterns.sddevelopment.be/practices/define_test_boundaries/) thesis and agree the prior text conflated "black-box" (a technique — no structural knowledge of the system under test) with "integration / E2E" (a scope).

Pushed d5a18acf to address it:

- Added an explicit first step: **Select the responsibility boundary** — pairs the tactic with the existing `test-boundaries-by-responsibility` tactic (your pattern, basically). The responsibility boundary, not the deployment topology, now defines the system under test.
- Generalized "stand up at deployable boundary" to allow in-process module composition at the responsibility boundary — compose stack is an example, not a requirement.
- Broadened the substitution rule: any outside-boundary dependency (other services, third-party systems, filesystem, clock, network) may be substituted when the substitute preserves the observable contract. Previously this only licensed third-party substitution.
- Added two failure modes: (1) conflating black-box with full-stack/E2E, (2) excessive substitution of inside-boundary collaborators.
- Directive now requires the test to **state its responsibility boundary** (in name, docstring, or supporting doc) so reviewers can tell what ran for real vs. what was substituted.

Net effect: the directive pushes agents toward specification-derived, observable-output assertions while leaving scope selection to `test-boundaries-by-responsibility` — black-box is available at module, acceptance, integration, and E2E levels, with the responsibility boundary decided per test.

Issue #654, opened 2026-04-15
Charter stack duplication: top-level charter and specify_cli.charter are both packaged, both live, and drifting


robertDouglass:
## Summary

The repository currently ships and uses two charter package surfaces:

- `src/charter/*`
- `src/specify_cli/charter/*`

This is **not doctrine duplication**. `src/doctrine/*` exists only once. The duplication is specific to the charter stack.

The current state appears to be a **backward-compatibility / staged-migration scaffold**, not an intentional long-term architecture. Both packages are still packaged, both are still imported by live `src/` code, some `specify_cli.charter` modules have already been collapsed into thin re-export shims over `charter.*`, and other modules remain separate implementations. Repo mission artifacts explicitly describe this as **"twin-package lockstep"** and note that deduplicating the twin packages was out of scope in prior work.

That means there is a real reason we cannot simply delete one tree today, but there is also no good architectural reason to keep two independent implementations long-term.

### 1. The duplication is real and intentionally shipped

`pyproject.toml` packages all of these:

- `src/kernel`
- `src/specify_cli`
- `src/doctrine`
- `src/charter`

The CLI entrypoint is still:

- `spec-kitty = "specify_cli:main"`

So the live CLI is `specify_cli`, but the top-level `charter` package is also intentionally distributed in the wheel.

### 2. Both import surfaces are live in production code

Live `src/` imports use **both** surfaces.

Examples importing top-level `charter.*`:

- `src/specify_cli/cli/commands/charter.py`
- `src/specify_cli/cli/commands/charter_bundle.py`
- `src/specify_cli/dashboard/charter_path.py`
- `src/specify_cli/upgrade/migrations/m_3_2_3_unified_bundle.py`
- `src/specify_cli/charter/context.py`
- `src/specify_cli/charter/sync.py`
- `src/specify_cli/charter/schemas.py`

Examples importing `specify_cli.charter.*`:

- `src/specify_cli/cli/commands/agent/workflow.py`
- `src/specify_cli/next/prompt_builder.py`
- `src/specify_cli/runtime/doctor.py`
- `src/specify_cli/charter/compiler.py`
- `src/specify_cli/charter/interview.py`
- `src/specify_cli/charter/extractor.py`
- `src/specify_cli/charter/generator.py`
- `src/specify_cli/charter/resolver.py`

This means the duplication is not dead source; removing either surface would currently break live code paths.

### 3. Only part of `specify_cli.charter` is already a shim

[...]

Issue #614, opened 2026-04-13
Separate integration boundaries inside src/specify_cli from core domain logic


stijn-dejongh:
## Parent

Part of #391 — Tech/Functional Debt Remediation

## Title intent

Separate integration boundaries inside `src/specify_cli` (orchestrator, sync, tracker, saas) from core domain logic.

## Status update

This issue remains **future work**.

The ownership-map tranche in **PR #683** establishes the slice boundaries that this extraction should use. This issue should stay open as the follow-on tracker for integration-boundary cleanup after the ownership map merges.

## Problem

Integration concerns currently sit close to core lifecycle/runtime behavior, which encourages ownership drift and makes it harder to tell whether a change affects core authority or external projection/transport.

## Goal

Clarify and enforce the separation between:

1. host-authoritative lifecycle / governance logic
2. external integration contracts and projections

## Acceptance criteria

- [ ] Orchestrator, sync, tracker, and saas boundaries have explicit ownership and dependency rules
- [ ] Core lifecycle/governance logic is not re-implemented inside integration packages
- [ ] Compatibility shims are documented where still needed
- [ ] Integration packages are easier to test independently
- [ ] Architectural docs reflect the resulting boundary split
- [ ] Any compatibility shims introduced by this extraction follow the registry/removal contract owned by #615

## Dependencies / sequencing

- Depends on the ownership map task under #610 being merged
- Should take into account accepted SaaS consolidation decisions such as the move toward `src/specify_cli/saas/`
- Benefits from functional CI splitting because integration slices can then be tested independently
- Architectural rule enforcement is mission-local using the existing `tests/architectural/test_layer_rules.py` infrastructure; there is no separate prerequisite issue for this

## Relationship to #461

Indirect but important.

Issue #461 pushes toward host-LLM-native execution, graph-backed governance, and clearer runtime boundaries. Integration slices should remain adapters and projections, not become alternate owners of execution/state authority.

stijn-dejongh:
**Safeguard scope update — folded from closed issues #393 / #394 / #395**

The ownership map (PR #683) previously referenced #393 (architectural tests), #394 (deprecation scaffolding), and #395 (import-graph enforcement) as external prerequisites for this extraction. Those issue IDs were incorrect — they pointed at unrelated closed bugs.

After review, all three safeguard concerns are clarified as follows:

[...]

Issue #594, opened 2026-04-11
Backlog: reduce Sonar cognitive-complexity debt in charter and agent workflow modules


robertDouglass:
## Problem
SonarCloud still reports a large backlog of maintainability issues, with the biggest contributors being cognitive-complexity violations in a small set of high-churn modules. During PR #592, the repo still showed critical complexity findings such as:

- `src/charter/catalog.py` (`python:S3776`)
- `src/charter/compiler.py` (`python:S3776`)
- `src/charter/resolver.py` (`python:S3776`)
- `src/specify_cli/cli/commands/agent/tasks.py` (`python:S3776`, very large function)
- repository helpers under `src/doctrine/**/repository.py`

## Why this matters
- These functions are expensive to change safely.
- They keep SonarCloud maintainability debt high even when security findings are addressed.
- The worst offenders are concentrated enough that targeted decomposition should pay down a meaningful share of the backlog.

## Proposed scope
- Split large branching functions into smaller parsing/decision helpers.
- Preserve behavior with narrow regression tests before each refactor.
- Prioritize the biggest complexity offenders first rather than broad stylistic cleanup.

## Suggested acceptance
- Reduce the highest-severity `S3776` findings in the files above.
- Keep refactors behavior-preserving with targeted tests.
- Re-run SonarCloud on the affected modules to confirm debt reduction.

stijn-dejongh:
## Partially addressed in mission `complexity-code-smell-remediation-01KP15HB` — WP03/T032

**Branch**: `feat/complexity-debt-remediation`

The charter cognitive-complexity backlog described here is being addressed in WP03 of the active complexity remediation mission. Coverage:

**Directly addressed (WP03)**:

- `src/charter/catalog.py` — **T032**: Characterize and decompose S3776-violating functions. The approach: extract each independent loading concern (`_load_paradigms`, `_load_tactics`, `_load_directives`, etc.) into named helpers so the top-level `load_doctrine_catalog` function becomes a pure composition with no deeply nested branches. Validated with `ruff check --select C901` (CC ≤ 10 per function) and a manual S3776 re-check.
- `src/charter/resolver.py` — **T015**: Decompose `resolve_governance` (CC 20) by extracting `_resolve_paradigms`, `_resolve_tools`, and `_resolve_directives` as named helpers. Conditional on absence of an active DRG rebuild mission (C-001 gate); implementer must verify at WP03 start.
[...]

Issue #391, opened 2026-04-05
EPIC: Tech/Functional Debt Remediation: 3.x development inconsistencies and behavioural problems


stijn-dejongh:
## Tech Debt Remediation

During 3.x development, we noticed existing and newly introduced inconsistencies and behavioural problems. This epic tracks their resolution.

### Scope

Structural and tooling issues that cause silent failures, degrade test coverage, or produce incorrect results during normal development workflows. These are not specific to any single mission — they affect any work that involves shell automation, bulk edits, module restructuring, or architectural enforcement.

stijn-dejongh:
I added a focused functional-module-extraction subtask set under this tech-debt umbrella, based on the current assessment of `src/specify_cli` ownership drift and duplication:

- #610 — Define canonical functional ownership map for `src/specify_cli` extraction work
- #611 — Consolidate charter ownership into `src/charter` and thin `specify_cli` charter adapters
- #612 — Extract runtime and mission decisioning into a clear functional runtime boundary
- #613 — Establish glossary as a clearly owned functional module
- #614 — Separate integration boundaries inside `src/specify_cli` from core domain logic
- #615 — Centralize migration and compatibility ownership for extracted functional modules

Recommended dependency / sequencing model:

1. **Safety nets first**
   - #393 bulk-edit occurrence classification
   - #394 validation for stringly-typed patch/logger/module references
   - #395 fail-loud architectural layer enforcement
2. **Ownership map first**
   - #610 defines canonical ownership before any large structural move
3. **Functional extraction in this order**
   - #611 charter consolidation
   - #612 runtime boundary extraction
   - #613 glossary boundary extraction
   - #614 integration-boundary cleanup
   - #615 migration / compatibility consolidation alongside or immediately after the first extractions

How this relates to #461:

- This stream is **not a duplicate of #461**.
- It is structural remediation under #391 that makes #461 easier to implement cleanly.
- #461 assumes clearer separation between governance, runtime, glossary, and integration concerns than the current `specify_cli` monolith actually provides.
- In particular:
  - #611 supports the governance / charter / DRG direction
  - #612 supports `ProfileInvocationExecutor`, `StepContractExecutor`, and mission composition
  - #613 supports glossary-as-runtime-middleware / DRG-resident glossary work
  - #615 reduces the risk that #461 introduces a second generation of compatibility cruft

Related supporting debt already on this umbrella:

- #549 (functional CI split) is the CI/testing expression of the same decomposition pressure
[...]

PR #661, opened 2026-04-16
Feat/complexity debt remediation

stijn-dejongh:
Addresses complexity/code-smell debt and duplicate compatibility surfaces introduced during earlier restructures/rebases.

Primary issue:
- #264

Umbrella:
- #391

Also addresses:
- #571
- #594

Partially addresses:
- #654
- #614


stijn-dejongh:
Drafting it, auto merge applied in github, rather than using the proper `rebase` approach.

stijn-dejongh:
## Boyscout pass — mypy, bandit, pip-audit, and CycloneDX SBOM

This PR landed a boyscout sweep across the static analysis stack. No logic changes — all fixes are type annotations, security suppressions with justifications, version floor bumps, and the SBOM pipeline.

---

### mypy (24 files)

Targeted the errors that were safe to fix without touching runtime behaviour:

- `dict` → `dict[str, Any]` annotations across 7 files to silence `type-arg`
- `TYPE_CHECKING` guard + `SkillRegistry | None` return types in two migration files (avoids circular import at runtime)
- `__all__` on `runtime/resolver.py` re-exports so mypy can follow them
- `cast(dict[str, Any], _run_async(...))` at three call sites in `local_service.py` — `_run_async` returns `Any` by design (thin asyncio bridge), cast is the right tool here
- Removed two stale `# type: ignore` comments that no longer matched the actual errors
- Fixed a `CompletedProcess[bytes]` vs `CompletedProcess[str]` variable collision in `stale_detection.py` by renaming the second binding
- Tightened `sync_data` annotation in `daemon.py` to resolve a union-attr chain

Deliberately **not** fixed: `agent/mission.py` (29 errors), `interview.py` (26 errors), and several architectural type mismatches (`Lane` vs `str`, `int | None` vs `str | None` in status models). Those need proper feature work, not a band-aid cast.

---

### bandit (17 findings → 0 unaddressed)

All suppressions carry inline justification comments:

| Rule | Files | Reason |
|---|---|---|
| B602 `shell=True` | `acceptance_matrix`, `git_ops`, `review/baseline` | Commands come from user-authored spec/config files; shell is intentional for pipes/redirects |
| B310 `urlopen` | `dashboard/handlers/api`, `dashboard/lifecycle` ×3, `sync/daemon` ×2, `sync/events` ×2 | All URLs are `http://127.0.0.1:<port>` localhost control-plane endpoints; scheme cannot be file:/ or custom |
| B314 XML parse | `review/baseline` | XML is JUnit output from a local test-runner subprocess, not network input |
| B608 SQL f-string | `sync/queue` ×4 | The interpolated value is `?,?,?` (count of placeholders), never user data; all values are parameterized |

---

### pip-audit (3 CVEs)

Bumped minimum version floors in `pyproject.toml`:

| Package | Old floor | New floor | CVE |
|---|---|---|---|
| `requests` | `>=2.31.0` | `>=2.33.0` | CVE-2026-25645 |
| `pytest` | `>=7.4` | `>=9.0.3` | CVE-2025-71176 |
| `pygments` | (transitive, unpinned) | `>=2.20.0` | CVE-2026-4539 |

`pygments` is a transitive dep via `rich` — added an explicit floor so `pip-audit` can verify the constraint is satisfied.

---

### CycloneDX SBOM

Every GitHub Release now attaches `sbom.cdx.json` alongside the wheel, tarball, and SHA256SUMS. The SBOM is generated by `cyclonedx-py environment` after the wheel builds, so it reflects the exact resolved dependency tree used for that release.

**Why this matters for enterprise evaluators:**

- **Compliance** — SOC 2, ISO 27001, and US EO-14028 all require a software inventory. An attached SBOM turns that into a checkbox rather than a manual exercise.
- **License clearance** — legal teams can auto-flag GPL/LGPL etc. from the SPDX identifiers in the SBOM without reading every transitive dep by hand.
- **Continuous CVE monitoring** — enterprises running Dependency-Track or similar tools can ingest the SBOM directly and get notified when a new CVE hits any of our deps, without filing a support ticket asking "are you affected?"
- **Procurement** — vendor risk processes increasingly gate approval on SBOM presence; shipping it removes friction for enterprise adoption.

PR #747, opened 2026-04-22
Fix Sonar propagator reliability bug and regex hotspots

robertDouglass:
## Summary
- retain fire-and-forget propagation tasks until completion so asyncio tasks cannot be garbage-collected early
- add a focused propagator test that exercises the running-loop path
- tighten directive-id regexes in charter synthesizer helpers to remove Sonar regex hotspot findings without changing valid matches

## Validation
- `python -m pytest tests/specify_cli/invocation/test_propagator.py -q`
- `python -m ruff check src/specify_cli/invocation/propagator.py tests/specify_cli/invocation/test_propagator.py src/charter/synthesizer/write_pipeline.py src/charter/synthesizer/resynthesize_pipeline.py`
- `PYTHONPATH=src python - <<'PY'`

## Notes
- `tests/charter/synthesizer/test_write_pipeline.py -q` is currently blocked on this machine because the repo test fixture venv bootstrap fails while creating `.pytest_cache/spec-kitty-test-venv` via `python -m venv`/`ensurepip`.

### DIFF
--- a/src/charter/synthesizer/resynthesize_pipeline.py
+++ b/src/charter/synthesizer/resynthesize_pipeline.py
@@ -121,7 +121,7 @@ def _rewrite_manifest(
     from datetime import UTC, datetime
 
     # Filename helper (mirrors write_pipeline._artifact_filename)
-    _DIRECTIVE_NUM_RE = re.compile(r"[A-Z]+_(\d+)")
+    _DIRECTIVE_NUM_RE = re.compile(r"^[A-Z]+_(\d+)$")
 
     def _filename(kind: str, slug: str, artifact_id: str | None) -> str:
         if kind == "directive":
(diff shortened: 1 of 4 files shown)


robertDouglass:
Closed instead of merged. The propagator intent was valid, but the anchored directive-id regex in this PR changes valid semantic IDs like  to , which is a real filename regression. The safe propagator/runtime pieces were carried in #751 instead.

robertDouglass:
Closed instead of merged. The propagator intent was valid, but the anchored directive-id regex in this PR changes valid semantic IDs like `RISK_APPETITE_003` to `000-*`, which is a real filename regression. The safe propagator/runtime pieces were carried in #751 instead.

PR #749, opened 2026-04-22
Fix safe upgrade cleanup and Gemini TOML coverage for 3.2.0a4

### LINKED ISSUE #673: [Bug] Gemini CLI shims are generated in Markdown instead of TOML
### Description

When running `spec-kitty init` for Gemini, the CLI generates command shims (like `/implement`, `/review`, etc.) in `.gemini/commands/` as pure Markdown files, even though the configuration specifies a `.toml` extension. This causes the Gemini CLI to fail to parse the commands because it expects a valid TOML structure with `description` and `prompt` keys.

### Root Cause

The `generate_shim_content` function in `src/specify_cli/shims/generator.py` was hardcoded to return Markdown format regardless of the agent's target extension. Additionally, `AGENT_ARG_PLACEHOLDERS` was missing Gemini and Qwen, causing them to use `$ARGUMENTS` instead of the required `{{args}}` syntax.

### LINKED ISSUE #674: `spec-kitty upgrade` silently deletes user-customized commands and skills
## Summary

Running `spec-kitty upgrade` deletes or clobbers files under `.claude/commands/` and `.claude/skills/` without checking whether the user has modified them. The most recent upgrade (3.0.3 → 3.1.4) **silently removed every file in `.claude/commands/`**, including ones I had heavily customized (e.g. `spec-kitty.pre-spec.md`, `spec-kitty.release.md`). It also gutted skill files in `.claude/skills/` — e.g. `spec-kitty-git-workflow/SKILL.md` went from ~290 lines to 1.

I only noticed because I re-read the commit that the upgrade auto-generated. Had I not reviewed the diff, the customizations would have been lost silently. On top of that, the upgrade auto-commits its own changes, which means the loss lands directly in git history instead of appearing as a local diff I can reject.

## The migration responsible

The upgrade commit body listed the applied migrations; the relevant one is **`3.1.2_globalize_commands`**. It appears to assume every file under `.claude/commands/` is canonical and replaceable by the global skill pack — it has no check against what the files currently contain. The same pattern (silent replacement) also affected `.claude/skills/*/SKILL.md` and their `references/` files during this upgrade.

## Why this is bad

1. **Silent data loss of user work.** No interactive confirmation, no backup, no diff preview of user-modified files. The migration just deletes or overwrites them.
2. **Auto-commit compounds the problem.** Because the upgrade auto-commits, there is no local "unstaged changes" prompt to catch the problem. Recovery requires the user to notice the deletion in the diff and `git show` the pre-upgrade tree to restore files by hand.
3. **`--dry-run` is not enough.** A dry run shows what *will* change, but the user still has to manually diff every modified file against canonical to know what is safe to accept. That is the job the tool should be doing.

## Proposed fix

Two parts: safer default behavior on upgrade, and a merge-assist command for resolving divergences after.

### Part 1 — upgrade should detect customizations and prompt

During `spec-kitty upgrade`, before touching any file in `.claude/commands/`, `.claude/skills/`, or `.agents/skills/`:

1. For each file the migration wants to replace or delete, compare its current content to the **canonical content of the user's current spec-kitty version** (i.e. the version we are upgrading *from*). The canonical content (or at least a checksum per file per version) should be shipped inside the CLI.
2. If the file matches canonical-for-old-version → safe to replace/delete silently.
3. If the file has diverged → the user has customized it. Do **not** silently clobber. Interactively prompt:
   ```
   .claude/commands/spec-kitty.pre-spec.md has been modified from its canonical 3.0.3 form.
   The 3.1.2_globalize_commands migration wants to delete it.
     [k] keep user version as-is (skip this migration step for this file)
     [b] back up to .claude/commands/spec-kitty.pre-spec.md.bak.3.0.3 and apply migration
     [o] overwrite with canonical (discard user changes)
     [d] show diff
     [s] save to .kittify/upgrade-conflicts/ for later merge
   ```
4. Respect `--force` by picking a sensible default (probably `[b]` — back up then apply) for non-interactive runs, and print a summary at the end listing every backed-up file.
5. Never auto-commit files that had conflicts without the user's explicit acknowledgement. The end-of-upgrade summary should list what was committed and what was left unstaged for review.

### Part 2 — a `spec-kitty merge-customizations` command

After an upgrade with conflicts, the user has:
- The current (new-canonical) version of each file.
- The backup of their customized version (`.bak.<old-version>` or staged in `.kittify/upgrade-conflicts/`).

Add a new command, e.g. `spec-kitty merge-customizations [--file <path>] [--agent claude]`, that launches an interactive merge session. Per conflicting file:

1. Show a three-way view: old-canonical, new-canonical, user-customized.
2. Ask the user to describe *what* their customization adds (one-liner).
3. Propose a merged version — either as a human-readable diff against new-canonical, or, if a Claude agent is attached, via an LLM-assisted merge that preserves the user's intent against the new template.
4. Let the user edit, accept, or reject the proposal.
5. Write the accepted result back to its original path and remove the backup.

This is the `git mergetool` pattern, scoped to spec-kitty assets and aware of which sections are canonical-meaningful vs. user-meaningful.

### Nice-to-haves

- `spec-kitty upgrade --backup-customizations-only` — run the customization-detection logic, copy divergent files to a backup directory, and **do not** apply any migrations. For users who want to review before committing to the upgrade.
- A `.kittify/customizations.yaml` manifest where users can explicitly declare "I have intentionally modified `.claude/commands/spec-kitty.pre-spec.md`" so the upgrade treats declared customizations as first-class rather than surprises.
- Include old-version canonical checksums for every shipped asset in `.kittify/skills-manifest.json` (or a sibling file) so the detection in Part 1 is cheap and offline.

## Acceptance criteria

- Running `spec-kitty upgrade` against a project with a modified `.claude/commands/spec-kitty.pre-spec.md` does **not** delete the file without prompting.
- Running `spec-kitty upgrade --force` backs up any user-modified asset before replacing it, and prints the backup paths.
- After an upgrade with conflicts, `spec-kitty merge-customizations` can walk the user through merging each backup into the new canonical.
- The upgrade commit does not include silently-deleted user files; or, if it does, the commit body lists them under a clear "CUSTOMIZATIONS OVERWRITTEN" section.

## Version

- `spec-kitty-cli 3.1.4`
- Upgrade path where I hit this: `3.0.3 → 3.1.4`
- Platform: macOS

robertDouglass:
## Summary
- harden the active `3.1.2_globalize_commands` migration so local command files are only removed when the global runtime exists, the exact global replacement exists, and the file is proven spec-kitty-generated
- add `3.2.0a4_safe_globalize_commands` for projects that already recorded `3.1.2_globalize_commands`, surface preserved/archive paths in upgrade results, and skip upgrade auto-commit when manual review is required
- lock down Gemini/Qwen command generation with runtime-path TOML tests and update the `3.2.0a4` changelog/docs to match the new migration path

## Testing
- `pytest tests/specify_cli/shims/test_generator.py tests/specify_cli/runtime/test_agent_commands_routing.py tests/upgrade/migrations/test_m_3_0_3_globalize_skill_pack.py tests/upgrade/migrations/test_m_3_2_0a4_safe_globalize_commands.py tests/upgrade/test_upgrade_auto_commit_unit.py -q`
- `pytest tests/upgrade/test_auto_discovery.py tests/upgrade/test_unified_bundle_migration.py -q`

Closes #673
Closes #674


robertDouglass:
Rebased this branch onto current `main`, retested the upgrade-focused changes locally, and pushed head `811b728c`. Required CI on this head is green. I also removed the `CHANGELOG.md` hunk so this PR no longer inherits the currently red release-readiness workflow from `main`. GitHub blocks self-approval for the PR author, so this is ready for an external approval pass.

PR #780, opened 2026-04-25
fix: unblock nightly patch-target and doctrine checks

robertDouglass:
## Summary
- make `specify_cli.next.runtime_bridge` importable for patch-target validation by lazily proxying runtime entrypoints instead of importing runtime modules eagerly at import time
- treat `spec_kitty_runtime.*` patch targets as external for `scripts/check_patch_targets.py`, matching the repo's existing skip strategy for non-repo modules
- remove generic `pytest` references from two shipped doctrine skills so the generic-language-bias doctrine check passes again

## Nightly findings addressed
- scheduled `CI Quality` on `main` for 2026-04-25 failed the enforced patch-target validator on `tests/next/*`
- scheduled `fast-tests-doctrine` on the same run failed `tests/doctrine/test_generic_artifact_language_bias.py`
- GitHub Dependabot alerts: none open
- GitHub code-scanning alerts: none open

## Validation
- Passed: `PYTHONPATH=src python scripts/check_patch_targets.py tests/next`
- Passed: `PYTHONPATH=src python scripts/check_patch_targets.py tests`
- Passed: `PYTHONPATH=src python -m pytest tests/doctrine/test_generic_artifact_language_bias.py`

## Local environment limits
- `uv run --python 3.12 pytest ...` could not be used in this fresh `/tmp` clone because `pyproject.toml` points `spec-kitty-events` to `../spec-kitty-events`, which does not exist under the fresh timestamped clone path
- broader `tests/next/*` execution in this clone is not representative because the local environment also lacks the sibling runtime dependency setup and hits a readonly local sync queue DB during CLI/integration paths
- GitHub CI is the source of truth for the full validation of this patch

### DIFF
--- a/scripts/check_patch_targets.py
+++ b/scripts/check_patch_targets.py
@@ -55,6 +55,7 @@ _SKIP_MODULE_PREFIXES = frozenset(
         "shutil",
         "tempfile",
         "unittest",
+        "spec_kitty_runtime",
     ]
 )
 
--- a/src/doctrine/skills/spec-kitty-implement-review/SKILL.md
+++ b/src/doctrine/skills/spec-kitty-implement-review/SKILL.md
@@ -782,14 +782,14 @@ does not work. This is the most common post-merge defect. Run
 `grep -r "from.*<new_module>" src/` to verify at least one live caller exists
 for every new module.
 
-**Test-DB collisions across parallel lane worktrees (Django/pytest-django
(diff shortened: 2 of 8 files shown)


robertDouglass:
Closing per SonarCloud PR triage: the patch is not merge-safe as submitted because it conflicts with current main and its latest CI has failed integration-tests-next, diff-coverage, and quality-gate. That does not meet the requested non-regression bar.

Issue #822, opened 2026-04-27
Epic: 3.2.0 stabilization and release readiness


robertDouglass:
## Goal

Ship `3.2.0` as a stabilization release from the current prerelease line. This epic is intentionally scoped to bugs, release hardening, workflow stabilization, and UX simplification. It should not absorb new product features.

The release should make the existing CLI reliable enough for stable use: no silent data loss, no confusing command success/failure output, no broken agent loop state, no unsafe intake behavior, and no release gates that miss contract drift.

## Status: 2026-04-28

Three large stabilization slices are now merged on `main` and GitHub hygiene has been reconciled against their issue matrices:

- PR #803 (`184b172e8c7d34bc341ce5805a2e66b7535b6c8a`) landed merge/lane safety, runtime/status correctness, intake safety, package contract gates, sync/auth hardening, governance hygiene, and e2e traceability.
- PR #837 (`daaee89596fafb9919d831e1c8c1fce284b2ba27`) landed 3.2.0a5 release-train reset and CLI surface cleanup.
- PR #864 (`d92d4292e5f5a5e53d61a170be1c8e8dc0921b7f`) landed 3.2.0a6 fresh-project, strict JSON, agent identity, review-cycle, invocation lifecycle, and charter parity cleanup.

Do not re-plan closed rows below unless a new repro appears after those merge commits.

## Explicitly in scope

- Fix release-train hygiene and dev-tooling blockers.
- Remove deprecated command surfaces while preserving canonical replacements.
- Stabilize merge, lane, worktree, runtime `next`, review, and status bookkeeping.
- Harden intake against traversal, symlink, prompt-injection, OOM, corrupt-state, and root-mismatch failure modes.
- Make package/contract/release gates hard enough to catch schema drift before signoff.
- Improve user-facing clarity where current command shapes or noisy sync errors make successful flows look broken.
- Preserve compatibility where possible, especially legacy aliases, while hiding/de-emphasizing old surfaces in help/docs.

## Explicitly out of scope for 3.2.0

These are good backlog items, but should not be pulled into this stabilization release unless they become direct blockers:

- Browser-mediated auth / PKCE auth expansion (#559, #561, #562, #564, #565).
- Repo-local tracker inbox and ticket intake as a product feature (#687, #695).
- Widen Mode expansion beyond already-landed work (#758 and related SaaS follow-ups).
- Dashboard/frontend decoupling and shared design system epics (#645, #650, #651).
[...]

PR #831, opened 2026-04-27
fix(upgrade): persist schema_version through metadata.save() and fix stale test args

stijn-dejongh:
## Summary

Two bugs in the upgrade pipeline found during a local `spec-kitty upgrade` + `spec-kitty verify-setup` alignment run.

### Bug 1 — `schema_version` silently dropped by `ProjectMetadata.save()`

`ProjectMetadata.save()` reconstructed the YAML dict from scratch from the dataclass fields, which did not include `schema_version`. Every `save()` call therefore stripped the field that the compat gate reads, leaving the project in a gated state (`verify-setup` / any non-upgrade command blocked with exit code 4) after any migration run.

The existing workaround (`_stamp_schema_version`) called a raw-YAML patch _after_ `save()` only in the "has migrations" code path. The "no migrations" early-return path (line 112–121 of `runner.py`) called `save()` and returned without ever stamping, so projects already at the target version would also lose `schema_version` on every upgrade.

**Root cause confirmed by reproducing the regression locally**: running `spec-kitty upgrade` on a clean checkout stripped `schema_version: 3` from `.kittify/metadata.yaml`, causing `verify-setup` to block with `PROJECT_MIGRATION_NEEDED`.

**Fix:**
- Add `schema_version: int | None = None` to `ProjectMetadata` dataclass
- Round-trip it through `load()` and `save()`
- Set `metadata.schema_version = REQUIRED_SCHEMA_VERSION` in both `upgrade()` code paths before `save()` — covers both "has migrations" and "no migrations" branches
- Remove the `_stamp_schema_version` post-save workaround (no longer needed)

**Test added:** `test_upgrade_preserves_schema_version_in_metadata` in `tests/cross_cutting/versioning/test_upgrade_version_update.py` — written failing first, then fixed.

Related to the 3.2.0 stabilization work tracked in #822.

---

### Bug 2 — Stale test args: `typer.Option` is truthy in direct function calls

Eight tests in `tests/upgrade/test_upgrade_auto_commit_unit.py` called `upgrade_cmd.upgrade()` directly (bypassing the CLI) without passing the `cli=` and `project=` kwargs added in WP09 (FR-014/FR-015). `typer.Option(False, "--cli")` returns an `OptionInfo` object which evaluates to `True` in a boolean context, causing the mutual-exclusion guard (`if cli and project: raise typer.Exit(2)`) to fire for every such call.

A ninth test (`test_upgrade_dry_run_skips_auto_commit`) additionally used `dry_run=True + json_output=True`, which routes to the compat-planner JSON path (line 449: `if json_output and (project or dry_run)`) instead of the auto-commit path the test was actually verifying.

**Fix:**
- Add `cli=False, project=False` to all 8 affected direct calls
- Change the dry-run test to `json_output=False` and assert only on the `safe_commit` spy

---

## Test plan

- [x] `test_upgrade_preserves_schema_version_in_metadata` — new regression test, confirmed failing before fix
- [x] `tests/upgrade/test_upgrade_auto_commit_unit.py` — all 33 tests pass (8 were failing before)
- [x] `tests/upgrade/ tests/specify_cli/migration/ tests/cross_cutting/versioning/` — 598 passed, 0 failures

### DIFF
--- a/src/specify_cli/upgrade/metadata.py
+++ b/src/specify_cli/upgrade/metadata.py
@@ -37,6 +37,7 @@ class ProjectMetadata:
     version: str
     initialized_at: datetime
     last_upgraded_at: datetime | None = None
+    schema_version: int | None = None
     python_version: str = ""
     platform: str = ""
     platform_version: str = ""
@@ -96,10 +97,17 @@ class ProjectMetadata:
         except ValueError:
             last_upgraded_at = None
 
+        raw_schema_version = spec_kitty.get("schema_version")
+        try:
+            schema_version: int | None = int(raw_schema_version) if raw_schema_version is not None else None
+        except (TypeError, ValueError):
+            schema_version = None
+
         metadata = cls(
             version=spec_kitty.get("version", "unknown"),
             initialized_at=initialized_at,
             last_upgraded_at=last_upgraded_at,
+            schema_version=schema_version,
             python_version=env.get("python_version", ""),
             platform=env.get("platform", ""),
             platform_version=env.get("platform_version", ""),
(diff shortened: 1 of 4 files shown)


robertDouglass:
Closing as superseded by the merged 3.2.0 stabilization work. The release line now includes the schema metadata fixes through PR #837/#864, and the related fixed issues have been closed with release evidence.

PR #1029, opened 2026-05-11
[Mission 6] Offline Queue Enqueue Performance

### LINKED ISSUE #352: Optimize offline queue enqueue path; COUNT(*) is only part of the problem
## Summary

`OfflineQueue.queue_event()` has a real production-side performance issue on the offline/local-queue path.

Important qualifier: `SELECT COUNT(*)` inside `queue_event()` is a legitimate inefficiency, but it is not the primary cost. The bigger problem is that each enqueue currently does a full SQLite round-trip and transaction.

## What I measured

Profiled `queue_event()` in `src/specify_cli/sync/queue.py`.

Observed hot spots during enqueue profiling:
- SQLite `commit` was the largest cumulative cost
- then SQLite `execute`
- then SQLite `connect`

A standalone `SELECT COUNT(*)` against a 10k-row queue on an already-open connection was effectively negligible.

So the slow path is not just the count query in isolation. It is the full per-event pattern in `queue_event()`:
- open connection at `src/specify_cli/sync/queue.py:332`
- `SELECT COUNT(*)` at `src/specify_cli/sync/queue.py:334`
- insert at `src/specify_cli/sync/queue.py:354`
- commit at `src/specify_cli/sync/queue.py:365`
- close at `src/specify_cli/sync/queue.py:370`

In short: `COUNT(*)` is extra work that should be removed, but the bigger win is reducing per-event connection/transaction overhead.

## Why this matters in real Spec Kitty usage

This is not just a test artifact.

`queue_event()` is on the real fallback path whenever events are queued locally instead of going out over WebSocket:
- missing `project_uuid` at `src/specify_cli/sync/emitter.py:585`
- unauthenticated / WebSocket unavailable / WebSocket send failure at `src/specify_cli/sync/emitter.py:720`
- runtime explicitly logs that events are queued locally when auth/WS is not active at `src/specify_cli/sync/runtime.py:129`

Impact expectations:
- low event volume: probably not noticeable
- bursty offline sessions: can become noticeable
- high-volume non-coalesced events: definitely noticeable

## Existing mitigation

There is already some mitigation via coalescing:
- `MissionDossierArtifactIndexed` and `MissionDossierSnapshotComputed` are coalesced in `src/specify_cli/sync/queue.py:26`

However:
- common events like `WPStatusChanged` do not coalesce
- those events still pay the full enqueue cost for every event

## Recommendation

Treat this as a real production optimization opportunity.

Do not scope the fix narrowly to `COUNT(*)` only. The issue is the entire per-event SQLite pattern.

Priority order:
1. Avoid one `connect + commit` cycle per event.
2. Remove the pre-insert `COUNT(*)` check, likely by tracking size separately or by restructuring eviction so recounting is unnecessary.

## Possible implementation directions

Potential approaches to evaluate:
- keep a longer-lived SQLite connection for the queue instance instead of opening/closing per enqueue
- batch multiple enqueue operations into fewer transactions where possible
- maintain queue size metadata instead of querying `COUNT(*)` on every enqueue
- only perform eviction bookkeeping when capacity is actually threatened
- preserve existing semantics for coalescing, FIFO eviction, and persistence across restarts

## Acceptance criteria

- local/offline enqueue throughput is materially improved for bursty event streams
- `queue_event()` no longer performs a full connect/commit/close cycle per event in the common path
- `COUNT(*)` is removed from the hot path or otherwise made non-recurring
- no regression in FIFO eviction behavior
- no regression in coalescing behavior
- no regression in persistence semantics

## Context

This came up while investigating slow CI tests. The CI side was improved by shrinking synthetic high-volume tests, but that only addressed the test harness. The underlying enqueue path is still a valid runtime optimization target.

robertDouglass:
## Summary

Removes the per-event ``SELECT COUNT(*) FROM queue`` from ``OfflineQueue.queue_event()`` and ``OfflineQueue.append()``. The cap check is now served from an in-process row-count cache that is maintained atomically by every mutation path, making the enqueue hot path **O(1) in queue depth** instead of O(n).

Closes #352.

## Changes

- **``src/specify_cli/sync/queue.py``**: new ``_row_count`` cache guarded by ``threading.Lock``. The hot path uses ``INSERT OR IGNORE`` + conditional ``UPDATE`` so the cache can distinguish a brand-new row from a duplicate-``event_id`` replacement (preserving the legacy ``INSERT OR REPLACE`` semantics). Every mutation path (``mark_synced``, ``clear``, ``remove_project_events``, ``process_batch_results``, ``drain_to_file``) keeps the counter coherent with disk. ``size()`` re-reads from disk so multi-instance scenarios (other ``OfflineQueue`` objects mutating the same DB file) stay correct.
- **``tests/sync/test_offline_queue_counter.py``**: 18 new tests covering cache initialization, increment/decrement on every mutation path, eviction, coalesce no-op, strict-cap ``OfflineQueueFull``, multi-instance refresh, and a white-box guarantee that the hot path never executes ``SELECT COUNT(*) FROM queue``.
- **``scripts/benchmarks/bench_queue_enqueue.py``**: reproducible before/after enqueue throughput benchmark.
- **``kitty-specs/offline-queue-enqueue-performance-01KRC4HN/benchmark.md``**: results — **1.87x speed-up** at 5k prefill (862→1608 ev/s).

## Benchmark

| Config | Before | After | Speed-up |
|--------|-------:|------:|---------:|
| prefill 5k, trials 1k, repeats 5 (median) | 862 ev/s | 1608 ev/s | **1.87x** |
| prefill 20k, trials 500, repeats 3 (median) | 1762 ev/s | 2054 ev/s | 1.17x |

(Numbers are single-machine, single-threaded. The smaller 20k speed-up reflects SQLite's page cache amortizing ``COUNT(*)`` — the structural fix removes the scan entirely, but the absolute savings shrink once the scan is page-cached. The cap-check cost is now O(1) in queue depth regardless.)

## Test plan

- [x] ``SPEC_KITTY_ENABLE_SAAS_SYNC=1 uv run pytest tests/sync/test_offline_queue.py tests/sync/test_offline_queue_counter.py tests/sync/test_edge_cases.py tests/sync/test_routing.py -q`` → **71 passed**.
- [x] Full ``SPEC_KITTY_ENABLE_SAAS_SYNC=1 uv run pytest tests/sync/ -q`` → **1503 passed, 7 failed**; all 7 failures are pre-existing on ``main`` (orphan_sweep × 5, daemon ``PackageNotFoundError`` × 1, tracker origin_integration × 1) and unrelated to this change.
- [x] ``uv run python scripts/benchmarks/bench_queue_enqueue.py`` reproduces the reported numbers.

## Out of scope (future work)

- Long-lived ``sqlite3.Connection`` + ``PRAGMA journal_mode=WAL`` to amortize ``sqlite3.connect()`` per call.
- Async batch flushing.

Mission slug: ``offline-queue-enqueue-performance-01KRC4HN``.


robertDouglass:
## [M6 fix] Reconcile cached counter against DB when near cap

Commit: `1d91ab4d`

### What was wrong
The cached `_row_count` is consulted on the fast path inside `queue_event()` (and `append()`), but a sibling `OfflineQueue` instance against the same DB invalidates that cache silently. Reproducer: two instances, `max_queue_size=2`, persisted depth reached 3.

### Fix
"Trust but reconcile near cap" — when `cached_count + 1 >= cap` on either insert path, force a `SELECT COUNT(*) FROM queue` before deciding. The hot far-from-cap path (`cached + 1 < cap`) is unchanged, so the Mission 6 perf win is preserved for the common case.

Sites touched in `src/specify_cli/sync/queue.py`:
- `queue_event()`: reconcile before the eviction decision
- `append()`: reconcile before raising `OfflineQueueFull`

### Tests
New regression class `TestMultiInstanceCapEnforcement` in `tests/sync/test_offline_queue_counter.py` covers:
- `test_queue_event_cap_holds_across_two_instances` — fills via instance A, then drives the second from instance B; persisted depth stays at cap
- `test_queue_event_cap_holds_when_sibling_fills_to_cap_minus_one` — the reviewer's exact `cap=2 → 3` scenario; persisted depth stays `<= cap`
- `test_strict_append_cap_holds_across_two_instances` — same hazard on the strict `append()` surface; second instance raises `OfflineQueueFull` instead of overflowing

```
$ SPEC_KITTY_ENABLE_SAAS_SYNC=1 uv run pytest \
    tests/sync/test_offline_queue.py \
    tests/sync/test_offline_queue_counter.py \
    tests/sync/test_edge_cases.py -q
70 passed in 5.91s
```

(21 counter tests, up from 18; the 3 new ones cover the multi-instance hazard.)

### Benchmark
Steady-state hot path (cache far from cap) is unaffected:

| prefill | rate (median, 3 repeats) |
|---|---|
| 5,000 | 1,744 ev/s |
| 50,000 | 1,580 ev/s |

At-cap branch (reconcile fires on every insert, cap=1,000, 500 trials after pre-fill):
- 929 ev/s (~1,076 µs/event)

The reconcile cost is paid only when callers are near the cap — exactly the bounded failure-mode trade we want.

Issue #1060, opened 2026-05-15
Legacy cleanup: remove hidden --feature aliases from the CLI surface


robertDouglass:
## Problem

`--mission` is the canonical selector, but many CLI commands still accept hidden `--feature` aliases. Hidden aliases reduce help clutter, but they keep the old vocabulary and parsing path alive indefinitely.

This is a bounded compatibility surface and should have an explicit removal plan.

## Evidence

Search from the #1056 review found hidden `--feature` usage across command modules including:

- `agent status`
- `agent tasks`
- `agent workflow`
- `context`
- `implement`
- `merge`
- `next`
- `research`
- `materialize`
- `validate-*`
- `mission_type` / lifecycle helpers

There is also a helper in `src/specify_cli/missions/_legacy_aliases.py` that explicitly preserves the alias.

## Plan

1. Inventory every command that still accepts `--feature`.
2. Classify each command by external compatibility risk.
3. Add a release-note/deprecation note that hidden `--feature` will be removed after the chosen cutoff.
4. Remove the alias in low-risk commands first.
5. Keep at most one temporary selector-resolution compatibility layer if needed for scripted users, with telemetry/logging if available.
6. Flip terminology guard tests from "hidden alias allowed" to "alias forbidden" after the cutoff.

## Acceptance Criteria

- No normal CLI command accepts `--feature` after the removal milestone.
- `tests/contract/test_terminology_guards.py` fails on any `--feature` Typer option, hidden or visible.
- `src/specify_cli/missions/_legacy_aliases.py` is deleted or moved to a historical migration fixture.
- Docs mention `--mission` only for active command examples.
- Any remaining `feature_slug` internal variable names are either renamed or explicitly scoped to historical data parsing.

## Related

- #1059 compatibility inventory and sunset policy

stijn-dejongh:
📋 **Specced** under mission `codebase-sanitization-1060-1622-01KV5F0B (mission_id 01KV5F0BPCVR42KCJX10ZQNB9D)` (epic #1797 slice), on branch `mission/codebase-sanitization-1060-1622`.

This mission addresses #1060 **partially** — the **internal/agent command cluster only**: `agent status/tasks/workflow/context/mission`, `charter lint`, `materialize`, `validate_encoding`, `validate_tasks`, `verify`. It removes the hidden `--feature` option from those commands, updates first-party templates/skills to `--mission`, and tightens the terminology gate to forbid the alias (hidden or visible) for that cluster.

[...]

Issue #1057, opened 2026-05-15
Legacy cleanup: retire pre-3.0 status and task readers from active runtime


robertDouglass:
## Problem

The docs say the canonical status model is append-only event logs plus derived snapshots, and that frontmatter lane is historical/migration-only. Active runtime code still contains readers for pre-3.0 task/status layouts, including legacy lane directories and frontmatter lane compatibility.

That keeps old project shapes in the hot path and makes it harder to reason about status authority.

## Evidence

- `docs/status-model.md` says `status.events.jsonl` is canonical and frontmatter lane is historical/migration-only.
- `src/specify_cli/task_utils/support.py` still locates WPs in legacy `tasks/{lane}/` directories.
- `src/specify_cli/legacy_detector.py` detects lane-directory projects for runtime helpers.
- Status/model/audit code still tolerates `feature_slug`, missing `mission_id`, and legacy lane aliases in several active paths.

## Plan

1. Define the post-cutover invariant for active runtime:
   - flat `tasks/WP*.md`
   - status from `status.events.jsonl`
   - `mission_id` present for machine identity
   - no frontmatter lane writes or reads for active decisions
2. Move lane-directory and frontmatter-lane readers behind migration/upgrade commands only.
3. Make active commands fail with a clear `run spec-kitty upgrade` message when they encounter unmigrated pre-3.0 shapes.
4. Remove or isolate tests that normalize legacy fixtures through active runtime paths.
5. Update docs that still describe lane frontmatter as a live workflow.

## Acceptance Criteria

- Normal task/status commands do not branch on `is_legacy_format()` for active execution.
- Pre-3.0 projects are handled by upgrade/migration commands, not silently by every runtime helper.
- Tests prove active commands reject unmigrated lane-directory projects with a migration instruction.
- Docs no longer tell users to rely on frontmatter lane state as the workflow authority.
- Audit/review can still inspect old artifacts, but that support is explicitly read-only and not part of normal mutation paths.

## Context

This is a concrete follow-up from the compatibility-cruft review around #1056.

robertDouglass:
## P1 verification - 2026-05-31

**Verification basis:** reviewed current open issue state, labels, title/body context, and the 2026-05-31 queue sweep. This is triage verification; implementation still must begin by reproducing or encoding the failure as a test/gate.

**Decision:** `KEEP` — Keep as an individual implementation issue; scope is narrow enough for one failing test and one remediation PR.

[...]

PR #1065, opened 2026-05-15
plan: enable hosted sync by default

robertDouglass:
## Summary

Planning PR for the TeamSpace launch path where hosted auth/tracker/sync are enabled by default instead of gated by `SPEC_KITTY_ENABLE_SAAS_SYNC`.

This keeps the old rollout symbols as compatibility shims for imports and test seams, but the canonical `is_saas_sync_enabled()` now always returns `True`. User-facing sync/tracker commands no longer no-op or hide themselves based on the env var; explicit checkout/repo opt-out remains through `spec-kitty sync opt-out` and existing routing config.

## Notes

- Draft on purpose; expected to be rebased as launch timing changes.
- This is not intended to land until we are ready for the release channel to treat SaaS sync as default-on.
- Docs were updated to describe `SPEC_KITTY_ENABLE_SAAS_SYNC` as deprecated compatibility only.

## Validation

- `uv run pytest tests/saas tests/sync tests/agent/cli/commands/test_tracker.py tests/agent/cli/commands/test_tracker_discover.py tests/agent/cli/commands/test_tracker_status.py tests/agent/cli/commands/test_sync.py tests/cli/commands/test_sync_routes.py tests/auth/test_auth_doctor_report.py tests/auth/test_auth_doctor_offline.py tests/auth/test_auth_doctor_repair.py -q`  
  Result: 1624 passed, 8 skipped, 1 warning.
- `uv run pytest tests/sync/test_background.py tests/sync/test_background_body.py tests/sync/test_batch_sync.py tests/sync/test_daemon_intent_gate.py tests/sync/test_dossier_trigger.py tests/sync/test_issue_598_hang_fixes.py tests/sync/test_runtime.py tests/sync/test_events.py tests/agent/cli/commands/test_tracker_discover.py tests/auth/test_auth_doctor_offline.py tests/auth/test_auth_doctor_repair.py -q`  
  Result: 257 passed.
- `uv run ruff check <touched focused files>`  
  Result: passed.
- `git diff --check`  
  Result: passed.
- Direct smoke: `SPEC_KITTY_ENABLE_SAAS_SYNC=0` still reports enabled and root help exposes `sync`, `tracker`, and `issue-search`.

### DIFF
--- a/MANUAL_TEST_PLAN.md
+++ b/MANUAL_TEST_PLAN.md
@@ -24,7 +24,7 @@
 
 - [ ] Python 3.11+ with `spec-kitty-cli` v2.0.2 installed (`pip install spec-kitty-cli`)
 - [ ] Git configured with SSH access to Priivacy-ai org
-- [ ] `SPEC_KITTY_ENABLE_SAAS_SYNC=1` environment variable set
+- [ ] Hosted auth/sync available by default; do not rely on `SPEC_KITTY_ENABLE_SAAS_SYNC`
 - [ ] `spec-kitty auth login` completed successfully
 - [ ] Local test repository initialized with `spec-kitty init`
 - [ ] Browser with DevTools available (Chrome/Firefox)
--- a/docs/migration/cross-repo-e2e-gate.md
+++ b/docs/migration/cross-repo-e2e-gate.md
@@ -46,7 +46,7 @@ least these scenarios. Future missions add more on top.
 |------|-------------|----------------|
 | `dependent_wp_planning_lane.py` | FR-001, FR-005, FR-038 | A mission with sequential dependent WPs plus a planning-lane WP merges with no silent omission of approved commits. |
 | `uninitialized_repo_fail_loud.py` | FR-032, FR-039 | `spec-kitty specify`/`plan`/`tasks` in a non-Spec-Kitty directory exit non-zero with `SPEC_KITTY_REPO_NOT_INITIALIZED` and write zero files into a sibling initialized repo. |
-| `saas_sync_enabled.py` | FR-040 | A full mission run with `SPEC_KITTY_ENABLE_SAAS_SYNC=1` against a configured dev SaaS endpoint produces sync emits at the endpoint, OR records a structured "endpoint unreachable" outcome that triggers the operator-exception path. |
+| `saas_sync_enabled.py` | FR-040 | A full mission run against a configured dev SaaS endpoint produces sync emits at the endpoint, OR records a structured "endpoint unreachable" outcome that triggers the operator-exception path. |
 | `contract_drift_caught.py` | FR-041 | Staging a fake `spec-kitty-events` candidate that drops a required envelope field causes `pytest tests/contract/` to exit non-zero with a missing-field diagnostic. |
 
 ## How to run the gate
@@ -54,8 +54,6 @@ least these scenarios. Future missions add more on top.
 From the spec-kitty repo:
 
 ```bash
-export SPEC_KITTY_ENABLE_SAAS_SYNC=1
-
 # 0. TeamSpace mission-state gate
 spec-kitty doctor mission-state --audit --fail-on teamspace-blocker
 
@@ -106,8 +104,7 @@ infra, not code."]
 ## Reproduction command
 
 ```bash
-SPEC_KITTY_ENABLE_SAAS_SYNC=1 \
-  pytest spec-kitty-end-to-end-testing/scenarios/saas_sync_enabled.py -v
+pytest spec-kitty-end-to-end-testing/scenarios/saas_sync_enabled.py -v
 ```
 
 ## Follow-up
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -72,13 +72,12 @@ spec-kitty merge
 
 ### SPEC_KITTY_ENABLE_SAAS_SYNC
 
-Opt in to hosted auth, tracker, and sync flows.
+Deprecated compatibility variable. Hosted auth, tracker, and sync flows are enabled by default.
 
-**Purpose**: Enables the SaaS-backed readiness path. Leave it unset for fully local CLI workflows.
+**Purpose**: Retained so old shells and scripts do not break. Its value is ignored in the release channel. Use `spec-kitty sync opt-out` to disable uploads for a checkout/repository.
 
 **Example**:
 ```bash
-export SPEC_KITTY_ENABLE_SAAS_SYNC=1
 spec-kitty auth login
 ```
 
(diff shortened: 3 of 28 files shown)


stijn-dejongh:
Heads-up, @robertDouglass — during a drift-triage of all open PRs we applied the **`pr:needs-refresh`** label to this one.

**What the label means:** *"PR branch drifted from `main`; rebase/refresh before review or merge."*

**Drift on this PR:** the branch is currently in **merge-conflict** with `main` and is roughly **1924 commits behind** (stale since May), so it can't be reviewed or merged as-is.

**Recommended next steps:**
1. **Rebase onto current `main`** to clear the merge conflict and catch the branch up.
2. **Cross-check the change against recently-merged work on `main`** — the surfaces this PR likely touches have shifted substantially. Worth reconciling against:
   - **Coord-topology surface-resolver / artifact-authority changes** — read/write routing now goes through `resolve_planning_read_dir` / `primary_feature_dir_for_mission` (#2106 plus the #2160 coord-authority cluster).
   - **`--feature` → `--mission` selector removal** (#1060).
   - **Pre-3.0 legacy-reader retirement + boundary guard** (#1057).
   - **Dead-symbol-gate hardening** (#2158).
   - **Resolution-authority gates** (#2173 lineage).
3. After rebasing, **re-run the architectural/CI gates locally** — several only run in CI, so a green local fast-suite isn't sufficient.

Given the size of the gap here, it may be worth assessing whether a fresh re-cut on top of current `main` is cheaper than a rebase. This is a courtesy heads-up, not a gate — nothing here blocks you, it just flags what to reconcile against before the branch goes up for review/merge.

robertDouglass:
## Fully dialectic adversarial review

**Verdict: reject as a merge candidate; keep as reference and re-cut on current main.** This is not safely recoverable by mechanical rebase: the branch is 2,889 commits behind, conflicting, and predates the current sync authority architecture.

### Safe fix pushed

Commit `24dbec4c8` fixes one independent false-green: `sync opt-out --delete-private-data --yes` now exits nonzero when remote share inspection or remote erasure fails, while preserving the completed local opt-out. Added both failure-path tests.

Validation: 9 route tests passed; the wider focused rollout/readiness/daemon/tracker/sync suite passed (129 tests); Ruff and diff checks passed. Hosted/sync tests were run with `SPEC_KITTY_ENABLE_SAAS_SYNC=1` per this machine’s rule.

### Report-only blockers

1. **Privacy/upgrade transition is unsafe.** The PR ignores explicit `SPEC_KITTY_ENABLE_SAAS_SYNC=0/false`, while unknown checkouts default to sync enabled. Existing users who deliberately used the variable as a network kill switch receive no persisted migration/opt-out before authenticated metadata/artifact delivery becomes eligible. A fresh design must preserve explicit false through deprecation or provide a versioned consent/policy migration, quarantine historical backlog, and test upgrades.

2. **Corrupt preference fails open.** `SyncConfig._load()` collapses unreadable/invalid TOML to `{}`; routing interprets missing preference as `True`. Probe: a valid checkout opt-out resolves false, then corrupting config makes the same checkout resolve true. Default-on turns this into a privacy failure. Policy loading needs a fail-closed state + diagnostic and conformance coverage through emitter/body/batch/daemon.

3. **Wrong current authority.** Main moved the canonical flag to `core/saas_sync_config.py`; `saas/rollout.py` is now a pure identity-preserving shim. Main also added target authority, capture-first journal, delivery ledger/policy, retention×delivery config, and honest opt-in semantics. Taking this PR implementation recreates split authority; taking main leaves the intended launch change inert. Re-cut against current CORE and enumerate every producer/dispatcher/readiness/daemon/tracker consumer.

4. **Revocation is not enforced at final drain authority.** Branch `batch_sync()` checks only the global gate, not per-checkout/repo eligibility. A machine-global drain can race opt-out; stale/UUID-less/multi-project backlog lacks durable target-scoped revocation. Current main’s target/journal/ledger boundary is where delivery eligibility and concurrency semantics must be enforced.

5. **Stable contract/ADR are contradicted.** The current rollout contract mandates truthy env parsing, shim object identity, fail-closed behavior, and conditional registration. The accepted ADR explicitly says the gate is not removed. A launch needs a superseding ADR + contract version, migration/privacy notice, and honest wording; retaining symbol names while reversing their semantics is not compatibility.

6. **Readiness compatibility is ceremonial.** `ROLLOUT_DISABLED` remains in the enum while its wording is removed; `_probe_rollout()` is called and discarded; `saas_sync_disabled_message()` says sync is enabled. Introduce a clearly named availability/policy API or preserve the old behavioral contract during migration.

### Concessions

- Intent is explicit and branch-local tests pin the requested old-tree mechanics.
- Valid checkout/repo opt-out purges project event/body queues and emitter/body paths honor valid policy.
- Auth and Private Teamspace safeguards remain; no credential leak/injection issue found.
- Always-visible read-only command surfaces and stable import re-exports are useful pieces to carry into the recut.

Historical May CI is green but predates the July CORE relocation and delivery-authority architecture, so it cannot establish current integration safety.

MOES-Media:
Flagging a coupling worth deciding **before** this lands, since it only becomes user-visible once hosted sync is default-on.

This PR makes `is_saas_sync_enabled()` always return `True`, so sync no longer no-ops for users who never set `SPEC_KITTY_ENABLE_SAAS_SYNC`. But the CLI's fallback target is still Fly:

```python
# src/specify_cli/sync/target_authority.py:51
DEFAULT_SERVER_URL = "https://spec-kitty-dev.fly.dev"
```

(mirrored at `sync/config.py:72` and `sync/queue.py:465-470`)

Meanwhile production Teamspace is Upsun — `spec-kitty-saas/README.md` documents project `67rt36f456a5m`, and the Fly runbooks are archived under `spec-kitty-saas/docs/archive/fly/`.

Today that mismatch is masked: sync is opt-in, and anyone who opts in follows the Upsun runbook, which sets both `SPEC_KITTY_SAAS_URL` and `spec-kitty sync server`. After this PR, a user with neither an env override nor a `[sync].server_url` in `config.toml` resolves to the Fly default via `_classify_override` (`target_authority.py:216`) and ships there silently.

Two things that make this more than a constant swap:

1. The resolved URL feeds `_derive_queue_scope` (`target_authority.py:267-282`) and therefore `scope_db_path` — so repointing the default changes **which SQLite queue DB** is selected for anyone relying on the fallback. Existing queued events under the old scope would need a migration story rather than being silently orphaned.
2. Explicit overrides are unaffected — I verified `resolve_sync_target()` returns `process_override` cleanly when env and config disagree, so runbook-configured users keep working either way. It is specifically the fallback path that changes behaviour.

Worth confirming whether repointing the default belongs in this PR's scope or wants its own change with a queue-scope migration. Happy to open the latter if that is the preference.

Context: surfaced while debugging a colleague whose events never reached SaaS. Root cause there was incomplete local setup, not this — but it is what made the stale default visible.

stijn-dejongh:
Closing against product direction ✅

Thank you, @robertDouglass — this planning PR is what forced the question into the open and let us make an explicit call. **Product direction is decided:** hosted sync consent is **explicit per-project opt-in with a separate sync ledger per project, and no default-on.**

That's the opposite of this PR's machine-global default-on (`is_saas_sync_enabled()` always `True`), and the reason is your PR's own collision point: since this was cut, #3030 made the sync-enabled signal a **structural consent primitive**, so defaulting it armed machine-wide would undo that boundary. Consent instead lives per project, granted explicitly.

The direction is now tracked canonically in #3262 (parented under the Team Kitty launch gate #1091), cross-linked to #3030 and #1800. Closing this PR since its premise is superseded — but the groundwork here directly shaped the decision. Thank you.

PR #1205, opened 2026-05-20
fix(acceptance): ignore matrix in grep_absence checks

### LINKED ISSUE #1147: Bug: `grep_absence` verification method in acceptance-matrix.json self-references (matches its own pattern in the file)
## Summary

The `grep_absence` verification method in `src/specify_cli/acceptance/matrix.py` runs `grep -r <verification_command> .` from the repo root — treating the `verification_command` field as a literal grep pattern. Because the matrix file itself is in the repo and contains that exact pattern text (as a JSON value), the grep finds a match in its own definition and flags the invariant as `still_present` forever. The matrix file's overall verdict becomes `fail` no matter what the actual mission state is.

## Reproduction

```bash
mkdir -p /tmp/sk-grepself && cd /tmp/sk-grepself
git init -q && git config user.email r@e && git config user.name r
SPEC_KITTY_HOME=/tmp/sk-grepself HOME=/tmp/sk-grepself spec-kitty init --non-interactive --ai claude >/dev/null
git add -A && git commit -qm init
SPEC_KITTY_HOME=/tmp/sk-grepself HOME=/tmp/sk-grepself \
  spec-kitty agent mission create probe --mission-type software-dev --json >/dev/null

# Author an acceptance matrix with a grep_absence invariant
cat > kitty-specs/probe-*/acceptance-matrix.json <<'EOF'
{
  "mission_slug": "probe-...",
  "criteria": [],
  "negative_invariants": [
    {
      "invariant_id": "INV-001",
      "description": "No legacy 'old_function' calls should exist in src/.",
      "verification_method": "grep_absence",
      "verification_command": "old_function",
      "result": "pending"
    }
  ]
}
EOF

# Run accept — the matrix re-runs the grep_absence check
SPEC_KITTY_HOME=/tmp/sk-grepself HOME=/tmp/sk-grepself \
  spec-kitty agent mission accept --mission probe-* --json
# Resulting matrix shows INV-001 result: "still_present"
# Because grep -r 'old_function' . found 'old_function' in the matrix file itself.
```

Real-world repro from mission `unblock-sync-identity-boundary-canary-01KRZJ07`: C-001 and C-005 negative invariants both flipped to `still_present` because their `verification_command` text (`"git log --oneline 45edd287a..HEAD --stat -- spec-kitty-end-to-end-testing/"` and `"git diff 45edd287a..HEAD -- src/specify_cli/audit/"`) appeared inside the matrix file the gate was scanning.

## Root cause

`src/specify_cli/acceptance/matrix.py:219-256`, function `_check_grep_absence`:

```python
def _check_grep_absence(repo_root: Path, ni: NegativeInvariant) -> NegativeInvariant:
    """Grep for pattern — zero matches means confirmed absent."""
    ...
    result = subprocess.run(
        ["grep", "-r", ni.verification_command, "."],
        cwd=str(repo_root),
        capture_output=True,
        text=True,
    )
    ...
```

Three design problems:

1. **The field name is `verification_command` but it's used as a grep pattern**, encouraging authors to write a command like `git diff -- src/foo/` (which gets fed to grep verbatim as a pattern).
2. **The grep target is `.`** — the whole repo, including the matrix file, evidence directories, and any other artifact that quotes the pattern. Self-reference is inevitable for any non-trivial pattern.
3. **No exclusion of well-known noise paths.** Even if the field were correctly used as a grep pattern, the matrix file, `canary-evidence/`, and `decisions/` artifacts all routinely echo back invariant text.

## Impact

- Any non-trivial `grep_absence` invariant flips to `still_present` and the overall_verdict to `fail`, blocking `spec-kitty agent mission accept`.
- Operators discover this only at accept time, after which they either rewrite as `custom_command` (boilerplate-heavy) or give up on the grep_absence method.
- The field name leads authors to write *commands* (`git diff ...`) instead of *patterns* (`old_function`), guaranteeing self-reference.

## Fix options (any subset)

1. **Rename the field semantics.** Either rename the field to `verification_pattern` and document that it's a grep pattern, or change the implementation to actually run the field as a shell command.
2. **Exclude the matrix file from its own scan.** At minimum `grep -r --exclude='acceptance-matrix.json' --exclude-dir='.git'`. Better: `--exclude-dir='kitty-specs/<slug>'`.
3. **Add a `repo_path_filter` field** so authors can scope the grep to e.g. `src/` only. Currently the method always scans the whole tree.
4. **Document `custom_command` as the safer default** in the schema docstring (`AcceptanceCriterion` / `NegativeInvariant` classes around `src/specify_cli/acceptance/matrix.py:24-46`).

(2) is a one-line fix and removes the most common footgun without breaking existing matrices.

## References

- `src/specify_cli/acceptance/matrix.py:43` — schema doc claims `verification_method: "grep_absence" | "route_check" | "custom_command"`.
- `src/specify_cli/acceptance/matrix.py:191` — docstring: "Run grep for pattern in repo, assert zero matches."
- `src/specify_cli/acceptance/matrix.py:219-256` — `_check_grep_absence` implementation.
- Discovered during mission `unblock-sync-identity-boundary-canary-01KRZJ07` acceptance gate. See `kitty-specs/<slug>/acceptance-matrix.json` C-001 and C-005 history (rewritten as `custom_command` as the workaround).

robertDouglass:
## Summary
- exclude acceptance-matrix.json from grep_absence scans so invariants do not match their own definitions
- pass grep patterns after `--` to avoid option-like patterns being treated as flags
- add regression coverage for the self-reference case

Fixes #1147

## Tests
- `uv run pytest tests/lanes/test_acceptance_matrix.py::TestNegativeInvariants -q`

### DIFF
--- a/src/specify_cli/acceptance/matrix.py
+++ b/src/specify_cli/acceptance/matrix.py
@@ -229,7 +229,15 @@ def _check_grep_absence(repo_root: Path, ni: NegativeInvariant) -> NegativeInvar
         )
 
     result = subprocess.run(
-        ["grep", "-r", ni.verification_command, "."],
+        [
+            "grep",
+            "-r",
+            "--exclude=acceptance-matrix.json",
+            "--exclude-dir=.git",
+            "--",
+            ni.verification_command,
+            ".",
+        ],
         cwd=str(repo_root),
         capture_output=True,
         text=True,
--- a/tests/lanes/test_acceptance_matrix.py
+++ b/tests/lanes/test_acceptance_matrix.py
@@ -163,6 +163,35 @@ class TestNegativeInvariants:
         results = enforce_negative_invariants(tmp_path, invariants)
         assert results[0].result == "still_present"
 
+    def test_grep_absence_ignores_matrix_file(self, tmp_path):
+        """grep_absence should not match its own acceptance-matrix definition."""
+        feature_dir = tmp_path / "kitty-specs" / "test-mission"
+        feature_dir.mkdir(parents=True)
+        pattern = "old_legacy_route"
+        (feature_dir / MATRIX_FILENAME).write_text(
+            "{\n"
+            '  "mission_slug": "test-mission",\n'
+            '  "negative_invariants": [\n'
+            f'    {{"verification_method": "grep_absence", "verification_command": "{pattern}"}}\n'
+            "  ]\n"
+            "}\n",
+            encoding="utf-8",
+        )
+
+        results = enforce_negative_invariants(
+            tmp_path,
+            [
+                NegativeInvariant(
+                    "NI-01",
+                    "No legacy route",
+                    "grep_absence",
+                    verification_command=pattern,
+                ),
+            ],
+        )
+
+        assert results[0].result == "confirmed_absent"
+
     def test_custom_command_pass(self, tmp_path):
         invariants = [
             NegativeInvariant(


robertDouglass (commented):
Approval rationale (GitHub does not allow me to approve my own PR): reviewed the focused fix for #1147. The grep_absence path now excludes acceptance-matrix.json from recursive scans, ignores .git, and sends the pattern after -- so option-like patterns do not become grep flags. Regression coverage exercises the self-reference case, and I also verified option-like pattern behavior locally. No blockers found.

PR #1209, opened 2026-05-20
fix(retrospective): materialize emitted events

### LINKED ISSUE #1183: Retrospective event emission leaves status.json stale after synthesize/apply
## Summary

Follow-up from PR #1167 review/merge: the retrospective post-mortem synthesize/apply workflow is still able to append valid `retrospective.*` events to `status.events.jsonl` without refreshing `status.json`.

This means a successful `spec-kitty agent retrospect synthesize --apply` can leave the mission snapshot stale until someone manually runs `spec-kitty agent status materialize`.

## What the investigation found

The workflow caused or exposed several issues seen while reviewing #1167:

1. **Absolute `provenance_ref` values were directly caused by synthesize/apply.**
   - Pre-#1167 code emitted `ProposalAppliedPayload(provenance_ref=str(prov_path))` from `src/specify_cli/doctrine_synthesizer/apply.py`.
   - That wrote machine-local absolute paths into committed `status.events.jsonl` rows.
   - #1167 fixed this by serializing repo-relative paths.

2. **`status.json` drift is still possible after retrospective apply.**
   - `src/specify_cli/cli/commands/agent_retrospect.py` calls `apply_proposals(...)`, renders output, and exits.
   - `apply_proposals()` emits `retrospective.proposal.applied` / `.rejected` events through `emit_retrospective_event(...)`.
   - `emit_retrospective_event(...)` appends to `kitty-specs/<mission>/status.events.jsonl` only; it does not call `status.reducer.materialize(feature_dir)`.
   - Normal status transition emitters do append + materialize, so retrospective event emission is inconsistent with the shared status log contract.

3. **The old `agent status validate` failure was a validator bug exposed by retrospective events.**
   - Retrospective rows intentionally share `status.events.jsonl`.
   - Before #1167, the validator treated raw `retrospective.*` rows as status lane-transition rows.
   - #1167 fixed that by validating status rows via `read_events()` while still using `read_events_raw()` where raw-log awareness is needed.

4. **The Sonar/coverage issue was unrelated to runtime behavior.**
   - That was a CI path-filter coverage gap for `src/specify_cli/doctrine_synthesizer/**`, not a synthesize/apply logic bug.

## Focused reproduction

On current `main` after #1167, a minimal temp repo repro did this:

- created a mission with one valid lane event and materialized `status.json`
- called `apply_proposals(..., dry_run=False)` with one accepted `add_glossary_term` proposal
- inspected `status.events.jsonl`, `status.json`, and `validate_materialization_drift(...)`

Observed output:

```text
events_emitted 1
status_json_changed False
status_json_has_retrospective False
last_event_name retrospective.proposal.applied
last_provenance_ref .kittify/glossary/.provenance/workflow_probe.yaml
drift_findings ['Materialization drift: retrospective snapshot differs between status.json and reducer output']
```

That shows the provenance path fix is working, but the event append still does not refresh `status.json`.

## Impact

- Valid retrospective apply operations can immediately leave `status.json` stale.
- `spec-kitty agent status validate` now correctly detects the stale snapshot as drift.
- Operators may need to manually run `spec-kitty agent status materialize` after retrospective synthesis.
- Committed mission artifacts can miss the derived `retrospective` block even when `status.events.jsonl` contains retrospective proposal events.

## Recommended fix

Centralize the fix in `emit_retrospective_event(...)`:

1. Append the retrospective event as it does today.
2. Best-effort call `status.reducer.materialize(feature_dir)` after the append, mirroring normal status transition behavior.
3. Log a warning if materialization fails, but do not roll back the already-appended event.

This covers all retrospective event families, not just synthesize/apply:

- `retrospective.requested`
- `retrospective.started`
- `retrospective.completed`
- `retrospective.skipped`
- `retrospective.failed`
- `retrospective.proposal.generated`
- `retrospective.proposal.applied`
- `retrospective.proposal.rejected`

## Suggested regression tests

- `apply_proposals(..., dry_run=False)` emits `retrospective.proposal.applied` and leaves `status.json.retrospective.proposals_applied == 1`.
- Conflict rejection emission materializes `status.json.retrospective.proposals_rejected`.
- If materialization raises, the retrospective event remains persisted and a warning is logged.
- Idempotent re-apply that emits no event should not dirty `status.json` unnecessarily.

## Related

- PR #1167 fixed the absolute `provenance_ref` bug and validator raw-row handling.
- This issue tracks the remaining workflow consistency gap.

robertDouglass:
## Summary
- refresh `status.json` after appending retrospective events
- keep retrospective event append durable if materialization fails, logging a warning instead of rolling back
- add regression coverage that a proposal-applied event updates the materialized retrospective snapshot

Fixes #1183

## Tests
- `uv run pytest tests/retrospective/test_reducer_integration.py::TestEmitRetrospectiveEvent tests/retrospective/test_reducer_integration.py::TestStatusSnapshotRetrospectiveField::test_materialize_attaches_non_absent_retrospective_snapshot -q`

### DIFF
--- a/src/specify_cli/retrospective/events.py
+++ b/src/specify_cli/retrospective/events.py
@@ -219,5 +219,15 @@ def emit_retrospective_event(
         fh.write(line + "\n")
 
     logger.debug("Appended retrospective event %s (%s) to %s", event_id, event_name, events_path)
+    try:
+        from specify_cli.status.reducer import materialize
+
+        materialize(feature_dir)
+    except Exception as exc:  # noqa: BLE001 - append succeeded; keep event durable
+        logger.warning(
+            "Retrospective event %s was appended, but status.json materialization failed: %s",
+            event_id,
+            exc,
+        )
 
     return event_id
--- a/tests/retrospective/test_reducer_integration.py
+++ b/tests/retrospective/test_reducer_integration.py
@@ -19,6 +19,7 @@ import pytest
 
 from specify_cli.retrospective.events import (
     CompletedPayload,
+    ProposalAppliedPayload,
     RequestedPayload,
     StartedPayload,
     emit_retrospective_event,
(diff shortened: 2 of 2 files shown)


robertDouglass:
Approval rationale (GitHub rejected formal approval because this token is the PR author): reviewed end-to-end. The implementation appends retrospective events first and then refreshes status.json through the existing reducer boundary, so status projection follows the event log without making materialization part of event durability. I added regression coverage for proposal-applied materialization, proposal-rejected materialization, and the failure path where materialize() raises after append. Local checks passed: focused reducer integration (29 tests), status/retrospective slice (1,039 tests), and touched-file ruff. GitHub checks are green on 85215de10.

PR #1352, opened 2026-05-28
fix: CI dependency resolver consistency and test surface hardening

stijn-dejongh:
## Summary

- **CI environment contract**: All test and lint jobs now use `uv sync --frozen --all-extras` instead of `pip install -e .[test]`, making the lockfile the single truth for both local and CI dependency resolution. This eliminates the class of failures where CI silently resolves a different version of a transitive dependency than `uv.lock` pins.
- **Typer 0.26+ compatibility**: `_JSONErrorGroup` exception handlers now catch both `click.exceptions.*` and `typer._click.exceptions.*` via `_CLICK_USAGE_ERRORS` / `_CLICK_ABORTS` tuples. Typer 0.26 vendors click internally; the old `except click.UsageError` silently missed every exception raised by that version.
- **Test surface alignment**: Charter-preflight fixtures and e2e conftest now use the canonical `charter.hasher.hash_content()` helper instead of raw `hashlib.sha256`, and the e2e conftest synthesises all required runtime state after `copytree` (no longer depends on gitignored machine state).
- **Lint enforcement**: `TID251` banned-api rules prevent future tests from re-implementing production hash algorithms or asserting on `click.exceptions.*` directly.
- **Doctrine CLI remediation**: Restored `doctrine` Typer group registration in `register_commands()` (incorrectly removed in this mission's FR-004 by treating a stale test as a contract). Narrowed `test_doctrine_cli_removed.py` to guard only the retired pre-DRG curation commands (`curate`, `promote`) — not the parent group, which carries the DRG-era authoring surface (`fetch`, `new`, `validate`, `pack`, `org`). Restored all `spec-kitty doctrine` sections in `docs/reference/cli-commands.md`. See PR comment for full root-cause analysis.
- **ADR**: `architecture/adrs/2026-05-28-1-ci-dependency-resolution-and-test-surface-consistency.md` documents discovery, root-cause analysis, five structural gaps, chosen options, and confirmation criteria.

## Closes / relates

Related to issues #1302, #1304, #1305, #1306, #1307, #1308, #1309, #1310 — do **not** close these until after this PR merges.

## Test plan

- [x] `uv lock --check` passes locally after version bump to `3.2.0rc29`
- [x] All test jobs use `uv sync --frozen --all-extras` (verified in workflow YAML)
- [x] `tests/agent/test_json_group_typer_surface.py` smoke-tests the `_JSONErrorGroup` JSON-envelope path
[...]

PR #1395, opened 2026-05-30
fix: CI dependency resolver consistency and test surface hardening

robertDouglass:
Same-repo reroll of #1352 so required secret-backed CI can access `SPEC_KITTY_SAAS_READ_TOKEN` where repository policy permits it.

Supersedes #1352.

Includes the same current head commit (`3bd13c5f6d917308797eddd0e2dd581d061c62d2`) from the fork PR, pushed into `Priivacy-ai/spec-kitty:feat/pre-doctrine-stabilization-remediation`.

Original summary:
- CI dependency resolution now uses `uv sync --frozen --all-extras`.
- Typer 0.26+ JSON envelope compatibility fixes.
- Charter/e2e fixture alignment.
- TID251 lint enforcement for banned test APIs.
- Doctrine CLI remediation and ADR documentation.


robertDouglass:
## Remediation plan for TID251 enforcement

Goal: make banned test API usage block CI, while preserving legitimate raw SHA-256 uses through explicit local justification.

1. Add a dedicated enforced CI step in the `lint` job:

```bash
uv run python -m ruff check src tests --select TID251
```

Do this separately from the existing advisory full Ruff report. Full `uv run ruff check src tests` currently fails on unrelated legacy Ruff findings, so making all Ruff blocking would broaden this PR unnecessarily.

2. Fix the live TID251 violation at `tests/regression/test_issue_1348.py:111`.

If it is file-integrity hashing, keep raw SHA-256 and add a specific justification:

```python
return hashlib.sha256(path.read_bytes()).hexdigest()  # noqa: TID251 - file-integrity checksum, not charter freshness hashing
```

If it is charter freshness semantics, replace it with `charter.hasher.hash_content()`.

3. Correct the `pyproject.toml` policy mismatch.

Recommended rigorous fix: remove the broad test-dir `TID251` ignores and annotate legitimate existing raw `hashlib.sha256` calls with `# noqa: TID251 - <specific reason>`. This makes the "new calls require justification" rule real.

Lower-churn fallback: keep the directory exemptions, but change the comments to honestly say those dirs are exempt. This is weaker and should not be described as requiring `# noqa` in those dirs.

4. Add a regression guard.

Add an architectural/config test that proves the banned API command is present in CI and would fail on an unannotated banned API. If broad ignores are removed, include coverage for a formerly exempt path.

5. Verify before reroll:

```bash
uv run python -m ruff check src tests --select TID251
uv run pytest tests/specify_cli/audit/test_wp_files_classifier.py tests/specify_cli/next/test_runtime_bridge_documentation_composition.py tests/specify_cli/mission_step_contracts/test_documentation_composition.py tests/architectural/test_no_tracked_test_feature_missions.py tests/agent/test_json_group_typer_surface.py -q
```

6. Commit and push:

```bash
git commit -m "fix: enforce banned API lint gate"
git push
```

Success condition: the PR can only be green when TID251 passes, and any legitimate raw SHA-256 use is either outside the enforced surface or locally justified.

robertDouglass:
## Adversarial review finding: BLOCK

TID251 is not actually enforced despite the PR summary claiming "TID251 lint enforcement for banned test APIs."

Evidence:

- `pyproject.toml` adds `TID251` to Ruff selection and configures banned APIs.
- `.github/workflows/ci-quality.yml` runs Ruff as advisory with `continue-on-error: true`, then tells reviewers these are advisory and do not block the workflow.
- PR head is green, but the enforced command we would expect to pass fails locally:

```bash
uv run ruff check src tests --select TID251
```

Current failure:

```text
TID251 `hashlib.sha256` is banned
   --> tests/regression/test_issue_1348.py:111:12
```

There is also a scope hole in the policy text. `pyproject.toml` says new `sha256` calls in ignored test directories still require `# noqa: TID251`, but whole-directory per-file ignores make that unenforceable. For example:

```bash
uv run ruff check tests/charter/synthesizer/test_bundle_validate_extension.py --select TID251
```

passes despite unannotated raw `hashlib.sha256` calls.

Prior blockers were rechecked and appear resolved: documentation `accept` now has contract/profile/default/tests, corrupt event logs no longer crash `classify_wp_files`, and tracked `kitty-specs/test-feature-*` files are gone. Focused suite passed: `38 passed, 1 warning`.

robertDouglass:
## Adversarial BLOCK remediated — TID251 is now genuinely enforced

Validated the BLOCK with a five-paradigm parallel investigation (debugger-debbie); all five paradigms independently reproduced it. Verdict: **confirmed**. The implementation contradicted its own ADR (`2026-05-28-1`, Gap 3/Gap 5, *"automated enforcement rather than convention"*) — TID251 was advisory config, not a gate.

**Fixed in `0ccf835ca`:**

| Finding | Fix |
|---|---|
| TID251 advisory only (`continue-on-error`, command omitted `--select`) | New **`[ENFORCED] banned-API lint gate (TID251)`** step in the `lint` job — `ruff check src tests --select TID251`, no `continue-on-error`. Now gates the build via `needs.lint.result`. |
| Scope hole: 10 whole-dir `tests/**` per-file-ignores silently disarmed the rule | Removed all of them (incl. stale `tests/skills/**` for a non-existent dir). Annotated **61** legitimate raw `hashlib.sha256` sites across **26** files with justified `# noqa: TID251 — <reason>`. |
| Live violation `tests/regression/test_issue_1348.py:111` on green | Annotated (file-integrity checksum, not charter freshness). |
| ADR referenced non-existent rule `# noqa: SC001` | Corrected to `TID251`. |
| "enforcement" claim was false-as-shipped | Now true; CHANGELOG clarified. |

**Verification:** `ruff check src tests --select TID251` → *All checks passed!* (0). New guard `tests/architectural/test_tid251_enforcement.py` (5 passed) pins the enforced step + no-whole-dir-exemption + fire/suppress behaviour, so this cannot silently regress.

**Deliberately out of scope** (pre-existing debt in files this PR doesn't touch; full-ruff stays advisory): the 11 non-TID251 ruff findings, plus adjacent gaps. Tracked separately: #1398 (ruff backlog), #1399 (mypy `built-in/__init__.py`), #1400 (retired `spec-kitty-runtime` in CI), #1401 (typer 0.26 path unexercised), #1402 (Gap-5 `click.exceptions` dead config).

Issue #1589, opened 2026-06-01
Coordination/mission branch status desync: runtime reads empty canonical status; dependency cycle silently leaves WPs uninitialized


stijn-dejongh:
## Environment
spec-kitty **3.2.0rc32**, lane-based mission `org-doctrine-profile-integrity-activation-closure-01KT1TV1` (18 WPs, coordination branch `kitty/mission-…-01KT1TV1`).

## Symptoms
During the implement-review loop, every `move-task … --to for_review` (from lane worktrees AND from the coordination worktree) failed with:
> Error: WP WPxx has no canonical status in feature … Run `spec-kitty agent tasks finalize-tasks --mission …` to initialize.

…but `finalize-tasks` had already run and reported `bootstrap: {newly_seeded: 18}`. `spec-kitty next` / `agent status lifecycle` showed **0 work packages** and 0% while `agent status doctor` reported the mission **"Healthy"**.

### 1. A tasks.md dependency cycle silently leaves canonical status uninitialized
`tasks.md` contained a WP09↔WP10 cycle (WP09 deps WP10; WP10 deps WP09). `finalize-tasks --json` *does* report it (`{"error": "Circular dependencies detected: [['WP09','WP10','WP09']]"}`), but:
- the bootstrap step is **aborted by the cycle**, so no `planned` lane events are emitted;
- the **downstream `move-task` error does not mention the cycle** — it just says "no canonical status, run finalize-tasks", which loops forever because finalize keeps aborting on the same cycle.

The cycle error should be surfaced at the point of failure (move-task/next), not only in finalize's JSON. **Suggestion:** chain the real cause into the "no canonical status" message, or have `next`/`move-task` detect-and-report an unbootstrapped-due-to-cycle state.

### 2. `agent status doctor` reports "Healthy" while the mission has 0 materialized WPs
`agent status lifecycle` shows `Work Packages: 0` and `materialize(feature_dir)` returns an empty snapshot, yet `doctor` says "Healthy". Doctor should flag an uninitialized/0-WP mission that has a populated `tasks.md`/lanes.json.

### 3. Coordination vs mission branch status read desync
After fixing the cycle and re-running finalize, canonical status is healthy **only in the coordination worktree**:
- `.worktrees/…-coord/…/status.events.jsonl` → **60 lane-transition events**, `materialize` → **18 WPs**.
[...]

PR #1597, opened 2026-06-01
fix(#1589 facet 3): status reads resolve the coordination worktree

stijn-dejongh:
Addresses facet 3 of #1589 (the coordination/mission status read desync). Companion to #1590 (facets 1 & 2).

## Problem

Lane-based missions commit canonical status to the per-mission **coordination worktree** (via the transactional emit path), but several read sites resolve `feature_dir = repo_root/kitty-specs/<slug>` against the **primary checkout** and read its stale mission dir. Result: `move-task`/`next`/lane reads report "WP has no canonical status / run finalize-tasks" even when the coordination worktree has a fully materialized log. `workflow` and `acceptance` already resolved coordination-aware; `move-task`'s direct `read_events`, `next`, `lane_reader`, and `materialize` did not.

## Fix

New `specify_cli/status/coord_read.resolve_status_dir()` resolves a `feature_dir` to the coordination worktree when one exists on disk, and is a **strict no-op** otherwise (legacy/no-coord missions, and paths already inside a `-coord` worktree). It only redirects when a coordination worktree is actually found — never to a recomposed primary path — so non-coord missions are byte-for-byte unaffected. Applied at the store chokepoint (`_events_path`), so `read_events`/`append_event` and every reader converge on the canonical log; mid8 is derived cheaply (trailing slug token first, `meta.json` fallback) since the path is hot.

## Tests

`tests/status/test_lane_reader_coordination_aware.py`: coord-backed read resolves (RED before — reads the empty primary → `CanonicalStatusNotFoundError`); legacy no-coord unchanged; missing-everywhere still raises. 666 status-suite tests pass; ruff + mypy clean on touched files.

Boy-Scout: fixed a pre-existing mypy `no-any-return` in `store._event_matches_expected`.

## Scope note

This fixes the **read** resolution. A **separate, deeper** coordination-layer issue remains (status events written to the coordination worktree are not always durably committed to the coordination branch, so they can be lost on subsequent worktree git operations) — being filed separately; it is out of scope here.

### DIFF
--- a/docs/development/3-2-page-inventory.yaml
+++ b/docs/development/3-2-page-inventory.yaml
@@ -2003,6 +2003,13 @@
   current_target: false
   citation_refs: []
   notes: null
+- path: docs/development/facet3-status-read-desync-design.md
+  tag: internal
+  divio_type: explanation
+  owning_workstream: status-coordination
+  current_target: false
+  citation_refs: []
+  notes: Design note for PR #1597 status read coordination worktree desync.
 - path: docs/development/issue-1040-scope-assessment.md
   tag: internal
   divio_type: none
--- a/docs/development/facet3-status-read-desync-design.md
+++ b/docs/development/facet3-status-read-desync-design.md
@@ -1,0 +1,114 @@
+# Design: fix #1589 facet 3 — coordination/mission status read desync
+
+**Status:** proposed (design-first; no code yet)
+**Scope:** `spec-kitty` status read path for lane-based missions
+**Related:** [#1589](https://github.com/Priivacy-ai/spec-kitty/issues/1589) facet 3; builds on #1590 (facets 1 & 2)
+
+## 1. Root cause (precise, evidence-backed)
+
+Lane-based missions keep canonical status on a **coordination branch**, materialized in a
+per-mission **coordination worktree** (`.worktrees/<slug>-<mid8>-coord/kitty-specs/<slug>-<mid8>/`).
+
+- **Writes are already correct.** `finalize`, `implement`, `move-task`, `merge`, and
+  `bootstrap_canonical_state` all emit through
+  `coordination.status_transition.emit_status_transition_transactional`, which commits the
+  event to the **coordination branch** (via `BookkeepingTransaction`). Verified: the coord
+  worktree holds the full log (82 lane events → 18 WPs materialize).
+
+- **Reads are inconsistent.** Two commands were updated to read coordination-aware:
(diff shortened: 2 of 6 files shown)


robertDouglass:
Adversarial follow-up after docs freshness fix (`aa1250c17`): docs inventory/status issue is addressed and local `check_docs_freshness.py --ci --link-check none` now exits 0, leaving only the existing `HELP-DRIFT mission` warning.

Remaining blockers I would not merge over:

- `resolve_status_dir()` still misses callers whose `feature_dir` is inside a lane worktree. It derives repo root as `feature_dir.parent.parent`, so `repo/.worktrees/<mission>-lane-a/kitty-specs/<mission>` searches for `<lane-a>/.worktrees/*-coord` instead of sibling `repo/.worktrees/*-coord`. Minimal repro output from this branch: `primary resolves to coord: True`, `lane resolves to coord: False`, then `get_wp_lane` raises canonical status missing despite coord log existing. Add a regression for lane worktree -> sibling coord worktree, not only primary checkout -> coord.

- `diff-coverage` remains failing: `src/specify_cli/status/coord_read.py (87.2%): Missing lines 37,41-42,69-70`; total diff coverage `88%` vs required `90%`. The uncovered paths are meta fallback and resolver exception/no-op behavior.

- `Clean install verification (FR-010)` remains failing on NFR-003 latency: post-cutover median `1.056s`, delta `+41.8%`, ceiling `<=34.2%`, verdict `FAIL`. `quality-gate` fails because required upstream jobs failed.

I fixed only docs freshness per request; the above need code/test/perf follow-up before merge.

robertDouglass:
Closing as superseded by the narrower landed path in #1598 + #1606.\n\nTriage result:\n- #1598 landed the write-side fix: coordination commits no longer clobber canonical coord-branch event history.\n- #1606 landed the read-side fix: transactional status reads use the same target as transactional writes, and move-task now derives old_lane from transactional status rather than primary-checkout events.\n\nThis PR's design and tests correctly identified the read desync, but the implementation is too broad for the immediate bug fix: changing status.store._events_path() globally can redirect both reads and non-transactional writes into an existing coord worktree without necessarily committing those writes to the coordination branch. That is exactly the class of split-brain/write-durability risk we are trying to remove.\n\nPreferred future slice, if still needed: salvage the coord-aware read tests/design into explicit read APIs or command-level status-read adapters only; do not globally redirect the store write path unless write durability and branch commits are covered end-to-end.

PR #1598, opened 2026-06-01
fix: planning-artifact claim-blockers + coordination event-log clobber (#1602)

stijn-dejongh:
Two distinct, compounding defects in the WP-claim planning-artifact commit path (`cli/commands/implement.py`). Either one alone makes lane-based missions unable to progress past the first claim; together they were the real cause of the apparent coordination 'status reset / write-durability' problem.

## Bug 1 — porcelain parse truncates worktree-only paths

`_feature_dir_status_paths()` sliced `line[3:]` off `_git_stdout()`, which `.strip()`s the whole `git status --porcelain` output. Porcelain v1 emits a fixed 3-char prefix `XY<space>PATH`; for a modified-but-unstaged tracked file `X` is a space, so the first line is `" M path"`. The outer strip removed the leading space, shifting columns so `line[3:]` ate the first path char (`kitty-specs`→`itty-specs`). The bogus path failed `source.exists()`, the transaction staged nothing, and the claim died with 'commit() called with no events or artifacts to commit'.

**Fix:** parse from raw stdout (no outer strip), slice the fixed 3-column prefix per line; resolve rename `old -> new` to the new path.

## Bug 2 — non-idempotent commit vs the coordination branch

The coordination model commits a claim's planning-artifact edits to the coordination branch but leaves them uncommitted in the main checkout. The next claim re-discovers those edits via `git status` even though their content is already on the coord branch. Re-committing identical content is an empty commit, which `safe_commit` rejects ('git commit failed') — so only the first claim of a mission ever succeeded.

**Fix:** filter discovered files against the coordination branch HEAD (`git show <ref>:<path>`); skip those whose working-tree bytes already match, and skip the commit entirely as an idempotent no-op when nothing genuinely differs.

## Tests (ATDD, real git repos)

- `TestFeatureDirStatusPaths` — modified-but-unstaged tracked file; RED `'itty-specs/...'`, GREEN full path + `exists()`.
- `TestPlanningArtifactIdempotentCommit` — file dirty-in-main but identical-on-coord; RED `safe_commit 'git commit failed'`, GREEN no-op with no empty commit on the coord branch.

Full `test_implement.py` (7) passes; ruff + mypy clean on the touched file.

Verified end-to-end on a real mission: WP01/WP02/WP13 now all claim successfully, commit to the coordination branch, and their `planned → in_progress` transitions persist across materialization (previously only the first claim worked).

### DIFF
--- a/src/specify_cli/cli/commands/agent/workflow.py
+++ b/src/specify_cli/cli/commands/agent/workflow.py
@@ -34,6 +34,7 @@ contract.
 
 from __future__ import annotations
 
+import json
 import logging
 import re
 import subprocess
@@ -239,6 +240,54 @@ def _canonical_status_feature_dir(main_repo_root: Path, mission_slug: str) -> Pa
     return resolve_mission_read_path(main_repo_root, mission_slug, mid8)
 
 
+def _merge_event_log_bytes(existing: bytes, incoming: bytes) -> bytes:
+    """Append-only union of two JSONL event logs, keyed by ``event_id`` (#1602).
+
+    The coordination branch's canonical ``status.events.jsonl`` is authoritative
+    and append-only: a workflow commit must NEVER drop an event already recorded
+    there. But the main-checkout copy fed into the coordination commit can lack
+    lane transitions that only live on the coordination branch (they are emitted
+    straight to the coord worktree), while carrying non-lane lifecycle/decision
+    envelope events the coord copy lacks. Overwriting the coord log with that copy
+    wipes the lane history — ``read_events()`` then returns 0 and the loop wedges
+    with "no canonical status".
+
+    This merge preserves every ``existing`` (coord) line in order, then appends
+    only ``incoming`` lines whose ``event_id`` (or, when absent, raw text) is not
+    already present. Net effect: the canonical log can never be clobbered, yet a
+    genuinely new transition still lands. Appended events are the chronologically
+    newest (the just-emitted transition) or non-lane events the reducer skips, so
+    ordering stays correct.
+    """
+
+    def _key(line: str) -> str:
+        try:
+            obj = json.loads(line)
+        except (json.JSONDecodeError, ValueError):
+            return f"raw:{line}"
+        if isinstance(obj, dict):
+            eid = obj.get("event_id")
+            if isinstance(eid, str) and eid:
+                return f"id:{eid}"
+        return f"raw:{line}"
+
+    existing_lines = [
+        line for line in existing.decode("utf-8", "replace").splitlines() if line.strip()
+    ]
+    seen = {_key(line) for line in existing_lines}
+    merged = list(existing_lines)
+    for line in incoming.decode("utf-8", "replace").splitlines():
+        if not line.strip():
+            continue
+        key = _key(line)
+        if key not in seen:
+            seen.add(key)
+            merged.append(line)
+    if not merged:
+        return b""
+    return ("\n".join(merged) + "\n").encode("utf-8")
+
+
 def _commit_via_coordination_transaction(
     *,
     coord_branch: str,
@@ -280,7 +329,16 @@ def _commit_via_coordination_transaction(
                 if txn_path.resolve() == path.resolve():
                     txn.stage_path(path)
                 else:
-                    txn.write_artifact(txn_path, path.read_bytes())
+                    incoming = path.read_bytes()
+                    # #1602: the canonical event log is append-only. Never let a
+                    # main-checkout copy overwrite (clobber) the coordination
+                    # branch's lane history — union-merge instead so existing
+                    # coord events always survive.
+                    if path.name == _STATUS_EVENTS_FILENAME and txn_path.exists():
+                        incoming = _merge_event_log_bytes(
+                            txn_path.read_bytes(), incoming
+                        )
+                    txn.write_artifact(txn_path, incoming)
             receipt = txn.commit(message)
     except BookkeepingPolicyRefused as policy_exc:
         _record_receipt(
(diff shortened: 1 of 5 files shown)


robertDouglass:
Verdict: SHOULD NOT MERGE

Findings

- P1: Non-additive planning-artifact changes now fail open.
  Evidence: `src/specify_cli/cli/commands/implement.py:241` strips rename records to only the new path; `implement.py:345` silently drops deleted paths; `implement.py:377` then returns before `BookkeepingTransaction`.
  Repro deletion on PR:
  ```text
  status-before: D kitty-specs/demo-feature-01J6XW9K/tasks.md
  returned: ok
  status-after: D kitty-specs/demo-feature-01J6XW9K/tasks.md
  coord-has-tasks: True
  ```
  Same repro on `origin/main`: raised `Exit 1`.

  Repro rename on PR:
  ```text
  parsed: ['kitty-specs/demo-feature-01J6XW9K/tasks-renamed.md']
  returned: ok
  coord old exists: True
  coord new exists: True
  ```
  Same repro on `origin/main`: raised `Exit 1`.

  Violated invariant: claim preflight must not proceed while planning artifacts are uncommitted or incoherent on the coordination branch.
  Missing test: deletion and rename planning-artifact status cases. Either commit them transactionally, including old-path deletion, or fail closed.

CI

- Current PR checks: no failures.
- Relevant passes: `fast-tests-cli`, `integration-tests-cli`, `Windows critical`, `lint`, `diff-coverage`, `quality-gate`.
- Relevant skipped: lanes/status/sync/merge fast + integration suites, `slow-tests`, `e2e-cross-cutting`, `mutation-testing`, `SaaS consumer compatibility`.

Local verification: `.venv/bin/pytest tests/specify_cli/cli/commands/test_implement.py -q` passed, 7 tests.

Required before merge

- Parse porcelain into typed status records, not just paths.
- For `D` and `R/C`, either support deletion/rename in `BookkeepingTransaction` or refuse before claim.
- Add regression tests proving dirty deletion and rename cannot advance with stale coordination artifacts.

stijn-dejongh:
Thanks @robertDouglass — the P1 is valid and now fixed in 2a7257b92.

**Root of the fail-open:** the porcelain parser kept only a rename's *new* path, the write loop / idempotency filter `continue`d on a non-existent (deleted) source, and the early `return` then let the claim proceed while the stale deleted/renamed-from artifact stayed on the coordination branch. `BookkeepingTransaction.write_artifact` is write-only and cannot remove an old path.

**Fix — fail closed on structural changes (restores main's contract):**
- `_feature_dir_status_entries()` parses porcelain into typed records `(xy, path, is_structural)`; `is_structural` is true for deletions (`D` in either column) and renames/copies (`old -> new`).
- `_ensure_planning_artifacts_committed_git()` now refuses up front with `Exit 1` and an actionable message when any feature-dir change is structural, naming the offending entries and telling the operator to commit the `git rm`/`git mv` to the coordination branch and re-run. The write-only transaction never sees a deletion/rename.
- The no-op-identical idempotency skip is retained for the **modify/add** case only (the original bug this PR targets).
- `_feature_dir_status_paths()` now returns *writable* (non-structural) paths only.

**Regression tests added** (`TestStructuralPlanningArtifactsFailClosed`):
- `test_deleted_planning_artifact_fails_closed` — unstaged `D` → `typer.Exit`, HEAD unchanged.
- `test_renamed_planning_artifact_fails_closed` — staged `R old -> new` → `typer.Exit`, HEAD unchanged.

Full `test_implement.py` = 9 passed; ruff + mypy clean on the touched file.

On the two 'Required before merge' items I implemented the *refuse-before-claim* option rather than teaching the transaction to delete/rename, since the latter is a larger change to the transaction's staging/rollback surface and these structural cases are rare on the claim path — happy to follow up with full transactional deletion support in a separate PR if you'd prefer that direction.

stijn-dejongh:
Pulled in **#1602** (the lane-loop status desync) as a related fix on this branch — commit `9dae7dd7e`.

**Root cause (reproduced through the real code path).** Lane transitions live only on the coordination branch (the emit pipeline appends them to the coord worktree). The workflow commit `_commit_via_coordination_transaction` (reached from `agent action review` / `move-task`) re-committed the **main-checkout** copy of `status.events.jsonl` onto the coordination branch via `write_artifact`. That copy carries only the bootstrap + lifecycle/decision **envelope** events (which the reducer skips) and **none** of the lane transitions — so the overwrite wiped the lane history, `read_events()` returned 0, and the loop wedged with 'WP has no canonical status'.

**Fix.** Enforce the append-only invariant at the commit site: when committing `status.events.jsonl` to the coordination worktree, union-merge with the coord copy (dedup by `event_id`, raw-line fallback for non-canonical lines, preserve coord order + append genuinely-new events). Coordination events can never be dropped; a just-emitted transition still lands.

**Reproduction proof** (`tests/regression/test_issue_1602.py`): seeds `planned→claimed→in_progress` on the coord branch via the **real** `BookkeepingTransaction`, writes an envelope-only main copy, drives the actual `_commit_via_coordination_transaction`, and asserts the coord lane history survives.
- **RED without the fix**: `reduce(read_events(coord)) → KeyError 'WP01'` — the exact 'no canonical status' wedge.
- **GREEN with the fix**: WP01 stays `in_progress`, envelope event carried.
Plus unit coverage of `_merge_event_log_bytes` (clobber / no-op subset / new-transition-carried).

Verified: full implement→review integration flow (15) green; coordination/status/agent/regression sweep = 2301 passed (4 failures are pre-existing on the parent commit — environment-sensitive context-detection/kanban tests — confirmed unrelated). ruff + mypy clean.

Note: this is the commit-site guard. It composes with the read-resolution fix in #1597 (facet 3); together they close the desync from both the read and write sides.

Issue #1693, opened 2026-06-05
design(sync): no_team drain-block should never occur for authenticated workspace users — audit across spec-kitty, spec-kitty-events, spec-kitty-saas


robertDouglass:
## Summary

`_classify_drain_blocked_reason()` in `src/specify_cli/sync/emitter.py` can return `\"no_team\"` for an authenticated user, causing queued events to be held indefinitely in the local `OfflineQueue`. On workspace, every authenticated user has a private workspace that is their default team. Therefore **`no_team` after a successful rehydrate should be treated as a bug, not a legitimate persistent drain-blocked state.**

This surfaced during investigation of issue #1688, where routing `ProfileInvocationStarted`/`ProfileInvocationCompleted` through `OfflineQueue` (Option A) was identified as the correct fix for Failure 2. The `no_team` drain-block creates a new hazard: governance proof events could be queue-held indefinitely for a class of users who should never hit this condition.

---

## Resolution chain today

```python
# emitter.py:1688
team_slug = self._get_team_slug() if is_saas_sync_enabled() else None

# emitter.py:665-685 → _team.py:24-80
# _current_team_slug() calls resolve_private_team_id_for_ingress():
#   1. require_private_team_id(session)         — checks session.teams for is_private_teamspace=True
#   2. if None → rehydrate_membership_if_needed() + retry
#   3. if still None → log warning, return None  ← treated as normal "pending" state

# emitter.py:1642-1643
if team_slug is None:
    return "no_team"   ← event stamped drain_blocked_reason="no_team", held in queue indefinitely
```

```python
# session.py:81-97
def require_private_team_id(session: StoredSession) -> str | None:
    # NEVER returns session.default_team_id as fallback
    # NEVER returns session.teams[0].id as fallback
    # Returns None if no team has is_private_teamspace=True
    return get_private_team_id(session.teams)
```

The rehydrate path (`rehydrate_membership_if_needed()`) is a single-flight call to `/api/v1/me`. If the server returns a membership list where no team has `is_private_teamspace=True`, `no_team` fires and **there is no further recovery**.

---

## The invariant that should hold

> **Every authenticated workspace user has a private workspace.** It is provisioned at account creation. `is_private_teamspace=True` should always be present on exactly one team in the `/api/v1/me` response for any authenticated user.

If this invariant holds server-side, then `no_team` after a successful rehydrate is a bug — either:
[...]

Issue #1688, opened 2026-06-04
arch(ops): introduce Op as a first-class artifact type — replace ephemeral invocation records with durable, git-tracked, queue-backed Op records


robertDouglass:
## Supersedes

Original framing (durability fix for two invocation failures) is correct but undershoots. This issue now carries the full architectural proposal. The two failure modes are symptoms of a missing concept, not isolated bugs.

---

## Root Cause: The Missing Concept

`spec-kitty ask`, `spec-kitty advise`, and `spec-kitty do` have no proper home. They were built as governance machinery and their records were stored as local scratch files in a gitignored directory. The framing was wrong from the start.

The true nature of these commands: **a bounded, doctrine-governed agent action dispatched immediately, without mission ceremony, that still deserves a permanent trace.** The governance records are a byproduct. The thing itself is a unit of work — call it an **Op**.

An Op is to `ask`/`advise`/`do` what a Mission is to the full spec-kitty planning loop. Lighter, immediate, no ceremony — but real work, real doctrine, real record.

This reframing resolves both original failures structurally, not as patches.

---

### Failure 1 — Local file is ephemeral

`EVENTS_DIR = ".kittify/events/profile-invocations"` (writer.py:16) is covered by `.gitignore:65` (bare `.kittify/events/` ignore, no re-include). No auto-commit wiring exists in `executor.complete_invocation()`. Every record is git-invisible and lost on `git clean` or re-clone.

### Failure 2 — SaaS propagation bypasses the queue

`InvocationSaaSPropagator` uses `ThreadPoolExecutor` + direct WebSocket. When offline or unauthenticated, the record exists only in the gitignored local file. `do_cmd.py` has no propagator wired at all — zero SaaS propagation, not even best-effort.

### Additional finding (Paula Patterns, independent scout)

`do_cmd.py:_build_executor` omits `InvocationSaaSPropagator` entirely. The issue title mentions `advise/ask`; `do` has a deeper gap that is not addressed unless the fix applies at the executor level, not the call-site level.

---

### Concept: Op

An **Op** is a bounded, doctrine-governed agent action with a permanent record. Properties:

| Property | Value |
|---|---|
| Identity | ULID (`op_id`) |
| Profile | The agent profile under which the Op runs |
| Governance context | Charter + doctrine loaded at dispatch time (hash recorded) |
| Lifecycle | `started` → `completed` (append-only JSONL) |
[...]

PR #1714, opened 2026-06-05
feat(ops): Op record git durability — storage, auto-commit, doctor ops (Step 1 of 3)

robertDouglass:
## Summary

Implements issue #1688 Step 1: moves Op record storage from the gitignored `.kittify/events/profile-invocations/` directory to the git-tracked `kitty-ops/` directory, wires an auto-commit after every successful Op completion, and adds orphan detection via `spec-kitty doctor ops`.

- **WP01**: Redirect 5 constants across 4 files (`writer.py`, `lifecycle.py`, `propagator.py`, `record.py`) to `kitty-ops/`. Add `mission_id`/`wp_id` optional fields to `InvocationRecord`.
- **WP02**: Wire `_commit_op_record()` in `complete_invocation()` (direct subprocess git, `--no-verify`, best-effort). New `src/specify_cli/doctor/ops.py` with `list_orphan_ops()`. New `spec-kitty doctor ops` CLI subcommand. 27 new tests.

## Changes

- `src/specify_cli/invocation/writer.py` — `EVENTS_DIR` → `kitty-ops`, `INDEX_PATH` → `kitty-ops/ops-index.jsonl`, fix `_append_to_index` inline path
- `src/specify_cli/invocation/lifecycle.py` — `LIFECYCLE_LOG_RELATIVE_PATH` → `kitty-ops/lifecycle.jsonl`
- `src/specify_cli/invocation/propagator.py` — `PROPAGATION_ERRORS_PATH` → `kitty-ops/propagation-errors.jsonl`
- `src/specify_cli/invocation/record.py` — add `mission_id: str | None`, `wp_id: str | None`; update `MVTP.tier_1.storage_path`
- `src/specify_cli/invocation/executor.py` — add `_commit_op_record()`, wire in `complete_invocation()`
- `src/specify_cli/doctor/__init__.py` — new package
- `src/specify_cli/doctor/ops.py` — new: `list_orphan_ops()`, `_has_completed_event()`
- `src/specify_cli/cli/commands/doctor.py` — add `@app.command(name="ops")` subcommand
- `CHANGELOG.md` — document `.kittify/events/` abandonment (data loss notice)
- Tests: `test_writer.py`, `test_record.py`, `test_executor.py`, `test_doctor_ops.py` (new)

## Test plan

- [ ] `uv run pytest tests/specify_cli/invocation/ -q` — all pass
- [ ] `spec-kitty doctor ops` lists orphan ops (started-only files in `kitty-ops/`)
[...]

PR #1720, opened 2026-06-05
[codex] arch(ops): persist completed op records

LynnColeArt:
## Summary

Implements the self-contained Step 1 slice from #1688 for Op durability:

- routes completed Op record commits through `safe_commit` instead of executor-local git plumbing
- adds an explicit `safe_commit` policy path for completed Op records on protected branches, limited to one `kitty-ops/<op-id>.jsonl` path and still covered by staging preservation/backstop checks
- preserves `mission_id` and `wp_id` correlation fields on completed records
- keeps started-only Ops as uncommitted orphans while committing closed Ops under `kitty-ops/`
- keeps `kitty-ops/ops-index.jsonl` local-only as a performance cache
- adds focused writer/executor/git-helper regression coverage for correlation, git restore, protected-branch Op commits, local-only index behavior, and orphan behavior

## Notes

This intentionally does not implement Step 2 OfflineQueue/SaaS propagation, because that depends on #1693 and SaaS `OpStarted` / `OpCompleted` handlers.

Detached HEAD behavior: Op auto-commit remains best-effort and is skipped when the executor cannot resolve a current branch for `safe_commit.destination_ref`. The completed Op JSONL is still written locally; a branch is required for the durable git commit path.

## Validation

- `uv run --extra test python -m pytest tests/specify_cli/git/test_commit_helpers.py tests/specify_cli/invocation/test_executor.py tests/specify_cli/invocation/test_writer.py`
- `uv run --extra test python -m pytest tests/integration/git/test_safe_commit_backstop.py tests/specify_cli/invocation/test_invocation_e2e.py`
- `uv run --extra lint python -m ruff check src/specify_cli/git/commit_helpers.py src/specify_cli/invocation/executor.py tests/specify_cli/git/test_commit_helpers.py tests/specify_cli/invocation/test_executor.py tests/specify_cli/invocation/test_writer.py`
- `git diff --check`

Local note: `tests/specify_cli/invocation/cli/test_do.py` is obscured in this checkout by a pre-existing connected-Teamspace logged-out stderr diagnostic mixed into `CliRunner.output`; the command's `stdout` remains valid JSON.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,9 @@ pyproject.local.toml
 .kittify/.migration-backup/
 .kittify/encoding-provenance/
 .kittify/migrations/mission-state/
+
+# Local performance cache; durable Op records live in kitty-ops/<op-id>.jsonl.
+kitty-ops/ops-index.jsonl
 kitty-specs/**/.kittify/config.yaml
 kitty-specs/**/.kittify/dossiers/
 
--- a/src/specify_cli/git/commit_helpers.py
+++ b/src/specify_cli/git/commit_helpers.py
@@ -51,8 +51,9 @@ that legitimately mutate the protected branch:
 
 Completed Op records may also land on protected branches, but only when an
 internal caller explicitly passes ``allow_completed_op_on_protected_branch=True``
-and the changeset contains exactly one ``kitty-ops/<op-id>.jsonl`` file. This
-is not a commit-message prefix exception and is not exposed through the public
+and the changeset contains exactly one ``kitty-ops/<op-id>.jsonl`` file whose
+JSONL content includes a completed event for that same Op id. This is not a
+commit-message prefix exception and is not exposed through the public
 ``safe-commit`` command. The path still flows through the staging preservation
 and backstop checks below.
 
@@ -67,6 +68,7 @@ from __future__ import annotations
 
 from specify_cli.core.constants import KITTY_SPECS_DIR
 import contextlib
+import json
 import logging
 import os
 import subprocess
@@ -482,7 +484,10 @@ def _test_mode_allows_protected_branch() -> bool:
     )
 
 
-def _is_completed_op_record_exception(normalized_files: Sequence[str]) -> bool:
+def _is_completed_op_record_exception(
+    worktree_root: Path,
+    normalized_files: Sequence[str],
+) -> bool:
     if len(normalized_files) != 1:
         return False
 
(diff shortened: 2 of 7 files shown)


robertDouglass:
Review feedback: #1720 is cleanly based on merged #1714 and has no textual merge conflict, but it should not merge as-is until the protected-branch behavior is resolved.

Concrete issue:
- `ProfileInvocationExecutor._commit_op_record()` now uses `safe_commit()` for the normal path, which is the right boundary.
- But when `safe_commit()` raises `ProtectedBranchRefused`, the executor falls back to raw `git add --force` + `git commit --only` in `_commit_op_record_on_protected_branch()`.
- That preserves #1714 behavior, but it bypasses the protected-branch policy that `safe_commit` is intended to own. This leaves Op durability as an executor-local git exception instead of an explicit git-layer policy.

Required remedy before undrafting/merging: choose one of these and make it explicit.

1. Preferred architecture fix: move the Op-record protected-branch exception into the git abstraction.
   - Add a named, documented `safe_commit` policy path for completed Op records, or an explicit option whose semantics are limited to `kitty-ops/<op-id>.jsonl` commits.
   - Keep staging preservation/backstop checks in `safe_commit`.
   - Add/adjust tests proving completed Op records may commit under that policy while unrelated staged files remain untouched.

2. Simpler conservative fix: remove the protected-branch fallback.
   - If `safe_commit` refuses a protected branch, let the best-effort Op commit skip and log the warning.
   - Add a regression test that protected-branch refusal does not raise and does not bypass `safe_commit`.

Non-blocking but should be called out in the PR: #1720 changes detached-HEAD behavior. `_current_branch()` returns `None`, so Op auto-commit is skipped on detached HEAD; #1714’s raw git path would have attempted the commit. If intentional, document it in the PR body/test name.

The `ops-index.jsonl` change looks correct: keeping durable Op records tracked while ignoring the local performance index is consistent with the writer/doctor behavior.

LynnColeArt:
@robertDouglass thanks for the careful review. I addressed this in `05000c094`.

What changed:
- Removed the executor-local raw git fallback entirely; `_commit_op_record()` now only calls `safe_commit()`.
- Moved the protected-branch Op exception into the git layer as `allow_completed_op_on_protected_branch=True`.
- Kept that policy narrow: exactly one `kitty-ops/<op-id>.jsonl` path, `op(...)` commit message, no index/lifecycle/extra paths.
- Kept staging preservation and the sparse-checkout/backstop checks inside `safe_commit`.
- Added git-helper regression coverage proving completed Op records can commit on protected branches while unrelated staged files remain staged, plus rejection tests for missing policy, extra paths, bad paths, and wrong messages.
- Updated the PR body to call out the detached-HEAD behavior explicitly.

Validation after the fix:
- `uv run --extra test python -m pytest tests/specify_cli/git/test_commit_helpers.py tests/specify_cli/invocation/test_executor.py tests/specify_cli/invocation/test_writer.py`
- `uv run --extra test python -m pytest tests/integration/git/test_safe_commit_backstop.py tests/specify_cli/invocation/test_invocation_e2e.py`
- `uv run --extra lint python -m ruff check src/specify_cli/git/commit_helpers.py src/specify_cli/invocation/executor.py tests/specify_cli/git/test_commit_helpers.py tests/specify_cli/invocation/test_executor.py tests/specify_cli/invocation/test_writer.py`
- `git diff --check`

robertDouglass:
@LynnColeArt thank you for the careful follow-through on this. Moving the Op commit exception into `safe_commit`, keeping the index local, preserving correlation fields, and covering the protected-branch/staging behavior made this much stronger than the original shape. I added one final hardening commit so the protected-branch Op exception verifies the JSONL actually contains a completed event for the matching Op id, then re-ran the focused tests and waited for CI to go green before merging.

Appreciate the clean iteration here.

PR #1786, opened 2026-06-08
feat(router): tool-gated profile selection safeguard

### LINKED ISSUE #1783: spec-kitty do/ask: print commit hint after writing op record
## Problem

`spec-kitty do` and `spec-kitty ask` write a durable `kitty-ops/<op-id>.jsonl` op record but print no hint that the file needs to be committed. The `.gitignore` explicitly calls these files durable (only `ops-index.jsonl` is excluded), but agents and users only notice the untracked file if they happen to check `git status`. Sessions that end without a status check silently lose the audit trail.

By contrast, `spec-kitty profile-invocation complete` auto-commits its record — the asymmetry shows the gap is real.

## Proposed fix

After `spec-kitty do` / `spec-kitty ask` writes the op record, print a one-liner to stdout:

```
Op record written — commit it: git add kitty-ops/<op-id>.jsonl
```

This closes the loop inline, at the moment the file is created, without requiring hook infrastructure or orientation-block changes.

## Scope

- `src/specify_cli/cli/commands/do_cmd.py` — `do` command
- `src/specify_cli/cli/commands/ask_cmd.py` (or equivalent) — `ask` command
- Unit tests for the hint output

## Related

Identified during rc40 upgrade session. Architectural audit by `architect-alphonso` (invocation `01KTKBWR6WGSYJG2YA129HXNQ5`).
Sibling issue: #1782 (doctor mission-state --fix SNAPSHOT_DRIFT bug)

robertDouglass:
## Summary

- Adds `available-tools: [...]` field to `AgentProfile` (opt-in) so profiles can declare what external tools they're equipped to use
- `ActionRouter.route()` gains a `required_tools: frozenset[str]` parameter — when set, only profiles whose `available_tools` covers all required tools are eligible
- Profiles that omit `available_tools` are excluded from tool-gated routing (conservative opt-in)
- If no profile covers the required tools, router returns `profile_id=None` with `confidence="no_tool_match"` — executor raises `ROUTER_NO_TOOL_MATCH`, signalling the caller to proceed with full tool access
- `TOOL_KEYWORD_MAP` in the router maps request tokens (`github`→`gh`, `docker`, `aws`, etc.) so the executor derives `required_tools` automatically from request text
- 7 new unit tests covering: no_tool_match fallback, matching profile selected, empty/None required_tools are backwards compatible, keyword map correctness, match_reason content

## Motivation

Before this change, `spec-kitty do "get the GitHub security warnings"` routed to a profile whose charter only declared `git, mypy, pytest, ruff, spec-kitty` — no `gh`. The agent then either broke the governance loop to fetch data directly, or failed silently. This safeguard makes the mismatch explicit: no governed profile → no forced governance → full tool access with a clear signal.

## To use

Add `available-tools: [git, gh, mypy, pytest, ruff, spec-kitty]` to any profile YAML that should handle GitHub-related requests.

## Test plan

- [ ] `pytest tests/specify_cli/invocation/test_router.py::TestToolGatedRouting` — 7 passed
- [ ] `pytest tests/specify_cli/invocation/` — full suite green
- [ ] `ruff check` — clean
- [ ] `mypy` — no new errors (one pre-existing `no-any-return` in `_derive_action_from_tokens`)

Closes #1783 (follow-on from the op-record commit hint work)


robertDouglass:
Unhappy with the direction here.

Issue #1782, opened 2026-06-08
doctor mission-state --fix: RetrospectiveCaptured events cause SNAPSHOT_DRIFT to persist after repair


robertDouglass:
## Bug

`spec-kitty doctor mission-state --fix` silently fails on missions whose `status.events.jsonl` contains non-status-transition events (e.g. `RetrospectiveCaptured`) that appear beyond the first quarantine batch. The validator runs **before** the quarantine pass for those rows, so it raises `missing required to_lane` and exits with an error instead of quarantining the offending row. The mission is left in `SNAPSHOT_DRIFT` permanently.

## Reproduction

Mission: `agent-harness-install-audit-follow-through-01KT5YTQ`  
`kitty-specs/agent-harness-install-audit-follow-through-01KT5YTQ/status.events.jsonl`  
Line 32: a `RetrospectiveCaptured` event (has no `to_lane` field)

```
spec-kitty doctor mission-state --fix \
  --mission agent-harness-install-audit-follow-through-01KT5YTQ --json
```

**Output (exit code 1):**
```json
{
  "status": "error",
  "validation_errors": [
    "kitty-specs/agent-harness-install-audit-follow-through-01KT5YTQ/status.events.jsonl:32: missing required to_lane"
  ],
  "quarantined_rows": 10,
  "missions_updated": 0
}
```

After this, `doctor mission-state --teamspace-dry-run` still fails with `SNAPSHOT_DRIFT` for this mission.

The first `--fix` run (repo-wide) reported `updated=1, unchanged=162, errors=7` — this mission was one of the 7 errors.

## Root Cause

The repair logic validates event rows before deciding whether to quarantine them. Non-status-transition event types (`RetrospectiveCaptured`, and likely others with `type` fields) are structurally valid for their own schema but lack `to_lane`, which the lane-event validator treats as malformed. The fix should:

1. Identify and quarantine all non-status-transition event types **before** running lane-field validation, OR  
2. Make the validator tolerate events whose `type` field indicates they are not lane transitions

## Environment

- spec-kitty-cli: 3.2.0rc40
- spec-kitty-events: 6.0.0
- Discovered during: `spec-kitty upgrade` (rc39 → rc40) + `doctor mission-state --fix` sequence

LynnColeArt:
I pushed the fix for #1782 on branch codex/issue-1782-doctor-retrospective and opened PR #1852.

[...]

PR #1852, opened 2026-06-11
fix(doctor): quarantine retrospective lifecycle rows

LynnColeArt:
Refs #1782.

Summary:
- quarantine `event_type`/`event_name` side-log rows before lane validation (matching exactly what `_scan_raw_status_rows` flags), keeping the existing quarantine behavior intact
- **preserve** retrospective lifecycle rows (`type: RetrospectiveCaptured/CaptureFailed/Skipped`) in place rather than quarantining them — they are contracted data written by design per `contracts/retrospective-events.contract.md` and read back by `retrospect`, `retrospective.summary`, and the runtime bridge, so removing them on `doctor mission-state --fix` would strip retrospective provenance from healthy missions
- promote `_should_skip_status_event` to the public `is_non_lane_event` / `is_retrospective_lifecycle_event` API on `specify_cli.status.store` (no more private cross-package import)
- add repair and rule regressions covering both dispositions: `DecisionPointOpened` quarantined, `RetrospectiveCaptured` preserved untouched

Note: rebased onto main; the previously bundled CI gating commit was dropped as it landed separately via #1851, and the spec-draft commit was already on main.

Checks:
- `pytest tests/unit/migration/test_canonicalization_rules.py tests/migration/test_mission_state_repair.py tests/next/test_retrospective_terminus_wiring.py` (103 passed)
- `pytest tests/migration tests/unit/migration tests/specify_cli/status tests/status tests/retrospective` (5,550 passed)
- `ruff check` and `mypy` clean

### DIFF
--- a/src/specify_cli/migration/mission_state.py
+++ b/src/specify_cli/migration/mission_state.py
@@ -40,6 +40,7 @@ from specify_cli.migration.canonicalization import (
 )
 from specify_cli.status import ULID_PATTERN, Lane, StatusEvent
 from specify_cli.status import materialize_snapshot, materialize_to_json
+from specify_cli.status.store import _should_skip_status_event
 
 MIGRATION_SCHEMA_VERSION = "1.0.0"
 CANONICAL_ENVELOPE_SCHEMA_VERSION = "3.0.0"
@@ -1233,7 +1234,7 @@ def _rule_reject_non_status_event(
     row: _Row, _ctx: MigrationContext
 ) -> CanonicalStepResult[_Row]:
     """Rule 1: quarantine rows that carry event_type or event_name (not status events)."""
-    if "event_type" in row or "event_name" in row:
+    if "event_name" in row or _should_skip_status_event(row):
         return CanonicalStepResult(
             state=row,
             actions=("quarantined_non_status_event",),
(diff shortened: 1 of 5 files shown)


robertDouglass:
Thank you @LynnColeArt for tracking down #1782 and landing the fix at exactly the right spot in the canonicalization pipeline — rule 1 ahead of the `to_lane` validator is precisely where this discrimination belongs, and the regression coverage at both the rule and full-repair levels made the follow-up easy to build on. 🙏

A four-profile review panel (Debugger Debbie, Paula Patterns, Reviewer Renata, Architect Alphonso) surfaced one refinement, which I've pushed as a follow-up commit (3b2abe3): retrospective lifecycle rows are contracted data that `retrospect`, `retrospective.summary`, and the runtime bridge read back from `status.events.jsonl`, so instead of quarantining them (which removes them on rewrite), repair now **preserves them in place** — matching the runtime reader's skip-in-place semantics. `event_type`/`event_name` rows keep the existing quarantine behavior, which now mirrors the dry-run scan's flag set exactly. The commit also promotes the shared predicate to a public API (`is_non_lane_event` / `is_retrospective_lifecycle_event`) and restores the `DecisionPointOpened` row in the integration test alongside the retrospective row so both shapes stay covered end to end.

Merging once CI is green. Thanks again for the contribution!

PR #1853, opened 2026-06-11
test(intake): extract shared TTY-test fixture to eliminate repeated setup

### LINKED ISSUE #719: test(intake): extract shared TTY-test fixture to eliminate repeated setup
## Context

Surfaced during `/review` of PR #714.

The three new TTY tests in `tests/specify_cli/cli/commands/test_intake.py` (`test_auto_tty_valid_selection_ingests_correct_file`, `test_auto_tty_non_numeric_input_exits_1`, `test_auto_tty_out_of_range_number_exits_1`, `test_auto_tty_zero_input_exits_1`) each duplicate the same four-patch context manager block and `mock_sources` literal verbatim.

Similarly, `test_directory_at_candidate_path_expands_md_children` and `test_non_md_files_in_directory_are_excluded` in `tests/specify_cli/test_intake_sources.py` duplicate the same `mock_sources`, `plans_dir` mkdir, and `patch()` block.

## Suggested fix

Extract a pytest fixture or shared helper, e.g.:

```python
@contextmanager
def _auto_tty_context(tmp_path, mock_sources):
    with (
        patch("specify_cli.cli.commands.intake.Path.cwd", return_value=tmp_path),
        patch("specify_cli.cli.commands.intake._resolve_repo_root", return_value=tmp_path),
        patch("specify_cli.intake_sources.HARNESS_PLAN_SOURCES", mock_sources),
        patch("specify_cli.cli.commands.intake.sys") as mock_sys,
    ):
        mock_sys.stdin.isatty.return_value = True
        yield mock_sys
```

No correctness impact — pure readability improvement.

LynnColeArt:
Refactors the repeated intake test patch/setup blocks into shared helpers.

- adds a small shared helper for intake command repo-root and TTY setup
- reuses a shared HARNESS_PLAN_SOURCES patch helper in scan tests
- keeps the behavior unchanged; this is test-only cleanup

Verification:
- uv run --with pytest pytest tests/specify_cli/cli/commands/test_intake.py tests/specify_cli/test_intake_sources.py -q
- uv run --with ruff ruff check tests/specify_cli/intake_test_helpers.py tests/specify_cli/cli/commands/test_intake.py tests/specify_cli/test_intake_sources.py
- git diff --check

Closes #719

### DIFF
--- a/tests/specify_cli/cli/commands/test_intake.py
+++ b/tests/specify_cli/cli/commands/test_intake.py
@@ -11,6 +11,7 @@ from typer.testing import CliRunner
 
 from specify_cli.cli.commands.intake import intake
 from specify_cli.mission_brief import BRIEF_SOURCE_FILENAME, MISSION_BRIEF_FILENAME
+from tests.specify_cli.intake_test_helpers import patched_intake_command_environment
 
 pytestmark = [pytest.mark.fast, pytest.mark.non_sandbox]
 
@@ -47,11 +48,7 @@ def _make_plan_file(tmp_path: Path, rel: str = "plan.md", content: str = "# Plan
 def test_auto_no_matches_exits_1(intake_app: typer.Typer, tmp_path: Path) -> None:
     """--auto with no plan files found exits 1 and leaves .kittify/ untouched."""
     mock_sources: list = []
-    with (
-        patch("specify_cli.cli.commands.intake.Path.cwd", return_value=tmp_path),
-        patch("specify_cli.cli.commands.intake._resolve_repo_root", return_value=tmp_path),
-        patch("specify_cli.intake_sources.HARNESS_PLAN_SOURCES", mock_sources),
-    ):
+    with patched_intake_command_environment(tmp_path, mock_sources):
         result = runner.invoke(intake_app, ["--auto"], catch_exceptions=False)
 
     assert result.exit_code == 1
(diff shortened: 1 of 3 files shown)


robertDouglass:
Thanks @LynnColeArt — this is a genuinely useful cleanup. The extraction is faithful: we verified line-by-line that `patched_intake_command_environment` preserves the original patch targets, ordering, and teardown semantics (including the subtle module-level `sys` patch for the TTY tests), and all 24 affected tests pass.

A four-reviewer panel went over the PR; here's what we changed and why before merging:

**Why we rebased your branch:** it was based on the `v3.2.0rc41` tag, so it carried a duplicate of the `kitty-specs/pre-and-post-mission-lifecycle-support-01KTNK1G/` spec commit that had already landed on `main` (`270b95004`). The content was byte-identical, so it was harmless — but it made the diff look like 10 files / +362 lines when your actual change was 3 files / ~+68/−73, and CI was validating against rc41-era code. Rebasing onto current `main` dropped the duplicate commit and left the clean test-only diff your description promised. (Tip for next time: branch from current `origin/main` rather than a release tag.)

**Follow-up commit (`c1d0cc1a`) applying the panel's non-blocking suggestions:**
- `patch_cwd` knob — `test_manual_intake_no_source_agent` originally avoided the `Path.cwd` patch (which is process-global, since it patches the shared `pathlib.Path` class); the knob restores its narrower patch surface and documents the blast radius.
- Tri-state `tty: bool | None` — the non-TTY case now also lives in the helper (same mechanism as before), so the whole TTY contract is in one place.
- Moved the load-bearing "CliRunner replaces sys.stdin, so patch the module-level `sys`" rationale comment into the helper so a future simplification doesn't mysteriously break the four TTY tests.
- `patched_intake_command_environment` now composes `patched_harness_plan_sources` so the patch-target string is named once.
- Widened `mock_sources` to `Sequence` — the touched files now pass `mypy --strict` standalone.

Verified locally: 24/24 intake tests, ruff clean, mypy clean, full architectural suite green (340/340). Merging once CI passes. Thanks again for closing #719!

PR #1859, opened 2026-06-11
fix(dashboard): prefer coord worktrees for kanban scans

### LINKED ISSUE #1824: Bug: dashboard scanner reads main checkout instead of coord worktree during active coord-topology missions (stale kanban state)
## Summary

In spec-kitty 3.2.0rc41, the dashboard's `/api/kanban/<mission>` endpoint silently shows stale mission state for any mission running under the coordination-worktree topology. WPs that have advanced past `planned` on the coord worktree (to `for_review`, `approved`, etc.) are reported as `planned` by the dashboard, because the scanner reads main's `kitty-specs/<mission>/status.events.jsonl` (which contains only the bootstrap events) instead of the coord worktree's event log (which has the authoritative live state).

`spec-kitty agent tasks status` shows the correct state for the same mission at the same instant, because it uses a different (coord-aware) resolver via `MissionStatus.load()`. The two surfaces produce different answers for the same mission with no warning to the operator.

`.kittify/derived` is also stale during this window, but the kanban endpoint does not consult it for lane placement.

A dashboard restart does NOT fix it — every request recomputes via the same scanner, and there is no cache to invalidate.

### Prerequisites

- A spec-kitty 3.2.0rc41 project initialized via `spec-kitty init`.
- A mission running under the coordination-worktree topology (i.e., with a `.worktrees/<mission>-coord/` checkout).

### Steps

1. Create a coord-topology mission; let main's `kitty-specs/<mission>/status.events.jsonl` accumulate only the bootstrap events.
2. Advance a WP past `planned` on the coord worktree (e.g., via `spec-kitty agent action implement WP01`). Verify the new event lands in `<coord>/kitty-specs/<mission>/status.events.jsonl` and NOT in main's log.
3. With the dashboard running, open the mission view in a browser.
4. Run `spec-kitty agent tasks status --mission <slug>` in a terminal.

### Expected behavior

Both surfaces show the same WP lane distribution — preferably the coord-worktree state, since that is where authority lives during execution per the 3.2 coordination-topology model.

### Actual behavior

Dashboard kanban shows all WPs in `planned`. CLI shows the real distribution (advanced WPs reflected). Restarting the dashboard does not change the displayed state.

## Root cause

Two stated design intents in spec-kitty's source code contradict each other:

| Surface | Source comment | Resolver |
|---|---|---|
| `cli/commands/agent/status.py:142-149` (= `spec-kitty agent tasks status`) | "fail-closed coord authority... `MissionStatus.load()` to resolve the coord-aware read path so that coordination-topology missions read from the coord worktree" | coord-aware ✓ |
| `dashboard/scanner.py:302-329` (= `/api/kanban`) | "Main repo (kitty-specs/) paths take priority over worktree copies. Worktrees may have stale data from when they were created..." | main-priority ✗ for 3.2 coord-topology |

The dashboard scanner's main-priority design predates the 3.2 coordination-worktree mission model and was never updated to use `MissionStatus.load()` (the new coord-aware resolver). The original "worktrees may have stale data" assumption is correct for **lane** worktrees but inverted for the **coord** worktree, which is authoritative during execution.

## Workaround Applied

During any active coord-topology mission, operators must use `spec-kitty agent tasks status --mission <slug>` (or `--json`) for status instead of the dashboard. After mission merge, the coord worktree is removed and main has the full event log, so the dashboard becomes accurate again — the bug only affects the in-flight window.

## Environment

- OS: macOS (Darwin 25.5.0)
- Python: 3.13.13
- spec-kitty-cli: 3.2.0rc41 (installed via pipx)
- Reproduced in an isolated test repo with manufactured divergent main/coord state, against the same pipx-installed binary used in production.

---

**Authored by**: Kent Gale (kentonium3/kg-automation) & Claude Code (Claude Opus 4.7), 2026-06-10.
**Submission approved by**: Kent Gale (kentonium3/kg-automation), 2026-06-10.
**Local tracking**: kentonium3/kg-automation#577.

LynnColeArt:
Fix the dashboard scanner so active coordination-topology missions read the authoritative coord worktree before the main checkout.

This keeps /api/kanban/{feature_id} and the dashboard feature list from showing stale mission state when the coord worktree has advanced beyond the bootstrap log in kitty-specs/<mission>/.

Verification:
- uv run --with pytest pytest tests/test_dashboard/test_scanner.py -q
- uv run --with pytest pytest tests/test_dashboard/test_api_contract.py -q
- uv run --with ruff ruff check src/specify_cli/dashboard/scanner.py tests/test_dashboard/test_scanner.py
- git diff --check

Closes #1824


stijn-dejongh (approved):
Went through the code, I noticed there is some logic that depends entirely on the worktree/branch naming convention. It is accurate, but warrants a randy reducer scan for similar "enforcement by convention" blocks in the codebase, as these are likely to cause more of the split-brain problems we have been battling recently.

Non-blocking for this PR, issue seems genuinely resolved ( priority for this release - debt accepted ).

I like the spec, @LynnColeArt , looking forwards to that pre/post mission run landing. ( Okay to include it, no need to strip it out of PR - genuine useful planning and documentation in our tracker ).

PR #1869, opened 2026-06-12
fix: suppress localhost-only Sonar HTTP hotspots

robertDouglass:
## Summary
- route dashboard and sync-daemon `serve_forever()` calls through localhost-only helpers
- add explicit Sonar suppression rationale for 127.0.0.1 control planes
- clean up adjacent `mypy` nits in `sync/daemon.py` so touched files stay green

## Why suppression is necessary
SonarCloud flagged two `pythonsecurity:S5332` hotspots on loopback-only control planes. These servers bind exclusively to `127.0.0.1` and are intentionally unreachable from remote peers, so TLS would add ceremony without reducing exposure. The suppression is scoped to the exact `serve_forever()` calls and documents that the transport never leaves localhost.

## Validation
- `.venv/bin/pytest tests/test_dashboard/test_server.py tests/sync/test_daemon_self_retirement.py`
- `.venv/bin/ruff check src/specify_cli/dashboard/server.py src/specify_cli/sync/daemon.py`
- `.venv/bin/mypy src/specify_cli/dashboard/server.py src/specify_cli/sync/daemon.py`

### DIFF
--- a/src/specify_cli/dashboard/server.py
+++ b/src/specify_cli/dashboard/server.py
@@ -58,6 +58,11 @@ def _build_handler_class(project_dir: Path, project_token: str | None) -> type[D
     )
 
 
+def _serve_loopback_http_forever(server: HTTPServer) -> None:
+    """Serve the localhost-only dashboard control plane."""
+    server.serve_forever()  # NOSONAR(pythonsecurity:S5332) -- binds to 127.0.0.1 only; no remote transport exposure
+
+
 def run_dashboard_server(project_dir: Path, port: int, project_token: str | None) -> None:
     """Run the dashboard server forever (used by detached child processes)."""
     try:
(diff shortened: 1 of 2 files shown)


robertDouglass:
Superseded by #1870, which replaces the per-module NOSONAR-wrapped helpers with a shared `core/loopback_http.py` module that enforces 127.0.0.1 binding by construction (no suppressions needed). This PR's two unrelated fixes — the stale psutil `type: ignore[import-untyped]` removal and `text=True` on `_spawn_sync_daemon_process`'s Popen — have been ported into #1870 (commit 43e13f0f2), so nothing here is lost. Disposition decided after dual-profile governed review (Reviewer Renata + Architect Alphonso) of both PRs.

Issue #1868, opened 2026-06-12
Epic: Canonical seams exist in name only — bind authority to type/owner for package layering, mission identity, guard capability, daemon identity, the CI suite map, and versioned contracts


robertDouglass:
## Summary

PR #1850 (branch `fixups/code-engine-stabilization`, remediation commits `8582e4df` → `aec70048`, CI 59/59 green) fixed every verified P0/P1 behavioral defect from its review: the `TEST_MODE` guard bypass, the daemon-reaper over-kill, the mid8 handle regression, `StatusSurfaceFragment` threading, and the CI blind spot for `src/runtime/next` / `src/mission_runtime`. **This epic does not block any release.**

What the review established — and what the remediation by design did not attempt — is that the defect *class* survives: the codebase repeatedly declares a canonical authority (a package layer, an identity resolver, a guard capability, a daemon owner record, a CI suite map, a schema version) and then enforces it with prose, hand-maintained lists, or nothing. Every P1 in the review was an instance of an authority existing in name while call sites kept the freedom to bypass it. This epic binds each of six seams to a type or an owner so the next instance is a test failure, not a production incident.

All evidence below was re-verified against the code at `aec70048` — not against the (partially stale) review prose. Each workstream is independently executable as its own mission; the checkboxes are its done-conditions.

## Already closed by PR #1850 + remediation (not re-litigated here)

| Closed | Evidence at `aec70048` |
|---|---|
| `TEST_MODE` capability bypass: all 3 production `TEST_MODE` asserts and all 6 borrowed `MERGE_BOOKKEEPING` asserts now assert `STANDARD`; only `cli/commands/merge.py:2363` (MERGE_BOOKKEEPING) and `upgrade.py:218` (UPGRADE_BOOKKEEPING) remain non-STANDARD | `cli/commands/agent/workflow.py:421`, `:1490`; `cli/commands/agent/mission.py:3164`; repo-wide grep |
| `SPEC_KITTY_TEST_MODE` precheck waivers deleted; only the documented operator hatch waives prechecks | `cli/commands/agent/tasks.py:810`; `cli/commands/implement.py:55` |
| AST capability ratchet (per-member allowlists + staleness assertion) and behavioral pinning added | `tests/architectural/test_guard_capability_call_sites.py`; `tests/git/test_guard_capability_regression.py` |
[...]

PR #2018, opened 2026-06-17
fix: re-harden merge bookkeeping rollback paths

robertDouglass:
## Summary
- re-assert merge rollback snapshot paths against trusted bookkeeping roots before any restore write/unlink
- thread `main_repo` through rollback callers so restore validation can prove the trust boundary to Sonar
- add regression coverage for both successful rollback restore and path-escape rejection

## Validation
- `.venv/bin/ruff check src/specify_cli/cli/commands/merge.py tests/merge/test_merge_done_recording.py`
- `.venv/bin/pytest tests/merge/test_merge_done_recording.py -q`
- `.venv/bin/pytest tests/specify_cli/cli/commands/test_merge.py -q`

## Sonar / security notes
- Addresses the open `pythonsecurity:S2083` / GitHub code-scanning alert on `src/specify_cli/cli/commands/merge.py` by making rollback restores fail closed outside trusted repo bookkeeping roots.
- Previous night's scheduled `CI Quality` run on `main` was `#2734` / `27602156830`; Sonar quality gate failed with this merge-path issue plus unreviewed loopback-only hotspots.
- Remaining Sonar hotspots in `src/specify_cli/core/loopback_http.py` are localhost-only `http://127.0.0.1` findings and still need Sonar UI review/rationale, not a code change in this PR.

### DIFF
--- a/src/specify_cli/cli/commands/merge.py
+++ b/src/specify_cli/cli/commands/merge.py
@@ -853,12 +853,18 @@ def _assert_bookkeeping_snapshot_path_is_trusted(
     """Reject rollback snapshot paths outside merge bookkeeping roots."""
     repo_resolved = get_main_repo_root(repo_root).resolve(strict=False)
     resolved_candidate = candidate.resolve(strict=False)
-    trusted_roots = (
+    trusted_dirs = (
         (repo_resolved / KITTY_SPECS_DIR).resolve(strict=False),
         (repo_resolved / WORKTREES_DIR).resolve(strict=False),
         (repo_resolved / KITTIFY_DIR / "runtime" / "merge").resolve(strict=False),
     )
-    if not any(resolved_candidate.is_relative_to(root) for root in trusted_roots):
+    trusted_files = (
+        (repo_resolved / KITTIFY_DIR / "merge-state.json").resolve(strict=False),
+    )
+    if not (
+        any(resolved_candidate.is_relative_to(root) for root in trusted_dirs)
+        or resolved_candidate in trusted_files
+    ):
         raise ValueError(
             f"Refusing bookkeeping snapshot path outside trusted repo roots: {candidate}"
         )
@@ -871,12 +877,16 @@ def _restore_final_bookkeeping_snapshots(
 ) -> None:
     """Best-effort restore for final merge bookkeeping rollback."""
     for path, original in snapshots.items():
-        with contextlib.suppress(OSError):
+        try:
             trusted_path = _assert_bookkeeping_snapshot_path_is_trusted(
                 repo_root=repo_root,
                 candidate=path,
             )
             _restore_optional_bytes(trusted_path, original)
+        except ValueError as exc:
+            logger.warning("Skipping untrusted bookkeeping rollback snapshot: %s", exc)
+        except OSError:
+            continue
 
 
 def _target_branch_still_at_baseline(
--- a/src/specify_cli/cli/commands/merge.py
+++ b/src/specify_cli/cli/commands/merge.py
@@ -858,9 +858,7 @@ def _assert_bookkeeping_snapshot_path_is_trusted(
         (repo_resolved / WORKTREES_DIR).resolve(strict=False),
         (repo_resolved / KITTIFY_DIR / "runtime" / "merge").resolve(strict=False),
     )
-    trusted_files = (
-        (repo_resolved / KITTIFY_DIR / "merge-state.json").resolve(strict=False),
-    )
+    trusted_files = (repo_resolved / KITTIFY_DIR / "merge-state.json",)
     if not (
         any(resolved_candidate.is_relative_to(root) for root in trusted_dirs)
         or resolved_candidate in trusted_files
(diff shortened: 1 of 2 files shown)


stijn-dejongh:
## Review — adversarial squad (correctness + bypass + cross-PR patterns)

**Verdict: one change requested before merge.** The intent (never write rollback bytes outside bookkeeping roots) is right, but raising during best-effort rollback introduces a behavioral regression that's worse-in-kind than the (currently unreachable) traversal it defends against.

### 🔴 Blocking: `ValueError` escapes `contextlib.suppress(OSError)` during rollback
`_assert_bookkeeping_snapshot_path_is_trusted` raises `ValueError`, but the restore loop body is `with contextlib.suppress(OSError):` — `ValueError` is **not** an `OSError`, so it propagates out. `_restore_final_bookkeeping_snapshots` is called from `except: _restore_final_bookkeeping_snapshots(main_repo, snapshots); raise` rollback blocks, so a tripped guard would:

1. **Mask the original merge exception** — the guard's `ValueError` propagates *before* the bare `raise`, so the operator sees "Refusing bookkeeping snapshot path outside trusted repo roots" instead of the real failure (e.g. `BaselineMergeCommitError`, the working-tree-invariant error). The diagnostic the operator needs is destroyed.
2. **Abort restoring the remaining snapshots** — the loop dies on the first offending key; keys iterated before it are restored, keys after are not → partial/corrupted bookkeeping, exactly the class this rollback exists to prevent.

Reproduced empirically on 3.11.15: with a trusted key ordered after an untrusted one, the original exception is replaced by the `ValueError` and the trailing trusted key is left un-restored.

**Why this is pure downside today:** every snapshot key in both dicts is already constructed from trusted roots (`resolve_status_surface` → `.worktrees/`/`kitty-specs/`; `get_state_path` → `.kittify/runtime/merge/<id>/…`; `_target_bookkeeping_status_paths` → `kitty-specs/`). So the guard has **no live attack surface** — it's defense-in-depth over an already-trusted producer, which means the masking/partial-rollback risk buys nothing in exchange.

**Suggested fix:** don't *raise* inside best-effort rollback — make it log-and-skip. Broaden to `contextlib.suppress(OSError, ValueError)` (or an explicit try/except that warns and `continue`s). That keeps the pre-PR guarantees — all keys attempted, original exception propagates — while still refusing the untrusted write.

### 🟡 Latent false-reject: trusted-root set misses the legacy state path
`get_state_path` has a legacy branch (`mission_id is None`) returning `.kittify/merge-state.json` — directly under `.kittify/`, **not** under the `.kittify/runtime/merge` trusted root. Not reachable from `_run_lane_based_merge_locked` today (mission_id always set), but any future caller snapshotting that path would trip a false-reject → the cascade above. Add `.kittify/merge-state.json` (or the bare `.kittify` boundary) to the trusted set, or assert mission_id-keyed.

### Test note
`test_final_bookkeeping_rollback_rejects_paths_outside_trusted_roots` asserts the guard raises `ValueError` — which actually **codifies** the regression rather than catching it. Recommend instead: a snapshots dict with a trusted key ordered *after* an untrusted key, called from inside an `except` re-raising an original exception; assert the **original** exception still propagates **and** the trusted key was still restored. The three trusted roots are each exercised by the happy-path test (good), though only indirectly.

### Mechanics (all correct)
All ~9 call sites thread the new `main_repo`/`repo_root` positional arg; `resolve(strict=False)` + `is_relative_to` semantics are right (symlinked trusted roots resolve consistently — intended logical containment).

<sub>Reviewed by reviewer-renata (correctness/tests), debugger-debbie (adversarial, live-reproduced the rollback-masking sequence), paula-patterns (cross-PR/seam consolidation). The `_assert_*_trusted` idiom now appears 3× in merge.py — a behavior-preserving consolidation into one parameterized `_assert_path_under_trusted_roots(roots=…)` is worth a follow-up ticket, separate from this fix.</sub>

stijn-dejongh:
Filed the consolidation follow-up referenced above: **#2022** — *consolidate merge.py path-trust into two canonical seams* (parent epic #1868).

It captures Goal B: collapse `_assert_status_path_within_target_surface`, `_assert_status_surface_path_is_trusted`, and this PR's `_assert_bookkeeping_snapshot_path_is_trusted` into one parameterized `_assert_path_under_trusted_roots(roots=…)`. Behavior-preserving, separate from the rollback-masking change requested in this PR.

robertDouglass:
Addressed Stijn's review in the reroll.

What changed:
- `_restore_final_bookkeeping_snapshots` now keeps rollback best-effort: untrusted snapshot paths are logged/skipped instead of escaping as `ValueError`, so the original merge exception is preserved and later trusted snapshots are still restored.
- Legacy `.kittify/merge-state.json` is trusted as an exact in-repo file target, without resolving the trusted target itself. That avoids accidentally blessing a symlink escape.
- Replaced the old raises-`ValueError` regression with coverage for original-exception preservation, trailing trusted snapshot restoration, untrusted-path non-write, and legacy state restore.

Local validation on head `73df9b774`:
- `.venv/bin/ruff check src/specify_cli/cli/commands/merge.py tests/merge/test_merge_done_recording.py`
- `.venv/bin/pytest tests/merge/test_merge_done_recording.py -q`
- `.venv/bin/pytest tests/specify_cli/cli/commands/test_merge.py -q`

Running final CI gate now before merge.

PR #2019, opened 2026-06-17
fix: guard merge bookkeeping mission slug paths

robertDouglass:
## Summary
- reject unsafe mission slug path segments before merge bookkeeping composes target checkout paths
- keep the final merge bookkeeping flow unchanged for valid slugs while failing closed on traversal, backslash, and non-ASCII inputs
- add regression coverage for the new guard and existing bookkeeping path slices

## Validation
- uv run pytest tests/specify_cli/cli/commands/test_merge.py -q
- uv run pytest tests/merge/test_merge_done_recording.py -q -k bookkeeping
- uv run ruff check src/specify_cli/cli/commands/merge.py tests/specify_cli/cli/commands/test_merge.py

## Context
- fixes the only open SonarCloud-backed code-scanning alert from the scheduled `CI Quality` run on `main` at 2026-06-16T07:42:11Z
- no suppressions added

### DIFF
--- a/src/specify_cli/cli/commands/merge.py
+++ b/src/specify_cli/cli/commands/merge.py
@@ -99,6 +99,7 @@ TARGET_BRANCH_NOT_SYNCHRONIZED = "TARGET_BRANCH_NOT_SYNCHRONIZED"
 TARGET_BRANCH_SYNC_INVARIANT = "local_target_branch_must_match_tracking_branch"
 _STATUS_EVENTS_FILENAME = "status.events.jsonl"
 _STATUS_FILENAME = "status.json"
+_MISSION_SLUG_PATH_SEGMENT_RE = re.compile(r"^[A-Za-z0-9_-]+$", re.ASCII)
 
 # T011 — FR-009: push-error parser tokens (locked tuple — do not reorder or extend without a spec change)
 LINEAR_HISTORY_REJECTION_TOKENS: tuple[str, ...] = (
@@ -770,6 +771,21 @@ def _paths_have_status_changes(repo_root: Path, paths: list[Path]) -> bool:
     return bool((out_status or "").strip())
 
 
+def _validate_mission_slug_path_segment(mission_slug: str) -> str:
+    """Reject mission slugs unsafe for direct path composition."""
+    candidate = mission_slug.strip()
+    if (
+        not candidate
+        or candidate in {".", ".."}
+        or "/" in candidate
+        or "\\" in candidate
+        or not candidate.isascii()
+        or _MISSION_SLUG_PATH_SEGMENT_RE.fullmatch(candidate) is None
+    ):
+        raise ValueError(f"Mission slug must be a single safe path segment: {mission_slug!r}")
+    return candidate
+
+
 def _target_bookkeeping_status_paths(
     *,
     main_repo: Path,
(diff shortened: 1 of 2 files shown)


stijn-dejongh:
## Review — adversarial squad (correctness + bypass + cross-PR patterns)

**Verdict: APPROVE-WITH-NITS.** The guard is correct and unbypassable at the seam it patches; the notes below are non-blocking, with one structural follow-up worth filing.

### What's solid
- `_validate_mission_slug_path_segment` correctly rejects every malformed-segment shape. Live-tested: `..`, `.`, `/`, `\`, whitespace, control chars, and non-ASCII (`naïve`, ligatures, zero-width) are all rejected.
- `.strip()`-vs-original is **not** a discrepancy here — the function returns the stripped `candidate` and that stripped value is what's passed to `primary_feature_dir_for_mission`; the original `mission_slug` isn't re-used after validation. Upstream `_resolve_mission_slug` already strips, so the values agree by the time they arrive.

### Non-blocking nits
- **Redundant predicate clauses.** `^[A-Za-z0-9_-]+$` with `re.ASCII` + `.fullmatch` already subsumes `not candidate`, `in {".",".."}`, `"/" in`, `"\\" in`, and `not candidate.isascii()` — those branches can never be the deciding check. Belt-and-suspenders; harmless, but Sonar may flag them as unreachable. Either trim to the regex or add a one-line "intentional fast-path / defense-in-depth" note.
- **`re.ASCII` is redundant** against an explicit `[A-Za-z0-9]` class (it only affects `\w`/`\d`/`\s` shorthands). Harmless but slightly misleading.
- **Test note:** the pre-existing `ensure_within_directory` already rejected `../../escape` (the old assertion), so that case was passing for the *other* guard. The genuine new value of this guard is rejecting `..` and `foo/bar` shapes that `ensure_within_directory` **accepts** (`kitty-specs/..` resolves back to repo root, which *is* `relative_to` the root; `kitty-specs/foo/bar` stays under root). A test asserting those two specific shapes are now rejected would pin the real win.

### Structural finding (recommend a follow-up, not a blocker for this PR)
The same operator-controlled `mission_slug` reaches **unguarded sibling composers** in `merge.py` — `primary_feature_dir_for_mission(main_repo, mission_slug)` at ~`:2332` and ~`:811`, plus the `f"kitty-specs/{mission_slug}/…"` interpolations at ~`:2696`. The canonical primitive `primary_feature_dir_for_mission` (`src/specify_cli/missions/_read_path_resolver.py` ~`:415`) composes `repo / KITTY_SPECS_DIR / mission_slug` with no validation. This PR closes one of several masks of the same class.

Suggested follow-up: move single-safe-segment validation into `primary_feature_dir_for_mission` (and its `candidate_*` sibling) so every caller inherits it, and have this merge-side helper delegate. That makes the resolver's "blessed owner of `KITTY_SPECS_DIR` assembly" docstring actually true. Out of scope for a Sonar point-fix — worth a ticket.

<sub>Reviewed by reviewer-renata (correctness/tests), debugger-debbie (adversarial bypass, live-reproduced), paula-patterns (cross-PR/seam consolidation). Verified against on-main code, not just the diff.</sub>

stijn-dejongh:
Filed the structural follow-up referenced above: **#2022** — *consolidate merge.py path-trust into two canonical seams* (parent epic #1868).

It captures Goal A specifically: move the single-safe-segment slug validation into `primary_feature_dir_for_mission` (`_read_path_resolver.py`) so the sibling composers (`merge.py` ~`:2332`/`:811`/`:2696`) inherit it, rather than guarding one call site. This PR's point-fix stands on its own; the consolidation is deliberately scoped out of the security fix.

PR #2036, opened 2026-06-19
fix: harden sonar security paths

robertDouglass:
## Summary
- validate and resolve merge bookkeeping rollback snapshot paths at capture-time so only trusted repo-derived paths reach restore I/O
- reduce generated Claude wrapper permissions from `0755` to `0700` to remove world/group execute bits
- extend regression coverage for trusted snapshot capture and wrapper permissions

## Investigation
- GitHub Dependabot alerts: none open as of 2026-06-19
- GitHub code-scanning alerts: 2 open SonarCloud-backed alerts on `main`
  - `src/specify_cli/cli/commands/merge.py` path-injection warning
  - `src/specify_cli/tool_surface/bundles/claude_wrapper.py` world-accessible chmod warning
- SonarCloud hotspots: 2 open entries in `src/specify_cli/core/loopback_http.py`
  - reviewed as loopback-only HTTP helpers binding strictly to `127.0.0.1`; no code change in this PR

## Validation
- `.venv/bin/ruff check src/specify_cli/cli/commands/merge.py src/specify_cli/tool_surface/bundles/claude_wrapper.py tests/specify_cli/cli/commands/test_merge.py tests/specify_cli/tool_surface/bundles/test_claude_wrapper.py tests/core/test_loopback_http.py`
- `.venv/bin/pytest tests/specify_cli/cli/commands/test_merge.py`
- `.venv/bin/pytest tests/specify_cli/cli/commands/test_merge.py -k 'bookkeeping_snapshot'`
- `.venv/bin/pytest tests/specify_cli/tool_surface/bundles/test_claude_wrapper.py tests/core/test_loopback_http.py`


stijn-dejongh:
### Follow-up pushed — squad-reviewed hardening of the path-injection class

A maintainer adversarial squad (implement → mutation-verify → independent re-verify) reviewed this PR for missed opportunities in the two finding-classes it hardens, then implemented the gaps. Pushed on top of the existing commits.

**Permissions class:** confirmed COMPLETE codebase-wide — `claude_wrapper.py` is the only generated-executable site; all secret/token writes are already `0o600`, the git hook `0o700`.

**Path-injection class — extended beyond `merge.py`.** The same untrusted `mission_slug` (read from `status.events.jsonl` event content) reached **filesystem sinks in the `status/` package unguarded** — including three *write* sinks (worse than the read sink), which `mkdir`+write under `.kittify/derived/`. A crafted event `{"mission_slug": "../../../../evil"}` + `spec-kitty status materialize` wrote **outside** `derived/`. Fixes:
- `status/store.py` — `_SlugResolver.resolve` now validates via the canonical `assert_safe_path_segment` before the `meta.json` join (fail-closed → `None`), matching the sibling `aggregate.py` resolver.
- `core/paths.py` — new `safe_mission_slug(slug, fallback)` helper.
- `status/reducer.py` — sanitizes `snapshot.mission_slug` at the **single reduce seam**, so the three write sinks (`progress`/`lifecycle`/`views`) fall back to the trusted `feature_dir.name`. One chokepoint; future sinks covered too.

**Tests:** added `../`-traversal + **symlink-escape** coverage for the bookkeeping seam (empirically confirmed the seam already rejects symlink escapes via `resolve()`-containment — secure, not a bypass); planted-attacker-file resolver tests; per-sink + reduce-seam fail-closed tests. **All security tests are mutation-verified un-fakeable** (each fails when its guard is removed, with the escaped dir/read witnessed). `ruff` + `mypy` clean; **727** status/merge tests pass.

**Also:** the two touched test files were made `mypy`-clean (boy-scout) in an earlier commit.

#### Notes for the reviewer / Sonar UI
- **Residual (intentionally not fixed):** `store.py:resolve` uses `.exists()`/`read_text()` (segment-guard only, no `resolve()`-containment), so a slug that is a *valid single segment* but names a **symlink dir planted under `kitty-specs/`** could still escape. This requires write access inside the repo's specs dir (a much higher bar than controlling event-log text) and matches the existing `aggregate.py` precedent — classified as a residual hardening gap, not a defect. Flagging for a future toolguide/hotspot decision rather than scope-creeping this PR.
- **Pre-existing `mypy` finding (not introduced here):** `tests/status/test_store.py` has a `status_store.os does not explicitly export attribute "os"` error from a #946-era `monkeypatch.setattr` idiom. It is **config-suppressed** under the standard gate (`follow_imports=skip` for `specify_cli.*`) and does not surface in CI; left as-is to avoid changing an unrelated atomic-write test's patch semantics.

stijn-dejongh:
### 📋 Stacking a governed mission on this branch — heads-up for co-maintainers

This PR has grown beyond its original two Sonar fixes, so we're putting it under a Spec Kitty mission for traceability. **No action needed from you** — this is informational.

**What happened:** an adversarial maintainer squad reviewed this PR for missed opportunities in the two finding-classes it hardens. It found that the `merge.py` path-injection fix and the `store.py` follow-on were not isolated bugs but **instances of one vulnerability class** — an untrusted path segment (`mission_slug`/`feature_slug`/`wp_id`, read from `status.events.jsonl`, `meta.json`, or frontmatter) reaching a filesystem sink without a containment check, including three *write* sinks that `mkdir`+write under `.kittify/derived/`. Those were closed here; but a class deserves a governed close-out, not point fixes.

**What we stacked:** mission **`untrusted-path-containment-hardening`** (`mission_id 01KVFTFVQCQS88M0S78M59APB0`). Its `spec.md` is committed on this branch (`kitty-specs/untrusted-path-containment-hardening-01KVFTFV/`), so the PR now tells the whole story: the landed hardening **is** the mission's first increment (FR-007).

**Why stack rather than open a fresh PR:** the forward work is a direct continuation of what's already on this branch and shares its threat model; keeping spec + landed code together preserves the audit trail. Where the remaining work lands (continue stacking here vs. a follow-on PR) is a decision we'll take at `/spec-kitty.plan`.

**Forward scope the mission governs (not in this diff yet):**
- `store.py` + `aggregate.py` `resolve()`-containment — closes the residual symlink-dir escape (valid-label slug naming a symlink dir under `kitty-specs/`).
- A codebase-wide audit of every untrusted-string→FS sink in `src/specify_cli`, routing reachable ones through the canonical seam, documenting the rest, plus a regression guard banning new ad-hoc joins.
- `core/loopback_http.py`: document the loopback-only (127.0.0.1) rationale and retain its regression tests — **explicitly not** forcing HTTPS (that stays out of scope); the two hotspots are recorded for Sonar UI review.

Per repo policy this targets local `main` via PR; nothing here pushes to `origin/main` directly.

Issue #2034, opened 2026-06-19
CI gates select tests by marker; no gate selects `unit` or `contract`, so committed-and-failing tests never run (12 failures invisible on main)


kentonium3:
## Summary

On a clean `main` checkout, the full local test suite reports **12 failures**, yet `main` CI is
green. The cause is structural, not flaky: every job in `ci-quality.yml` selects tests by **marker**,
and **no job selects `-m unit` or `-m contract`** (nor is there any unfiltered full-suite run). Tests
marked only `unit`, `non_sandbox`, or `contract` are therefore selected by **zero gates** and never
execute in CI — so committed, deterministically-failing tests stay invisible.

This report concerns the **9** failures that are deterministic, committed defects on `main` and are
never run by any gate. (Three further local failures are environment-specific to our macOS runner —
timing, tmp-path length, and a non-git fixture — and are excluded here.)

Likely belongs under EPIC #1931 (Test quality & suite hygiene); #454 (post-merge test-assertion
staleness) is closely related to the kwarg-drift items below.

## Reproduction

Target: `main` @ `6ec393c5d`.

```bash
git worktree add --detach /tmp/sk-clean main
cd /tmp/sk-clean
uv sync --frozen --all-extras
# serial run — the documented parallel command has nondeterministic collection on macOS
HOME=$(mktemp -d) .venv/bin/python -m pytest tests/ -q -p no:cacheprovider
# → 12 failed
```

Then observe that none of the following are selectable by any `ci-quality.yml` gate.

**Gate inventory (from `ci-quality.yml`):**
- Fast shards: `-m "fast and not windows_ci"`
- Integration-core-misc matrix (covers `tests/tasks`, `tests/contract`, `tests/docs`,
  `tests/specify_cli`, …): `-m "not windows_ci and (git_repo or integration or architectural)"`
- Slow gate: `-m "slow and not windows_ci"`
- Agent integration gate: `tests/agent -m "not windows_ci and (git_repo or integration)"`
- Timing gate: a single file, `-m timing`
- e2e gate: `tests/e2e` + `tests/cross_cutting` by path

No gate matches `-m unit` or `-m contract`, and there is no unfiltered `pytest tests/` run. The
nightly `schedule:` trigger collapses the **path** filters but not the **marker** filters;
`release-readiness.yml` runs no pytest; `ci-windows.yml` runs only `windows_ci`.

**The 9 hidden, deterministic failures (committed on `main`):**

[...]

PR #2047, opened 2026-06-20
chore(test): dedup pytest marker registry to single source (#2034 workstream 2)

kentonium3:
Addresses **workstream 2** of #2034 (the marker-registry duplication confirmed as *"point 2 seems good"*).

`pytest` reads only one config file, preferring `pytest.ini` over `[tool.pytest.ini_options]` when both exist. The `markers` list in `pyproject.toml` was therefore dead config that never took effect and had silently drifted ~10 markers out of sync with the live `pytest.ini` registry.

## Change

- Remove the dead `[tool.pytest.ini_options]` markers block; replace with a NOTE documenting why it must not return.
- `pytest.ini` remains the single source of truth (and is a strict superset of the removed markers, so runtime behavior is unchanged — pytest already ignored the pyproject block).
- Add `tests/architectural/test_marker_registry_single_source.py` (`architectural`-marked, runs in the existing architectural gate) to pin the invariant so the duplication cannot silently come back.

Independent of the workstream-1 gate-gap design still under discussion on #2034.

---
*Authored by Kent Gale (kentonium3) & Claude Code (Claude Opus 4.8).*

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -178,30 +178,15 @@ exclude = [
     "src/specify_cli/**/tests/**",
 ]
 
-[tool.pytest.ini_options]
-markers = [
-    "fast: Pure unit tests with no I/O or subprocess calls (CI fast gate)",
-    "integration: Integration tests using real filesystem or in-process I/O, no subprocess/git",
-    "slow: Tests taking >10 seconds or requiring heavy setup",
-    "git_repo: Tests that need a real git repository (integration gate)",
-    "adversarial: Adversarial/security tests for 0.13.0",
-    "doctrine: Doctrine package smoke and integration checks",
-    "distribution: Tests requiring wheel install (slow, no SPEC_KITTY_TEMPLATE_ROOT)",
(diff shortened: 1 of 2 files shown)


kentonium3:
**Heads-up on the red CI Quality check — it's not from this diff.** This PR changes only `pyproject.toml` (removing the dead `[tool.pytest.ini_options].markers` block) and adds one test. The 3 failures are in `tests/architectural/test_untrusted_path_containment.py` (×2) and `tests/architectural/test_no_raw_mission_spec_paths.py`, flagging path handling in `cli/commands/merge.py` (e.g. `merge.py:1103`, authored 2026-04-12) — untouched by this PR; the new `test_marker_registry_single_source.py` passes.

They appear to be pre-existing on `main`: `main`'s CI is green, but adding a file under `tests/architectural/` flips `core_misc=true` and runs the architectural shard, which normal `main` pushes skip — so this PR surfaced latent failures rather than introducing them. (Incidentally a concrete example of the marker/path-gated coverage gap in #2034.) Flagging here so the red isn't read as a problem with the dedup.

PR #2051, opened 2026-06-21
fix: coord-topology orchestration — start-implementation, append-history, lane allocation, merge target

rayjohnson:
## Summary

Four fixes (one PR) to the **coordination-topology** path that an external orchestrator drives via `orchestrator-api`, plus the user-facing `spec-kitty merge`. Found while taking a real coord-topology mission end-to-end: WPs reached `done` but nothing was committed or integrated, and the merge landed on the wrong branch. Each fix has a fails-before/passes-after regression test against a real git repo; validated end-to-end on a live mission.

The common thread: several resolvers read mission metadata via the **topology-aware** surface, which under coordination topology resolves to the coordination worktree — whose mission dir has no `meta.json`. They must read the **primary checkout** (where `coordination_branch` / `target_branch` live).

## What's fixed

1. **`start-implementation` crashed** with `TypeError: transactional status batch only supports one feature/mission/wp`. The transactional batch guard compared each request against the resolved *primary* anchor, which legitimately differs from the coord-worktree request surface. Now anchors on the first request's canonicalized dir, matching the non-transactional sibling `emit_status_transition_batch`.

2. **`append-history` hit `SAFE_COMMIT_PATH_POLICY`.** It committed the WP prompt file from the primary checkout while the file lives in the coordination worktree. Now commits from the worktree the file actually lives in, via the canonical `resolve_placement_only` target (mirrors `events/decision_log.py`).

3. **`start-implementation` never allocated a lane worktree** — it returned a bare legacy path, so `merge-mission` found no lane branches and dependent WPs never saw their dependencies' code. Now allocates the real lane worktree (lane branch on the coordination branch, dependency-lane tips merged) when the mission has a lanes manifest; legacy/non-lane missions keep the prior behavior. The response carries `lane_id`, `lane_branch`, `lane_base_ref`, and `workspace_path` now means that lane worktree. The `for_review` transition is gated on a real commit beyond the lane base (shared with the native `move-task` gate), so "done without a commit" is impossible via the API too.

4. **`spec-kitty merge` and `merge-mission` merged into `main`** when the mission's `target_branch` was set (e.g. `feat/x`) but read via the coord surface. Both now resolve the target through one shared authority — `core.paths.resolve_merge_target_branch` (`--target` > primary-meta `merge_target_branch` > primary-meta `target_branch` > repo default) — so they can never disagree. This also removes a downstream `SafeCommitHeadMismatch`.

## Contract

`orchestrator-api` `CONTRACT_VERSION` `1.0.0 → 1.1.0`: additive `start-implementation` response fields (`lane_id`, `lane_branch`, `lane_base_ref`) + a changed meaning for `workspace_path` (now the lane worktree). New error code `LANE_ALLOCATION_FAILED`. **Pairs with `spec-kitty-orchestrator` O1–O3**, which were built against this contract.

## Known / out of scope

- **Commit gate base:** the `for_review` gate counts commits beyond the lane's base ref (the coordination branch), byte-identical to the base the native `move-task` gate uses. For a lane with `depends_on_lanes`, the dependency-merge commits already sit beyond that base, so the gate can pass on dep-merges alone — this is a **pre-existing property of the native gate**, preserved here intentionally for parity, not a new gap.

## Testing

- New fails-before/passes-after regression tests for all four fixes against real coord-topology git repos, plus focused unit tests for the new helpers/branches.
- ruff + `mypy --strict` clean (zero new findings); terminology and primary-anchored-gate architectural guards pass.
- Validated end-to-end on a live coordination-topology mission (orchestrate → lanes committed → `merge-mission` integrates into the correct target).

## Sonar

New branches/helpers are covered by focused tests in this PR. No code-level hotspots expected; loopback/localhost semantics are untouched.


stijn-dejongh:
@rayjohnson Thank you for furthering the SSOT / contradictory resolution issue. 

The PR tackles the topology issues nicely, but leaves a few gaps open. Gicen the recent work that landed on main, I will fold this PR into a new branch, which will also contain some remediations, and further strangling of the topology split-path issues. 

There is no need to follow this up further, I will land it for you.

PR #2039, opened 2026-06-19
test(#2034): fix 9 un-gated test failures (symptoms, part 1 of 2)

kentonium3:
## What & why

Part 1 of 2 for #2034. This PR fixes the **9 broken tests** that are currently
slipping through CI un-gated, leaving the repo's test suite green by construction.

Per @StijndeJongh's triage, #2034 has two separable workstreams:

1. **This PR — fix the symptoms** (mechanical, high-confidence). These tests run
   in *zero* CI gates today, so they could drift from source unnoticed; this PR
   re-syncs them.
2. **Follow-up PR — close the gate gap (root cause):** add an unfiltered
   `pytest tests/` gate so no test can leak by construction, and dedupe the
   marker registry. It lands *after* this PR, because an unfiltered gate landing
   first would run these 9 broken tests and go red — the gate can only merge
   green once the suite is clean.

## The 9 fixes

| Area | Fix | Source of truth |
|------|-----|-----------------|
| `tests/tasks/test_tasks_2x_unit.py` | `_find_mission_slug(explicit_feature=)` → `explicit_mission=` (2 tests) | `src/specify_cli/cli/commands/agent/tasks.py:794` |
| `tests/specify_cli/test_active_mission_removal.py` | `verify_setup(feature=)` → `mission=` | `verify_setup` renamed the option |
| `architecture/test-suite-acceleration-plan.md` | Scrub 5× `/Users/robert/…` machine-specific home literals; make references repo-relative | `test_command_template_cleanliness.py` |
| `docs/how-to/install-claude-code-plugin.md` | Add required `title`/`description` YAML front matter | `tests/docs/test_docs_seo.py` |
| `tests/specify_cli/skills/__snapshots__/{codex,vibe}/{analyze,plan}.SKILL.md` | Regenerate stale snapshots | drift traced to committed source: #1996 (analyze caution block), #2007 (`setup-plan --mission` flag) |

The snapshot regeneration was verified to contain **only** the expected drift —
the source-template edits were already committed upstream; the snapshots simply
weren't regenerated at the time.

## Scope notes

- 3 further failures in the original report are macOS-runner-specific (timing,
  tmp-path length, non-git fixture) and are intentionally **out of scope** here.
- The root-cause gate fix follows in a separate PR.

## Verification

- Targeted tests for all 9 fixes pass; full `tests/docs/test_docs_seo.py` and
  full `tests/specify_cli/skills/test_command_renderer.py` (102) pass.
- `ruff check` clean on touched files.

[...]

Issue #1933, opened 2026-06-14
Run slowest CI test groups nightly instead of every PR


robertDouglass:
## Summary

Move the slowest PR CI test groups out of default PR execution and into an automatic nightly run. PR CI should stay focused on fast, high-signal validation, with explicit full-CI opt-in when a PR needs the heavy suite.

## Timing research

Full CI logs pulled here:

- `/Users/robert/spec-kitty-dev/spec-kitty-20260613-095746-qM0w01/github-ci-logs/zips`
- `/Users/robert/spec-kitty-dev/spec-kitty-20260613-095746-qM0w01/github-ci-logs/logs`
- Job metadata used for timing: `/Users/robert/spec-kitty-dev/spec-kitty-20260613-095746-qM0w01/github-ci-logs/jobs/*.json`

Analyzed ten recent PR `CI Quality` runs from 2026-06-12 and 2026-06-13. Durations below are from GitHub job `started_at` -> `completed_at`, excluding skipped jobs.

| CI group | Observed timing | Evidence / conclusion |
| --- | ---: | --- |
| `integration-tests-core-misc` | 7 full non-cancelled runs at 27m50s-29m11s, avg ~28m27s | Dominant PR critical path. One additional cancelled run had already consumed 16m20s. Run `27460591709` was still `in_progress` after starting at `2026-06-13T07:46:12Z` when pulled. |
| `fast-tests-charter` | 4 non-skipped runs at 8m48s-9m24s, avg ~8m58s | Secondary slow group. Keep on PR only if still required/targeted; otherwise nightly/full-CI candidate. |
| `integration-tests-next` | 4 non-skipped runs at 8m37s-9m14s, avg ~8m51s | Secondary slow group. Keep on PR only if still required/targeted; otherwise nightly/full-CI candidate. |
| `fast-tests-core-misc` | 9 non-skipped runs at 3m28s-5m52s, avg ~5m21s | Not the main blocker, but still a catch-all group to monitor. |
| next tier | `integration-tests-agent` avg ~3m44s, `fast-tests-cli` avg ~3m28s, `fast-tests-doctrine` avg ~2m36s, `integration-tests-cli` avg ~2m30s | These are not the slowest-of-slow PR blockers in the pulled logs. |

[...]

Issue #1931, opened 2026-06-14
EPIC: Test suite friction — test quality, suite hygiene & anti-pattern remediation


stijn-dejongh:
## Scope

Top-level epic owning **test suite friction — test quality, suite hygiene, and test infrastructure**, merged from the former #1931 and #2071 on 2026-08-02.

### Operator theory: the test suite can degrade from SCAFFOLD into FRICTION

Tests are supposed to be **scaffold** — a trustworthy structure that supports correct change and catches real regressions. When the suite drifts into **friction**, it obstructs correct work more than it protects it, via three anti-pattern classes:

1. **Inaccurate / intent-misaligned tests** — assertions that no longer reflect the actual contract or intended behavior; they pass or fail for reasons unrelated to whether the code is correct.
2. **Duplicate knowledge** — the same fact/contract encoded in many places, so a legitimate design change requires hand-editing N tests, making the suite brittle and expensive to evolve.
3. **Monkey-patch reliance / accidental-pass** — tests green *for the wrong reason* (over-mocked, monkey-patched around the real seam, asserting on stubs), giving false confidence without exercising the real path.

The most damaging consequence: a correct, intended change trips a mis-aligned or duplicate-knowledge test, and under time pressure the *change* gets reverted instead of the *test* being fixed — the suite actively blocks correct evolution.

### Also in scope (from the original #1931)

- Post-merge test-assertion staleness detection.
- Test-suite state leaks — atomicity, idempotency, self-cleaning (fake missions, /tmp litter, hardcoded shared paths).
- test-feature-* scratch-mission leakage into live mission state.
- Broader opt-in test gate before WP moves to for_review (catch stale tests earlier).
- CI-integrity: gates that silently stop enforcing (skipped checks, lost signal) — a false-green risk, not routine test debt.
- Playwright bootstrap for dashboard UI regression coverage.

## Goal

Restore and keep the test suite as a **trustworthy scaffold**: audit + remediate the three anti-pattern classes, establish a test-hygiene directive/styleguide, and add ratchets so regressions cannot silently recur — alongside the broader suite-hygiene and CI-infrastructure work already tracked here.

## Prior art to build on

- **Post-merge AST stale-assertion analyzer** (mission 068, `src/specify_cli/post_merge/`).
- **Live-evidence-over-static principle** — a bug witnessed in a real run is not fixed because the code "looks fixed"; tests must exercise the real path, not a monkey-patched stub.
[...]

PR #2067, opened 2026-06-21
test(#2034): static CI gate-coverage checker + orphan ratchet (folds in #1933)

kentonium3:
## Summary

Closes the root cause of #2034 (folding in #1933) with a *static* meta-checker, per @stijn-dejongh's direction in the issue thread. No monolithic full-suite run, no CI re-sharding: it reads the existing per-gate `-m`/path structure and statically asserts coverage over it.

CI selects tests by `(paths, -m marker_expr)` per gate, sharded across jobs. The **authoring** taxonomy diverges from the **selection** taxonomy: `pytest.ini` documents `unit` as "the category default for module-scoped tests" (and `contract` for contract tests), but **no gate selects `-m unit` or `-m contract`**, and several test directories are touched by no gate at all. Net effect today (figures from the baseline refreshed onto current `main`):

| | count | % |
|---|---:|---:|
| total tests | 25,762 | |
| **selected by zero gates** | **9,895** | **38.4%** |
| selected by exactly one gate | 14,424 | 56.0% |
| selected by ≥2 gates (duplicate) | 1,443 | 5.6% |

The orphan set is overwhelmingly **untested-but-green** — a coverage/regression-risk gap, not a pile of hidden failures (the 9 original symptoms were the failing tip; those are already fixed in #2039). Dominant cause is unchanged: the large majority carry `unit` (the module-scoped default no gate selects), a couple hundred carry `contract`, and the rest are whole ungated dirs (`tests/context`, `tests/prompts`, `tests/runtime`, …).

This PR adds the **enforcement substrate** so no *new* test can leak into zero gates by construction. It deliberately does **not** re-tier or re-shard CI — that migration is out of scope here; this is the guardrail and the baseline is the worklist.

## What's in it

All under `tests/architectural/`:

- **`_gate_coverage.py`** — parses every `pytest` invocation across the four suite-running workflows (`ci-quality`, `ci-windows`, `drift-detector`, `release`), expands the `integration-tests-core-misc` shard matrix, and evaluates each gate's `(paths, --ignore globs, -m expr)` against a one-pass `--collect-only` using pytest's own marker-expression evaluator (`_pytest.mark.expression.Expression`) — so selection semantics are identical to a real `-m` run. Also a small CLI (`--check` / `--update-baseline`).
- **`_gate_collect_plugin.py`** — collection-only plugin that dumps `{nodeid, relpath, markers}` per test and suppresses execution.
- **`_gate_coverage_baseline.json`** — the frozen orphan **file** worklist (448 files) — today's backlog made explicit (a ready-made #1931 worklist).
- **`test_gate_coverage.py`** — the drift guards (below).

## Policy (matches the thread)

- **Orphans → ratchet from baseline.** The recorded 448 files are allowed; any **new** ungated file goes red immediately. Closes the gap for all *future* tests without blocking on a ~9.9k-test cleanup.
- **Duplicates → report-only.** 1,443 surfaced as a warning, never failed — the fast↔integration domain split intentionally overlaps.

## Drift guards (`test_gate_coverage.py`)

Cheap structural guards (no collection): the four workflows still parse into a well-formed gate model; the core-misc shard matrix and the `windows_ci`/`quarantine`/`timing`/`slow` selectors stay modellable; the `ci-windows` gate is asserted to model `-m windows_ci` (its paths are dynamic `git grep` → whole-tree fallback, coverage-safe *only* because that marker narrows it); selection logic is unit-tested (a `unit`-only test in a misc-shard dir is an orphan, the same path marked `git_repo` is covered; the `--ignore` exclusion branch is exercised directly). A synthetic-universe test pins `analyze()`'s orphan/covered classification deterministically, and the live ratchet asserts the orphan set stays a **non-empty subset** of the baseline — so the checker silently going blind is itself caught. Plus the ratchet (new ungated file → fail) and the report-only duplicate surface.

## Scope

The CI re-tiering, domain re-sharding, and shift-left replacement of expensive e2e/smoke tests are **out of scope** for this PR. This adds the static guarantee that makes re-tiering safe; the baseline is the worklist to migrate against.

**Taxonomy/tiering is deliberately scoped out, not deferred for lack of a decision.** The checker is **agnostic to the tier count** — completeness (ratchet) and duplicate-report need zero taxonomy input. When the canonical tiers are settled they just relabel a *derived* column (a test's tier = the set of gates that select it); this code doesn't change. So the 448-file baseline is the #1931 worklist to migrate against, and re-tiering is a clean follow-up. This is the one call with scope implications — flag it on review if you'd rather fold re-tiering in here.

**Tracker disposition on merge:** this PR *fences* the hole, it does not fill it — the ~9.9k ungated tests still don't run in CI. So **#2034 should not auto-close on merge**; rather, re-scope it to the coverage-closure / re-tiering work under the #1931 epic, and keep #1933 open (the enforcement substrate lands here, the cost-side migration does not). The 448-file baseline becomes #1931's concrete worklist.

## Design decisions

These are decided as below, with the rationale; each is cheap to change on review if you'd steer differently.

1. **Ratchet runs as an `architectural` test** (~2 min: one full `--collect-only`), **not a new CI job — `ci-quality.yml` is untouched**, keeping CI job/shard structure out of scope. Reversible: a dedicated collection-only job (cheaper, isolated from the shard) is a drop-in follow-up if you'd prefer to isolate it.
2. **Baseline keys on orphan _files_, not nodeids** — far less churn (test renames don't thrash it) while still catching any new ungated file. Known tradeoff: a new `unit`-only test added to an *already-orphaned* file isn't caught — and **67 of the 448 baselined files are active churn magnets** (hot seams like `test_commit_router*`, `test_mission_topology_seam`), so that residual exposure is real, not theoretical. Reversible: tightening to nodeid-level is a one-line change if you want that gap closed.
3. **Uses pytest's private `_pytest.mark.expression`** to guarantee identical `-m` semantics. pytest is **floored (`>=9.0.3`), not upper-pinned** — a breaking move of this private import fails loudly at import time, and the import contract is pinned by a focused test; `uv.lock` pins the exact resolved version for reproducible runs. Adding a pytest upper bound is a dependency-policy call I've left to you.

## Validation

- Model reproduces the issue's known orphans (`test_tasks_2x_unit.py`, `test_active_mission_removal.py`, `test_command_renderer.py`, …).
- **Falsification:** a synthetic `unit`-only test in ungated `tests/context/` made `--check` exit 1 and list it as a new ungated file; green again on removal.
- Hardened against a partial-collection false-negative (a broken new test file aborts collection → the checker fails loudly instead of trusting an incomplete dump).
- `ruff` + `mypy` clean; the existing `tests/architectural/` guards still pass; 18/18 of this PR's tests pass.

## Sonar / CI notes

New branches/helpers are covered by focused tests in this PR (structural, selection-logic incl. the `--ignore` branch, parser-edge, the marker-expression import contract, and the collection-soundness guard). No localhost/transport hotspots introduced. No `# noqa` / `# type: ignore` suppressions added.

---


kentonium3:
**Process note (no action needed).** The orphan-ratchet in this PR runs in the `integration-tests-core-misc (architectural)` shard, which is gated out of draft PRs — so to capture a real-runner timing + green result I'm going to briefly add `ready-for-ci` (or flip this PR to ready), let that shard run, then **convert it straight back to draft**. If you see a ready→draft blip here, it's purely to collect CI data, **not** a review request. A full walkthrough of the approach — and the still-open, non-blocking taxonomy question — will follow once I've confirmed it on the runner.

kentonium3:
@stijn-dejongh — implementing the static gate-coverage checker you called for in #2034 — flipping this to ready for review. The calls I made on the open design points, up front so they're not buried; all cheap to change if you'd steer differently:

- **Ratchet placement:** runs as an `architectural` test (~2 min collect-only), `ci-quality.yml` untouched — kept CI shard structure out of scope. Trivially movable to a dedicated collection-only job if you'd rather isolate it.
- **Baseline granularity:** keyed on orphan *files* (440), not nodeids — renames don't thrash it. Known tradeoff: a new `unit`-only test in an already-orphaned file isn't caught; nodeid-level is a one-line tightening if you want it.
- **Taxonomy/tiering:** scoped **out** — the checker is tier-count agnostic, so I treated the 440-file baseline as the #1931 worklist rather than re-tiering here. This is the one call with scope implications; say the word and I'll fold re-tiering in.

(Earlier ready→draft blip was just CI-data collection — this one is a genuine review request.)

stijn-dejongh:
## Maintainer review — 4-lens adversarial squad (debbie live-repro / alphonso model-correctness / renata test-quality / priti scope)

First: this is strong, well-documented work, and the design was maintainer-steered in the #2034 thread (static meta-checker, ratchet-not-zero, duplicates report-only) — not a unilateral narrowing. Rebased onto current `upstream/main`; the rebase surfaced **8 new orphan files** from missions merged during the PR's lifetime (topology/meta-reader/lang-guard tests) → baseline refreshed 440→448 in a separate attributed commit. That refresh is itself evidence the leak is live.

### ✅ debbie (live reproduction) — APPROVE
Verified by execution, not code-reading: `--check` runs green on the rebased tree (25762 tests / 9895 orphans / 448 files); a planted `@pytest.mark.unit`-only test in `tests/context/` makes `--check` **exit 1 naming the file**, green on removal (the ratchet fires); the static `(paths, -m)` model matches real `pytest --collect-only -m "<expr>"` selection **cell-for-cell** across covered / orphan / duplicate probes (no false coverage, no false orphan); and a broken new file that aborts collection makes `collect_universe()` **raise rather than trust a partial dump** (the #2034 P2 hardening is real). Marker semantics are pytest's own `_pytest.mark.expression.Expression`, so they're identical to CI by construction.

### ✅ alphonso (model/parser correctness) — faithful, with one flake to chase
Parser fidelity is strong: all 56 pytest invocations across the 4 workflows are correctly enumerated (env-prefix strip, `uv run` vs `-m`, `git grep`/`pipx` rejected as non-commands, `| tee`/`&&` splitting, continuations, `--ignore`, `::`-nodeid selectors, whole-tree default). The `integration-tests-core-misc` 6-shard matrix expands correctly; marker evaluation runs **after** the auto-marker conftest hooks (confirmed: 1374 charter tests carry the auto `fast` marker in the dump); path/ignore boundary traps (`tests/git` vs `tests/git_ops`) handled.
- **[HIGH] non-deterministic failure of `test_selection_logic_matches_marker_and_path`** — `Expression.evaluate({git_repo})` returned `False` once where it should select, then passed on 4 reruns + always outside pytest. Couldn't repro deterministically (debbie saw a similar anomaly she traced to **bytecode-cache contamination**). Since the same eval path backs `analyze()`, a transient False could in principle inflate the orphan set. **Recommend: reproduce under the real `architectural` shard (pytest 9.0.3, `-n auto`) and root-cause — don't retry-to-green.** Likely a `__pycache__`/xdist isolation artifact, but worth confirming on a gate.
- **[MEDIUM]** the private `_pytest.mark.expression` import is **floored (`pytest>=9.0.3`), not pinned** — the "pinned" claim is inaccurate; fails loud on a breaking bump (acceptable) but add an upper bound or soften the docstring.
- **[MEDIUM]** `ci-windows.yml` paths can't be parsed (dynamic `git grep`) → modeled as whole-tree, coverage-safe *only* because `-m windows_ci` narrows it. Document the invariant + assert the windows gate models `marker_expr == "windows_ci"`.

### ⚠️ renata (test non-fakeability) — request changes (2 HIGH)
The checker works, but its **own tests don't fully guard its core** (the exact "untested-but-green" failure mode this PR exists to catch):
- **[HIGH]** `test_no_new_orphan_surfaces` / `test_orphan_backlog_does_not_grow` only compute `current − baseline` and never assert the orphan set is non-empty/matches baseline → mutating `analyze()` to **never detect orphans** leaves all 10 tests GREEN (verified). Add a real-universe floor: `set(report.orphan_files) == set(baseline['orphan_files'])` (or `>=` count floor).
- **[HIGH]** the `--ignore` exclusion branch in `CompiledGate.selects` (line ~295) has **no focused test** → disabling it (duplicate count 1443→9294) leaves all 10 GREEN. Add a `selects`-with-`--ignore` test (29 real ignores make it load-bearing).
- **[MEDIUM]** `_path_matches` nodeid-prefix branch + the string-parsing helpers (`_substitute_matrix`/`_join_continuations`/`_strip_to_command`, used 323× in ci-quality.yml) are only transitively exercised — add focused parser tests (project rule: every new branch needs a test in-PR).
- **[MEDIUM, honest tradeoff]** the file-keyed baseline means a new `unit`-only test appended to an **already-orphaned** file isn't caught — **67 of the 448 files are active churn magnets** (incl. hot seams like `test_commit_router*`, `test_mission_topology_seam`). It's documented, but the 67-hot-file residual exposure should be in the PR body.

### ✅ priti (scope/sequencing) — KEEP RE-TIERING SCOPED OUT (decisive); READY on scope
- **Fold-in verdict:** keep #1933 re-tiering / re-sharding OUT. The ratchet is useful **without** the cleanup (prevention + a measured 448-file worklist); folding in CI-surgery (incl. nonexistent `ci:full`/`ready-for-ci` labels) against a moving matrix would make a 0-deletion test-infra PR un-reviewable. Guardrail-first against a frozen baseline is the correct sequence.
- **Tracker disposition on merge:** **do NOT auto-close #2034** — the 9,895 ungated tests still don't run; the PR *fences* the hole, doesn't fix it. Re-scope #2034 to the coverage-closure/re-tiering work and keep it OPEN under #1931. Keep **#1933** open (substrate delivered, migration not). **#1931** epic gains the 448-file baseline as its concrete worklist.

### Synthesis
The checker is sound and the scope is right. Before merge I'd want renata's **2 HIGH** closed (a CI gate whose own core can silently stop detecting orphans is the one regression this PR mustn't allow) and alphonso's **flake** reproduced/root-caused on the real shard. All are bounded, in-PR additions consistent with the project's "every new branch needs a test" rule — happy to add them or leave them to you, @kentonium3. Thanks for the thorough write-up and the falsification evidence.

kentonium3:
@stijn-dejongh — thanks for rebasing onto current `main` and refreshing the baseline (440→448).

I took the whole list. Pushed as `399f6ff81` on top of your baseline-refresh commit — no behaviour change to the checker, all bounded in-PR test/doc additions consistent with the "every new branch needs a test" rule. 18/18 green (`-n0`), `ruff` + `mypy` clean, Codex review clean.

**renata HIGH — the checker's own tests were fakeable**

- **Orphan floor.** Added `test_analyze_detects_orphan_and_covered_records`: a synthetic-universe unit test that pins `analyze()`'s orphan/covered classification deterministically (no collection). Mutating `analyze()` to never detect orphans now fails this outright. I also floored the live ratchet (`test_no_new_orphan_surfaces`) to assert the current orphan set is a **non-empty subset** of the baseline. One deliberate divergence from the literal suggestion: I did *not* use strict `set(current) == set(baseline)`, because that would fight the intentional shrink-warns design — `test_orphan_backlog_does_not_grow` treats legitimate backlog shrinkage as a *warning*, not a failure, so a strict `==` would hard-fail the moment someone fixes an orphan. Subset + non-empty preserves "growth fails, shrinkage warns" while still catching the checker going blind. Happy to tighten further if you'd rather.
- **`--ignore` branch.** Added `test_selection_respects_ignore` — a focused guard on the exclusion branch in `CompiledGate.selects`. Disabling that branch (which you measured flipping the duplicate count 1443→9294) now fails.

**renata / alphonso MED**

- Focused tests for the previously-only-transitively-exercised helpers: `_substitute_matrix`, `_join_continuations`, `_strip_to_command`, and the `_path_matches` nodeid-prefix branch.
- `test_pytest_marker_expression_import_contract` + corrected the docstring: you're right that "pinned" was inaccurate — pytest is **floored (`>=9.0.3`), not upper-pinned**. The import is private, so I've made it fail loudly at import time and pinned the contract with a test. An upper bound in `pyproject.toml` is a dependency-policy call I've left to you (flagged in the PR body).
- `test_windows_gate_models_windows_ci_marker` asserts the `ci-windows` gate models exactly `-m windows_ci`, and I documented the invariant in `CompiledGate.__init__`: dynamic `git grep` paths → whole-tree fallback, which is coverage-safe *only* because the marker narrows it.

**alphonso HIGH — the flake**

I hardened `test_selection_logic_matches_marker_and_path` with diagnostics, but I want to be straight about the root cause rather than retry-to-green. `selects()` is a **pure, deterministic** function of its arguments — the marker expression compiles once and `Expression.evaluate` has no data-dependent branch that can flip for a fixed marker set. So a transient `False` there is necessarily an *environmental* artifact (stale `__pycache__` / xdist worker isolation — the same class debbie traced on a sibling test), not a logic fault. I couldn't reproduce it deterministically locally either. **This push is a `synchronize` on a non-draft PR, so it runs the real `integration-tests-core-misc (architectural)` shard — that's the real-shard evidence you asked for.**

**priti — scope**

Agreed and reflected in the PR body: re-tiering stays **out**, this is guardrail-first against a frozen baseline. I added an explicit **tracker disposition** note — **don't auto-close #2034 on merge** (the ~9.9k ungated tests still don't run; this fences the hole, doesn't fill it); re-scope it under #1931 with the 448-file baseline as the worklist, and keep #1933 open. I also documented the 67-hot-file file-keyed residual exposure under design decision #2.

Let me know if you'd prefer any of these tightened (nodeid-level baseline, strict `==` floor, or the pytest upper bound) — all are small. Thanks again.

btw — this is my first-ever code PR and second PR behind that one doc change I submitted a week or so ago. Thanks for facilitating!

kentonium3:
Heads-up on the CI red here: it's **inherited from `main`, not introduced by this PR**. This diff only touches `tests/architectural/_gate_coverage*`, but the rebase landed it on `aeb8dfc31`, whose own `ci-quality` run is red ([run 28044067632](https://github.com/Priivacy-ai/spec-kitty/actions/runs/28044067632)) with the same jobs failing. This PR's architectural shard came back **3 failed / 537 passed**, and the 3 are exactly the pre-existing `main` failures:

- **`architectural`** — `test_untrusted_path_containment.py::test_audit_passes_on_fixed_tree`, `::test_all_discovered_rows_appear_in_inventory`, and `test_no_dead_symbols.py::test_no_public_symbol_in_all_is_unimported`.
- **`specify-cli-rest`** — `tests/specify_cli/test_standalone_tasks_cli_canonical.py::test_src_task_helpers_require_canonical_status_for_lane` (`CanonicalStatusNotFoundError` for `060-standalone-test`).
- **`quality-gate`** — aggregates the above.

No *new* failure beyond the inherited set — every gate-coverage test passed (they're in the 537). Two things worth calling out:

- **alphonso's flake didn't reproduce.** `test_selection_logic_matches_marker_and_path` passed on the real `architectural` shard, consistent with the read that the transient `False` was an environmental `__pycache__`/xdist artifact, not a logic fault.
- I'm flagging these rather than touching them — they're unrelated to this change and look like in-flight work on `main`. The PR can't go fully green until `main` is, but nothing here is blocked on this diff.

stijn-dejongh:
Maintainer rebase + baseline refresh (with your review items already in): rebased onto current `main` (now includes #2106 write-surface-coherence and **#2108**, which gated 64 hot churn-magnet orphans). Regenerated the frozen baseline accordingly — the orphan set dropped **448→384 files (9902→7869 tests)**, so the ratchet floor now reflects the reduced backlog rather than grandfathering the now-gated files. Local: `_gate_coverage.py --check` clean, `test_gate_coverage.py` 18 passed. Note: #2108 surfaced 4 *red* orphans (tracked in #2109) — proof your checker's thesis holds; those stay orphan until fixed-then-gated. Merging once CI is green. Thanks for this — it's the ratchet that keeps the orphan backlog from regrowing.

Issue #2096, opened 2026-06-23
Tiered coding standards (DDD) — doctrine-only slice: styleguide + DRG edge


stijn-dejongh:
Child slice of #1843, delivered by mission `org-pack-subdir-and-doctrine-qol-01KVSRJ6` (Thread C, fork branch `feat/doctrine-qol-2083`).

## Scope (doctrine-only)
- Add one canonical **`styleguide`** doctrine artifact defining domain-importance tiers (core vs glue) **mapped to named, currently-existing `src/` areas**, with a per-tier rigour table (coverage / duplication / smell / lint / typing).
- Register it in the DRG via the graph generator (`spec-kitty doctrine regenerate-graph`), and make it **non-orphan**: ≥1 inbound DRG edge (a `suggests`/`requires` from a directive/paradigm — doctrine-only, no CI/agent-effort change) + a non-orphan test.

## Explicitly deferred to parent #1843
- CI per-tier enforcement gates.
- Agent-effort routing by tier.

Rationale: gives the tiered standard a canonical, resolvable home now without the epic-scale enforcement work. Parent: #1843.

Issue #2092, opened 2026-06-23
Doctrine catalog silently drops artifacts with `applies_to_languages: [any]` — `any`/`all` treated as literal languages, not wildcards


stijn-dejongh:
## Symptom

A doctrine artifact (directive/tactic/styleguide/etc.) that declares `applies_to_languages: [any]` — a natural way to say "applies to all languages" — is **silently filtered out** of the active catalog under any concrete language scope, and is reported as a MISSING artifact when referenced. `spec-kitty doctrine validate` PASSES for the file (it is schema-valid), so the drop is invisible at authoring time.

## Concrete Instance (Already Hit)

The tactic `delete-the-assertion-not-the-test` (referenced by DIRECTIVE_041 and the `python-pedro` agent profile) declared `applies_to_languages: [any]`. During `spec-kitty dispatch --profile python-pedro` (which builds the tactics repo with `active_languages=['python']`), the intersection `{"any"} ∩ {"python"} = ∅` caused the tactic to be filtered out. The charter renderer's `repo.get(...)` returned `None` and printed:

```
(catalog entry not found for tactic:delete-the-assertion-not-the-test) Cause: missing_artifact
```

The specific tactic was unblocked in PR #2089 by removing the `[any]` field — but the **root gap remains**. The next author who writes `[any]` or `[all]` hits the same silent drop with no authoring-time signal.

## Root Cause (Anchors — Live Code)

- **`applies_to_languages_match()`** in `src/doctrine/shared/scoping.py` (~line 24): requires the artifact's language set to _overlap_ the active set. An artifact with the field **omitted** is unscoped and always loads. But `[any]` / `[all]` are treated as **literal language tokens** — they carry no wildcard semantics.
- **Active-language set construction**: `src/doctrine/service.py` (~lines 75–83).
- **Missing → warning seam**: `src/charter/context.py` (~line 1851) — the `MISSING_ARTIFACT` diagnostic path.

## Two Fixes Requested

1. **Wildcard semantics or validate-time rejection** — either:
   - Treat `[any]` / `[all]` as wildcard values in `applies_to_languages_match()` (equivalent to omitting the field), OR
   - Reject them at `spec-kitty doctrine validate` time with an actionable message: _"Use of `any`/`all` as a language token is not supported — omit the field to mean always-applicable."_
   
   A **validate-time guard** is preferred: it fails loud where the author can see it, without silently filtering at query time.

[...]

Issue #2080, opened 2026-06-22
Audit + remediate the DRG and doctrine artefacts (daphne-led, post-paradigm-references enablement)


stijn-dejongh:
## Summary

Now that the `doctrine-daphne` curator profile is in place to own doctrine integrity, run a **systematic daphne-led audit + analysis + remediation pass** over the DRG (doctrine relationship graph) and the doctrine artefacts. This consolidates several concrete gaps surfaced while authoring the test-hygiene + adversarial-squad enablers (2026-06-22) into one functional doctrine-debt epic with a remediation plan as its deliverable.

Owner profile: **`doctrine-daphne`** (curator). This is the first substantial workload for that profile.

### 1. DRG generator gaps (#1755 territory — #1755 is CLOSED but residual)
The extractor (`src/doctrine/drg/migration/extractor.py`) only NODE-walks `tactics/built-in/`. Top-level `tactics/*.tactic.yaml` (e.g. `acceptance-test-first`) become DRG nodes **only via inbound references** — a brand-new top-level tactic is invisible to the graph until something references it. Audit which artefacts are silently absent from the DRG because of asymmetric node-walking, and close the visibility gap.

### 2. FR-028 retired-field inconsistency (mission `excise-doctrine-curation-and-inline-references-01KP54J6`, #463)
Inline `tactic_refs` / `paradigm_refs` were excised in favour of structured references / DRG edges — **but**:
- `directive_refs` survived inline on directives/paradigms.
- The paradigm `references` field had to be **(re)enabled this cycle** to support paradigm-level references.

The retired-field set vs the surviving-field set is **inconsistent across artifact kinds**. Reconcile: decide per-kind which inline reference fields are canonical vs DRG-edge-only, and align schema/model/extractor accordingly.

### 3. Orphan-residual mechanism freshness
The DRG carries a documented orphan ceiling (`DOCUMENTED_ORPHAN_RESIDUAL`) plus a per-orphan rationale doc. Audit whether:
- the residual list is **current**,
- any "valid orphan" should instead be **wired** into the graph,
- the ceiling has **drifted** vs the rationale doc.

### 4. Charter-catalog sync
[...]

Issue #1843, opened 2026-06-11
Tiered coding standards by domain importance (DDD): per-tier coverage/duplication/smell/lint/typing rigour in doctrine + CI + agent effort


stijn-dejongh:
## Intent (operator decision, 2026-06-11)
Not every part of the codebase is equal, and it should not be treated equally. Per DDD ideology: **core domain / shared kernels / multi-consumer libraries** are held to a HIGHER standard of rigour (coverage, duplication, smells, lint, typing) than **supporting/glue/duct-tape/POC** code. This must reflect in:
1. **How CI treats the code** — tier-aware quality gates, and
2. **How much effort agents put into it** — implementers/reviewers calibrate depth to the tier of what they touch.

### A. Doctrine: a set of related built-in artifacts
- A **tier taxonomy artifact** (e.g. `core` / `supporting` / `generic` / `experimental`, names TBD) with per-tier standards declarations (min coverage, mypy strictness, smell budget, duplication budget, review depth).
- A **module→tier map** declared per project (charter-resolved; likely DRG-linked so tier assignments are first-class governed artifacts with `requires`/`suggests` edges to the standards they imply).
- **Directives/tactics** consuming the tier: implementer profiles ("tier=core ⇒ mypy --strict, property tests where applicable"), reviewer profiles ("tier=generic glue ⇒ don't demand core-tier polish"), and the WP-prompt generator threading the tier of `owned_files` into prompts (effort calibration, both directions — prevents BOTH under-rigour on the kernel AND gold-plating of glue).

### B. Enforcement: spec-kitty repo CI as first consumer (dogfood)
- **Coverage:** per-tier thresholds (e.g. core ≥90%, supporting ≥75%, experimental advisory-only) — per-path coverage gating likely needs a CI-side check over coverage.xml, since SonarCloud quality gates are project-scoped (feasibility note below).
- **mypy:** strict-mode module list = core-tier modules (mypy supports per-module config natively).
- **ruff:** per-tier rule sets via per-path configuration.
- **Sonar:** investigate per-path issue severity/exclusion mapping vs a CI-side tier report; smell budget per tier.
- Tier map seed for this repo (illustrative, needs analysis): `core/commit_guard`, `mission_runtime/*`, `status/*` reducer+store, DRG/merge = **core**; CLI command shells, render helpers = **supporting**; scripts/, demos = **generic/experimental**.

## Open questions (for the in-depth analysis)
[...]

PR #2098, opened 2026-06-23
Org-pack subdir source + doctrine QoL (#2083, #707, #2096, #2092)

### LINKED ISSUE #2083: Org-pack git source can't consume packs in a subdirectory (no subdir/pack_path field)
## Summary

`doctrine.org.packs[]` with `source_type: git` cannot consume a pack that lives in a **subdirectory** of its repository. The git source clones the repo **root** into `local_path`, and the org-pack schema has no `subdir`/`pack_path` field, so packs whose root is e.g. `pack/` inside the repo can't be referenced via a git source.

## Environment

- spec-kitty-cli `3.2.2` (also reproducible on current `main`)

## Where it shows up

- `doctrine/drg/org_pack_config.py` — `OrgPackConfig` fields are `name`, `local_path`, `source_type` (`git`/`https`/`api`), `url`, `ref`. There is no `subdir`/`pack_path`.
- `specify_cli/doctrine/sources/git_source.py` — `GitSource.fetch(target_dir)` runs `git clone <url> <target_dir>` then `git checkout <ref>`. `target_dir` **is** `local_path`, i.e. the clone root becomes the pack root.
- `doctrine/drg/org_pack_config.py::resolve_org_roots` returns `pack.local_path` verbatim (no `pack/` auto-detection).

## Repro

A repo whose doctrine pack lives at `pack/` (repo root is not a pack root — `org-charter.yaml` is at `pack/org-charter.yaml`):

```yaml
# .kittify/config.yaml
doctrine:
  org:
    packs:
      - name: my-pack
        source_type: git
        url: ssh://git@example.com/org/repo.git
        ref: main
        local_path: .doctrine-cache/repo   # clone root
```

`spec-kitty doctrine fetch` clones the repo into `.doctrine-cache/repo`, but the actual pack root is `.doctrine-cache/repo/pack`. `spec-kitty doctor doctrine` / pack validation then fail to find a valid pack at `local_path`.

## Proposed enhancement

Add an optional `subdir` (a.k.a. `pack_path`) field to `OrgPackConfig`. When set on a git/https source, the effective pack root resolves to `local_path / subdir` after fetch, while the clone/extract still targets `local_path` (repo root). This keeps `local_path` as the cache/clone location and lets the pack live in a subdirectory.

```yaml
- name: my-pack
  source_type: git
  url: ssh://git@example.com/org/repo.git
  ref: main
  local_path: .doctrine-cache/repo
  subdir: pack          # effective pack root = .doctrine-cache/repo/pack
```

(Sparse-checkout of `subdir` would be a nice optimization but isn't required for correctness.)

## Workaround today

Clone manually and point `local_path` at the clone's subdir (`.doctrine-cache/repo/pack`), treating `source_type: git` + `url` + `ref` as documentary metadata rather than a `doctrine fetch`-driven source.

## Impact

Common for monorepos / repos that ship a doctrine pack alongside other content (tools, docs, CI). Without subdir support these packs can only be consumed via published archive extraction, not directly from their git source.

stijn-dejongh:
## Summary

Doctrine-stack QoL mission (`org-pack-subdir-and-doctrine-qol-01KVSRJ6`) bundling one P1 source-resolution bug, one P1 doctrine-catalog reliability bug, and two doctrine quality improvements. Five clean commits, one per deliverable + an admin/planning bundle.

| Issue | Thread | What |
|-------|--------|------|
| **#2083** (P1) | A | Org-pack `source_type: git` can now consume a pack in a **subdirectory**: new `OrgPackConfig.subdir` + a single canonical `effective_root` seam that **every** pack-root consumer adopts (incl. the `doctor doctrine` health path), containment validation, round-trip, and effective-root fetch reporting. |
| **#707** | B | Honest `ruamel.yaml` vs `PyYAML` library-choice guide — names the real call sites and the `config.yaml` dual-read contradiction (current-state, not a fabricated invariant). |
| **#2096** (slice of #1843) | C | Doctrine-only tiered coding-standards `styleguide` mapping core/glue/scripting tiers to named `src/` areas; **non-orphan** via a `suggests` edge from `DIRECTIVE_030`. No CI/agent-effort changes (deferred to epic #1843). |
| **#2092** (P1) | D | Doctrine catalog no longer silently drops `applies_to_languages: [any]`/`[all]` — kind-agnostic **validate-time guard** (fail-loud at authoring) + a `SCOPE_FILTERED` catalog-miss diagnostic wired end-to-end (FR-013). |

Closes #2083, #707, #2096, #2092. References #1843 (epic — this delivers slice ① only) and #2080 (daphne-led DRG audit — deliberately **not** folded; left as a follow-up).

## How it was built

- Two **adversarial review squads** (post-spec + post-tasks) hardened the plan. The post-tasks squad caught a partial-fix BLOCKER: `resolve_org_roots` (the registry fan-in feeding ~9 `DoctrineService` consumers) had to adopt the seam too, not just the direct readers.
- Each WP went through an independent `reviewer-renata` review (one real reject/fix cycle on #2083's fetch-reporting int-contract).
- Post-merge architectural sweep caught and fixed cumulative-gate debt the per-WP reviews missed: two schema-validity regressions (a directive `references` `when:` field and a styleguide `tiers:` field that made both artifacts silently skip at load) and two dead-symbol-gate issues.

## Test status

- New WP tests + `tests/doctrine/` and `tests/architectural/` sweeps green on the integrated branch (426 architectural tests pass).
- **Note for CI:** three architectural tests (`test_uv_lock_pin_drift`, `test_real_home_isolation_guard`, `test_ci_quality_path_filters`) failed only in the local dev clone because its venv had test tooling installed manually (uv.lock drift) / shared HOME; they should pass on clean CI. Flag if CI shows otherwise.
- No Sonar UI work is expected; the code is self-contained.


stijn-dejongh:
✅ **CI green** (15 pass / 0 fail; quality-gate CLEAN).

Three first-run failures were triaged and fixed in `dac9d6692`:
- **docs-freshness** (mission gap): registered `docs/development/yaml-libraries.md` in the docs page-inventory — a CI-only gate the local checks don't cover.
- **fast-tests-charter** — two `*_shipped_invariant_*override*` tests (`test_org_drg_cannot_override_shipped_invariants.py`, `test_org_drg_loader.py`) asserted a **retired** hard-fail `OrgDRGConflictError`. **These are pre-existing on `main`, not introduced here** — proven by running them against a clean `upstream/main` worktree with its own `src` on `PYTHONPATH` (both fail identically; this PR touches neither `merge.py` nor those tests). The merge intentionally **warns-not-raises** now (`doctrine.drg.merge._warn_builtin_override`: "permitted by the merge but visible by design"). I remediated both to assert the current permitted-with-warning contract (non-vacuous — they assert the same-kind-override WARNING naming the overridden node). Flagging since this is a small out-of-mission test change; happy to split it to a separate PR if preferred.
- **quality-gate** cleared once the above went green.

History was compressed to 5 deliverable commits (admin bundled, one per issue) + the two remediation commits.

Issue #2102, opened 2026-06-23
record-analysis dirty-tree preflight blocks on spec-kitty's own uncommitted artifacts (meta.json + .kittify global.jsonl) under coord topology


rayjohnson:
## Summary

Under **`topology: coord`** (protected-main / `pr_bound`) missions, `record-analysis`
refuses to run because its dirty-tree preflight blocks on artifacts that **spec-kitty itself
left modified-and-uncommitted on the primary checkout**. The tool's own commit flow routes
those writes to the *coordination* branch, so they linger as uncommitted changes on the
primary/target branch — and then `record-analysis`'s preflight blocks on that self-created
dirt.

This is **distinct** from the feature_dir resolver split in #2101 (different mechanism: a
dirty-tree allowlist gap, not a path resolver). They share the coordination-topology theme.

## Environment

- **Exact commit:** `aeb8dfc31a88d2ae779e7b32f5340e90159733e6` (`upstream/main`, 2026-06-23)
- Reported version string: `3.2.2` (identical across recent commits; use the SHA)
- Mission: `software-dev`, `pr_bound: true`, `topology: coord`, primary branch protected,
  coordination worktree materialized

## Symptoms

`record-analysis` aborted with a `DIRTY_WORKTREE` error. The dirty paths were two files
spec-kitty had modified during the mission but not committed to the primary checkout:

- `kitty-specs/<mission>/meta.json`
- `.kittify/encoding-provenance/global.jsonl`

Both are spec-kitty-managed bookkeeping, not user edits. We resolved it by committing those
two files to the target branch ourselves.

## Root-cause analysis

The preflight (`_enforce_analysis_report_write_preflight` in
`src/specify_cli/cli/commands/agent/mission.py`) drops "coordination residue" paths from the
dirty set via `is_coordination_artifact_residue_path`
(`src/mission_runtime/artifacts.py:141`). But:

1. **`meta.json` is never residue-allowed.** It is classified `PRIMARY_METADATA` with
   `ignores_primary_coord_residue=False` (`artifacts.py:120`). That is correct *if* the
   normal flow commits `meta.json` to the primary checkout — but the mission left it modified
   and uncommitted, so the guard (by its own rules) blocks.
2. **`.kittify/encoding-provenance/global.jsonl` can't even be classified.**
   `_artifact_kind_for_path` only recognizes paths under `kitty-specs/`
   (`artifacts.py:164-190`); this file lives outside that tree, so the predicate returns
[...]

Issue #1716, opened 2026-06-05
Make coordination topology coherent from mission create through planning


robertDouglass:
## Summary

> **SUPERSEDED (2026-07-08, operator decision, mission `coord-primary-partition-lock-01KWZ46V`):**
> The "Locked Architecture Decision" and remediation items 1-4 below (materialize the
> coordination worktree at `mission create`; route `setup-plan`/`finalize-tasks` through
> `BookkeepingTransaction` onto the coordination branch) are **rejected**, not adopted.
> PRs #2106/#2113 already shipped and ratified the opposite model: **planning stays on
> primary**. The ratified partition is now locked:
>
> - **coord** = lifecycle surfaces only — status, notes, trace, issue-matrix, `move-task`.
> - **primary** = stable planning — spec/plan/WP outlines (never routed to coord).
> - **no-coordination topology** (`SINGLE_BRANCH` / `LANES`) → everything resolves to primary.
>
> Materializing the coord worktree at create is explicitly out of scope (see
> `coord-primary-partition-lock` spec C-003/C-004). Remediation item 5 (retire "planning
> happens in main" language) and the general spirit of item 6 (one placement seam, no
> duplicate resolvers) **remain valid** and are delivered by that mission's placement seam
> (`mission_runtime.resolve_action_context`) and its FR-010 docs truth-up. Read the rest of
> this issue as **historical investigation evidence** for the split-authority symptom, not as
> the current design — see research `coord-primary-partition-lock-01KWZ46V/research.md` D2 for
> the full rationale.

Modern missions can currently enter a split-authority state: `mission create` writes `coordination_branch` into `meta.json`, so later read/status/decision paths treat the coordination worktree as authoritative, but spec/setup-plan/bootstrap planning paths can still write and commit through the primary checkout (`main` / `target_branch`) before the coordination worktree exists.

This produces the confusing operator-visible behavior we saw:

- spec/setup-plan appears to operate on `main` successfully;
- later plan/status/decision commands resolve the mission as coordination-topology;
- read-side code refuses to fall back to stale primary checkout state when the coord worktree is missing;
- the operator has to manually create/use `.worktrees/<mission>-<mid8>-coord/` to proceed.

Architectural root cause: `coordination_branch` is used as the topology activation signal, but coord worktree/materialized authority is deferred until later lifecycle commands. Bootstrap planning commands still carry older “planning happens in main” semantics.

### Mission that introduced the topology

[...]

Issue #2085, opened 2026-06-22
A mission completed by the orchestrator can't pass spec-kitty accept


rayjohnson:
I ran a mission fully through `spec-kitty-orchestrator` — all 9 work packages implemented, agent-reviewed, and lane-merged, and the orchestrator's own `accept-mission` was happy. But when I ran `spec-kitty accept` to actually get it merge-ready, it blocked — and the blockers are things the orchestrator never does and seemingly was never expected to:

- **Every tasks.md subtask checkbox is still `- [ ]`.** In the hand-driven flow the implementing agent ticks these off with `mark-status` as it goes; the orchestrator's agent never does. So accept reports ~41 "unchecked tasks" even though all 9 WPs are approved/done.
- **The acceptance matrix isn't filled in.** `accept` requires a populated, verified `acceptance-matrix.json` (a non-pending pass verdict); nothing in the orchestrate flow produces or verifies one.
- (Plus a dirty-tree trip on a global bookkeeping file — separate; see #2084.)

So an orchestrated mission is, structurally, not acceptable via `spec-kitty accept` without a human going back to run `mark-status` and fill in the matrix by hand.

What I can't tell is the **intent**. The orchestrator drives to `orchestrator-api accept-mission`, which is a *lighter* bar (WPs done + workflow evidence) and doesn't check checkboxes or the matrix. `spec-kitty accept` is the heavier merge-readiness gate. Nothing bridges the two — they were built to different finish lines. So: is `spec-kitty accept` meant to be a **separate post-orchestrate verification step** (in which case the docs/UX should say so, ideally with a helper), or is the orchestrator supposed to leave a mission that `accept` can pass?

One concrete, low-risk improvement regardless of that answer: **the unchecked-tasks gate could derive completion from WP status.** `accept` already requires every WP to be approved/done; if they are, the subtasks are done, and *separately* demanding the tasks.md checkboxes be ticked is redundant bookkeeping that only the interactive flow happens to produce. Deriving from WP terminal status (or downgrading unchecked-tasks to a warning when all WPs are approved/done) would clear that gate for orchestrated missions. The acceptance matrix is the harder one — that's genuine verification evidence, so it probably can't be auto-satisfied; it mainly needs a clear answer on whose job it is.

[...]

Issue #2058, opened 2026-06-21
tech-debt: decompose the `cli/commands/agent/tasks.py` god-module


stijn-dejongh:
## Problem

`src/specify_cli/cli/commands/agent/tasks.py` is a **god-module: 4525 LOC, maxCC ~178**.

It is one of the **proven crime-scene god-modules** (`merge` / `mission` / `tasks`) the 3.2.x forensic squad identified by metric shift, not by intuition:

- `docs/engineering_notes/3-2-x-goal-corroboration/caacs-delta-robbie.md` (consumer-hotspot table) classes it **HEAT — severe** and notes it was *already a god-module*: SLOC `3041 → 4539`, maxCC `118 → 178` (`+1498` SLOC, `+60` defs), with **47 bugfix commits** in the `v3.1.10..v3.2.0` range — heavy bugfix-commit clustering on this exact surface.

This is the highest maxCC of the three crime-scene god-modules, and the bugfix-commit density confirms it is where defects actually cluster.

## Designation

This module is hereby **designated a god-module**. It will receive a **top-of-file decomposition pointer** comment referencing this issue (matching the convention already applied for #2056 `agent/mission.py` and #1623 `doctor.py`). The **full decomposition is deferred to this ticket** — a research squad should propose cohesive seams (tasks-outline / package-materialization / finalize-validation / dependency-graph wiring / parsing-validation helpers / command-registration) and identify which slice, if any, folds into an active mission versus is deferred here. The decomposition must preserve the public `agent tasks` CLI surface (no command-name or flag changes) and keep each extracted seam independently testable.

## Acceptance (initial, to be refined by research)

- [ ] Top-of-file decomposition pointer added to `agent/tasks.py` referencing this issue.
- [ ] Cohesive seams proposed (outline / package-materialization / finalize-validation / parsing-validation helpers / command-registration).
- [ ] `agent/tasks.py` reduced toward a thin command-registration shim; each extracted seam carries focused tests.
- [ ] No regression to the `agent tasks` CLI contract.

## Relationships

- Parent: #1797 (3.2.0 codebase sanitization — dead-code & LOC reduction)

stijn-dejongh:
**Scope note from mission 01KVMBD6 / PR #2060 adversarial review**

De-god is incomplete: the planning-commit tails in `tasks.py` were NOT migrated to the new `commit_for_mission` router.

**Specific sites to route:**
[...]

PR #2104, opened 2026-06-24
fix: tasks phase deadlocks on protected-main missions — work-package files created in the wrong folder (#2101)

### LINKED ISSUE #2101: Coordination-worktree missions: planning/tasks commands resolve different feature_dir (tasks phase deadlocks; not fixed by #2070)
## Summary

For a **protected-main / `pr_bound` / `topology: coord`** mission, the planning/tasks
subcommands resolve **different feature directories** for the same mission. The commands
that *write* work-package files target the coordination worktree, while the commands that
*validate* them read the main checkout. No single file location satisfies both, so the
`/tasks` phase deadlocks and cannot be completed.

**This is not fixed by #2070.** We updated to current `main` specifically hoping the
"single-authority topology cleanup" would resolve it, and reproduced the bug on that exact
build.

## Environment

- **Exact commit:** `aeb8dfc31a88d2ae779e7b32f5340e90159733e6` (`upstream/main`, 2026-06-23)
- Reported version string: `3.2.2` (note: the version string is identical across recent
  commits, so the SHA above is the only unambiguous marker)
- orchestrator-api contract: `1.1.0`
- Mission: `software-dev`, `pr_bound: true`, `topology: coord`, primary branch protected,
  coordination worktree materialized
- Verified the running binary executes the editable source at the commit above (not a stale
  install) — the split is independent of shell/CWD.

## Background (for context)

When the primary branch is protected, a mission is created PR-bound with `topology: coord`
and a **coordination worktree** is materialized. The mission feature directory then exists
in two places:

- **Main checkout:** `<repo>/kitty-specs/<mission>/`
- **Coordination worktree:** `<repo>/.worktrees/<mission>-coord/kitty-specs/<mission>/`

The planning agent authors and commits artifacts to the **coordination branch** (inside the
coordination worktree).

## What went wrong (user-facing)

Specify and plan worked. During `/tasks`, `map-requirements` **succeeded** — it wrote
`requirement_refs` into each WP and committed them to the coordination branch. The very next
step, `finalize-tasks`, then reported there were **no work packages at all** and every
functional requirement as unmapped. There was no location we could place the WP files that
satisfied both commands: placing them where `map-requirements` wanted made `finalize` blind,
and vice-versa. The phase could not complete.

## Symptoms (reproducible)

For the **same** mission handle, the four commands disagree:

| Command | Resolves to |
|---|---|
| `agent context resolve --action tasks` | **coordination worktree** |
| `agent tasks map-requirements` | **coordination worktree** (only finds WP files there) |
| `agent mission check-prerequisites --include-tasks` | **main checkout** |
| `agent mission finalize-tasks --validate-only` | **main checkout** |

`finalize-tasks --validate-only` against the real mission:

```json
{
  "error": "Requirement mapping validation failed",
  "requirement_refs_parsed": {},
  "dependencies_parsed": {},
  "unmapped_functional_requirements": [ "...all functional requirements..." ]
}
```

`requirement_refs_parsed: {}` / `dependencies_parsed: {}` means **zero WP files were
parsed** — `finalize` globbed `tasks/*.md` on the **main checkout** (scaffold only), while
the actual WP files were on the **coordination worktree**.

## Root-cause hypothesis

There appear to be **two families of feature-dir resolvers that have drifted apart**:

1. **Topology-aware** (returns the coordination worktree when it exists):
   `resolve_action_context(action="tasks")` -> `_resolve_mission_slug` ->
   `resolve_handle_to_read_path` -> `_resolve_existing_for_slug` -> `coord_feature_dir(...)`,
   gated by `_topology_routes_through_coord(stored_topology)`.
   Used by `context resolve` and `map-requirements` (via `resolve_feature_dir_for_mission`);
   `setup_plan` reaches the same surface via `_find_feature_directory`.
   (files: `src/mission_runtime/resolution.py`, `src/specify_cli/missions/_read_path_resolver.py`)

2. **Topology-blind** (always the main checkout):
   `primary_feature_dir_for_mission(...)` (= `get_main_repo_root()/kitty-specs/<slug>`).
   Used by `check-prerequisites` (via `_primary_anchored_feature_dir`) and `finalize-tasks`,
   which then does `tasks_dir.glob("WP*.md")` on the primary path.
   (file: `src/specify_cli/cli/commands/agent/mission.py`)

This looks like a **half-finished migration**. #2087 deliberately moved `check-prerequisites`
onto the **primary** surface to agree with `finalize-tasks`, and #2091 moved identity reads
onto primary. But the *authoring/tasks-read* commands (`context resolve --action tasks`,
`map-requirements`, `setup_plan`) were left on the **coord-aware** resolver. So the commands
that **write** the WP files target the coordination worktree, while the commands that
**validate** them read the main checkout. #2070 reorganized this area but did not bring these
specific commands onto one surface.

## Suggested fix (defer to maintainers on direction)

Converge the lagging commands onto whichever surface is canonical for planning artifacts.
Per #2087/#2091's direction that is the **primary** surface, with commits staged into the
coordination worktree at commit time (the pattern already used by `record-analysis` /
`_stage_finalize_artifacts_in_coord_worktree`):

- Make `context resolve --action tasks`, `map-requirements`, and `setup_plan` resolve
  `primary_feature_dir_for_mission` for reading/writing artifacts.
- Keep their commits routed through `commit_for_mission` so they still land on the
  coordination branch (protected-main safe).
- Add a regression test asserting all four commands return the **same** `feature_dir` for a
  materialized coord mission. `tests/missions/test_surface_resolution_equivalence.py` is the
  natural home.

(We have not verified the write/commit-routing side of this direction end-to-end; that
should be proven with a test rather than assumed.)

## Workaround (verified)

1. Copy the WP files + `tasks.md` from the coordination worktree's `tasks/` into the **main
   checkout's** `kitty-specs/<mission>/tasks/` (and `tasks.md`).
2. **Do not `git add` them on the target branch** (the main checkout sits on the target
   branch; committing there collides when the coordination branch merges in).
3. Run `finalize-tasks`. It reads the main-checkout filesystem and routes its own commit to
   the coordination branch.
4. Delete the uncommitted copies from the main checkout afterward.

This unblocked the phase for us (and correctly surfaced genuine WP-content validation errors,
e.g. `owned_files` paths needing `create_intent`, which are unrelated to this bug).

rayjohnson:
## What happened

We were running a normal mission on a repo where the `main` branch is protected. Spec Kitty handles that by giving the mission its own separate **coordination folder** to work in. Spec and plan went fine. Then the tasks phase got stuck:

- The step that **creates** the work-package files put them in the **coordination** folder.
- But the next steps — the one that **validates** the tasks, and the one that **starts implementation** — looked in the **main** folder, which only had empty scaffolding.

So every tool reported "no work packages found," even though we'd just created six of them. And there was no single folder that made both sides happy: put the files where the creator wanted, and the validator went blind; put them where the validator looked, and the creator couldn't find them. The mission couldn't move forward.

## Why this was hard to pin down

This took us a while, and it's worth being honest about why: the design notes, the actual code, and the tests don't all agree on which folder is the "real" one. We kept concluding the bug was in one step, then finding evidence it was somewhere else, and going back and forth. That's not a knock on anyone — it lines up with the cleanup work already open in #1716.

What finally settled it was to stop looking for a hidden "right answer" and instead look at the one part of the system that simply *has* to work a certain way: the code that **saves and commits** the files. That code only knows how to take files from the **main** folder and copy them into the coordination folder to commit them — and it even checks for changes by looking in the main folder. The steps that validate and implement also read the main folder. In other words, almost everything already assumes: **you author in the main folder, and the coordination folder is just where things get committed.**

The one piece that disagreed was the step that **tells the agent where to create files** — it pointed at the coordination folder. That single mismatch is the whole bug.

## The fix

We pointed the file-creation steps at the same **main** folder everything else already uses:

- `context resolve` now reports the main folder for the authoring actions (specify, plan, tasks, etc.).
- `map-requirements` now reads the main folder too.

We did **not** change the validate, implement, or save steps — they were already correct. That's why the change is small, and a big part of why we think it's right: it makes the odd-one-out match everything else, instead of rewriting the parts that already work.

## Why we think this honors the original intent

We didn't invent a new model — we made the authoring step agree with the model the rest of the code already enforces. If we've misread the intent and the coordination folder was meant to be the authoritative one, the alternative would be to change the save logic to commit *from* the coordination folder instead; we're happy to go that way if a maintainer prefers it. But every working piece pointed in the direction we chose.

One thing worth flagging: there's a test that was supposed to catch this exact bug, but it compared two helper functions instead of actually running the validate step — and those two helpers happened to agree, so the test passed while the real steps disagreed. We've rewritten it to exercise the real behavior.

## Testing

- Full test suite: no new failures (the pre-existing ones are unchanged).
- Ran a real protected-`main` mission end-to-end through the tasks phase and into implementation — no deadlock, no manual workarounds.

## Related, but not fixed here

The `record-analysis` step has a related version of the same folder confusion, plus a separate problem where it refuses to run because of files Spec Kitty itself left uncommitted (#2102). We left it alone for now — it's tangled up with tests that assume the older behavior and deserves its own change once the folder model is confirmed.

Fixes #2101.

### DIFF
--- a/src/specify_cli/cli/commands/agent/context.py
+++ b/src/specify_cli/cli/commands/agent/context.py
@@ -27,6 +27,16 @@ app = typer.Typer(
 
 console = Console()
 
+# #2101: actions whose ``feature_dir`` the agent uses to AUTHOR planning
+# artifacts (spec/plan/tasks/WP files, analysis report). These are authored on
+# the PRIMARY checkout and staged to the coordination branch at commit time, so
+# the reported ``feature_dir`` for these actions must be the primary surface —
+# unlike read-only actions (accept/status) which legitimately resolve the
+# coord-committed surface.
+_AUTHORING_ACTIONS: frozenset[str] = frozenset(
+    {"specify", "plan", "tasks", "tasks_outline", "tasks_packages", "analyze"}
+)
+
 
 def _find_feature_directory(
     repo_root: Path,
@@ -135,8 +145,23 @@ def resolve_context(
             cwd=Path.cwd(),
         )
 
+        payload = context.to_dict()
+        # #2101: planning artifacts are authored on the PRIMARY checkout and only
+        # staged to the coordination branch at commit time (commit_for_mission
+        # stages primary->coord; finalize's change-detection runs in the primary
+        # repo, where .worktrees/ is gitignored). The read-path resolver returns
+        # the coord worktree once materialized — but the step contracts tell the
+        # agent to author at this ``feature_dir``, so for the AUTHORING actions it
+        # must point at the primary checkout, matching where finalize/implement/
+        # check-prerequisites read and where the commit machinery stages from. The
+        # placement/commit target (branch_ref) stays the resolved coordination ref.
+        if action in _AUTHORING_ACTIONS:
+            from specify_cli.missions._read_path_resolver import primary_feature_dir_for_mission
+
+            payload["feature_dir"] = str(primary_feature_dir_for_mission(repo_root, mission_slug))
+
         if json_output:
-            print(json.dumps({"success": True, **context.to_dict()}, indent=2))
+            print(json.dumps({"success": True, **payload}, indent=2))
         else:
             console.print(f"[green]✓[/green] Resolved {action} context")
             console.print(f"  Mission: {context.mission_slug} ({context.detection_method})")
--- a/src/specify_cli/cli/commands/agent/tasks.py
+++ b/src/specify_cli/cli/commands/agent/tasks.py
@@ -379,19 +379,19 @@ def _map_requirements_feature_dir(main_repo_root: Path, mission_slug: str) -> Pa
     slug-only ``mid8_from_slug`` heuristic missed the coord worktree when the
     operator handle carried no mid8 tail, while finalize resolved into it.
 
-    The unified resolver raises ``ActionContextError`` when the directory cannot
-    be located; ``map-requirements`` historically surfaced this as its own
-    ``Mission directory not found: …`` message via an existence guard on the
-    returned path. To preserve that user-facing contract (Risk #1) the typed
-    error is translated back into the non-existent candidate directory so the
-    caller's existing existence guard fires the unchanged message.
+    #2101: anchor on the PRIMARY checkout. Planning artifacts (including
+    ``tasks/WP*.md``) are authored on the primary surface (``context resolve``
+    reports the primary ``feature_dir`` for authoring actions) and only staged to
+    the coordination branch at commit time. ``map-requirements`` reads/writes the
+    WP frontmatter and must therefore see the SAME primary surface ``finalize``,
+    ``check-prerequisites`` and ``implement`` read — the coord-aware
+    ``resolve_feature_dir_for_mission`` (the prior #2064 behaviour) selected the
+    coord worktree, where the commit machinery (which stages primary->coord and
+    change-detects in the primary repo) cannot commit. ``primary_feature_dir_for_mission``
+    is a pure path constructor; a non-existent dir is returned as-is so the
+    caller's existence guard fires the unchanged ``Mission directory not found``.
     """
-    from mission_runtime import ActionContextError
-
-    try:
-        return resolve_feature_dir_for_mission(main_repo_root, mission_slug)
-    except ActionContextError:
-        return candidate_feature_dir_for_mission(main_repo_root, mission_slug)
+    return primary_feature_dir_for_mission(main_repo_root, mission_slug)
 
 
 def _self_review_fallback_option_error(
--- a/tests/specify_cli/test_requirement_mapping_coord_surface.py
+++ b/tests/specify_cli/test_requirement_mapping_coord_surface.py
@@ -1,25 +1,22 @@
-"""#2064 read-surface desync regression: map-requirements and finalize-tasks
-must resolve the WP ``tasks/`` directory through ONE seam-resolved surface.
+"""#2101 read-surface authority: map-requirements and finalize-tasks must resolve
+the WP ``tasks/`` directory on the SAME (PRIMARY) surface.
 
-WP06 / FR-008 / SC-005. The bug: ``tasks.py::map_requirements`` was the lone read
-path on ``resolve_feature_dir_for_slug`` (whose slug-only ``mid8_from_slug``
-heuristic misses the coordination worktree when the operator handle carries no
-mid8 tail), while ``tasks.py::finalize_tasks`` (and the rest of ``tasks.py``)
-resolve through ``resolve_feature_dir_for_mission`` (the
-``resolve_action_context(action="tasks")`` seam, which reads the declared mid8
-from primary ``meta.json``). On a coord topology the two return DIFFERENT
-``tasks/`` directories, so ``map-requirements`` writes/reads one and finalize
-reads the other → spurious ``unmapped_functional_requirements``.
+This supersedes the #2064 framing. #2064 pointed ``map_requirements`` at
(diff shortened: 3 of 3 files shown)


rayjohnson:
**End-to-end validation on a real protected-`main` coordination mission.**

- `finalize-tasks` committed a real diff (`50721f3`) to the coordination branch (`kitty/mission-…`) **only** — the target branch was left untouched. (This is the commit-with-diff path that's awkward to reproduce in a unit test; confirming it live.)
- All **6 work packages** ran implement → review → done with **no manual workaround**: WP files were located on the primary surface, dependency order was honored, and 3–4 lanes ran concurrently throughout.

One unrelated, non-blocking signal worth surfacing (NOT touched by this change): each `review → done` step logs

```
[TRANSITION_REJECTED] Illegal transition: for_review -> done
```

The state machine rejects a direct `for_review -> done`; the orchestrator then re-claims the review lane and the WP completes. It fired on all 6 and recovered every time — so it's cosmetic here, but the loop is papering over an illegal transition on each WP. Flagging for visibility; happy to open a separate issue if useful.

rayjohnson:
**Scope note (known gap).** This PR deliberately stays scoped to the **tasks-phase authoring** surface (`context resolve` + `map-requirements` → main folder) — the part validated end-to-end. Two related steps are **not** covered here and are tracked in #2107: the **plan-phase setup** (`setup-plan`) still creates `plan.md` in the coordination folder, and **accept** reads a mix of surfaces. Both have a different shape than this change — they author on the main folder but must *verify committed-state on the coordination branch* — so they belong in a focused follow-up rather than bloating this one.

stijn-dejongh:
Coordination note: **PR #2106** (`write-surface-coherence`) structurally **Closes #2101** (the same coord-worktree planning/tasks surface-resolution split this quick-fix targets) — it makes the write-side placement kind-aware and re-partitions planning artifacts onto the primary `target_branch` across all write sites (commit_for_mission callers + the `safe-commit`/`append-history` bypass writers + `_planning_commit_worktree`), with the FR-008 protected-primary refusal.

This is **not a request to close #2104** — rather, please **retest the #2101/#2104 deadlock scenario against #2106 once it lands** and assess whether this quick-fix is still needed or is superseded by the structural fix (to avoid a double-fix / merge conflict). Happy to coordinate on sequencing.

rayjohnson:
Thanks for the fast, thorough triage — and for the heads-up on #2106.

Agreed on sequencing. #2106 is the structural cure; our #2104 is a narrow quick-fix for the same tasks-phase deadlock. We'll **retest the #2101 deadlock scenario against #2106 once it lands**, and if it's fully covered we'll **close #2104** rather than risk a double-fix or merge conflict. If you'd rather, we're equally happy to close #2104 now and let #2106 carry it — whatever sequencing is easiest on your side.

A couple of notes that may help:

- The independent repro from @kentonium3 matches ours exactly (same build, same block), so the tasks-phase deadlock should reproduce cleanly for verifying #2106.
- On #2107: glad the `setup-plan` facet lined up with your live repro (setup-plan reading coord after the writes moved to primary), and splitting the accept-gate facet to #2085 makes sense to us.

Thanks again — happy to help verify once #2106 / #2107 land.

rayjohnson:
Retested on latest `main` (now at the 3.2.2 split-brain surface-resolution remediation): #2106 (kind-aware write-surface placement) and #2113 (gate-command read+write completion) have landed, and the coord-surface / write-surface tests are green. The structural fixes supersede this quick-fix, and #2058 has since refactored `tasks.py`, so there's nothing left for #2104 to add.

Closing in favor of the structural fixes. Thanks for the fast, thorough triage — and for folding our repros into the remediation. 🙏

Issue #2103, opened 2026-06-24
Global launcher fails with ModuleNotFoundError: tomli_w — resolves a stale/mismatched Python env


LynnColeArt:
## Problem

The machine-wide launcher at `~/.local/bin/spec-kitty` fails on startup with:

```
ModuleNotFoundError: No module named 'tomli_w'
```

The **same machine's** updated Spec Kitty repo venv imports `tomli_w` without issue. So the global launcher is resolving a **stale or mismatched Python environment** rather than the intended (repo-local / freshly-installed) one.

## How it surfaced

Found while dogfooding a real mission (a forked OpenShot project) against an updated Spec Kitty install. The repo venv was healthy; only the global `~/.local/bin/spec-kitty` entry point was broken.

## Likely cause

The `uv tool`-installed global launcher points at an environment that is missing `tomli_w` (either a stale install that predates the dependency, or the launcher resolving a different interpreter/env than the one that has it). Either the dependency isn't being installed into the launcher's environment, or the launcher isn't pinned to the environment that has it.

## Why it matters

This is an **install-friction** failure: the first command a freshly-set-up user runs can die before doing anything, with an opaque `ModuleNotFoundError`. That's squarely the "easy install on all platforms" concern (and an argument for a self-contained/bundled runtime, which would eliminate this whole env-resolution failure mode).

## Relation to existing work

Distinct from the raw-import failure noted in #2007 (item 9: `python -c "from specify_cli..."` → `ModuleNotFoundError: specify_cli`). That's a raw interpreter invocation; this is the **installed global launcher** resolving the wrong env. Filing separately; cross-linking #2007 as the install/DevEx training-bug epic.

## Suggested fixes (for discussion)

- Ensure `tomli_w` (and the full runtime dependency set) is declared/installed into the launcher's environment.
- Have the launcher pin to / resolve its own managed environment deterministically, rather than whatever Python it currently lands on.
- A clear preflight/self-check that names a missing-runtime-dep situation instead of a bare traceback.

stijn-dejongh:
Parented under **#1945** (ToolSurfaceContract epic) — the global launcher resolving a stale/mismatched Python env is an install/tool-surface integrity gap. Triage by planner-priti (Op 01KVW11HZ28KTNBET2ST5DVSMP).

robertDouglass:
[...]

PR #2126, opened 2026-06-25
fix: accept gate resolves mid8/ULID handles for its primary-partition reads

kentonium3:
Fixes #2122.

## What was broken

`collect_feature_summary` (the accept gate) passed the **raw mission handle** to its PRIMARY-partition reads — `_iter_work_packages` → `_wp_tasks_read_dir`, and `_planning_read_dir`. For a mid8 / ULID / numeric handle that composes a nonexistent `kitty-specs/<handle>` slug dir, so the gate fails with `AcceptanceError: Feature '<handle>' has no tasks directory`. The summary's identity anchor and status reads were already handle-correct; only the primary-partition reads weren't.

## The fix

Key the primary-partition reads on the canonical **primary slug**, derived from the already-resolved `_primary_anchor_feature_dir` (`feature_dir.name`) rather than the raw handle. That name is deliberately *not* recomposed from the coord read-dir (it can differ for backfilled legacy missions whose primary dir lacks the `-<mid8>` suffix). The STATUS reads stay coord-aware on the raw handle, preserving the per-partition split (C-002).

Targeted and caller-local: this does **not** touch the shared `resolve_planning_read_dir` seam — the broader handle-canonicalization cure at that entry point remains the deliberate follow-on. Additive; no change to the shared resolver.

## Testing

- The pre-existing (test-first) regression `test_collect_feature_summary_anchors_feature_dir_on_primary_for_mid8_handle` flips from red to green; full acceptance suite green (56), `ruff` + `mypy` clean.
- **Validated on real mission state:** against a clone of a real coordination-topology mission (mid8 handle `01KVXNDC`), pristine `main` errors `Feature '01KVXNDC' has no tasks directory`; with this fix the gate resolves to the real primary mission dir and reads its work packages.

## Note

This is the pre-existing red check on #2121 (`integration-tests-core-misc` → `test_collect_feature_summary_anchors_feature_dir_on_primary_for_mid8_handle`); merging this greens that shard.

---

*AI-assistance disclosure: diagnosed, implemented, and tested with AI assistance (Claude Code), with the completed code independently reviewed via Codex and validated on real mission state; all changes are human-reviewed.*

Issue #2122, opened 2026-06-25
v3.2.2 regression: accept gate (collect_feature_summary) breaks for a mid8-handle mission — handle passed to resolve_planning_read_dir as a slug


stijn-dejongh:
**Confirmed regression shipped in v3.2.2** (red on `main`/`f853934b6`, NO_COLOR, not a flake): `tests/specify_cli/test_acceptance_regressions.py::test_collect_feature_summary_anchors_feature_dir_on_primary_for_mid8_handle` fails with `AcceptanceError: Feature '01ABCDEF' has no tasks directory at .../kitty-specs/01ABCDEF/tasks`.

**Two sites in `src/specify_cli/acceptance/__init__.py` pass `feature` (which can be a mid8 HANDLE) into `resolve_planning_read_dir(repo_root, feature, kind=…)` as if it were a resolved SLUG:**
- `_iter_work_packages:406` → `_wp_tasks_read_dir(repo_root, feature)` → joins `kitty-specs/<mid8>/tasks` literally
- `normalize_feature_encoding:823` → same `resolve_planning_read_dir(repo_root, feature, …)` pattern

**Root cause:** the gate-read mission accept-gate fix (#2113, commit `b4085eaad`) switched from `resolve_feature_dir_for_mission` (handle-aware: resolves mid8→slug) to `resolve_planning_read_dir` (slug-only) without first resolving the handle. **Why it slipped:** the new fixtures used a full slug (passed); the existing `..._for_mid8_handle` test caught it but lives in the `specify-cli-rest` CI shard that was conflated with the pre-existing `test_standalone_tasks_cli_canonical` flake, so it rode through #2113's merge into v3.2.2.

**Severity:** real but narrow — the accept gate fails when invoked with `--mission <mid8>` (a supported selector); full-slug invocations work.

**Fix:** resolve the handle → canonical slug BEFORE the `resolve_planning_read_dir` calls at both sites (restore the handle-awareness `resolve_feature_dir_for_mission` had); existing `..._for_mid8_handle` test is the red-first guard. Check whether any other gate-read site has the same handle-vs-slug bug. Relates #2113 / epic #1716 / #1868 (mission-identity).

stijn-dejongh:
[...]

PR #2132, opened 2026-06-25
fix(tool-surface): lazy-import tomli_w so a stale env can't brick the CLI (#2103)

### LINKED ISSUE #2: spec-kitty merge fails with worktrees due to branch checkout deadlock
## Problem

`spec-kitty merge` is fundamentally broken when using git worktrees, which is the recommended workflow in the documentation.

## The Three-Way Deadlock

1. **Run from worktree → Error:**
   ```
   fatal: 'main' is already used by worktree at '/path/to/main-repo'
   ```
   The command tries to `git checkout main` but main is locked in the parent repo.

2. **Run from main repo (on main branch) → Error:**
   ```
   Error: Already on main branch. Switch to a feature branch first.
   ```
   The command demands you be on the feature branch before merging.

3. **Try to checkout feature branch in main repo → Error:**
   ```
   fatal: '001-feature' is already used by worktree at '/path/to/.worktrees/001-feature'
   ```
   Can't checkout the feature branch because it's locked in the worktree.

## Repository Structure

```
my-project/                      # main branch checked out
├── .worktrees/
│   └── 001-feature/            # 001-feature branch checked out
└── ...
```

## Expected Behavior

`spec-kitty merge` should work seamlessly with worktrees by either:
1. Detecting it's in a worktree and automatically operating on the main repo
2. Using `git merge <branch>` without requiring a checkout (you can merge from any branch)
3. Providing clear instructions: "Run this command from the main repo"

## Current Behavior

The command has no escape hatch - it's impossible to merge when using worktrees.

## Reproduction

```bash
# Setup
cd /path/to/main-repo
spec-kitty specify "feature"
cd .worktrees/001-feature
# ... do work ...

# Try to merge - FAILS
spec-kitty merge

# Try from main repo - FAILS  
cd /path/to/main-repo
spec-kitty merge
```

## Workaround

Manual merge:
```bash
cd /path/to/main-repo  # Already on main
git merge 001-feature
git push
git worktree remove .worktrees/001-feature
git branch -d 001-feature
```

## Environment
- Following documented worktree workflow from merge.md template

LynnColeArt:
Addresses #2103.

**Why this one jumped the queue:** it's been a daily annoyance on my machine — the global launcher dies on `ModuleNotFoundError: tomli_w` on *every* command, so the first thing I run is a traceback. It's been forcing me to drive everything through `uv run spec-kitty`. Worth fixing.

## Root cause
`tomli-w` *is* a declared runtime dependency (`pyproject.toml`), so this isn't a missing-dep bug — it's that `import tomli_w` sat at the **top of `codex_renderer.py`**, which gets pulled in during command registration. So an install whose env predates the dependency (the stale `uv tool` launcher case in #2103) crashes the entire CLI at startup, before any command runs, with an opaque traceback.

`tomli_w` is used in exactly one place — the Codex profile render path (`render()`).

## Fix
- Move `import tomli_w` out of module scope and into `render()`.
- Turn the failure into an actionable message ("reinstall/upgrade spec-kitty") instead of a bare `ModuleNotFoundError`.

Result: the CLI runs normally for every command even if `tomli_w` is missing from the env; only Codex profile rendering errors, and it tells you why.

## Tests
- `test_codex_renderer_does_not_import_tomli_w_at_module_level` — regression guard: the module must not bind `tomli_w` at import time.
- `test_codex_render_missing_tomli_w_raises_actionable_error` — `render()` fails with guidance, not an opaque traceback.

55/55 renderer tests pass; ruff + mypy clean on the touched files.

## What this does *not* do
It doesn't change how `uv tool` resolves the launcher's environment — a genuinely stale install should still be upgraded. This fix just stops a missing optional-path dependency from being catastrophic. If you'd rather also pin/verify the launcher env (suggested fix #2 in the issue), happy to do that as a separate change.

### DIFF
--- a/src/specify_cli/tool_surface/profiles/codex_renderer.py
+++ b/src/specify_cli/tool_surface/profiles/codex_renderer.py
@@ -17,8 +17,6 @@ from __future__ import annotations
 
 from pathlib import Path
 
-import tomli_w
-
 from charter.profiles import AgentProfile
 
 # Native format identifier (stable string recorded in the manifest).
@@ -72,7 +70,25 @@ class CodexProfileRenderer:
         )
 
     def render(self, profile: AgentProfile) -> str:
-        """Return valid TOML text for the Codex per-project agent format."""
+        """Return valid TOML text for the Codex per-project agent format.
+
+        ``tomli_w`` is imported lazily here rather than at module load: it is a
+        declared runtime dependency but is only needed on this Codex-rendering
+        path, so a stale/mismatched install that is missing it must not brick
+        every CLI command at import time (issue #2103). The guard turns the
+        opaque ``ModuleNotFoundError`` into an actionable message.
+        """
+        try:
+            import tomli_w
+        except ModuleNotFoundError as exc:
+            raise ModuleNotFoundError(
+                "tomli_w is required to render Codex agent profiles but is not "
+                "available in this environment. spec-kitty declares it as a "
+                "runtime dependency, so this usually means a stale install — "
+                "reinstall or upgrade spec-kitty (e.g. `uv tool upgrade "
+                "spec-kitty`) so its declared dependencies are present."
+            ) from exc
+
         description = profile.description or profile.purpose or ""
         doc: dict[str, object] = {
             "name": profile.name or profile.profile_id,
--- a/tests/specify_cli/tool_surface/profiles/test_renderers.py
+++ b/tests/specify_cli/tool_surface/profiles/test_renderers.py
@@ -259,6 +259,41 @@ def test_codex_renderer_profile_with_empty_description_falls_back_to_purpose() -
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


stijn-dejongh:
**Review: ✅ SHIP. Thank you, @LynnColeArt** — a clean, correctly-scoped fix for #2103. 🙏

Moving `import tomli_w` to a lazy import inside the Codex-rendering path (with a try/except that turns the opaque `ModuleNotFoundError` into an actionable 'reinstall/upgrade' message) is exactly right: a declared-but-stale runtime dep on a single rendering path should never brick *every* CLI command at module import. Narrow, well-documented, with a test.

The only CI red was `test_collect_feature_summary_anchors_feature_dir_on_primary_for_mid8_handle` — the **pre-existing #2122 test** that was red on your (older) base. I rebased the branch onto current `main` (now carrying #2126's #2122 fix); CI should come back green. Ready to land once it confirms.

PR #2137, opened 2026-06-25
fix(run_tests): detect host python interpreter (#2097)

### LINKED ISSUE #2097: Recommend: detect host python interpreter in shell scripts (run_tests.sh breaks on python3-only hosts)
## Problem

`run_tests.sh:11` invokes a bare `python`:

```sh
python -m pytest tests/ "$@"
```

On Ubuntu/Debian — which are python3-only by default and ship no `python` symlink — this fails with `python: command not found`. It's the same class of bug we just fixed in `spec-kitty-saas` (Priivacy-ai/spec-kitty-saas#324), where a host-side `python` heredoc was failing the local sync canary on python3-only machines.

A quick sweep found `run_tests.sh` as the clear executable instance in this repo. (The other `python` references — language scopes in doctrine YAML, agent profiles, and `verify_occurrences.py` command strings in `kitty-specs/` — are config/doc text, not host invocations, so they're out of scope here.) Worth a second look at CI workflows and any git hooks for other bare-`python` entry points.

## Reasoning

Detecting the interpreter is correct on all three setups (`python`-only, `python3`-only, both), where hard-coding either one breaks one of them. We landed on **prefer `python`, fall back to `python3`** in the saas fix: preferring `python` honors an activated venv/pyenv, and the `python3` fallback covers the bare Ubuntu case.

## Proposed change

Apply the same small resolver to host-side shell scripts (starting with `run_tests.sh`):

```sh
PYTHON_BIN="${PYTHON_BIN:-$(command -v python || command -v python3 || true)}"
if [[ -z "${PYTHON_BIN}" ]]; then
  echo "error: no python or python3 interpreter found on PATH" >&2
  exit 1
fi
# ...
"${PYTHON_BIN}" -m pytest tests/ "$@"
```

This is a recommendation, not a merge-ready change — filing it so the reasoning and the `spec-kitty-saas#324` precedent are captured for the team. Happy to put up the PR if it's wanted.

shashankcm95:
On python3-only hosts (Ubuntu/Debian, no `python` symlink), `run_tests.sh:11` failed with `python: command not found`. Resolve the interpreter once, preferring `python` (honors an activated venv/pyenv) and falling back to `python3`, erroring if neither is on PATH. Same resolver as spec-kitty-saas#324.

Closes #2097.

### DIFF
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,11 +1,9 @@
 #!/bin/bash
-# Test runner for the worktree that ensures correct imports
+# Test runner for the worktree (delegates to the canonical uv-managed test path).
 
-# Change to worktree directory
+# Change to the worktree directory so uv resolves the project venv.
 cd "$(dirname "$0")"
 
-# Add the src directory to PYTHONPATH
-export PYTHONPATH="${PWD}/src:${PWD}:${PYTHONPATH}"
-
-# Run tests with explicit path settings
-python -m pytest tests/ "$@"
+# Run the suite via uv (the canonical authority; see the Makefile `test` target and
+# CONTRIBUTING). uv sync's editable install handles src/ imports, so no manual PYTHONPATH.
+uv run pytest tests/ "$@"


stijn-dejongh:
Thank you, @shashankcm95 — the underlying defect is real and your fix resolves it. 🙏 We reviewed it with two lenses (architect-alphonso on install-plan adherence + debugger-debbie live across host scenarios).

**Live (debbie): the resolver is functionally correct** — witnessed across PATH-shim sandboxes: python3-only host → resolves `python3` (the #2097 bug is fixed); venv host → prefers `python` (= the venv interpreter); neither present → clean `error: ... exit 1`; `PYTHON_BIN` override wins; `$@` passthrough intact.

**Install-plan adherence (alphonso): a few changes would align it with our canonical setup.** Our intended dev/test path is a **uv-managed venv** (`CONTRIBUTING.md`: `uv sync` + `pytest`; `Makefile`: `uv run pytest`; `requires-python >=3.11`), and `run_tests.sh` is an off-canonical hand-run convenience helper (no Makefile/CI/docs caller). Two recommended adjustments:

1. **Flip the order to `python3`-first** (`command -v python3 || command -v python`): it matches the repo's other interpreter scripts (`scripts/prepare-mutmut-env.sh`, the handoff `generate.sh`), honors the venv equally, and avoids picking a **legacy python2 `python`** on bare hosts (which fails `>=3.11`) — debbie confirmed prefer-`python` does select python2 in that case.
2. **Prefer delegating to `uv run pytest tests/ "$@"`** (our canonical test authority) over interpreter-guessing — and drop the `spec-kitty-saas#324` citation, which is a cross-repo resolver rather than an in-repo precedent.

Neither is a functional blocker — the fix works as-is. These just bring it into line with the intended installation/run plan. If you'd prefer, a one-line `uv run`-based version is the cleanest landing. Thanks again for catching the python3-only-host break.

shashankcm95:
Thanks for the thorough review, and good catch on the python2 trap. You're right that prefer-`python` selects a legacy `python2` on a bare host and breaks `requires-python >=3.11`.

I've taken the cleaner landing you suggested rather than flip the resolver order: `run_tests.sh` now delegates to `uv run pytest tests/ "$@"`, matching the Makefile `test` target and CONTRIBUTING as the canonical authority. That drops the interpreter-guessing entirely (no python2 risk), and uv's editable install handles the `src/` imports so the manual `PYTHONPATH` is gone too. `$@` passthrough is preserved, and I've dropped the cross-repo spec-kitty-saas#324 citation.

Thanks again.

stijn-dejongh:
✅ **Perfect — this is the preferred fix.** Delegating to `uv run pytest tests/ "$@"` brings it fully into line with our canonical uv-managed test path (the Makefile `test` target + CONTRIBUTING), drops the interpreter-guessing entirely (so the python/python3 order + legacy-python2 concerns are moot), and correctly removes the manual `PYTHONPATH` since `uv sync`'s editable install handles src/ imports. Verified live: `uv run pytest tests/ --collect-only` resolves the venv and collects the full suite. Thanks @shashankcm95 — landing.

Issue #2160, opened 2026-06-26
Coord topology: unify artifact authority for task/status surfaces during the implement/review loop


stijn-dejongh:
## Root problem

In the coord-topology execution model the **writer** and the **validator/committer** of key artifacts disagree on which physical location is authoritative. Two distinct manifestations are active P0 blockers today:

- **`safe_commit` status-write authority split** (#2155): `safe_commit` is invoked from the repo root but refuses to stage status writes whose paths resolve inside the coord worktree. Status transitions complete in memory but are never committed, leaving the event log permanently behind.
- **`tasks.md`/`mark-status` vs `move-task` validation split** (#2154): `mark-status` writes and commits the `tasks.md` inside the coord worktree, but `move-task` validates against the planning-repo copy. Every WP blocks on "unchecked subtasks" because the validator never sees the write.

Both defects share the same structural root: there is no single canonical authority path that all writers, validators, and committers agree on at runtime.

### Seed children (live P0 — original scope)
> ✅ Linked as native sub-issues of this epic. **Core authority split RESOLVED via PR #2181.**

- [x] #2155 — `safe_commit` refuses to stage coord-owned status writes from repo root
- [x] #2154 — `mark-status` writes coord-worktree `tasks.md` but `move-task` validates planning-repo copy

### Newly added children (coord-authority cluster)
> ⚠️ GitHub enforces a single-parent constraint on native sub-issues. Issues with a listed native parent cannot be simultaneously parented here — they remain tracked as checklist references. #2183 / #2185 / #2186 / #2187 are linked as native sub-issues of this epic.

**Read-loop cluster (implement/review) — resolved via PR #2194:**
- [x] #2115 — implement-loop task-read twin (current native parent: #1716)
- [x] #2140 — is_committed spec-read (current native parent: #1878; also child of another epic by design)
- [x] #2183 — Phase 2: `is_def_use_canonical` / `_canonicalize_bare_modern_handle` fold seam (native sub-issue)

**Out-of-loop cluster (merge / lanes / identity) — resolved via PR #2212:**
- [x] #2185 — route merge/+lanes/ PRIMARY reads off coord-aware resolvers (native sub-issue)
- [x] #2186 — route/guard meta.json identity reads off coord-aware resolvers (native sub-issue)
- [x] #2187 — route agent_utils/status.py show_kanban_status tasks-read off coord (native sub-issue)

**Checklist references only (single-parent constraint; other epics hold native parentage):**
[...]

Issue #2158, opened 2026-06-25
Dead-symbol gate parser bug: _extract_all_literal blinds 57 modules / 117 dead symbols (split from #2049)


MOES-Media:
## Context
Split out of #2049 (ratchet allowlist burn-down). During #2049 implementation, fixing the `_extract_all_literal` parser bug in `tests/architectural/test_no_dead_symbols.py` proved to have a far larger blast radius than the burn-down mission could absorb, so it is deferred here.

## The bug
`_extract_all_literal` (≈line 910 of `tests/architectural/test_no_dead_symbols.py`) walks a module AST for its `__all__`. In the `ast.AnnAssign` branch, when the target is **not** `__all__`, `value` stays `None` and the code falls through to `if value is None: return frozenset()` — so a module whose first top-level node is a non-`__all__` annotated assignment (e.g. `_X: int = ...`) is treated as having an **empty** `__all__`, and the dead-symbol gate inspects **none** of its public symbols.

**Fix**: in the `AnnAssign` branch, `continue` when the target isn't `__all__`; only return the empty frozenset for an `__all__` AnnAssign declared without a value (`__all__: list[str]`).

## Why it's its own mission (the blast radius)
Applying the fix un-blinds **~57 modules** across `charter.*`, `kernel.*`, `doctrine.*`, `glossary.*`, `mission_runtime.*`, `runtime.next.*`, `specify_cli.*`, surfacing **~117 pre-existing dead symbols** the gate was silently hiding. Each needs triage (wire a real caller, remove from `__all__`, delete, or justifiably allowlist). That is a large, standalone burn-down — bundling it into #2049 would have **grown** `category_b_grandfathered_legacy` ~286→392 (the opposite of #2049's purpose).

## Scope
1. Fix `_extract_all_literal` (+ a focused unit test: non-`__all__` AnnAssign followed by a real `__all__`).
2. Triage the ~117 newly-surfaced dead symbols — prefer wiring/removal over allowlisting; only grandfather with justification where genuinely deferred.
3. Net goal: the gate inspects the previously-hidden modules **without** a large net allowlist growth.

## Cross-reference
- Split from #2049 (`shrink-ratchet-allowlists`), which delivers the clean allowlist removals without this fix.

stijn-dejongh:
**Triage — Parent assignment**

Parent: #1797 (Epic: 3.2.0 codebase sanitization — dead-code & LOC reduction) via its parent #391.

[...]

PR #2134, opened 2026-06-25
refactor(2056): decompose agent/mission.py god-module

MOES-Media:
## Summary

Decomposes the `agent/mission.py` god-module (#2056, parent #1797). Behavior-preserving refactor: public CLI surface byte-identical, each seam independently testable, `maxCC <= 15`.

- `mission.py`: **4125 → 362 LOC**.
- `finalize_tasks` (1227 LOC) decomposed into phase helpers coordinated by dataclasses (`_BootstrapState`, `_DependencyResolution`, `_CommitOutcome`); side-effect ordering, exception propagation, and the INV-6 zero-mutation-in-validate-only invariant all preserved.
- `_planning_commit_worktree` / `_resolve_planning_placement` / `_stage_finalize_artifacts_in_coord_worktree` relocated to `coordination/commit_router.py`, with `tasks.py` repointed to the new home (cross-file move, exercised by `test_commit_router_planning_residue.py`).
- One-way leaf imports (INV-8) enforced by an AST no-CLI-import test.

## Type-checking

The canonical CI mypy gate is the whole-tree `mypy --strict src/specify_cli src/charter src/doctrine`. The decomposition takes it from a **46-error baseline to `Success: no issues found`** — including dropping the transitional `cast(...)` calls that an adversarial pre-PR review flagged as `[redundant-cast]` under the whole-tree gate (they were only needed under a non-gate single-file invocation; the whole-tree gate resolves the real signatures).

## Safety net

WP01 golden CLI characterization test + phase tests (41 passing) pin behavior. Adversarial review confirmed the golden contract has real teeth (bidirectional set-equality over the live Click tree) and the WP08 relocation resolves and is exercised.

## Sonar / follow-ups

None outstanding. Mission artifacts under `kitty-specs/decompose-mission-god-module-01KVXHF8/`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2070, opened 2026-06-22
Mission B — CommitTargetKind type eradication + richer-API resolve_context_for_mission adoption (behavior-neutral follow-on to #2069 seam)


stijn-dejongh:
**Follow-on / blocked-by:** the landing of mission `single-planning-surface-authority-01KVPR00` (the MissionTopology SSOT seam, design driver #2069). Do not start until 01KVPR00 lands.

## What

The behavior-neutral follow-on to the MissionTopology SSOT seam (#2069) landed by mission `single-planning-surface-authority-01KVPR00`. Two parts:

### 1. `CommitTargetKind` TYPE eradication
Delete the `CommitTargetKind` enum (`src/mission_runtime/context.py:51`) and its ~143 value-literal references:
- ≈63 `CommitTarget(ref=…, kind=…)` constructions
- ≈24 imports / re-exports
- ≈56 test references

across ~41 files.

After the seam mission, the 9 `.kind is COORDINATION` DECISION sites already route through the `routes_through_coordination(target)` predicate (so `.kind` is no longer read to decide). This ticket removes the now-vestigial type and its constructions / imports / tests.

### 2. Richer-API adoption
Migrate the **14** real call sites of `resolve_placement_only` (6) + `resolve_action_context` (8) to call `resolve_context_for_mission(mission_id, topology)` explicitly. These call sites currently pass identity handles, not topology; they become correct UNCHANGED once the seam mission retires the two derivations, so this is incremental adoption of the topology-explicit API over an already-correct door (the #2065 read-side strangler pattern) — NOT a correctness fix.

## Why behavior-neutral / why a separate mission

The seam mission (01KVPR00) closes the coord/primary desync death-spiral entirely: it retires BOTH live `coordination_branch is None ⇒ FLATTENED` derivations — `resolution.py:705-718` AND `runtime_bridge.py:144-211` — and routes all 9 decision sites through stored topology.

This follow-on carries zero correctness gain; it is mechanical cleanup + API uptake, deferred so the focused seam mission stays reviewable.

Sizing by architect-alphonso (2026-06-22).

## Guard for safety

The differential-equivalence gate (`tests/missions/test_surface_resolution_equivalence.py`, extended by 01KVPR00 with the pure stored-topology cell) makes the eradication sweep verifiable.

## Cross-references (do NOT parent under a meta rollup)

- Design driver: #2069
- Epic facets: #1716, #2007
- Execution-context epic: #1619
[...]

PR #2172, opened 2026-06-26
Enforce CORE↛INTEGRATION boundary in-place (#614)

### LINKED ISSUE #614: Separate integration boundaries inside src/specify_cli from core domain logic

MOES-Media:
Closes #614.

Enforces the **CORE ↛ INTEGRATION** boundary inside `src/specify_cli` **in-place** (no physical extraction — that's deferred per ADR 2026-05-11-1) by inverting the 3 verified leaks and adding an AST-scan enforcement test.

## What changed (11 src files)
- **Leak #1** (`core/mission_creation.py` → `sync`/`tracker`) inverted: dossier sync via `status.fire_dossier_sync`, SaaS fan-out via the existing `emit_mission_created_local` lifecycle path, origin-binding moved verbatim to `tracker/origin_consumer.py` + registered as an observer (new `core/adapters.py`). `mission_creation.py` now has **zero** sync/tracker imports.
- **Leak #3** (`invocation/propagator.py` → `sync`) inverted via new `invocation/adapters.py` registry (safe-degrade to None when unregistered; sync registers concretes at startup).
- **Leak #2** (`readiness/coordinator.py` → `saas.rollout`) — a pure feature-flag read — kept with a single documented allowlist exemption.
- New `tests/architectural/test_integration_boundary.py` — AST-walk scan (catches module/`TYPE_CHECKING`/lazy imports), injection-proof, 3-field diagnostics. Also fixed a WP02-introduced `status`-facade violation (`fire_dossier_sync` now exported via the `status` facade).
- ADR `architecture/3.x/adr/2026-06-26-1-core-integration-boundary.md` documents the rule, sets, exemption, and deferred scope.

## Review
5 WPs, each opus-reviewed; post-assembly **mission-review** (FR-by-FR PASS, projections verified behavior-preserving — dossier/SaaS/origin all fire, single `status.events.jsonl` write, daemon sync fires once for MissionCreated only) + a **behavior-preservation squad lens (clean)**.

## Non-blocking notes
- FR-005 (collapse duplicate `emit_mission_created`) satisfied via a single observer-driven path rather than literal deletion (both defs kept as integration-internal API; documented).
- Cosmetic: the boundary-test allowlist uses prefix-`startswith` without a dot boundary (single-entry, harmless); AST scan doesn't see dynamic `importlib` imports (none exist in CORE today).


stijn-dejongh:
@MOES-Media : parking this untill a few coordination-focussed missions land, then a careful rebase will be in order.

stijn-dejongh:
@MOES-Media — applied **`pr:needs-revision`** (unresolved review findings to address before merge) and **`pr:needs-refresh`** (branch drifted from `main`; rebase before landing). Heads-up, not a gate.

A three-lens coherence check (architecture / planning / patterns) against recent merges + in-flight work came back **positive on the big question and land-independent** — with a few in-place fixes first.

**✅ Coherent — land independently of #2173.** The concern was that this adapter-inversion would collide with #2173 (infra-logic separation via *ports*). It doesn't: this governs **package import-direction** (CORE ↛ INTEGRATION) while #2173 injects **infra ports into pure core** (FS/Clock/Git/Resolver) — orthogonal axes. The new `core/adapters.py` + `invocation/adapters.py` reuse the **established `status/adapters.py` seam** (not a new mechanism), and #2173 explicitly *drops* the SaaS port ("already seamed via `fire_saas_fanout`"), i.e. blesses the very seam you extend. **Zero file collision** with the in-flight coord-authority missions (#2160/#2194/#2185/#2186 — disjoint surfaces). Nice work; the boundary even hardens what #2173's port work will lean on.

**Before merge:**
1. **Rebase onto current `main`** (~15 behind). The 3 leaks are verified still-live, but post-branch merges reshaped sibling files — **#2158** (dead-symbol gate, `sync/__init__.py`), **#2134** (status decompose, `status/__init__.py`), lifecycle/topology #2070/#1793. FR-005 keeps both `emit_mission_created` defs as observer-invoked-only — exactly what the #2158 dead-symbol ratchet scrutinizes — so re-green CI after rebase, and **assert MissionCreated daemon-publish + dashboard-sync fire exactly once** (no double-fire/drop) against main's rewired lifecycle path.
2. **`consume_pending_origin` silent-skip** (`core/adapters.py`): it returns the same "nothing" for *no pending origin* **and** *no consumer registered* — benign for best-effort SaaS/sync, but **correctness-adjacent for origin-binding** (a future `create_mission_core` entrypoint that forgets the `import specify_cli.tracker` side-effect would silently drop a staged ticket binding, and the gate guards imports, not registration timing). Fail loud on cold-miss when a pending-origin file actually exists (aligns with #2173's "fail-closed-loud" doctrine).
3. **The new arch-gate re-rolls the centralized scanner**: `test_integration_boundary.py` re-implements its own `_collect_imports` + `rglob` + `ast.parse` instead of the session-scoped `src_source_tree()` fixture 5 gates already share; the self-mutation test checks a *synthetic string* rather than a real injected on-disk violation (so it'd miss an enforcement-loop regression); and there's no `len(allowlist) <= 1` count-ratchet. Consume the shared scanner, drive the injection through the real scan, add the count-ratchet.
4. **Hygiene:** the new ADR doesn't cross-reference #2173/#1619 — add a one-line "Relationship to #2173" axis-split note; and the Leak#2 allowlist exemption (relocate `is_saas_sync_enabled`) has no issue number — file a follow-up so it can't ossify.

Suggested sequencing: **rebase once #2194 has landed** (it's landing soon and touches main heavily), so you rebase onto the final surface. Drop both labels when revised + rebased.

MOES-Media:
@stijn-dejongh — revised and rebased onto current `main`. All four "before merge" findings addressed; new head is `e4021c5cc`.

### Findings
1. **Rebase + behavior preservation.** Rebased onto current `main` (resolved the `architecture/3.x/adr` → `docs/adr/3.x` rename). New `tests/specify_cli/core/test_mission_creation_fire_once.py` drives the real `status/adapters` registry and asserts MissionCreated **dashboard-sync (dossier)** and **daemon-publish (lifecycle SaaS fan-out)** each fire **exactly once** — no drop, no double-fire on resume — against main's rewired lifecycle path.
2. **`consume_pending_origin` cold-miss.** Now fails loud (`PendingOriginConsumerNotRegisteredError`) when a staged `.kittify/pending-origin.yaml` exists but no consumer is registered (would otherwise silently drop a ticket binding); benign no-op preserved when nothing is staged. Tests added.
3. **Arch-gate hardened (kept).** `test_integration_boundary.py` now consumes the shared session-scoped `src_source_tree()` cache (no independent re-walk/re-parse), drives the injection sanity-check through a **real on-disk file** via the *same* `_scan_trees` function the gate uses (so an enforcement-loop regression is caught, not just a re-implemented copy), and adds a `len(ALLOWLIST) <= 1` count-ratchet. Live red-check confirmed a real injected CORE→INTEGRATION leak fails the gate.
4. **Hygiene.** ADR gains a "Relationship to #2173" axis-split note (referencing #2173/#1619); the Leak #2 `saas.rollout` allowlist exemption now references follow-up issue #2252 (relocate `is_saas_sync_enabled`), in both the ADR and the test allowlist rationale.

### Re-green after rebase
New gates landed on `main` during the ~95-commit gap; this PR's own diff tripped them, now fixed: WP-test markers, moved the git-using tracker observer test to its canonical home `tests/sync/tracker/` (`[integration, git_repo]`), dropped a `/tmp/` literal, re-keyed the single-mission-surface-resolver allowlist (337→336, from the Leak #1 import removal), repointed a stale `emit_mission_created` patch onto the canonical `status` facade, and added ADR frontmatter + page-inventory entry + census bump (117→118).

### CI
Green across the board — all six `integration-tests-core-misc` shards, `docs-freshness`, **`diff-coverage`**, and the full fast/integration suites.

**One remaining red is pre-existing and not from this PR:** `integration-tests-sync` → `test_mission_create_json_strict_when_sync_skips_ingress`. It fails **identically on a clean `origin/main` worktree** (which doesn't contain this PR's changes) with `Not authenticated: no valid access token` — i.e. fork-PR CI lacks the SaaS auth/secrets the base-repo CI has. I deliberately did not touch it; hacking it green would mask a main-level/auth-environment issue rather than fix anything in this PR. Happy to file a separate tracking issue if you'd like.

Believe this clears `pr:needs-revision` + `pr:needs-refresh` — over to you to drop them.

stijn-dejongh:
@MOES-Media — we're picking this one up now and intend to land it. Given it's an architectural-boundary change (#614), we're running an architecture-led, profile-loaded review squad on it. We've rebased it onto current `main` (it had drifted — resolved one conflict in `tests/docs/test_adr_content_invariance.py`: the ADR census now reconciles to 119 with your new integration-boundary ADR on top of the two that landed since you opened this). We'll handle the rebase + merge via maintainer-edit, so nothing is needed on your side. We'll follow up on this thread with the aggregated review verdict.

stijn-dejongh:
@MOES-Media — **reviewed, rebased, and landing. Thank you for this.** An architecture-led, four-lens profile-loaded squad cleared the CORE↛INTEGRATION boundary change:

- **Boundary enforcement (lead arch) → ALIGNED.** The AST-scan gate is sound and injection-proof (catches module / TYPE_CHECKING / lazy imports; dot-bounded; non-vacuous). All three leaks are genuinely inverted — exactly one INTEGRATION import remains in CORE (the documented, lazy, count-ratcheted `saas.rollout` feature-flag exemption). ADR content aligned.
- **Consolidation → CLEAN-INVERSION.** Each side-effect has one canonical path; the adapter registries are one consistent idiom; the FR-005 "two defs" decision is a sound single-seam.
- **Behavior preservation → PRESERVED (434 tests).** Dossier sync, SaaS fan-out, and origin-binding all still fire exactly once; one `status.events.jsonl` write. Note: the SaaS fan-out was actually *double-queued* on main before this — your change collapses it to one, a correctness improvement worth calling out.
- **Adapter safe-degrade → SAFE.** `core/adapters.py` fails loud when a staged origin binding would be dropped; the `invocation/adapters.py` degrade-to-None is by-design for additive SaaS events.

**We rebased onto current `main` and folded the squad's findings for you** (force-pushed to your branch via maintainer-edit — nothing needed on your side):
- Fixed 3 ADR cross-reference paths that 404'd after the docs-consolidation move (they now resolve to `docs/adr/3.x/…` and `docs/architecture/…`).
- Corrected the "module scope" comments in `tracker/__init__.py` + `origin_consumer.py` — the registration-triggering import is in the `mission_create` CLI command body; flagged it load-bearing so it isn't removed as "redundant" (would silently break origin-consumer registration).
- Dot-bounded the boundary-test allowlist `startswith` (so a `saas.rollout` exemption can't cover a future sibling module).
- Re-pinned two stale resolution-authority allowlist line-pins (`validate_workflow`, `list_dependents`) — pre-existing main debt that was masked by the core-misc path-filter and only surfaced because this PR touches `tests/architectural/` + `src/`. (The one real CI red — the sync strict-JSON test — was the separate #2254, now fixed by #2257 on main.)

**Follow-ups we did NOT fold (tracked separately, not blockers):** the pre-existing dual `MissionCreatedPayload` construction (lifecycle vs emitter) and a coverage test for the relocated daemon-push branch under `SPEC_KITTY_ENABLE_SAAS_SYNC=1`.

CI is re-running on the rebased head; once green it's ready to merge.

Issue #2175, opened 2026-06-26
fast-tests-doctrine red on main: spk-team-upsun-cli-sync skill unregistered + violates 3 pack invariants


MOES-Media:
## Summary

`tests/doctrine/test_spk_skill_pack.py::test_320_spk_skill_pack_is_complete` is **failing on `main`** (confirmed on the latest CI Quality run for HEAD `618a3fbc6`, run [28221032956](https://github.com/Priivacy-ai/spec-kitty/actions/runs/28221032956)):

```
FAILED tests/doctrine/test_spk_skill_pack.py::test_320_spk_skill_pack_is_complete
  Extra items in the left set: 'spk-team-upsun-cli-sync'
```

This reddens `fast-tests-doctrine` (and the aggregate `quality-gate`) on every branch that merges current `main`.

## Root cause

Commit `2bce35171 docs(skills): add Upsun CLI sync skill` added the directory `src/doctrine/skills/spk-team-upsun-cli-sync/` but never registered it in the `SPK_SKILLS` expected set in `tests/doctrine/test_spk_skill_pack.py`. `test_320` asserts the on-disk `spk-*` skill dirs exactly equal `SPK_SKILLS`, so the unregistered dir fails the equality.

The skill cannot simply be added to `SPK_SKILLS` as-is, because its `SKILL.md` also violates three sibling pack invariants enforced by the same test module:

| Invariant (test) | Limit | Actual |
|---|---|---|
| `test_spk_skill_frontmatter_matches_directory_name` — description length | ≤ 200 chars | **343** |
| `test_spk_skill_bodies_stay_concise` — body length | ≤ 80 lines | **125** |
| `test_spk_skill_map_mentions_every_public_skill` — listed in `spk-meta-skill-map/references/spk-skill-map.md` | required | **absent** |

## Fix (one of)

**(a) Conform the skill** (preferred — it's a legitimate skill):
- Trim `description` to ≤ 200 chars
- Trim the body to ≤ 80 lines
- Add `` `spk-team-upsun-cli-sync` `` to `src/doctrine/skills/spk-meta-skill-map/references/spk-skill-map.md`
- Add `"spk-team-upsun-cli-sync"` to `SPK_SKILLS`

**(b) Remove** `src/doctrine/skills/spk-team-upsun-cli-sync/` if the skill is not meant to ship.

## Notes

[...]

Issue #2049, opened 2026-06-20
Shrink architectural ratchet exception allowlists (burn-down: _baselines.yaml + dead-module/symbol frozensets)


stijn-dejongh:
## Context

`tests/architectural/_baselines.yaml` is a **SHRINK ratchet** (FR-111, Slice F C-004 burn-down policy): every category is supposed to trend downward toward zero, not upward. Mission **01KVJPEQ** (read-side surface-resolver adoption) bumped `category_4_backcompat_shims: 8 → 9` by allowlisting `specify_cli.mission_read_path` — a deliberately temporary growth, pending a separate backcompat decision on retiring the shim.

More broadly, the `_SYMBOL_ALLOWLIST` / `_ALLOWLIST` frozensets across the architectural test suite have accumulated exception entries that should be revisited and burned down.

### Immediate (paired with sister issue #2048)

- Reverse the 01KVJPEQ bump once `specify_cli.mission_read_path` is deleted:
  - `tests/architectural/_baselines.yaml` `category_4_backcompat_shims: 9 → 8`
  - Drop `_CATEGORY_C_BACKCOMPAT_SHIM_REEXPORT` entry `"specify_cli.mission_read_path::resolve_mission_read_path"` from `tests/architectural/test_no_dead_symbols.py`
  - Drop `"specify_cli.mission_read_path"` from `_CATEGORY_4_BACKCOMPAT_SHIMS` in `tests/architectural/test_no_dead_modules.py`

### Audit sweep (broader burn-down)

Audit all categories in `tests/architectural/_baselines.yaml` and the corresponding frozensets and verify whether entries are still justified:

| File | Frozenset / key | Current baseline | Burn-down owner |
|------|-----------------|------------------|-----------------|
| `test_no_dead_modules.py` | `_CATEGORY_4_BACKCOMPAT_SHIMS` | 9 (target ≤ 8 after #2048) | This issue / #2048 |
| `test_no_dead_modules.py` | `category_7_grandfathered_orphans` | 6 | C-006: must shrink ≥ 2/major release; target 0 by 4.0 |
| `test_no_dead_symbols.py` | `_CATEGORY_C_BACKCOMPAT_SHIM_REEXPORT` | 1 entry | This issue / #2048 |
| `test_no_dead_symbols.py` | `category_b_grandfathered_legacy` | 271 | C-007 / FR-121 follow-on mission |
[...]

PR #2176, opened 2026-06-26
feat(2158): harden dead-symbol gate — parser fix + 4 caller detectors, flat ratchet

### LINKED ISSUE #2158: Dead-symbol gate parser bug: _extract_all_literal blinds 57 modules / 117 dead symbols (split from #2049)

MOES-Media:
Closes #2158.

## What

Fixes the `_extract_all_literal` parser bug in the dead-symbol gate
(`tests/architectural/test_no_dead_symbols.py`): a non-`__all__` top-level
`AnnAssign` fell through to `return frozenset()`, silently **blinding the gate to
57 modules**. The `else: continue` fix makes those modules inspected, surfacing
~119 previously-hidden `__all__` symbols in a single gate run.

Rather than bulk-allowlisting the surfaced symbols (which would grow the ratchet
~107 — the regression #2049 deferred FR-006 to avoid), the gate's caller
detection is enhanced with **4 structurally-anchored detectors**, each binding
proof-of-life to a **RESOLVED declaring module** via the file's import-alias map
(never a bare-name substring match):

- **(a)** module-style attribute access `alias.symbol` — also subsumes Typer
  `app.command()(mod.fn)` registration
- **(b)** lazy `__getattr__` package-facade re-exports (static `(submodule, "name")` tuples)
- **(c)** `getattr(mod, "name")` dynamic dispatch

A **no-false-negative regression test** (NFR-001 / C-001) proves a synthetic dead
`__all__` symbol with no caller of any recognized kind is **still flagged** — the
detectors did not blind the gate.

## Residue disposition

Defs are retained; only `__all__` membership changes, so explicit
`from x import y` (including tests) and `import *` consumers are unaffected:

- **DELETE** `sync.owner::_daemon_root` re-export
- **DEMOTE** ~88 genuinely-internal symbols across ~26 modules
- **ALLOWLIST (deferred, 6 entries)** the `auth.transport` public client surface

## Security

Wires the `BANNED_FLAGS` check in `orchestrator_api/envelope.py` —
`parse_and_validate_policy` now rejects any `dangerous_flags` entry that is a
banned flag (previously defined but never enforced). Stale "not invoked" comment
corrected. Added a `--yolo`-rejected unit test.

## Result

The gate now inspects the previously-blind 57 modules with **NO net ratchet
growth** (`category_a`/`category_b` frozenset counts flat).

## CI note — two red checks are PRE-EXISTING and unrelated

`fast-tests-doctrine` and the aggregate `quality-gate` are red **only** because of
`test_320_spk_skill_pack_is_complete`, which **also fails on `main` right now**
(latest CI Quality run for main HEAD `618a3fbc6`:
[28221032956](https://github.com/Priivacy-ai/spec-kitty/actions/runs/28221032956)).
The cause is the unregistered `spk-team-upsun-cli-sync` skill — tracked in **#2175**.
This branch touches only the dead-symbol gate + `__all__` demotions; it does not
touch any skill. Every other check is green.

## Sonar

No new code-correctness debt expected from this change. The gate hardening adds
small AST helpers with focused unit tests; `envelope.py` adds one validated
rejection branch with a test.


stijn-dejongh:
## Adversarial squad review — verdict: **FIX-BEFORE-MERGE** (light; de-blinding is a real win, residue is verified safe)

Four profile-loaded lenses (architect / debugger / reviewer / reducer), read-only against `feat/harden-dead-symbol-gate`. Headline: **the residue disposition everyone would worry about — the ~88 `__all__` demotions, the `_daemon_root` deletion, the security wire — is verified safe, not just asserted.** The blockers are not there; they're in the new detector edges quietly re-introducing a narrower copy of the very blinding class this PR exists to kill.

### ✅ Verified sound (so reviewers don't re-litigate these)
- **Parser fix is correct and red-first proven.** `else: continue` precisely closes the `AnnAssign` fall-through; `test_extract_all_literal_skips_non_all_annassign` returns `frozenset()` against base code → genuinely fails pre-fix, would have caught the original 57-module blinding. Edge cases (`__all__: list[str]` no-value, tuple form, non-`__all__` AnnAssign before `__all__`) all hold. (Pre-existing gap, follow-up only: `__all__ += [...]` AugAssign still unhandled.)
- **"`import *` consumers unaffected" is TRUE, not assumed.** Only 4 real `from X import *` in `src/` (`charter_lint`/`charter_freshness`/`charter_preflight` → `charter_runtime.*`; `glossary` → package `__init__`) — **none** target a module whose `__all__` shrank. Sampled demoted symbols (`resolve_context_for_mission`, `check_daemon_owner_match`, `ReapResult`, `_DASHBOARD_UNSAFE_FLAGS`) — every consumer uses explicit `from x import y` or fully-qualified monkeypatch, unaffected by `__all__`.
- **`_daemon_root` deletion is provably dead** (imported only to re-export; zero consumers; real users go through `sync/daemon.py` where the def is retained).
- **Security wiring is real, fail-closed, non-vacuous.** Base `parse_and_validate_policy` truly never enforced banned flags ("NOT invoked anywhere" comment confirmed); the new check raises `ValueError` before `PolicyMetadata` is built; the test parametrizes over **every** `BANNED_FLAGS` entry (not just `--yolo`) and the happy-path test was retargeted to a non-banned flag. (Minor: enforcement is JSON-ingress only; direct `PolicyMetadata(...)` bypasses — acceptable, internal-trusted; worth one PR-body line.)
- **"No net ratchet growth" holds** — total `::` allowlist entries **365 → 363 (net −2 shrink)**; the new 6-entry `auth.transport` set is offset by removals + the emptied charter-split-legacy set. The "category_a/b counts flat" wording is imprecise (cat A shrank ~8, a new cat B of 6 was added) but the load-bearing non-growth claim is correct. The 6 `auth.transport` symbols are tested public surface (exercised by `tests/auth/`, `tests/integration/`) deferred for SaaS migration — legitimate allowlist use, just needs burn-down tracking.

### 🔴 Must-fix 1 — the new detector edges can re-blind the gate, and it's untested (converged: architect HIGH + debugger MEDIUM, + reviewer LOW variant)

The new detectors add **attribute-access edges** to `per_symbol[resolved_module]`, but the pre-existing `_symbol_has_caller` **rule 3** (submodule re-export proof-of-life) marks a package-level `__all__` symbol alive if *any* submodule has a same-named entry in `per_symbol`. Concrete trace (architect): `from M import SomeClass as C; C.NAME` keys `per_symbol["M.SomeClass"]`; `_submodule_index("M")` treats `"M.SomeClass"` as a submodule; a genuinely-**dead** `M::NAME` is then rescued. That is a narrower copy of the exact blind spot #2158 fixes — and the `test_no_false_negative_*` cases only use flat, unrelated module names (`other_pkg`, `declaring_module`), so rule 3 against the enlarged `per_symbol` is never exercised. Reviewer found the symmetric variant (a local var shadowing an import alias → wrong-module edge → false negative).

CI is green, so nothing is *actively* mis-rescued today — this is a latent, bounded hole, not a live break. But the fix is cheap and surgical, and a gate-hardening PR should not ship a fresh re-blinding vector. **Fix:** gate edge-recording to real modules only — `known = set(path_to_dotted.values()); record only if resolved in known` — and add a no-false-negative test for the aliased-symbol-import / submodule-attribute-collision case.

### 🔴 Must-fix 2 — no end-to-end "teeth" self-test (architect MEDIUM; gate-can't-self-validate)

Four detectors that can only *add* callers means the aggregate path could, in principle, rescue everything and silently no-op the gate. The unit negatives prove `_symbol_has_caller` returns False for an *unrelated* module, but nothing drives a constructed dead-symbol fixture through the **full** `test_no_public_symbol_in_all_is_unimported` aggregate pipeline. A hardened gate's teeth must be proven *in this merge* (it only bites future diffs). **Fix:** a positive-control test injecting a module with a dead `__all__` entry into the real `path_to_tree`/`decls` and asserting it surfaces as an offender.

### 🟡 Worth confirming — facade detector (b) may be dead weight (reducer MAJOR, empirical but reimplemented)

An empirical detector-toggle found the lazy `__getattr__` facade detector (b) — `_record_facade_edges` + helpers, ~80 lines incl. the complexity hotspot (radon 19) — has **zero *unique* beneficiaries**: every symbol it rescues (the `sync/__init__.py` facade) is already covered by the pre-existing parent-package re-export rule, so dropping (b) produced 0 new offenders with no allowlist cost. It's also a latent gate-*weakener* (a symbol in the lazy-dict but imported nowhere is dead, yet (b) marks it alive). **Caveat:** this rests on a faithful *reimplementation* of the gate outside the real harness (pytest was sandbox-blocked, and the probe had an acknowledged merge-offenders artifact) — so **confirm by running the actual gate with (b) disabled** before cutting it. Detector (a) (module-attr, incl. Typer registration) is strongly justified — 8 sole beneficiaries and it let the PR delete a whole legacy allowlist set; keep it. Detector (d) (getattr) rescues exactly 1 real symbol (`MIN_SUPPORTED_SCHEMA`) — keep-or-allowlist is a toss-up.

### Disposition
The 57-module de-blinding is strictly better than the silently-blinded status quo, and the scary residue is clean. Land the two cheap must-fixes (known-module edge gating + rule-3 no-false-negative test; end-to-end teeth test), confirm-then-decide on facade detector (b), and this ships.

---
*Squad note: convergent evidence across four independent lenses. The architect/debugger agreed on the rule-3 re-blinding mechanism from different angles; the reviewer independently cleared the demotion/security residue (verdict SHIP); the reducer's (b)-is-redundant finding is empirical but reimplementation-based, hence flagged as confirm-before-cut rather than a hard ask.*

stijn-dejongh:
**Disposition: we're landing this.** 

The 57-module de-blinding is a real correctness win, and the part that would normally gate a "demote ~88 `__all__` symbols" change — the residue safety — was independently verified clean by the review squad: no `import *` consumer is affected (only 4 star-imports in `src/`, none target a shrunk module), `_daemon_root` is provably dead, every demoted symbol's real consumers use explicit imports, the `BANNED_FLAGS` security wire is real/fail-closed/non-vacuously tested, and the allowlist ratchet actually *shrank* (365→363). The parser fix is red-first proven.

**Plan:**
1. Rebase the branch onto `upstream/main` and land it.
2. The squad's must-fixes and follow-ups are **not blocking this merge** — they'll be tracked and run as a series of **post-mission ops** on top of the landed change:
   - close the new-detector re-blinding vector (gate edge-recording to resolved real modules only) + add the rule-3 / aliased-symbol-import no-false-negative test;
   - add the end-to-end "teeth" positive-control test (prove the hardened gate still flags a truly-dead symbol through the full aggregate path);
   - confirm-then-decide on facade detector (b) (empirical toggle suggested zero unique beneficiaries — verify against the real gate harness before cutting);
   - minor follow-ups: `__all__ += [...]` AugAssign handling, a one-line note that `BANNED_FLAGS` enforcement is JSON-ingress-only, and burn-down tracking for the deferred 6-entry `auth.transport` allowlist.

Landing as-is is strictly better than the silently-blinded status quo; the remediations harden it further without holding up the fix. Thanks @MOES-Media.

stijn-dejongh:
**Folded the two must-fixes and re-homed for landing → #2184** (rebased onto current `main`).

Your original work is preserved verbatim as the first commit; the second commit folds in the squad's two must-fixes:

1. **Closed the detector re-blinding vector.** The edge-detectors recorded an edge to any alias target, including a synthetic *symbol* path: `from M import Cls as C; C.NAME` made rule 3 of `_symbol_has_caller` falsely rescue a dead `M::NAME` — a narrower copy of the 57-module blinding this PR fixes. Edges are now gated on `known_modules` (real src modules only). Added `test_no_false_negative_aliased_symbol_import_does_not_reblind` (with a control proving the guard is the difference); the real `test_no_public_symbol_in_all_is_unimported` still passes, so no legitimate rescue was dropped.
2. **Added end-to-end gate teeth.** Extracted `_compute_offenders` + `test_gate_still_flags_a_truly_dead_symbol` driving a constructed dead-symbol fixture through the real aggregate path — proves the four additive detectors didn't turn the gate into a silent no-op.

Facade-detector-(b) drop, AugAssign handling, and the BANNED_FLAGS/auth.transport notes remain follow-ups (the `known_modules` guard already de-risks (b)). Test-only changes; ruff + mypy clean; `tests/architectural/` green modulo pre-existing env failures (verified identical vs base).

#2184 supersedes this PR — suggest closing this one in its favour once #2184 merges. Thanks again @MOES-Media, the parser-fix + detector approach is the right call.

PR #2163, opened 2026-06-26
fix(accept): resolve mission-artifact path conventions on the primary surface, not repo root (#2162)

rayjohnson:
Closes #2162.

## Summary

`spec-kitty accept`'s path-convention check resolved **every** mission-declared path against `repo_root`. For non-research missions `_path_prefix_for_mission` returns `None`, so a mission-artifact path — software-dev's `deliverables: contracts/` (also an `artifacts.optional` entry) — was sought at `<repo_root>/contracts/` and reported missing, telling the operator to `mkdir -p contracts/` even though `contracts/` was committed at `kitty-specs/<mission>/contracts/` (live dogfood false positive).

This is a residual of the **#1716 / #2113** "no resolution to the repo primary for mission artifacts" cluster — #2113 fixed the artifact *reads*, not the path-convention check.

## Fix (aligned with #2113 C-002 "no fallback" / FR-009 "single canonical surface")

`validate_mission_paths` now resolves a declared path against the mission's **primary `feature_dir`** when that path is a declared mission artifact (member of `mission.config.artifacts`), else `repo_root`. The accept gate passes `feature_dir=planning_read_dir` — the same canonical primary surface `_missing_artifacts` already uses — **not** a repo-root fallback.

- Keyed on **artifacts-membership**, not the `deliverables` name (which is repo-relative `docs/` for plan/documentation missions, so the name would misfire).
- Build paths (`src/`/`tests/`/`docs/`) stay repo-root.
- Research's `path_prefix` routing is untouched.
- Defensive: a config without `artifacts` → no artifact paths (prior behavior); backward-compatible for other `validate_mission_paths` callers (new kwarg is optional).

## Tests

- New red-first validator unit tests: an artifact path (`contracts/`) resolves under `feature_dir`; a non-vacuity test shows that *without* `feature_dir` it misses at `repo_root` (the bug); a non-artifact path (`tests/`) stays repo-root even with `feature_dir`.
- 6 accept/tasks fixtures that created `contracts/` at the repo root (to satisfy the old buggy check) now create it under `feature_dir`.
- Accept/validator/tasks sweep: **784 passed**. Broad sweep: **11999 passed**; the 5 failures there are pre-existing on `upstream/main` (stash-verified, unrelated — mid8-routing / workspace-seam / feature-metadata). `ruff` + `mypy` clean.

## Note for reviewers

[...]

PR #2149, opened 2026-06-25
fix: native-flow PRIMARY reads off the primary surface under coord topology (#2115)

rayjohnson:
## Summary

Native-flow twin of the merged #2118. On a `coord`/`pr_bound` mission rooted on a writable target branch, planning artifacts (`lanes.json`=LANE_STATE, WP `tasks/`=WORK_PACKAGE_TASK, `meta.json`=PRIMARY_METADATA) live on the primary `target_branch` (#2090), but the native commands still read them off the **coordination worktree** (status-only) — hard-blocking `merge` and `tasks status`. This routes the PRIMARY-partition reads/writes through the kind-aware seam (`resolve_planning_read_dir` / `primary_feature_dir_for_mission`, primary for every topology); STATUS reads/writes stay on the coord surface.

## Fixed (10 functions)

| Surface | Function(s) | Change |
|---|---|---|
| `tasks status` | `status` | WP `tasks/` → primary seam |
| `tasks finalize-tasks` | `finalize_tasks` | tasks.md + WP `tasks/` read **and write** + identity → primary (closes a latent split: writes used the coord-aware resolver while `mission.py finalize-tasks` already anchored `lanes.json` to primary) |
| `merge` | `_run_lane_based_merge`, dry-run, `_run_lane_based_merge_locked`, `_merge_state_key_candidates` | `require_lanes_json` + `resolve_mission_identity` (+ resume keys) → primary |
| `merge` | `_mark_wp_merged_done` | WP lookup + status-meta source → primary (status emission still routes to coordination via the transactional helpers) |
| `merge` | `mission_number` baking | falls back to baking on the **target** branch (where `meta.json` lives post-#2090) when the coord/mission branch has none — no more null number |
| review | `_resolve_review_context` | lanes.json + tasks/ → primary |
| implement (auto-claim) | `preview_claimable_wp` (+ `_preview_claimable_wp_for_mission`) | backward-compat `status_dir` kwarg: tasks/dep-graph read PRIMARY, lanes read the coord status surface |

Removed `status`, `finalize_tasks`, `_resolve_review_context`, `_mark_wp_merged_done`, `_preview_claimable_wp_for_mission` from `_DIR_READ_KNOWN_RESIDUALS`.

## Deferred (still pinned, tracked, non-blocking)

[...]

Issue #2173, opened 2026-06-26
Epic: Infrastructure-to-logic separation — inject ports for FS, Clock, GitOps, SaaSQueue, Renderer; keep the core pure and stub-testable


stijn-dejongh:
## Thesis

A large and recurring family of bugs traces to a single missing architectural move: **logic reaches directly into infrastructure** instead of depending on an **injected port** (an object/function reference/protocol supplied at the constructor or API boundary).

The remedy is well-known: **define ports for infrastructure, inject adapters at the builder/shell boundary, keep the core pure and stub-testable** (hexagonal / ports-&-adapters / functional-core-imperative-shell). The codebase already embodies the pattern — `resolution.py`'s core is explicitly pure ("zero filesystem or git I/O") with I/O threaded in by an imperative shell; `RuntimeEventEmitter(Protocol)` + `NullEmitter` is a shipped port-and-adapters seam; `tracker_client_glue` already has a Clock port via default-param (`sleep or time.sleep`). **#2173 generalizes an existing idiom, not imports a foreign architecture.**

> **Framing note (corrected from original):** ~~"`context` is a proto-DI container — formalize it"~~ is **factually wrong**. `MissionContext` (`context/models.py:17`) is a **frozen value snapshot** — `context.mission_resolver` is a module path / free function, not a live collaborator. A frozen value object MUST NOT carry a mutable I/O adapter (category error, breaks the immutability invariant). **Correct framing:** "formalize infra ports at the **builder/shell seam** — the shell that *builds* the context (`_assemble_core_fragments`, `build_execution_context`). Ports inject into the shell; the frozen `MissionExecutionContext` keeps carrying resolved *values*, never live adapters."

### DI mechanism (settled)

The existing idiom: `def fn(..., *, port: Proto | None = None)` → `port = port or RealAdapter()`. Precedents: `RuntimeEventEmitter`/`NullEmitter`/`JsonlEventLog` (`events.py:67`), `_resolve_wp_bearing_fields` already takes injected `Callable` params (`resolution.py:465`), tracker-glue clock, the `build_connector` factory. **No framework, no god-object, no constructor-ification.** Backward-compatible → an un-ported site still works via the default; no broken half-state, no partial-adoption tax.

---

## Evidence — The Family of Issues

Each row is an independent symptom of the same structural cause: logic coupling to infrastructure instead of a defined, injectable port.

| Symptom | Infrastructure coupling |
|---|---|
[...]

PR #2106, opened 2026-06-24
feat: kind-aware write-surface placement — planning artifacts to primary (#2090)

stijn-dejongh:
Closes #2090. Closes #2087. Closes #2101. (Full split-brain issue linkage — deferred/adjacent/the #2107 gap + the #2101↔#2104 coordination — in the planner-priti survey comment below.)

---

## Summary

The **write-side twin** of the merged read-path single-authority work (mission `single-authority-topology-cleanup`, #2099 / PR for #147). That mission made every *read* path resolve a mission's surface from one authority. This one closes the symmetric **write** gap.

**The bug (an asymmetry):** the canonical `MissionArtifactKind` placement model already exists on the **read** side (`mission_runtime/artifacts.py` — `artifact_home_for` / `_PLACEMENT_ARTIFACT_KINDS`), but the **write** side (`resolve_placement_only` / `commit_for_mission`, the bypass writers `safe-commit` / `append-history`, and the 2nd routing authority `_planning_commit_worktree`) was **kind-blind** — routing *all* artifacts by topology. So a coordination-topology mission committed its `spec.md`/`plan.md`/`tasks.md` to the coordination branch while `finalize-tasks` read the primary checkout → "all requirements unmapped", forcing manual coord-worktree creation.

**The fix:** make the write side **kind-aware** and **re-partition** the planning + identity kinds onto the primary `target_branch` for every mission shape:

| Partition (→ destination) | Kinds |
|---|---|
| **PRIMARY** (planning + identity) | `SPEC` `DATA_MODEL` `RESEARCH` `CHECKLIST` `FINALIZED_EXECUTION_PLAN` `TASKS_INDEX` `WORK_PACKAGE_TASK` `LANE_STATE` `PRIMARY_METADATA` |
| **COORD** (status / bookkeeping) | `STATUS_STATE` `ISSUE_MATRIX` `ACCEPTANCE_MATRIX` `ANALYSIS_REPORT` |

The frozenset partition `_PRIMARY_ARTIFACT_KINDS` is the **single swappable locus** — flipping where a kind lands is moving it between two sets, with no change to the `CommitTarget` shape or the public signatures. `kind` is a **required** parameter on `resolve_placement_only` so no caller can silently regress coord→primary (a compile error forces enumeration of all 7 callers).

## What's in it (7 WPs, all reviewer-approved)

- **Kind-aware authority + re-partition** (`mission_runtime/artifacts.py`, `resolution.py`).
[...]

PR #2179, opened 2026-06-26
fix(cli): silence stderr diagnostics under --json so 2>&1 captures stay pure JSON (#2178)

### LINKED ISSUE #2178: --json output corrupted when captured with 2>&1 (stderr diagnostics merge into the JSON)
## What happens

Agents/automation commonly invoke `spec-kitty <cmd> --json 2>&1` and parse the merged stream. The CLI correctly puts the JSON object on **stdout** and diagnostics on **stderr** — but any stderr line then interleaves with the JSON in the `2>&1` capture, so `jq` / `json.load` fails on the first line. Reported from a live dogfood run (`finalize-tasks --json` repeatedly, under coord topology where charter/deprecation warnings fire).

## Confirmed mechanism

- `spec-kitty agent tasks status --json` → valid JSON on stdout, empty stderr → `2>&1` clean.
- But the logging bootstrap routes warnings/logs (`CharterCatalogMissWarning`, `DeprecationWarning: specify_cli.next is deprecated`, `Legacy asset resolved`, `No meta.json found`) to **stderr** (correct for humans). On a run where any of those fire, `--json 2>&1 | json.load` → `JSONDecodeError: Expecting value: line 1 column 1`.

Reproduced end-to-end via the real package: a `warnings.warn` / `logger.warning` during a successful `--json` run breaks `2>&1` parsing; with the fix it stays pure JSON.

## Fix direction

`--json` is machine mode → diagnostics should be silent so the merged stream is parseable. Single chokepoint: the logging bootstrap runs in a silent mode when `--json` is present (raise all root handlers above real records + install a `NullHandler` so `lastResort` never fires; `captureWarnings` keeps routing warnings through the now-silent handlers). Genuine command errors are still emitted as JSON on stdout by the commands.

Note: Typer *usage* errors (malformed invocation) still print to stderr pre-dispatch — a separate surface, out of scope here.

PR incoming.

rayjohnson:
Closes #2178.

## Summary

The CLI puts the JSON object on stdout and diagnostics on stderr (correct) — but agents commonly invoke `spec-kitty … --json 2>&1` and parse the **merged** stream, so any stderr warning/log line corrupts the JSON (`JSONDecodeError` on line 1). The warnings *do* fire under real conditions (`CharterCatalogMissWarning`, `DeprecationWarning`, `Legacy asset resolved`, `No meta.json found`) — e.g. the live `finalize-tasks --json` dogfood under coord topology.

## Fix (single chokepoint, `--json`-aware)

`install_cli_logging_bootstrap()` gains a `json_mode` flag, set from `"--json" in sys.argv` at the entry point (`main()`). In `json_mode` it:
- raises every root handler above any real record (`CRITICAL+1`), and
- installs a `NullHandler` when none exist, so Python's `lastResort` (WARNING→stderr) never fires.

`captureWarnings(True)` (already called) keeps routing `warnings.warn` through those now-silent handlers, so warnings are dropped too — **without** mutating the global `warnings` filter (test-friendly). A successful `--json` run therefore emits only the JSON object on stdout **and** on a merged `2>&1` stream. Genuine command errors are still emitted as JSON on stdout by the commands themselves.

**Default (human) mode is unchanged** — WARNING+ still reaches stderr via the Rich handler.

## Tests

- `json_mode` silences a real pre-installed handler (`logger.warning`/`error` + a captured `warnings.warn` → empty stderr); installs a handler to suppress `lastResort`; **default mode still emits** (non-vacuity).
- End-to-end via the real package: `json_mode=True` → `--json 2>&1 | json.load` **VALID**; `json_mode=False` → `JSONDecodeError` (proves the fix is the difference). All 11 bootstrap tests pass; `ruff` + `mypy` clean.

## Scope

- One CLI-side chokepoint (no per-command changes) — aligned with the single-seam preference.
- Out of scope (flagged): Typer **usage** errors for malformed invocations still print to stderr pre-dispatch; making those JSON is a separate, larger change.
- Independent of #2149 (native-flow coord-read) and the merged #2163 (accept-contracts).


stijn-dejongh:
## Review

The logging-side mechanism is correct: `setLevel(CRITICAL+1)` on existing handlers + a sentinel `NullHandler` when none exist correctly suppresses both real records and Python's `lastResort` (`callHandlers` counts the NullHandler so `found != 0` → `lastResort` never fires), `captureWarnings(True)` runs before the silence block, and the three new tests are non-vacuous. 👍

One completeness gap worth addressing before merge:

### `--json` purity guarantee leaks through direct-stderr writes

The fix silences **logging-routed** diagnostics only — but the legacy-asset path the PR cites as its motivating case has a second branch that bypasses logging entirely.

In `src/specify_cli/runtime/resolver.py`, `_warn_legacy_asset()` splits:
- global runtime **not** configured → `logger.warning(...)` + `warnings.warn(...)` → **silenced** ✅
- global runtime **is** configured → `_emit_migrate_nudge()` (resolver.py:107), which does a raw `print("Note: Run \`spec-kitty migrate\` …", file=sys.stderr)` → **not silenced** ❌

**Repro:** a user with `~/.kittify/cache/version.lock` present runs `spec-kitty finalize-tasks --json 2>&1` in a project that resolves a legacy-tier asset (the exact "live `finalize-tasks --json` dogfood under coord topology" named in the PR body). The nudge line lands on the merged stream ahead of the JSON → `json.load` → `JSONDecodeError` on line 1. The bug this PR targets still reproduces on this branch.

`_emit_migrate_nudge`'s own docstring — *"Output goes to stderr so it never interferes with `--json` output on stdout"* — is precisely the premise this PR invalidates.

Same class of gap (lower-traffic): `plan_validation.py:104` (`print(f"Warning: …", file=sys.stderr)`) and the various `Console(stderr=True)` sites under `readiness/`, `sync/`, `charter_runtime/`.

**Suggested action — either:**
1. Route `_emit_migrate_nudge` (and the plan-validation warning) through `logger.warning` so the single chokepoint actually catches them, **or**
2. Narrow the CHANGELOG/PR claim from "emits only the JSON object on the merged `2>&1` stream" to "only logging/warnings-routed diagnostics are silenced; direct-stderr nudges remain out of scope" — mirroring the existing Typer-usage-error carve-out.

Minor / non-blocking: `"--json" in sys.argv` would false-positive if a literal `--json` were ever an option *value*, but it matches the established pattern at `__init__.py:301,313`, so no regression.

stijn-dejongh:
Heads-up, @rayjohnson — during a drift-triage of all open PRs we applied the **`pr:needs-refresh`** label to this one.

**What the label means:** *"PR branch drifted from `main`; rebase/refresh before review or merge."*

**Drift on this PR:** the branch is currently in **merge-conflict** with `main` and sits roughly **13 commits behind**, so it can't be reviewed or merged as-is.

**Recommended next steps:**
1. **Rebase onto current `main`** to clear the merge conflict and catch the branch up.
2. **Cross-check the change against recently-merged work on `main`** — the surfaces this PR likely touches have shifted substantially. Worth reconciling against:
   - **Coord-topology surface-resolver / artifact-authority changes** — read/write routing now goes through `resolve_planning_read_dir` / `primary_feature_dir_for_mission` (#2106 plus the #2160 coord-authority cluster).
   - **`--feature` → `--mission` selector removal** (#1060).
   - **Pre-3.0 legacy-reader retirement + boundary guard** (#1057).
   - **Dead-symbol-gate hardening** (#2158).
   - **Resolution-authority gates** (#2173 lineage).
3. After rebasing, **re-run the architectural/CI gates locally** — several only run in CI, so a green local fast-suite isn't sufficient.

This is a courtesy heads-up, not a gate — nothing here blocks you, it just flags what to reconcile against before the branch goes up for review/merge.

rayjohnson:
Refreshed against current `main` (force-pushed `fix/json-quiet-stderr` → `dba9fc7bf`). Clean rebase, no conflicts — single commit, `mergeable`. Rides the `[Unreleased] - 3.2.3` line (no version bump needed; CHANGELOG entry is in the `🐛 Fixed` section). The `pr:needs-refresh` drift is resolved — should be clear to re-review. @stijn-dejongh

PR #2233, opened 2026-06-28
Add CLI autocompletion

### LINKED ISSUE #2232: Add CLI autocompletion
## Summary
Add shell completion support to `spec-kitty` so users can tab-complete root commands and nested subcommands.

## Acceptance Criteria
- `spec-kitty` exposes Typer completion options such as `--install-completion` and `--show-completion`.
- Root command completion includes user-facing commands such as `agent`, `doctor`, and `mission`.
- Nested command completion is scoped to the active group, for example `spec-kitty agent <TAB>` completes agent subcommands without leaking unrelated root commands.

## Validation
- `pytest tests/specify_cli/cli/commands/test_root_completion.py -q`

zohar:
## Summary
- Enable Typer completion support for `spec-kitty`.
- Add focused tests for root and nested command completion behavior.
- Include the mission artifacts for the CLI autocompletion improvement.

Closes #2232

## Test plan
- [x] `pytest tests/specify_cli/cli/commands/test_root_completion.py -q`

Made with [Cursor](https://cursor.com)


robertDouglass (dismissed):
P1 blocker: completion works functionally, but it misses this mission's own latency contract.

The spec added in this PR says completion suggestions for common command paths must be available within 500 ms (`kitty-specs/cli-autocompletion-01KW6N5G/spec.md`, NFR-001/SC-003). The implementation only flips the root Typer flag (`src/specify_cli/__init__.py`), so every TAB completion still constructs the full root app and imports the full command tree. On a warm local venv, the generated Bash completion path is 3-4x over budget:

```text
run=1 exit=0 ms=2189.4 stdout_lines=43 stderr=''
run=2 exit=0 ms=1681.4 stdout_lines=43 stderr=''
run=3 exit=0 ms=1655.0 stdout_lines=43 stderr=''
run=4 exit=0 ms=1693.8 stdout_lines=43 stderr=''
run=5 exit=0 ms=1662.4 stdout_lines=43 stderr=''
min=1655.0 median=1681.4 max=2189.4
```

Repro:

```bash
.venv/bin/python - <<'PY'
from __future__ import annotations
import os, subprocess, tempfile, time, statistics
cmd=['.venv/bin/spec-kitty']
times=[]
for i in range(5):
    tmp=tempfile.mkdtemp()
    env=os.environ.copy()
    env.update({
        'HOME': f'{tmp}/home',
        'XDG_CONFIG_HOME': f'{tmp}/config',
        'XDG_CACHE_HOME': f'{tmp}/cache',
        'XDG_DATA_HOME': f'{tmp}/data',
        'SPEC_KITTY_HOME': f'{tmp}/sk',
        'SPEC_KITTY_NO_UPGRADE_CHECK': '1',
        'SPEC_KITTY_NO_BANNER': '1',
        'COMP_WORDS': 'spec-kitty ',
        'COMP_CWORD': '1',
        '_SPEC_KITTY_COMPLETE': 'complete_bash',
    })
    start=time.perf_counter()
    p=subprocess.run(cmd, cwd='.', env=env, text=True, capture_output=True, timeout=15)
    elapsed=(time.perf_counter()-start)*1000
    times.append(elapsed)
    print(f'run={i+1} exit={p.returncode} ms={elapsed:.1f} stdout_lines={len(p.stdout.splitlines())} stderr={p.stderr.strip()!r}')
print(f'min={min(times):.1f} median={statistics.median(times):.1f} max={max(times):.1f}')
PY
```

The added tests only call `command.shell_complete(...)` on an already-constructed command object, so they don't exercise the real generated-shell protocol or catch this startup/import cost. Please either add a lightweight completion fast path that avoids full CLI construction for candidate generation, or explicitly change/drop the 500 ms requirement and acceptance criterion before merging.

zohar:
Thanks for the careful review and for catching this — you're right that the original implementation missed the mission's own latency contract (NFR-001 / SC-003). I went with option 1 and added a lightweight completion fast path.

## Root cause
Completion was functionally correct, but every TAB press ran `main()`, which builds the **entire** Typer application — importing the whole command tree (`init`, `upgrade`, `status`, `events`, migrations, tool-surface, ...) just to enumerate candidate names. That import cost is the latency: ~**765–1398 ms** (~**812 ms median**) on a warm venv, well over the 500 ms budget.

## Fix
A new `specify_cli.completion` module intercepts the `_SPEC_KITTY_COMPLETE` protocol early in `main()`, **before any heavy import**. It:

- builds a throwaway `TyperGroup`/`TyperCommand` tree (command/subcommand names, help, hidden, deprecated, and exact child order) from a pre-generated manifest — **no implementation modules are imported**;
- hands that tree to **Typer's own `shell_complete`**, so the emitted output is **byte-for-byte identical** to the full app across bash/zsh/fish;
- **falls back to the full application as soon as an option token (`-…`) appears** on the line. Because the only thing the fast path ever serves is command/subcommand names from a faithful mirror of the real tree, it can never emit a wrong suggestion — anything involving options/values defers to the real app.

SaaS-gated commands (`tracker`, `issue-search`) are filtered by environment so coverage matches the real surface in both configs.

## Measured before/after
| | latency |
|---|---|
| Before (full app build) | ~765–1398 ms (~812 ms median) |
| After (fast path, real binary) | **~76 ms median, max 134 ms** |
| After (in-process generation) | **~4.8 ms** |

Comfortably under the 500 ms contract.

## New tests (`tests/specify_cli/cli/commands/test_completion_fast_path.py`)
- **Byte-identical parity** vs the full app across bash + zsh — **0 mismatches over 41 group paths** (fish spot-checked too).
- **Manifest drift guard**: rebuilds the real CLI and fails if the manifest and the live command surface diverge (covers the "remain accurate when commands are added/removed/renamed" edge case). Regenerate with `SPEC_KITTY_ENABLE_SAAS_SYNC=1 python -m specify_cli.completion --regenerate`.
- **No-heavy-import structural guarantee**: asserts the fast path serves candidates without importing the heavy command modules — the structural backstop behind the latency contract.
- **<500 ms latency assertion** (min-of-N, subprocess level, per SC-003).
- **No-file-mutation safety test** (NFR-003 / SC-004).

I kept the 500 ms requirement and acceptance criterion as-is, since the fast path now meets it.

stijn-dejongh:
@zohar : please rebase onto latest `main`, and remove the version revert commit. The mainline branch is on `3.2.4` and should remain so.

stijn-dejongh:
**Held under the 3.2.x merge-hold (non-critical enhancement).** Per the [3.2.x delivery plan](../blob/main/docs/planning/3-2-x-approach.md), non-critical community enhancements are held during the doctrine-swap focus (band 3) so maintainer attention and merge bandwidth stay on the brain-surgery epics (#2652/#2721/#2519). This change is welcome and does **not** conflict with the doctrine surface — it's simply deferred by prioritization, not by any problem with the PR. Labeled `pr:deferred`; **CI is now skipped on deferred PRs** to save compute. It'll be unblocked (one clean rebase) once the freeze lifts. — 3.2.x merge-hold triage

zohar:
Superseded by #2919 (combined CLI UX PR).

Issue #2146, opened 2026-06-25
Clarify sync target ownership: config.toml canonical, queue scope derived, no env split-brain


robertDouglass:
## Problem

Local sync targeting now has three surfaces that can disagree:

- `~/.spec-kitty/config.toml` sync server URL: intended canonical configured sync target.
- Queue scope string: derived identity `server_url|user|team`, used to choose the correct scoped SQLite queue DB.
- `SPEC_KITTY_SAAS_URL`: env var that can still point at another server, including a local development server.

There should be two concepts, not competing URL sources:

1. **Configured sync target**: canonical user/operator setting.
2. **Derived queue scope**: deterministic identity used only to isolate SQLite queues.

Including the server URL in queue scope is correct. It prevents events queued for one target from being accidentally drained into another target. The bug is that URL ownership and diagnostics are now ambiguous.

## Concrete Evidence

Observed shape, with environment, user, team, URL, and digest values redacted:

```text
stale marker shape:
<local-dev-url>|<user-id>|<old-team-slug>
digest: <stale-scope-digest>
```

Current sync status reported a different active queue:

```text
queue-<derived-scope-digest>.db
```

That active queue corresponded to a different derived scope:

```text
<hosted-teamspace-url>|<user-id>|<current-team-slug>
```

So `active_queue_scope` was confusing, but it was not selecting the active queue. Current queue selection derives scope from session/credentials + configured server and `default_queue_db_path()`.

## Code Anchors

- `src/specify_cli/sync/queue.py:383-384` defines `~/.spec-kitty/active_queue_scope` path.
- `src/specify_cli/sync/queue.py:391-396` builds canonical queue scope as `server|user|team`.
- `src/specify_cli/sync/queue.py:427-436` reads `~/.spec-kitty/config.toml` `[sync].server_url` for queue-scope derivation.
- `src/specify_cli/sync/queue.py:439-488` derives queue scope from auth session + configured server.
- `src/specify_cli/sync/queue.py:921-938` selects the active queue DB via `default_queue_db_path()`, falling back to credentials, not `active_queue_scope`.
- `src/specify_cli/sync/queue.py:941-957` defines `read_active_scope()` / `write_active_scope()` marker helpers.
- `src/specify_cli/auth/config.py:1-35` still treats `SPEC_KITTY_SAAS_URL` as the SaaS base URL source for auth code paths.

## Root Cause Model

[...]

PR #2131, opened 2026-06-25
mission(event-sync): preserve local events & per-target drains — implementation (#2124)

LynnColeArt:
## ✅ Post-review remediation — addressed an adversarial "SHOULD NOT MERGE" review

An independent review flagged **1 P0 + 5 P1**. Every finding was **independently reproduced with a live repro**, then **fixed with a regression test**:

| Finding | Fix (commit) |
|---|---|
| **P0** draft · 45 behind main · research-note path conflict | merged current `main`, relocated note to `docs/plans/research/`, marked ready (`a0d85fb`) |
| **P1 invalid batch envelope** — journal stored only the inner payload, so dispatched events were missing all 8 contract envelope fields (`schema_version`, `event_id`, …) | store the **full validated envelope** at capture; dispatcher/receiver replay it (`7361356`) |
| **P1 dual destructive drain** — `sync now` ran the legacy delete-on-success drain **and** the new dispatch (2× POST per event) | retire the legacy **event** drain; the journal dispatcher is the sole event path; body-upload preserved (C-006) (`067af71`) |
| **P1 gc destroys re-drainability** — purged on `delivered_anywhere` (any target) | purge **only** events delivered to **all known targets** (`48cfbaa`) |
| **P1 migration unwired + non-atomic** — `migrate_queues_to_journal` had no caller; append-before-provenance orphaned rows | added `spec-kitty sync migrate`; made journal+provenance **atomic** per source (`1ee154d`, `067af71`) |
| **P1 status sections silently drop / migration audit never passed** | emit **stable default sections** on failure; open + pass `MigrationAudit` so `migration_conflicts` is real (`0922a70`, `067af71`) |
| follow-on: preserve Issue #829 exit-1 UX + migrate 5 stale exit-code tests | dispatch-derived exit keeps exit-1 for logged-out; tests driven via `DispatchSummary` (`3176ac5`, `6a4a2ce`) |

**Post-fix validation:** `tests/delivery`+`event_journal` 207 · `tests/sync` 1777 · `tests/architectural` 600 · `tests/cli -k sync` 56 — all green; mission **re-accepted** at `d334a10`. Remaining CI red is **external/not this PR**: a `spec-kitty-events 6.1.0` contract snapshot a *foreign* mission bumped (this PR changed no deps) + 2 other missions' contract frontmatter — needs a repo-level snapshot regen before tagging.

---

[...]

PR #2246, opened 2026-06-29
fix(sync): honor SPEC_KITTY_SAAS_URL in get_server_url; drop hardcoded default (#2146)

LynnColeArt:
**Draft / proposal** — the immediate, surgical slice of #2146 (canonical sync target authority). Folds cleanly into #2131 WP01; opening standalone because it's a small fix that enforces an *existing* architectural decision and resolves a reproduced, real-data bug. Take it as-is or absorb into WP01.

## Problem (reproduced with real data — see #2131)

`auth/config.py` documents **decision D-5**: *"there is NO hardcoded SaaS domain anywhere in the codebase — callers must set `SPEC_KITTY_SAAS_URL`."* The auth subsystem obeys it. But `sync/config.py:get_server_url()` **violated it**: it read `[sync].server_url` only (ignoring `SPEC_KITTY_SAAS_URL`) and fell back to a hardcoded `https://spec-kitty-dev.fly.dev`.

Consequences, observed live:
- `SPEC_KITTY_SAAS_URL=… auth login` authenticated against one target while `sync` resolved a *different* one — the #2146 split-brain (SC-008).
- The hardcoded default (`spec-kitty-dev.fly.dev`) is **unreachable**, so an unconfigured checkout failed *silently* against a dead host.

## Change

`get_server_url()` now resolves in D-5 order:
1. `SPEC_KITTY_SAAS_URL` (the source of truth — takes precedence over config),
2. `[sync].server_url` config key,
3. the D-5 resolver `auth.config.get_saas_base_url()` — which raises a clear `ConfigurationError` when nothing is set, **instead of returning a hardcoded, dead domain**.

So auth and sync can no longer resolve different targets, and "nothing configured" fails loudly with guidance rather than silently hitting a dead host.

## Scope / tests

- One function changed (`sync/config.py`), `import os` added. New: `tests/sync/test_config_server_url.py` — env precedence, config fallback, trailing-slash strip, and raise-when-unconfigured (no hardcoded default).
- Ruff + mypy clean on the changed code; 17 existing `test_config_*` tests still pass. No existing test depended on the removed default (the `fly.dev` references in tests mock `get_server_url` or pass explicit URLs).

## Known sibling violation (NOT in this PR)

`saas_client/auth.py:18` carries its own rogue default (`https://api.spec-kitty.io`) — the same D-5 violation in a different subsystem. WP01's `ResolvedSyncTarget` should unify all of these (auth / sync / saas_client / readiness) through one resolver; flagging rather than fixing here to keep the diff focused.

### DIFF
--- a/src/specify_cli/sync/config.py
+++ b/src/specify_cli/sync/config.py
@@ -1,4 +1,6 @@
 """Sync configuration management"""
+
+import os
 import sys
 from enum import StrEnum
 from pathlib import Path
@@ -28,8 +30,8 @@ class SyncConfig:
     """Manage sync configuration"""
 
     def __init__(self) -> None:
-        self.config_dir = Path.home() / '.spec-kitty'
-        self.config_file = self.config_dir / 'config.toml'
+        self.config_dir = Path.home() / ".spec-kitty"
+        self.config_file = self.config_dir / "config.toml"
 
     def _load(self) -> dict[str, Any]:
         """Load config.toml, returning empty dict when missing or invalid."""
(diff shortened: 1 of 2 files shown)


robertDouglass:
Review feedback from a fresh `origin/main` checkout plus adversarial-squad pass.

This PR is pointing at a real bug class, but I do not think it should land as-is.

Evidence the problem is real on current `main`:

- `SPEC_KITTY_SAAS_URL=http://localhost:8000` makes `SyncConfig().resolve_runtime_target().resolved_server_url` resolve to `http://localhost:8000`.
- The raw accessor still returns `https://spec-kitty-dev.fly.dev`: `SyncConfig().get_server_url`.
- Some live consumers still use that raw accessor as a runtime target, notably `src/specify_cli/tracker/saas_client.py:200`, `src/specify_cli/sync/sharing_client.py:35`, and `src/specify_cli/sync/background.py:398/:458/:544`. So the underlying split-brain risk is not imaginary.

Why this PR is not the right fix now:

- Current `main` explicitly documents `SyncConfig.get_server_url()` as a config-file accessor only. Runtime target authority now belongs in `resolve_sync_target()` / `SyncConfig.resolve_runtime_target()`.
- This branch moves env precedence into `get_server_url()`, which conflicts with that target-authority contract and with current resolver tests.
- It also conflicts with current `main` in `src/specify_cli/sync/config.py`. If resolved by taking the PR side, it risks dropping the newer `SPEC_KITTY_HOME` / `get_runtime_root().base` behavior.

Recommendation: close or supersede this PR rather than merge it. Keep the real bug, but fix it by moving remaining live callers off `get_server_url()` and onto the canonical target authority (`resolve_runtime_target().resolved_server_url` / `resolve_sync_target()`), with focused tests proving tracker, sharing, and background sync all use the same resolved target as auth/readiness.

PR #2247, opened 2026-06-29
fix(sync): allow loopback HTTP in `sync server` URL validation (#2146)

LynnColeArt:
**Draft / proposal** — small `sync` target-hygiene fix in the #2146 family (sibling to #2246).

## Problem

`spec-kitty sync server <url>` rejected any non-HTTPS URL, so `http://localhost:8000` failed with *"Use a full HTTPS URL."* That blocks the documented local-dev workflow (a local Docker SaaS on loopback HTTP — Robert's own pattern is `SPEC_KITTY_SAAS_URL=http://localhost:8000`). It also contradicts the "loopback/local-only HTTP is a special case — don't force HTTPS" guidance. Hit live tonight: had to hand-edit `~/.spec-kitty/config.toml` because the CLI setter wouldn't accept a localhost target.

## Change

`sync_server` now allows **HTTP for loopback hosts** (`localhost`, `127.0.0.1`, `::1`) while still requiring HTTPS for everything remote. One block in `cli/commands/sync.py`; tests added for loopback-accept and non-loopback-HTTP-reject.

## Flag for WP01 (NOT fixed here)

`saas_client/auth.py:18` carries a hardcoded `https://api.spec-kitty.io` default that **contradicts D-5** ("no hardcoded SaaS domain") — but it's enshrined by `test_client.py:121` and has multiple callers. That's a real D-5-vs-tested-behavior tension that WP01 (#2131 target authority) should reconcile deliberately, not a drive-by flip. Flagging, not changing.

### DIFF
--- a/docs/api/cli-commands.md
+++ b/docs/api/cli-commands.md
@@ -3742,11 +3742,13 @@ _Synchronization commands_
  Show or set sync server URL.
 
  Examples:
- spec-kitty sync server
- spec-kitty sync server https://spec-kitty-dev.fly.dev
+spec-kitty sync server
+spec-kitty sync server https://spec-kitty-dev.fly.dev
+spec-kitty sync server http://localhost:8000
 
 ╭─ Arguments ──────────────────────────────────────────────────────────────────╮
-│   url      [URL]  Sync server URL to set (must be https://...)               │
+│   url      [URL]  Sync server URL to set (HTTPS, or loopback HTTP for local  │
+│                   development)                                              │
 ╰──────────────────────────────────────────────────────────────────────────────╯
 ╭─ Options ────────────────────────────────────────────────────────────────────╮
 │ --help          Show this message and exit.                                  │
--- a/src/specify_cli/cli/commands/sync.py
+++ b/src/specify_cli/cli/commands/sync.py
@@ -1799,7 +1799,7 @@ def _check_server_connection(server_url: str) -> tuple[str, str]:
 def sync_server(
     url: str | None = typer.Argument(
         None,
-        help="Sync server URL to set (must be https://...)",
+        help="Sync server URL to set (HTTPS, or loopback HTTP for local development)",
     ),
 ) -> None:
     """Show or set sync server URL.
@@ -1807,6 +1807,7 @@ def sync_server(
     Examples:
         spec-kitty sync server
         spec-kitty sync server https://spec-kitty-dev.fly.dev
+        spec-kitty sync server http://localhost:8000
     """
     from specify_cli.sync.config import SyncConfig
 
--- a/src/specify_cli/cli/commands/sync.py
+++ b/src/specify_cli/cli/commands/sync.py
@@ -1818,10 +1818,17 @@ def sync_server(
 
     normalized_url = url.strip().rstrip("/")
     parsed = urlparse(normalized_url)
-    if parsed.scheme != "https" or not parsed.netloc:
+    # HTTPS is required for remote targets, but loopback HTTP is a deliberate
+    # local-development special case (e.g. http://localhost:8000 against a local
+    # Docker SaaS) — don't force HTTPS on loopback.
+    host = (parsed.hostname or "").lower()
+    is_loopback = host in {"localhost", "127.0.0.1", "::1"}
+    scheme_ok = parsed.scheme == "https" or (parsed.scheme == "http" and is_loopback)
+    if not scheme_ok or not parsed.netloc:
         console.print(
-            "[red]Error:[/red] Invalid server URL. Use a full HTTPS URL, "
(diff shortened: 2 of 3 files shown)


stijn-dejongh:
@LynnColeArt : when ready to start landing: best to rebase your changes onto mainline, as the core loop has gone through some significant refactors.

LynnColeArt:
@stijn-priivacy handing this off for your merge call (core is your lane).

**Scope:** relaxes `sync server <url>` validation to accept loopback HTTP (`localhost`/`127.0.0.1`/`::1`) so a local Docker SaaS at `http://localhost:8000` is settable; remote targets still require HTTPS. Focused tests both ways; docs/help updated.

**Deliberately orthogonal to #2246.** Robert closed #2246 because it moved env precedence into `get_server_url()`, conflicting with the target-authority contract. This PR touches none of that — no `get_server_url()` change, no resolver change, just input validation on the setter. It also matches the repo's loopback Sonar carve-out (don't force HTTPS on intentionally loopback-only URLs).

Separately: Robert's #2246 review named the *real* #2146 bug worth fixing properly — live callers (`tracker/saas_client.py:200`, `sync/sharing_client.py:35`, `sync/background.py:398/458/544`) still read the raw `get_server_url()` (hardcoded fly.dev) instead of `resolve_runtime_target().resolved_server_url`. Flagging as a follow-up, not part of this PR.

stijn-dejongh:
Claiming this PR for today's landing pass: rebase onto upstream/main, adjudicate red checks, fold fixes as needed. Evidence to follow.

stijn-dejongh:
## Maintainer Remediation Summary

**Verdict: MERGE-READY pending green CI**

### What was done

**Rebase:** Rebased `fix/sync-target-hygiene` onto `upstream/main` (2 PR commits rebased cleanly onto `d63ec2152`).

### Folds

Three landing fold commits were pushed on top of the contributor's 2 commits:

1. **`landing fold: add @pytest.mark.unit to new loopback URL validation tests`**
   — The two new test methods (`test_set_server_url_accepts_loopback_http`, `test_set_server_url_rejects_non_loopback_http`) lacked the `unit` marker. Added `@pytest.mark.unit` to both. The module-level `pytestmark = pytest.mark.fast` was already in place.

2. **`landing fold: rename misleading test; add portless localhost + 0.0.0.0-reject coverage`** (adversarial squad finding)
   - **Renamed** `test_set_server_url_rejects_non_https` → `test_set_server_url_rejects_non_loopback_remote_http`. The old name was misleading: the system no longer rejects *all* non-HTTPS — it rejects non-loopback HTTP. The old test also coincidentally passed under the old logic (reject-all-HTTP) and would have passed equally under a buggy "reject-nothing" implementation, so it did not witness the new code path.
   - **Added** `http://localhost` (portless) to the loopback-accept parametrize list — pinning the portless contract that the help text implies (`http://localhost[:port]`).
   - **Added** `test_set_server_url_rejects_zero_addr_http` — `http://0.0.0.0` is not in the explicit loopback set and should be rejected; there was no test asserting this.

### Local evidence

- **30 + 2 (new) = 32 tests**, `TestSyncServerCommand` class: 9 pass
- **Red-first verification**: swapped pre-fix `sync.py` from `upstream/main` — 3/4 new tests fail as expected (`test_set_server_url_accepts_loopback_http[*]` all fail; `test_set_server_url_rejects_non_loopback_http` passes because the old code already rejected non-loopback HTTP — it was already correct). Restored the PR fix: all pass.
- **`ruff check` on changed files**: all checks passed
- **Architectural terminology test** (`tests/architectural/test_no_legacy_terminology.py`): 3 passed

### Process notes (no code action needed)

- `docs/api/cli-commands.md` was hand-edited by the contributor. The `check_docs_freshness` CI gate does not validate argument help text inside fenced blocks for `sync server`, so argument-text drift between `sync.py:help=...` and this file will go undetected. The current content is correct (matches the live `help=` string). Future changes to the `sync server` help text must update both files manually.
- The structural gap where `SyncConfig.set_server_url` has no server-side URL validation (all validation lives in the CLI command layer) is pre-existing, not introduced by this PR. Filed as a follow-up.

### Review: core logic

The URL validation logic is correct:
- `urlparse("http://[::1]:8000").hostname` → `"::1"` (Python strips brackets) — set membership check is sound
- `(parsed.hostname or "").lower()` handles the `hostname=None` edge case
- Set-based check prevents prefix attacks (`http://localhost.evil.com` → full FQDN, not in set → rejected correctly)
- `http://0.0.0.0` not in the loopback set → correctly rejected

**State: watching CI; merge-ready on green.** No landing-order constraints with other open PRs.

@LynnColeArt — note that `docs/api/cli-commands.md` requires a manual update when you change the `sync server` help text in the future, since no CI gate enforces that file's argument-text sections automatically.

Issue #2066, opened 2026-06-21
Opaque diagnostics when WP requirement_refs don't match spec.md FR IDs; no way to see the parsed FR set


rayjohnson:
## Environment
- spec-kitty 3.2.1 (api 1.1.0), mission type software-dev

## Summary
When a work package's `requirement_refs` reference an ID that doesn't exactly match what spec-kitty parsed from `spec.md`, the errors are opaque and don't tell you (a) which IDs spec-kitty actually parsed from the spec, or (b) which ref is the offender / why. This makes a simple formatting mismatch look like deep corruption, and is easy to misread as "spec-kitty invented FR IDs that don't match anything."

## Repro A — non-3-digit ID rejected with minimal context
1. Author `spec.md` with a sub-ID like `FR-003a` (a letter-suffixed requirement bullet).
2. Reference it in a WP and run `map-requirements --batch`.
3. Result: `{"error": "Stale or invalid refs in WP frontmatter", "stale_refs": {"WP05": ["FR-003a"]}, "hint": "Re-run with --replace ..."}`.
   - It flags `FR-003a` as invalid, but doesn't say *why* (not a recognized 3-digit `FR-###`) or list the valid IDs it parsed. The `--replace` hint doesn't address the actual cause (the ID shape).

## Repro B — "all unmapped" with zero diagnostics
- Separately (see related issue on map-requirements vs finalize), `finalize-tasks --validate-only` can report **every** FR as `unmapped` while emitting only `{"error": "Requirement mapping validation failed"}` plus the unmapped list — no indication of which WP dir/branch it scanned, how many WPs/refs it found, or which FR IDs it parsed from the spec. There's no way to tell a mapping bug from a parse bug from a path bug.

## Expected
- Reject (or clearly normalize) non-`FR-###` IDs with a message naming the rule ("requirement IDs must be 3-digit `FR-###`; got `FR-003a`").
- On any requirement-coverage failure, emit a diagnostic block: the FR IDs parsed from `spec.md`, the refs found per WP, and the specific unmatched pairs — so a formatting mismatch is obvious rather than looking like invented/orphaned IDs.

## Impact
Users (and agents) burn significant time guessing why coverage fails, and the failure mode reads like data corruption rather than a one-character ID-format mismatch.

## Suggestion
[...]

PR #2253, opened 2026-06-29
fix(#2066): surface parsed FR set + classify stale refs in map-requirements

rayjohnson:
## Summary

Closes the **`map-requirements` half of #2066** — "Opaque diagnostics when WP `requirement_refs` don't match spec.md FR IDs; no way to see the parsed FR set."

When the stale/invalid-refs gate trips, the error payload was only:

```json
{"error": "Stale or invalid refs in WP frontmatter", "stale_refs": {"WP05": ["FR-003a"]}, "hint": "Re-run with --replace ..."}
```

…which never says *why* a ref is bad or lists the IDs the parser actually found, so a one-character format mismatch (`FR-003a`) reads like invented/orphaned IDs or data corruption (the issue's Repro A).

## What changed

The stale-refs gate (`agent/tasks.py`) now emits:
- **`parsed_spec_ids`** — the FR/NFR/C IDs parsed from `spec.md`.
- **`stale_ref_reasons`** — per-WP classification of each offender into:
  - `malformed` — violates the `FR-NNN` / `NFR-NNN` / `C-NNN` format (e.g. letter-suffixed `FR-003a`, or an unfilled `<FR-XXX>` placeholder), and
  - `unknown_spec_id` — well-formed but not declared in `spec.md`.
- a **hint that names the format rule**.

Console (non-`--json`) output gains the same rule line + parsed-spec-IDs list.

Classification is a new pure helper `classify_stale_refs()` in `requirement_mapping.py` (keeps the command function's complexity down; directly unit-tested).

## Scope / relationship to prior work

- The **finalize-tasks** coverage-diagnostics half of #2066 (Repro B) already landed via the `single-planning-surface-authority` **FR-014 campsite fold** (`mission_finalize.py` now emits `requirement_refs_parsed`, `unmapped_functional_requirements`, etc.). This PR completes the *map-requirements* side the fold explicitly left as "future work."
- **Authoring-time** (`specify`/`plan`) FR-ID validation remains out of scope (also flagged as future work in the issue thread).

## Tests

- `requirement_mapping.py`: `TestClassifyStaleRefs` — malformed vs unknown split, `<FR-XXX>` placeholder → malformed, raw-case preserved + sorted.
- `test_map_requirements.py`: extended `test_reports_stale_invalid_refs` to assert the new payload keys; added `test_stale_refs_classify_format_vs_unknown` exercising Repro A (`FR-003a` malformed vs `FR-999` unknown).
- `ruff` + `mypy` clean on changed files; terminology guard green. No version bump (no `__init__.py` change); CHANGELOG entry under `[Unreleased] - 3.2.4`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - 3.2.4
 
-_No changes yet — entries land here as missions merge._
+### 🐛 Fixed
+
+- **`map-requirements` now explains *why* a WP `requirement_refs` entry is stale
+  instead of looking like data corruption (#2066).** When the stale/invalid-refs
+  gate trips, the `--json` payload (and console output) now surface the FR-ID set
+  parsed from `spec.md` (`parsed_spec_ids`), classify each offending ref per WP into
+  `malformed` (violates the `FR-NNN` / `NFR-NNN` / `C-NNN` format — e.g. a
+  letter-suffixed `FR-003a` or an unfilled `<FR-XXX>` placeholder) vs
+  `unknown_spec_id` (well-formed but not declared in the spec), and the hint names
+  the format rule. A one-character ID-format mismatch is now obvious rather than
+  reading like invented/orphaned IDs.
 
 ## [3.2.3] - 2026-06-29
 
--- a/src/specify_cli/cli/commands/agent/tasks.py
+++ b/src/specify_cli/cli/commands/agent/tasks.py
@@ -2489,6 +2489,7 @@ def map_requirements(
     from specify_cli.frontmatter import write_frontmatter
     from specify_cli.status import read_wp_frontmatter
     from specify_cli.requirement_mapping import (
+        classify_stale_refs,
         compute_coverage,
         normalize_requirement_refs_value,
         parse_requirement_ids_from_spec_md,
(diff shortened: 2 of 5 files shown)


stijn-dejongh:
Thanks @rayjohnson — picking this up now. We're kicking off a profile-loaded review squad on it, and we intend to land it. We'll handle the rebase onto current `main` and the merge ourselves (maintainer-edit), so nothing is needed on your side. We'll follow up on this thread with the aggregated review verdict.

stijn-dejongh:
@rayjohnson — **review complete: approved, landing as-is.** Two profile-loaded lenses ran on the branch rebased onto current `main`:

- **Correctness / coverage → APPROVE.** The `classify_stale_refs` partition is complete and disjoint with correct malformed-wins precedence; we traced the full `map_requirements` pipeline and confirmed no case-derivation mismatch between `stale_refs` (raw case) and the uppercased `malformed` set — the one way bucketing could have silently mislabeled. `parsed_spec_ids` is the genuine parsed FR/NFR/C set; the console and JSON paths agree. The 12 unit + CLI tests exercise the real behavior end-to-end (including the `FR-003a` malformed vs `FR-999` unknown split through the actual JSON payload).
- **Adversarial edge-cases → ROBUST.** 12 live probes + a De Morgan partition proof: lowercase↔uppercase bridging, `<FR-XXX>` placeholders (incl. all-placeholder WPs and empty `malformed`), non-string tokens, and empty inputs all bucket correctly; no token ever lands in both or neither bucket.

The rebase onto current `main` was clean (no conflicts), so we're merging it as-is — nothing needed on your side. Thanks for the focused fix and the direct unit test on the new helper.

**One cosmetic note for a possible future follow-up (non-blocking, not requested here):** duplicate raw tokens in a WP's refs aren't de-duplicated, so they'd appear twice in the same error bucket; and the console path omits the per-WP `malformed`/`unknown_spec_id` bucketing that the JSON payload exposes. Both are within the current contract — purely diagnostic polish.

Issue #2367, opened 2026-07-04
spec-kitty merge blocked by coordination worktree: uncommitted VCS-lock at claim + non-transactional rollback of coord status writes


stijn-dejongh:
## Summary

`spec-kitty merge` was blocked by the **coordination worktree** (`.worktrees/<mission>-coord`) holding uncommitted changes that a resync (`git reset --hard`) would destroy — the #1826 / NFR-002 safe-resync guard doing its job, but tripping on **spec-kitty's own tool-generated churn**, not real user work. Two distinct mechanisms, both encountered live.

Witnessing case: closeout of mission **`ci-suite-map-bind-01KWNPMP`** (Wave 0 CI suite-map, branch `tidy/ci-suite-map-2034`), 2026-07-04.

This is the same **class** as #1826 (dirty/stale coordination worktree blocks merge) but a **different mechanism** — #1826 (CLOSED, fixed by the ref-advance→worktree-resync fix) was *behind-HEAD after `update-ref`*. Neither mechanism below is that; both are **uncommitted / stale tool-writes** the resync guard then refuses to clobber.

## Mechanism A — VCS lock written at claim, never committed to the coordination branch

`meta.json` carried `vcs: git` + `vcs_locked_at`, written by `spec-kitty implement` at **CLAIM** time (the VCS lock), but **never committed** to the coordination branch. At merge time the coord worktree therefore has uncommitted `meta.json` churn; the resync guard refuses to `git reset --hard` over it (would destroy the lock), and merge is blocked. Had to **hand-commit `meta.json` in the coord worktree** to unblock.

**Root cause:** the VCS-lock write to coord `meta.json` is not auto-committed at claim.
**Ask:** auto-commit the VCS lock at claim (it is tool-generated metadata, not user work); OR have the resync guard treat spec-kitty's own generated metadata as auto-stashable/auto-committable churn (distinguish tool churn from real user changes).

## Mechanism B — rolled-back merge leaves stale status writes in the coord worktree

A merge attempt **rolled back** (on an unrelated squash conflict — see Note) and left **stale partial writes** in the coord worktree's `status.events.jsonl` + `status.json` (done-transition emissions from the aborted run). The next `--resume` then saw a dirty coord tree. Had to `git checkout -- status.events.jsonl status.json` in the coord worktree to revert the aborted run's emissions before `--resume` would proceed.

[...]

Issue #2222, opened 2026-06-27
Parallel-claim friction: implement writes vcs lock to meta.json then blocks next WP claim on the uncommitted self-write


stijn-dejongh:
## Symptom

`spec-kitty agent action implement` writes `vcs` / `vcs_locked_at` into the mission's `meta.json` when a WP is claimed, then **blocks the next WP claim** until that `meta.json` change is committed. With auto-commit disabled, claiming two dependency-free root WPs back-to-back fails after the first: the second claim refuses because the tree is dirty with the first claim's own `vcs` lock write.

Witnessed live on the **Doctrine Governance Fidelity** mission (context: PR #2211), where dependency-free roots were meant to be claimed in parallel.

## Why this is friction

The lock write is the lifecycle's **own** self-write, but it then trips the dirty-tree guard on the very next legitimate claim — the same self-write-then-self-guard pattern catalogued in #1914 / #2017.

## Suggested fix

- Auto-commit the `vcs` / `vcs_locked_at` lock write as part of the claim, **or**
- Don't gate sibling/parallel claims on the uncommitted lock (treat the lock as in-process state, not a dirty-tree blocker).

## Evidence

- Back-to-back parallel claim of dependency-free roots failed after the first on the Doctrine Governance Fidelity mission.
- PR #2211 context.

Parent epic: #1795 (Parallel-lane git mechanics & event-log merge semantics). Related: #1914 (governed/gate ops must be no-op-stable), #2017 (workflow guards blocking legitimate in-flight actions).

stijn-dejongh:
**Claimed by mission `lifecycle-tooling-friction`** (slug `lifecycle-tooling-friction-01KW4V6C`, branch `mission/lifecycle-tooling-friction`).

**Lane C** — Stop gating sibling claims on the vcs-lock self-write: exclude the lock file from the dirty-tree guard so a claim is not blocked by its own lock mutation. Only affects the `auto_commit=False` path.

Issue-matrix status: **`in-mission`** (addressed within this mission; per-WP `approved`, closes at mission `done`).

— Planner Priti

Issue #2105, opened 2026-06-24
Bug: spec-kitty `upgrade --project --yes` auto-commit stages only metadata.yaml


kentonium3:
# Bug: spec-kitty `upgrade --project --yes` auto-commit stages only metadata.yaml

## Summary

`spec-kitty upgrade --project --yes` (3.2.2) bumps the project version stamp and auto-commits, but the auto-commit stages **only** `.kittify/metadata.yaml`. Every other file the upgrade writes in the same run — `.kittify/*-manifest.json` updates plus newly-installed doctrine artifacts (e.g. the `adversarial-squad` skill and `doctrine-daphne` agent profile) — is left uncommitted in the working tree. The resulting `chore: apply spec-kitty upgrade changes` commit advances the version stamp ahead of the content backing it, leaving a dirty tree after a "successful" upgrade. Trusting the auto-commit yields a repo whose stamp claims 3.2.2 while the 3.2.2 artifacts are uncommitted — surprising for scripted/CI upgrade rolls.

### Prerequisites

- spec-kitty-cli 3.2.2, installed from `Priivacy-ai/spec-kitty@main` (SHA `aeb8dfc31`) — no PyPI 3.2.2 release exists yet
- A spec-kitty-initialized project at v3.2.1 with a clean working tree

### Steps

```bash
cd <project at v3.2.1>      # clean working tree
spec-kitty upgrade --project --yes
git status -s               # dirty: manifests + new artifacts uncommitted
git show --stat HEAD        # the upgrade commit contains ONLY metadata.yaml
```

### Expected Behavior

The auto-commit includes every file the upgrade created or modified, leaving a clean working tree.

### Actual Behavior

The auto-commit includes only `.kittify/metadata.yaml`; the manifest updates and the newly-installed skill/agent files remain uncommitted. Reproduced in 7 of 7 projects upgraded in a single roll (the uncommitted set varies only by each project's configured agent set).

```text
Current version: 3.2.1
Target version:  3.2.2
Created 7 tool surface(s)
  9 surface(s) not applicable, skipped.
Project is already up to date!
→ Auto-committed upgrade changes (1 files)

$ git show --stat HEAD
chore: apply spec-kitty upgrade changes (3.2.1 -> 3.2.2)
 .kittify/metadata.yaml | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

$ git status -s
 M .kittify/agent_profiles_manifest.json
[...]

PR #2387, opened 2026-07-04
fix(#2392): canonical per-checkout upgrade auto-commit seam (closes #2385, #1873; pins #2105)

### LINKED ISSUE #1873: upgrade: freshly synthesized worktree metadata may never be written after #1857
Follow-up from the panel review of #1857 (edge case, low likelihood).

In `_upgrade_worktrees` (`src/specify_cli/upgrade/runner.py:307-312`), when `ProjectMetadata.load(wt_kittify)` returns None, metadata is synthesized via `_create_initial_metadata(detected_version)`. Before #1857 the unconditional save guaranteed `metadata.yaml` was created. After #1857, if the detected version already equals the target and no migration records anything, the synthesized metadata is **never saved**; `_stamp_schema_version` then logs "schema_version stamp skipped: ... does not exist" and returns.

The common case is safe — `VersionDetector.detect_version` returns `"unknown"` when load fails, which never equals the target, so the dirty flag forces a save. The gap is only when detection succeeds *at* the target version. Still, this converts a previously self-healing path into a warning.

**Proposed fix:** set `worktree_metadata_dirty = True` when `wt_metadata` was synthesized (load returned None), so first-touch worktrees always get a metadata file.

Refs: #1857

### LINKED ISSUE #2385: spec-kitty upgrade leaves sibling worktrees dirty (auto-commits main only) → blocks coord-topology merge (#1826/NFR-002)
## Summary

`spec-kitty upgrade` applies migration changes to sibling worktrees (coord + lane) but **auto-commits only in the main project checkout**, leaving the worktrees dirty. A later `spec-kitty merge` on a coordination-topology mission then trips the `#1826 / NFR-002` "refusing to advance branch — worktree has uncommitted changes a `reset --hard` would destroy" backstop, and the merge cannot proceed.

## Severity

Blocks `merge` for any coordination-topology mission whose repo was `spec-kitty upgrade`-d (version bump) while its worktrees were live — i.e. the orchestrated → upgrade → accept → **merge** path.

## Root cause

- `spec-kitty upgrade` processes worktrees by default (`--no-worktrees` opts out). `MigrationRunner._upgrade_worktrees()` (`src/specify_cli/upgrade/runner.py`) iterates `.worktrees/*` and applies every migration whose `runs_on_worktrees` is true — but it has **no commit step**.
- The upgrade command's `_auto_commit_upgrade_changes()` (`src/specify_cli/cli/commands/upgrade.py`) is called with the **main `project_path` only** (upgrade.py:910, :1032). It never commits inside the worktrees it just modified.
- Result: after upgrade, each worktree carries uncommitted migration churn (e.g. `.gitignore`, `.kittify/metadata.yaml` version bump, the encoding-provenance untracking). The main checkout gets auto-committed; the worktrees do not.
- `spec-kitty merge` advances the coord/mission branch (checked out in the `…-coord` worktree) via `git reset --hard`. The `#1826 / NFR-002` guard correctly refuses to clobber the worktree's uncommitted changes, so the merge halts at the first lane.

## Reproduction (observed)

Coordination-topology mission (`topology: coord`, `pr_bound: true`) with live worktrees; repo upgraded `3.2.1 → 3.2.4`, then `spec-kitty merge`:

```
Checking and merging lane-a...
✗ lane-a: Refusing to advance branch 'kitty/mission-<slug>'
   (…-coord worktree has it checked out and holds uncommitted local changes
    that a resync (git reset --hard) would destroy) (#1826 / NFR-002).
  Dirty entries:
     M .gitignore
    D  .kittify/encoding-provenance/global.jsonl
     M .kittify/metadata.yaml
```

The three dirty entries are exactly the upgrade's own migration output — none are mission deliverables or status/event data — and they match what the main checkout auto-committed.

## Proposed fix

Make upgrade's worktree handling symmetric with the main checkout: after `_upgrade_worktrees` applies migrations in a worktree, auto-commit (or revert) the resulting changes **in that worktree**, reusing the same eligibility + `safe_commit` machinery `_auto_commit_upgrade_changes` uses for `project_path`. Committing onto the coord/lane branch (via `safe_commit` with the appropriate guard capability) leaves every worktree clean, so a subsequent `merge` proceeds without manual git.

(Alternatively / additionally: `merge` could offer to auto-drop *ignorable* `.kittify/` housekeeping dirt in the coord worktree with operator consent — but the root fix belongs in upgrade, which should not leave worktrees it mutated in a dirty state.)

## Workaround (for repos already hit)

The dirt is regenerable upgrade housekeeping; the merge target descends from the coord tip, so a `reset --hard` in the worktree just drops the churn (`git checkout`/`reset` also restores the untracked-then-deleted encoding-provenance log):

```
git -C .worktrees/<slug>-coord reset --hard HEAD
spec-kitty merge --resume
```

## Related

`#1826` (the NFR-002 worktree-clobber guard doing its job), and the orchestrated-mission → upgrade → accept → merge path more broadly (see #2369, #2376). Discovered at the merge gate immediately after those accept-path fixes.

rayjohnson:
Closes #2385. Closes #1873. Regression-pins #2105. Implements the canonical-seam spine of epic #2392 (architect-alphonso's high-level design posted on the epic).

## Scope — expanded per review

The original PR was a point-fix for #2385 at the CLI layer. Per @stijn-dejongh's review and the #2392 consolidation, it now lands the **one-seam fix for the three members the design shows genuinely collapse into one seam** (#2385 + #1873 + #2105), using the design's target shape.

## The canonical invariant (epic #2392)

> Every path an upgrade run writes or migrates, in every checkout it touches (main + each live `.worktrees/*`), must end in exactly one auto-commit — with the commit-set derived from that checkout's real `git status --porcelain` diff against a pre-write baseline, not a hardcoded file list — or be intentionally reverted before the command returns.

## The fix — one routine, one enumeration, two symmetric call sites

**Extraction (Slice A).** The commit routine (porcelain baseline diff, directory expansion, eligibility filter, `safe_commit(UPGRADE_BOOKKEEPING)`) moves from `cli/commands/upgrade.py` to `specify_cli/upgrade/autocommit.py` as `commit_touched_checkout(checkout, baseline_paths, from_version, to_version)` — parameterized purely on a checkout path, exactly as the design specifies. The guard-capability allowlist moves with it (mirroring the MERGE_BOOKKEEPING one-seam consolidation).

**Per-checkout application (Slice B).**
- *Main*: unchanged call, from the CLI (both the no-migrations and migrations paths).
- *Each worktree* (#2385): `MigrationRunner._upgrade_worktrees` — the owner of the canonical worktree enumeration and of the worktree writes — captures a per-worktree porcelain baseline **before** that worktree's writes and calls the extracted routine **after** them, landing the commit on that worktree's own branch. No second enumeration in the CLI.
- *#1873*: metadata synthesized from `None` is marked dirty at synthesis, so it is saved (and committed) even when the detected version already equals the target — restoring the #1857 self-healing path.
- *#2105*: as the design found, the commit-set on this tree is already porcelain-derived, so the residual defect was scope (main-only), not derivation. Now regression-pinned: a test asserts manifest updates + newly-installed skills land in the commit-set, guarding against regression to the hardcoded-list past.

**Edge cases from the design's risk list.**
- Pre-existing uncommitted work in a live lane worktree (in-flight WP edits) is never swept in — per-worktree baseline diff (test: real git worktree with WIP + README edits).
- Per-worktree `manual_review_required` migrations skip **that worktree's** commit with a warning (finer-grained than the main-only global gate).
- Detached-HEAD checkout → skip the commit with a warning; never guess a ref.
- `--dry-run` remains a strict no-op per checkout (test: HEAD unchanged, tree clean).
- Root `.gitignore` is now commit-eligible: the gitignore-backfill migrations write it, and the old root-level-file filter silently dropped it — leaving exactly the merge-blocking churn #2385 reports. (Caught by the new invariant test, which runs real migrations.)

**Invariant test (Slice C).** After `runner.upgrade(..., auto_commit=True)` on a repo with a live worktree: the worktree carries no porcelain dirt beyond what pre-existed; everything the run wrote to main is within the CLI commit's reach.

## Deliberately fenced OUT — #2367 (per the design's §7, honestly)

The design's verdict is that this cluster is *one shared invariant realized at three seams*, and forcing #2367 through the upgrade helper would be the split-brain outcome in the opposite direction:
- **#2367-A** (VCS-lock at implement-claim) is a genuine policy decision — the current exclusion is a deliberate, race-motivated choice (#2222/C-003). Committing the lock here would silently reverse it.
- **#2367-B** (rolled-back merge leaves stale coord status writes) is the rollback dual, living in the merge executor's snapshot capture set.

Both stay open on #2392 as sibling members with their own seams (the design suggests a fenced Slice D / follow-up).

## Boundary noted (not changed here)

`_run_upgrade_surface_repair` runs **after** the auto-commit on both CLI paths, so surface-repair writes (healed skills/profiles) are intentionally left for the operator — that ordering pre-dates this PR and has its own interactive/drift-policy semantics. If the invariant should swallow repair writes too, that's a deliberate reordering decision I'd rather take separately.

## Tests

- `tests/upgrade/test_upgrade_worktree_commit.py` (real git worktrees, rewritten to the runner seam): #2385 commit-on-own-branch + main untouched; baseline protects pre-existing WIP; #1873 synthesis-save + commit; dry-run no-op; Slice-C invariant with real migrations.
- `tests/upgrade/test_upgrade_auto_commit_unit.py` retargeted to `upgrade.autocommit`; adds the #2105 regression pin and the detached-HEAD guard.
- Full sweep: 1127 upgrade-related tests green (`-k upgrade` across the repo), `tests/architectural` guard-capability + no-dead-modules gates green, `ruff` + `mypy` clean.

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -123,6 +123,17 @@ speed up how fast we catch one.
   `IGNORED` surfaces (fresh `spec-kitty init` gitignores them), and a sibling
   backfill migration (`3.2.4_runtime_dirs_gitignore_backfill`) adds them on
   `spec-kitty upgrade` for already-initialised projects.
+- **`spec-kitty upgrade` no longer leaves sibling worktrees dirty, which was
+  blocking coord-topology `spec-kitty merge` (#2385).** Upgrade applies migrations
+  across worktrees (coord + lane) by default, but its auto-commit ran on the main
+  checkout only — so each worktree was left with uncommitted migration churn
+  (`.gitignore`, `.kittify/metadata.yaml`, encoding-provenance untracking). A later
+  `spec-kitty merge` then tripped the #1826/NFR-002 guard (it refuses to
+  `reset --hard` a coord/lane worktree with uncommitted changes), halting the
+  merge. Upgrade now captures a per-worktree baseline up front and auto-commits
+  each worktree's *new* upgrade churn on its own branch, mirroring the main
+  checkout; the baseline ensures any pre-existing uncommitted work in a worktree
+  is never swept in.
 - **Guard/gate friction hotfixes (#2346 / #2324, #1834).**
   - **Subtask guard no longer misattributes a later WP's checkboxes (#2346, also closes #2324).** `_check_unchecked_subtasks` entered a WP's section on *any* heading that merely mentioned its id, so a dependent heading like `### WP03 — … (depends: WP01, WP02)` re-entered WP01/WP02's section and harvested WP03's unchecked `- [ ] T0xx` rows as the earlier WP's blockers — spuriously blocking that WP's lane transition. A heading now belongs to the WP named by its **first** `WPxx` token, not any mention.
   - **`grep_absence` negative invariants accept an optional path-scope (#1834).** The acceptance gate ran `grep -r <pattern> .` over the whole repo, so a negative-invariant pattern that a mission's own spec/plan/WP prose mentioned false-positived as `still_present`. `NegativeInvariant` now carries an optional `scope` (whitespace-separated repo-relative search roots); when set, the grep runs only under those paths. Default (unscoped) preserves the whole-repo search, and `scope` is omitted from serialization when unset so existing matrices are untouched.
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -123,17 +123,37 @@ speed up how fast we catch one.
   `IGNORED` surfaces (fresh `spec-kitty init` gitignores them), and a sibling
   backfill migration (`3.2.4_runtime_dirs_gitignore_backfill`) adds them on
   `spec-kitty upgrade` for already-initialised projects.
-- **`spec-kitty upgrade` no longer leaves sibling worktrees dirty, which was
-  blocking coord-topology `spec-kitty merge` (#2385).** Upgrade applies migrations
-  across worktrees (coord + lane) by default, but its auto-commit ran on the main
-  checkout only — so each worktree was left with uncommitted migration churn
-  (`.gitignore`, `.kittify/metadata.yaml`, encoding-provenance untracking). A later
-  `spec-kitty merge` then tripped the #1826/NFR-002 guard (it refuses to
-  `reset --hard` a coord/lane worktree with uncommitted changes), halting the
-  merge. Upgrade now captures a per-worktree baseline up front and auto-commits
-  each worktree's *new* upgrade churn on its own branch, mirroring the main
-  checkout; the baseline ensures any pre-existing uncommitted work in a worktree
-  is never swept in.
+- **Upgrade-worktree coherence: one canonical per-checkout auto-commit seam
+  (epic #2392 spine — closes #2385, #1873; regression-pins #2105).** The
+  `spec-kitty upgrade` commit routine (porcelain-baseline-derived commit-set,
+  directory expansion, eligibility filter) is extracted from the CLI into
+  `specify_cli.upgrade.autocommit.commit_touched_checkout()` and applied
+  symmetrically to **every checkout the run touches** over the runner's single
+  worktree enumeration:
+  - **#2385** — worktree migration churn was never committed (auto-commit ran on
+    the main checkout only), so each coord/lane worktree was left dirty and a
+    later `spec-kitty merge` tripped the #1826/NFR-002 guard. The runner now
+    captures a per-worktree porcelain baseline before that worktree's writes and
+    commits its *new* churn on its own branch; pre-existing uncommitted work
+    (e.g. in-flight WP edits) is never swept in, and per-worktree
+    `manual_review_required` migrations skip that worktree's commit with a
+    warning. Root `.gitignore` churn — exactly what the gitignore-backfill
+    migrations write — is now commit-eligible instead of being dropped by the
+    root-level-file filter.
+  - **#1873** — freshly synthesized worktree metadata is persisted (and
+    committed) even when the detected version already equals the target,
+    restoring the #1857 self-healing path.
+  - **#2105** — already porcelain-derived in this tree; now regression-pinned by
+    a test asserting manifest updates and newly-installed skills land in the
+    commit-set (guards against regressing to a hardcoded file list).
+  - A detached-HEAD checkout skips the auto-commit with a warning instead of
+    guessing a ref. `--dry-run` remains a strict no-op per checkout. An
+    invariant test asserts that after `upgrade`, no touched checkout carries
+    porcelain dirt beyond what pre-existed. **#2367's two mechanisms are
+    deliberately NOT routed through this seam** (per the #2392 architect
+    design): the implement-claim VCS-lock policy (#2222/C-003 race decision)
+    and the merge-rollback snapshot capture set are separate seams sharing the
+    same invariant.
 - **Guard/gate friction hotfixes (#2346 / #2324, #1834).**
   - **Subtask guard no longer misattributes a later WP's checkboxes (#2346, also closes #2324).** `_check_unchecked_subtasks` entered a WP's section on *any* heading that merely mentioned its id, so a dependent heading like `### WP03 — … (depends: WP01, WP02)` re-entered WP01/WP02's section and harvested WP03's unchecked `- [ ] T0xx` rows as the earlier WP's blockers — spuriously blocking that WP's lane transition. A heading now belongs to the WP named by its **first** `WPxx` token, not any mention.
   - **`grep_absence` negative invariants accept an optional path-scope (#1834).** The acceptance gate ran `grep -r <pattern> .` over the whole repo, so a negative-invariant pattern that a mission's own spec/plan/WP prose mentioned false-positived as `still_present`. `NegativeInvariant` now carries an optional `scope` (whitespace-separated repo-relative search roots); when set, the grep runs only under those paths. Default (unscoped) preserves the whole-repo search, and `scope` is omitted from serialization when unset so existing matrices are untouched.
--- a/src/specify_cli/cli/commands/upgrade.py
+++ b/src/specify_cli/cli/commands/upgrade.py
@@ -40,7 +40,6 @@ from __future__ import annotations
 
 import json
 import os
-import subprocess
 import sys
 from datetime import UTC, datetime
 from pathlib import Path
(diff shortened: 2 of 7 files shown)


stijn-dejongh:
Thanks for the fast, well-documented fix here, @rayjohnson — the root-cause writeup (worktree auto-commit step simply missing from `_upgrade_worktrees`) and the baseline-capture approach in `_auto_commit_worktree_upgrade_changes` are both correct and clearly reasoned.

Flagging for coordination before this lands, cc @MOES-Media:

We just opened umbrella epic #2392 ("upgrade-worktree coherence — one canonical fix for state/gitignore propagation across coord/lane/main") to consolidate this bug with three siblings that are the *same underlying defect* wearing different clothes — state changes not propagated/committed coherently across the main checkout, the coordination worktree, and lane worktrees:

- #2385 — this PR's target: worktree auto-commit step missing from `_upgrade_worktrees`
- #2367 — coord-worktree dirt from two further mechanisms (VCS lock never committed to the coord branch at claim time; rolled-back `merge` leaves stale partial `status.events.jsonl`/`status.json` writes)
- #1873 — synthesized worktree metadata never saved to disk when version already matches target (dirty flag not set)
- #2105 — even the main-checkout auto-commit only stages a subset of the files that actually changed

Left as four independently-authored point-fixes, each aware only of its own reproduction path, we're near-guaranteed to re-diverge — one PR fixes the main-checkout commit set, another fixes worktree iteration, a third fixes metadata persistence, none aware of the others' invariants — and reproduce exactly the split-brain / partial-fix regression class this repo has hit before. That's the concern here, not the quality of this specific patch.

A rudimentary high-level design for the canonical seam (the `spec-kitty upgrade` commit/propagation path: `_auto_commit_upgrade_changes()` + `MigrationRunner._upgrade_worktrees()` / metadata persistence in `runner.py`) is being posted to #2392 now.

Given that, could you pick one of:

1. **Expand this PR** to cover the whole cluster (#2385 + #2367 + #1873 + #2105) consistently, using the canonical-seam design landing on #2392 as the target shape — reusing the baseline-capture + per-worktree auto-commit machinery you've already built here as the foundation. This PR is a strong start on mechanism #2385; it just needs to grow to close the seam for all four rather than one.
2. **Close this PR** and relaunch the fix as a single mission run against #2392, so the whole cluster lands together under one design/review pass instead of four sequential partial patches.

Either is fine — we just want to avoid merging a correct-but-partial fix that leaves the other three mechanisms to regress into split-brain later. Happy to help scope whichever direction you pick.

stijn-dejongh (changes requested):
Direction and point-fix are well executed, but needs expansion to sibling issues ( new cluster logged ) to avoid parallel logic path regression.

rayjohnson:
@stijn-dejongh went with **option 1 — expanded this PR** to the #2392 canonical seam, following architect-alphonso's design as the target shape (commit 6d188d20f; PR title/body updated).

What landed, mapped to the design:
- **Slice A**: commit routine extracted to `upgrade/autocommit.py::commit_touched_checkout()`, #2105 regression-pinned (manifest + newly-installed skills asserted in the commit-set). Guard-capability allowlist follows the seam, mirroring the MERGE_BOOKKEEPING consolidation.
- **Slice B**: the runner owns per-worktree baselines and commits over its single enumeration (#2385), and the #1873 synthesis-dirty flag is fixed in the same change — including the design's one-integration-test shape (live worktree, version==target, both checkouts end clean, pre-existing WIP preserved).
- **Slice C**: invariant test — after `upgrade`, no touched checkout carries porcelain dirt beyond what pre-existed. Running it against real migrations caught a bonus gap: root `.gitignore` (exactly what the gitignore-backfill migrations write) was silently dropped by the eligibility filter's root-level rule; now commit-eligible.
- **§7 respected**: #2367-A/B are deliberately NOT routed through this seam — A is the #2222/C-003 race-motivated policy decision, B is the merge-rollback capture set. Both stay on #2392 for their own seams; happy to pick up B as a follow-up PR if you want it next.

One boundary flagged in the PR body: `_run_upgrade_surface_repair` runs after the auto-commit on both CLI paths (pre-existing ordering), so repair writes are still left for the operator — reordering that felt like a separate decision, not something to slip into this PR.

stijn-dejongh:
Picking this up for a maintainer landing pass — rebase + in-depth review against the #2392 seam design and the earlier guidance. Will post findings shortly.

stijn-dejongh:
## Maintainer landing pass — complete

Rebased onto current `main` (CHANGELOG conflict resolved — both the #2388 entry that landed while this PR was open and this PR's expanded #2392 entry are preserved in correct order). Adversarial squad run (architect-alphonso, debugger-debbie, reviewer-renata). Three folds applied and pushed to `rayjohnson/spec-kitty:fix/2385-upgrade-commits-worktrees`.

### Squad verdict

**Unanimous SHIP-WITH-FOLDS.** The design is architecturally sound: `upgrade/autocommit.py` is correctly placed in the upgrade bounded context, the "callers own the baseline capture" pattern is correctly respected at all call sites, the `UPGRADE_BOOKKEEPING` guard-capability allowlist migration to `autocommit.py` mirrors the `MERGE_BOOKKEEPING` consolidation pattern, and the three bug fixes (#2385, #1873, #2105) are correctly realized. No structural blockers.

### Folds applied

| Severity | Reviewer | Finding | Resolution |
|----------|----------|---------|------------|
| HIGH | debugger-debbie | `test_upgrade_invariant_every_touched_checkout_ends_clean` was vacuous with an empty registry — `runner.upgrade()` takes the no-migrations branch which skips `_upgrade_worktrees` when `from_version != target_version` (3.2.1 ≠ 3.2.9), so the test passed with no assertions exercised | Registered a stub migration with `runs_on_worktrees=True` + `try/finally MigrationRegistry.clear()` + inline comment clarifying the main-checkout seam split |
| MEDIUM | debugger-debbie | No registry isolation — tests that call `runner.upgrade()` inherit whatever registry state prior tests left, causing order-fragility | `MigrationRegistry.clear()` added at start of invariant test; `try/finally` ensures cleanup on failure |
| MEDIUM | debugger-debbie | `test_synthesized_worktree_metadata_is_saved_when_version_matches_target` checked cleanliness but not that the upgrade commit actually landed on the worktree branch — a regression where `commit_touched_checkout` returned early would still pass | Added `assert "spec-kitty upgrade" in _git_out(wt, "log", "-1", "--pretty=%s")` |

### Not folded

- Alphonso [MINOR] — CLI calls private `_upgrade_worktrees` directly (pre-existing, predates this PR). Tracked as tech-debt.
- Renata [LOW] — CHANGELOG `.gitignore` widening already documented in the PR's #2385 sub-bullet (`"Root .gitignore churn... is now commit-eligible"`). No additional entry needed.
- Debbie [LOW] — `check_output` exception fallback path (`destination_ref = "main"`) not unit-tested. Acceptable for a low-probability hardware/env failure path; the safe_commit call underneath it would still fail fast if the branch is wrong.

### Verification

568 tests pass (`tests/upgrade/` + `tests/architectural/test_guard_capability_call_sites.py`). `ruff check` clean on modified files.

CI is running on the updated branch. **Ready to merge on green.** Thanks @rayjohnson — excellent expanded implementation, both the baseline-capture design and the one-seam consolidation are exactly right.

stijn-dejongh:
## Follow-up ops — complete

Second landing fold pushed (`39584f8af`) covering the two unfixed findings from the first pass, plus the #2392 SSOT verification.

### #2392 seam adherence — FAITHFUL

Architect-alphonso verified the implementation against the design posted to #2392. Verdict: **FAITHFUL** — exact match on invariant, interface, call-site structure, and fenced exclusions (#2367-A/B cleanly absent). No deviations.

### Folds applied

| Severity | Finding | Resolution |
|----------|---------|------------|
| MINOR (alphonso) | `upgrade.py:728` called private `MigrationRunner._upgrade_worktrees()` directly, bypassing the public interface | Added `upgrade_worktrees_only(target_version, dry_run, auto_commit)` public delegating method to `MigrationRunner`; updated CLI call site; added 2 delegation-contract tests |
| LOW (debbie) | `commit_touched_checkout`'s `except Exception` fallback to `destination_ref = "main"` had no unit test — a checkout on a non-main branch where `check_output` fails would silently commit to `"main"` | Added `test_commit_falls_back_to_main_when_branch_detection_fails` asserting `committed=True`, `warning=None`, `target.ref == "main"` |

### Verification

571 tests pass (`tests/upgrade/` + `tests/architectural/test_guard_capability_call_sites.py`). `ruff` clean. **Ready to merge on green.**

PR #2414, opened 2026-07-06
fix(#2413): session banner compares versions by ordering, not inequality

### LINKED ISSUE #2339: upgrade dry-run JSON contract rejects dotted migration_ids (first live offender: 3.2.0rc45_retire_standalone_skill_surface)
## What

`tests/specify_cli/cli/commands/test_upgrade_command.py::test_project_migration_needed_project_dry_run_json_contract` fails on current main (`2c8e1dc43`):

```
jsonschema.exceptions.ValidationError: '3.2.0rc45_retire_standalone_skill_surface' does not match '^[a-z0-9_]{1,128}$'
On instance['pending_migrations'][57]['migration_id']
```

The dry-run JSON contract pins `migration_id` to `^[a-z0-9_]{1,128}$` (no dots), but the in-tree migration id convention is **split**: some ids are dotted (`3.2.0rc35_spk_skill_pack`, `3.2.0rc43_retire_profile_context_command`, `3.2.0rc45_retire_standalone_skill_surface`, all the `0.x`/`1.x` era ids), some underscored (`3_2_0rc39_refresh_orientation_block`, `3_3_0_op_record_schema_v2`).

Introduced by `2146506bd` ("Remove standalone governance verbs", 2026-07-03) adding `m_3_2_0rc45_retire_standalone_skill_surface.py` with `migration_id = "3.2.0rc45_retire_standalone_skill_surface"`.

## Reproduce

```
PWHEADLESS=1 pytest tests/specify_cli/cli/commands/test_upgrade_command.py::test_project_migration_needed_project_dry_run_json_contract
```

Reproduced on two independent main-based branches during the 2026-07-04 landing pass (it reds local runs of PR branches rebased onto current main, e.g. #2336/#2238 — it is unrelated to those diffs).

## Decision needed (two-authority conflict)

Either the **contract pattern** is wrong (should admit dots: the dotted convention predates it and is the majority) or the **id convention** is wrong (ids should be underscore-only, matching module names `m_3_2_0rc45_*`). One authority must win and the other side gets migrated/normalized — half-fixing only rc45 leaves the split-brain in place. Note `migration_id` values may already be persisted in consumer-project state (applied-migrations ledgers), so renaming shipped ids needs a compat read path.

Found during the landing pass while adjudicating a local test failure as pre-existing (campsite-cleaning standing order: pre-existing/orphan CI failures get root-caused and filed, not stepped around).

### LINKED ISSUE #2413: session banner says "Upgrade available: <older version>" when the installed CLI is newer than PyPI's cached latest — inequality instead of ordering
## Symptom

The session-presence orientation banner shows e.g. `⚠ Upgrade available: 3.2.2 — run spec-kitty upgrade --cli to update.` while the installed CLI is **3.2.4** — recommending a downgrade.

## Root cause

`src/specify_cli/session_presence/manager.py` decides the `upgrade-available` health with an **inequality**, not an ordering, in both the normal and the exception-fallback branch:

```python
elif avail and avail != current:
    health = "upgrade-available"
...
except Exception:
    health = "upgrade-available" if avail and avail != current else "healthy"
```

`avail` comes from `UpgradeChecker.get_available_version()` — the cached PyPI `latest_version` from `~/.kittify/last-cli-check.json` (1-hour TTL, but stale values are served indefinitely by design: "stale but better than None"). So whenever the installed version is **ahead of** the cached PyPI latest — freshly released version not yet in cache, pre-release/rc installs, dev installs from source — the banner claims an "upgrade" is available *to the older version*.

## Fix shape

Compare with `packaging.version.Version` ordering (the repo already depends on `packaging` — `upgrade/runner.py` uses it for downgrade refusal):

```python
def _upgrade_is_available(avail: str | None, current: str) -> bool:
    if not avail:
        return False
    try:
        return Version(avail) > Version(current)
    except InvalidVersion:
        return False
```

…used by both branches in `manager.py`. A unit test matrix: `avail > current` → upgrade-available; `avail == current` → healthy; `avail < current` (this bug) → healthy; unparseable → healthy.

Small, isolated fix — happy to send a PR.

### LINKED ISSUE #2417: Consolidate the 3 near-duplicate version-comparison primitives (upgrade.py / upgrade_probe.py / session_presence)
## Finding

Found by architect-alphonso during a lightweight review squad on PR #2414 (fix(#2413): session banner compares versions by ordering, not inequality), then independently confirmed against the source.

The codebase now has **three** independent implementations of "is version A newer than version B":

1. `src/specify_cli/cli/commands/upgrade.py:303-311` — `_version_is_newer(latest: str | None, installed: str) -> bool` (pre-existing).
2. `src/specify_cli/core/upgrade_probe.py:_classify()` (~lines 181-208) — a richer variant that classifies into `ALREADY_CURRENT` / `AHEAD_OF_PYPI` / `NO_UPGRADE_PATH` / upgrade-available states. Its `AHEAD_OF_PYPI` docstring literally describes the #2413 root cause (an RC/dev build ahead of what PyPI has published) — meaning this exact defect class had already been solved once in this codebase before #2413 was filed against a *different* call site.
3. `src/specify_cli/session_presence/manager.py:31-43` — the new `_upgrade_is_available(avail: str | None, current: str) -> bool`, added by PR #2414 to fix #2413.

All three wrap `packaging.version.Version(...) > Version(...)` with slightly different edge-case handling (guard-on-`None`-only vs. falsy-empty-string; `except InvalidVersion` vs. bare `except Exception`).

## Why this matters

Not a correctness bug today — all three call sites currently behave correctly for their own inputs (independently verified for #3 during the #2414 landing pass: every input shape checked out, including epoch/local-identifier versions via `packaging`'s own ordering). But it's the same shape of drift that already caused #2413: three copies of "the same rule enforced independently" is exactly the pattern this project's own governance charter calls out as a canonical-source violation, and the next edge case fix (or PyPI/`packaging` behavior change) risks being applied to only one of the three copies.

## Suggested approach (not prescriptive — the assignee should weigh)

Extract one shared `is_version_newer(candidate: str | None, current: str) -> bool` (or reuse/generalize `upgrade_probe._classify`, which is already the richest of the three) into a common module (e.g. `specify_cli.core.version_compare`), and point all three call sites at it. `upgrade.py` and `session_presence/manager.py` only need the boolean; `upgrade_probe.py` needs the richer classification and could either keep its own logic (if genuinely a distinct concern) or become the canonical implementation the boolean helper wraps.

Not blocking anything — filed as a tracked follow-up per the project's "any friction/debt discovered during landing gets filed immediately" practice, not left as a passing comment on PR #2414.

rayjohnson:
Fixes #2413.

## The bug

The session-presence orientation banner decided `upgrade-available` with a **bare inequality** in both health branches of `SessionPresenceManager._build_content` (`src/specify_cli/session_presence/manager.py`):

```python
elif avail and avail != current:          # compat-plan branch
...
health = "upgrade-available" if avail and avail != current else "healthy"   # exception fallback
```

`avail` is the cached PyPI `latest_version` (`~/.kittify/last-cli-check.json`, 1-hour TTL, stale values served by design). Whenever the installed CLI is **newer** than that cache — a fresh release not yet fetched, an rc/dev install — the banner says `⚠ Upgrade available: 3.2.2` while 3.2.4 is running, recommending a downgrade. With 3.2.4 released today, every machine with a stale cache shows this.

## The fix

One helper, used by both branches:

```python
def _upgrade_is_available(avail: str | None, current: str) -> bool:
    if not avail:
        return False
    try:
        return Version(avail) > Version(current)
    except InvalidVersion:
        return False
```

`packaging` is already a dependency (`upgrade/runner.py` uses it for downgrade refusal). Unparseable/None/empty versions are treated as no-upgrade, preserving the module's never-block-session-start contract.

## Tests

- Behavioral regressions in `TestBuildContent`: cached latest **older** than installed → `healthy`, in both the compat-plan branch and the exception-fallback branch (the #2413 report).
- `TestUpgradeIsAvailable` matrix: newer → True; equal → False; older (the bug) → False; rc ordering both directions (`3.2.4` vs `3.2.5rc1` → False, `3.2.4` vs `3.2.4rc39` → True); None/empty/unparseable → False.
- Full `tests/specify_cli/session_presence/` + `test_session_start.py`: 170 passed. `ruff` + `mypy` clean.

---

Closes #2417 (folded into this PR by a maintainer landing-pass op — see comments below).

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -15,7 +15,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - 3.2.5
 
-_No changes yet — entries land here as missions merge._
+### 🐛 Fixed
+
+- **Session banner no longer recommends a downgrade (#2413).** The
+  session-presence health check decided "upgrade-available" with a bare
+  inequality (`avail != current`), so any machine whose installed CLI was
+  *newer* than the cached PyPI latest — a fresh release not yet in the 1-hour
+  cache, or an rc/dev install — got "⚠ Upgrade available: 3.2.2" while running
+  3.2.4. Both health branches now use `packaging.Version` ordering
+  (`avail > current`); unparseable versions are treated as no-upgrade so
+  session start never breaks on a weird cache value.
 
 ## [3.2.4] - 2026-07-05
 
--- a/kitty-specs/cli-upgrade-nag-lazy-project-migrations-01KQ6YDN/contracts/compat-planner.json
+++ b/kitty-specs/cli-upgrade-nag-lazy-project-migrations-01KQ6YDN/contracts/compat-planner.json
@@ -193,7 +193,7 @@
         "required": ["migration_id", "target_schema_version", "description"],
         "additionalProperties": false,
         "properties": {
-          "migration_id": { "type": "string", "pattern": "^[a-z0-9_]{1,128}$" },
+          "migration_id": { "type": "string", "pattern": "^[a-z0-9_.]{1,128}$" },
           "target_schema_version": { "type": "integer", "minimum": 0, "maximum": 1000 },
           "description": { "type": "string", "maxLength": 256 },
           "files_modified": {
--- a/src/specify_cli/cli/commands/upgrade.py
+++ b/src/specify_cli/cli/commands/upgrade.py
@@ -61,6 +61,7 @@ from specify_cli.cli.commands._teamspace_mission_state_gate import (
 )
 from specify_cli.core.commit_guard import GuardCapability
 from specify_cli.core.env import is_truthy
+from specify_cli.core.version_compare import is_version_newer
 from specify_cli.git.commit_helpers import safe_commit
 
 
@@ -300,17 +301,6 @@ def _run_cli_mode(
 # ---------------------------------------------------------------------------
 
 
-def _version_is_newer(latest: str | None, installed: str) -> bool:
-    if latest is None:
-        return False
-    try:
-        from packaging.version import Version
-
-        return Version(latest) > Version(installed)
-    except Exception:  # noqa: BLE001
-        return False
-
-
 def _agent_check_payload() -> dict[str, object]:
     """Return machine-readable upgrade readiness for agent prompt preambles."""
     from specify_cli.compat import (
@@ -363,7 +353,7 @@ def _agent_check_payload() -> dict[str, object]:
         payload["reason"] = "nag_disabled"
         return payload
 
-    if not _version_is_newer(latest_version, cli_status.installed_version):
+    if not is_version_newer(latest_version, cli_status.installed_version):
         return payload
 
     if is_truthy(os.environ.get(ENV_UPGRADE_DISABLED)):
--- a/src/specify_cli/core/upgrade_probe.py
+++ b/src/specify_cli/core/upgrade_probe.py
@@ -29,7 +29,8 @@ from enum import StrEnum
 from typing import Any
 
 import httpx
-from packaging.version import InvalidVersion, Version
+
+from specify_cli.core.version_compare import is_version_newer, try_parse_version
 
 PYPI_JSON_URL = "https://pypi.org/pypi/spec-kitty-cli/json"
 """PyPI's standard JSON metadata endpoint for the ``spec-kitty-cli`` package."""
@@ -190,21 +191,18 @@ def _classify(
     PEP 440 version. Network/parse failures are handled upstream in
     :func:`probe_pypi`.
     """
-    try:
-        installed_ver = Version(installed)
-    except InvalidVersion:
+    installed_ver = try_parse_version(installed)
+    if installed_ver is None:
         return UpgradeChannel.UNKNOWN
 
-    latest_ver: Version | None
-    try:
-        latest_ver = Version(latest)
-    except InvalidVersion:
-        # Latest is malformed — fall through to releases-membership check.
-        latest_ver = None
+    # ``latest`` may be malformed — try_parse_version falls through to the
+    # releases-membership check below (via is_version_newer returning False)
+    # rather than raising.
+    latest_ver = try_parse_version(latest)
 
     if latest_ver is not None and installed_ver == latest_ver:
         return UpgradeChannel.ALREADY_CURRENT
-    if latest_ver is not None and installed_ver > latest_ver:
+    if is_version_newer(installed, latest):
         return UpgradeChannel.AHEAD_OF_PYPI
 
     # installed != latest (or latest unparseable). Check releases membership.
--- a/src/specify_cli/core/version_compare.py
+++ b/src/specify_cli/core/version_compare.py
@@ -1,0 +1,58 @@
+"""Canonical version-comparison primitives shared across upgrade surfaces (#2417).
+
+Before this module existed, "is candidate version newer than current version"
+was implemented independently in three places:
+
+- ``cli/commands/upgrade.py::_version_is_newer`` — caught the broad
+  ``Exception`` (``# noqa: BLE001``) rather than the specific parse failure.
+- ``core/upgrade_probe.py::_classify`` — a richer PyPI-channel classifier
+  that needs version *parsing* and *ordering* as building blocks alongside
+  releases-membership checks; only its comparison sub-logic is unified here.
+- ``session_presence/manager.py::_upgrade_is_available`` — added to fix #2413,
+  where a bare inequality (``avail != current``) reported "upgrade available"
+  whenever the cached PyPI latest lagged the installed version (fresh release
+  not yet cached, rc/dev installs), instead of comparing ordering.
+
+This module is the single source of truth for both operations. It is pure
+(no CLI/click/typer imports) so it is safe to import from any layer — CLI
+commands, core probes, session-presence writers, or future callers.
+"""
+
+from __future__ import annotations
+
+from packaging.version import InvalidVersion, Version
+
+
+def try_parse_version(value: str | None) -> Version | None:
+    """Parse *value* as a PEP 440 version, or return ``None`` if it cannot be.
+
+    ``None``, the empty string, and any string ``packaging.version`` rejects
+    all resolve to ``None``. Callers that need a well-formed comparison
+    target should treat ``None`` as "unknown — do not compare".
+    """
+    if not value:
+        return None
+    try:
+        return Version(value)
+    except InvalidVersion:
+        return None
+
+
+def is_version_newer(candidate: str | None, current: str) -> bool:
+    """Return True iff *candidate* is a parseable version strictly newer than *current*.
+
+    - ``None``, empty, or unparseable *candidate* -> ``False`` (never blocks
+      on bad input).
+    - Unparseable *current* -> ``False`` (conservative: can't prove an
+      upgrade exists if we can't parse the baseline).
+    """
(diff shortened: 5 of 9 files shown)


stijn-dejongh:
Claiming this PR for today's landing pass: rebase onto upstream/main, adjudicate the red `arch-adversarial (arch_shard_1)` check, fold fixes as needed. Evidence to follow.

stijn-dejongh:
## Landing pass — summary

**Verdict: fold applied, merge-ready pending fresh CI.**

Rebase check: branch was already directly on current `upstream/main` (0 commits behind) — no rebase needed.

**Red check classified & folded:** `arch-adversarial (arch_shard_1)` — `test_census_worklist_matches_live_derivation`. Root cause: this PR's own `_upgrade_is_available` helper (+docstring) shifts `session_presence` from 1227 → 1246 LOC, staling the committed `tests/architectural/ci_topology_census.json` snapshot. This is bin 2 of the landing runbook ("a contract the PR legitimately crosses") — regenerated via the documented `uv run python -m tests.architectural._gate_coverage --emit-census` (1-line diff, no other census field changed). Pushed as `landing fold: regenerate stale CI-topology census after session_presence LOC shift`.

**Red-first verification** (this is a bugfix PR): swapped the pre-fix `manager.py` back in and reran `tests/specify_cli/session_presence/`. Both `TestBuildContent` regression tests go genuinely red for the exact #2413 symptom (`AssertionError: assert 'upgrade-available' == 'healthy'`), plus 6 more `ImportError`s on the new `_upgrade_is_available` helper (8 failed total). Restored the fix — 170 passed again. The PR's tests are not vacuous; they witness the bug.

**Local evidence after the fold:**
- `tests/specify_cli/session_presence/` + `test_session_start.py`: 170 passed
- Full `tests/architectural/`: 742 passed, 4 skipped
- `ruff check` on touched files: clean
- `mypy --strict src/specify_cli/session_presence/manager.py`: clean

No other findings — small, well-scoped, single-purpose fix; no squad warranted (not an architectural/API-surface change).

Watching fresh CI on the new head; merge-ready on green.

@rayjohnson — no action needed on your end, this was a one-line CI-topology census re-pin unrelated to your fix (a repo-wide snapshot test that any LOC change to a tracked directory needs to regenerate).

stijn-dejongh:
## Landing pass — update: #2417 consolidation folded in

Following up on the review squad's findings: the version-comparison duplication architect-alphonso flagged (issue #2417) is now folded into this PR as a third commit, `b04ce1ca2` — `refactor: consolidate version-comparison primitives (closes #2417)`.

**What changed:** extracted a single canonical `src/specify_cli/core/version_compare.py` (`try_parse_version` + `is_version_newer`, pure — no CLI imports) and re-pointed all three prior independent implementations at it:
- `cli/commands/upgrade.py::_version_is_newer` — deleted, call site rewired.
- `session_presence/manager.py::_upgrade_is_available` (this PR's own fix for #2413) — deleted, call site rewired.
- `core/upgrade_probe.py::_classify` — its richer PyPI-channel state machine (`ALREADY_CURRENT`/`AHEAD_OF_PYPI`/`NO_UPGRADE_PATH`/`UPGRADE_AVAILABLE`) is preserved exactly; only its version-parsing/ordering sub-logic now delegates to the canonical helper.

Dispatched as a governed Op (python-pedro, model-routed to claude-sonnet-5), TDD red-first (the new test file was written and confirmed red against a nonexistent module before the helper existed). I independently re-verified before pushing: read the full diff line-by-line (confirmed `_classify`'s state-machine semantics are byte-for-byte preserved, not just "probably equivalent"), re-ran every test suite myself rather than trusting the transcript, and reproduced the one remaining failure on the pre-consolidation commit to confirm it's genuinely pre-existing.

**Verified test evidence (independently re-run, not paraphrased):**
- `tests/specify_cli/core/test_version_compare.py` (new) + `tests/specify_cli/session_presence/` + `test_session_start.py` + `tests/specify_cli/cli/commands/test_upgrade_command.py` + `tests/core/test_upgrade_probe_and_notifier.py` + `tests/architectural/test_no_dead_symbols.py`: **276 passed, 1 failed**.
- The 1 failure (`test_project_migration_needed_project_dry_run_json_contract`) is **#2339** (dotted `migration_id` vs. the dry-run JSON contract's `^[a-z0-9_]{1,128}$` pattern) — already tracked, reproduces identically on the pre-consolidation commit (`1d533e94c`), unrelated to this change.
- `ruff check` + `mypy --strict` on all touched files: clean, zero suppressions.

Closes #2417 on merge (PR body updated).

Watching fresh CI on the new head; merge-ready on green.

stijn-dejongh:
## Landing pass — update: #2339 folded in (campsite style), 2 new findings filed

Fourth commit pushed: `b6e04b282` — `fix(#2339): widen migration_id contract pattern to admit the dotted convention`.

**HiC decision** (posted on #2339 before implementing): 83 of 89 real migration IDs use a dotted convention the contract's `migration_id` pattern (`^[a-z0-9_]{1,128}$`) didn't admit at all. Widened to `^[a-z0-9_.]{1,128}$` rather than renaming any shipped ID (renaming would need a compat read path for already-persisted per-project migration ledgers — real risk for zero benefit). Backward-compatible widening, no `schema_version` bump needed.

**Verification note — the E2E test path was broken for unrelated reasons.** The dispatched Op correctly stopped rather than forcing a false "green": `test_project_migration_needed_project_dry_run_json_contract`'s `_CONTRACT_PATH` resolution only coincidentally finds the contract file when run from a `.worktrees/<name>` maintainer layout — in a real CI checkout it doesn't resolve at all, so the schema-conformance assertion has been a **silent no-op in CI** since the test was introduced. That's why #2339 was never caught. On top of that, an unrelated pre-existing contract violation (a migration description exceeding `maxLength`) would also block the test even with #2339 fixed. Both are genuinely pre-existing (traced to commits from 2026-04-27 and 2026-06-12, unrelated to this PR) — filed rather than folded in, per locality-of-change discipline:
- **#2419** (P1) — the dead-in-CI contract test itself, parented under #1931.
- **#2420** (P3) — the masked description-length violation, currently invisible because of #2419.

Verified the #2339 fix's correctness directly at the schema level (bypassing the broken E2E path): validated all 89 real `migration_id` values against the pattern. Before: 83/89 rejected (red, matches the exact defect). After: 0/89 rejected.

Dispatched as a second governed Op (python-pedro, claude-sonnet-5) — independently re-verified by me before pushing, same as the #2417 fold.

PR #2414 now carries 4 commits: the contributor's #2413 fix, the census re-pin, the #2417 consolidation, and the #2339 contract fix. Watching fresh CI on the new head.

stijn-dejongh:
## Landing pass — update: 2 self-inflicted findings from the last fold, fixed

The post-#2339 CI run (5th commit) surfaced 3 red arch-adversarial checks, both traceable to the #2417 consolidation commit rather than #2339 itself:

1. `tests/specify_cli/core/test_version_compare.py` (added by #2417) had no module-level `pytestmark`, making it invisible to every marker-based CI gate — caught independently by `test_pytest_marker_convention`, `test_no_new_orphan_surfaces`, and `test_split_preserves_zero_orphans`. Added `pytestmark = [pytest.mark.unit, pytest.mark.fast]`, matching its sibling test file's convention.
2. Same census-staleness class as before (#2416, filed earlier this session): #2417 *shrank* `session_presence` (1246 → 1229 LOC) by deleting the now-redundant `_upgrade_is_available` helper, staling the committed CI-topology census again. Regenerated via the documented command.

Both folded into commit `ddfe0a8ea`. Full `tests/architectural/` suite green (742 passed, 4 skipped) after both fixes.

PR #2414 now carries 5 commits total. Watching fresh CI on the new head.

PR #2426, opened 2026-07-06
fix(compat-planner): revive dead contract checks + fix the drift (#2419)

### LINKED ISSUE #2419: compat-planner.json contract-conformance test is dead in CI (hardcoded sibling-checkout path never resolves)
## What

Found while landing #2339 (migration_id contract pattern fix, PR #2414). `tests/specify_cli/cli/commands/test_upgrade_command.py::test_project_migration_needed_project_dry_run_json_contract` is meant to validate a real `spec-kitty upgrade --dry-run --json` payload against the committed JSON contract `kitty-specs/cli-upgrade-nag-lazy-project-migrations-01KQ6YDN/contracts/compat-planner.json` — described in that contract's own header as "Stable JSON contract... Stable across patch releases."

The test resolves the contract file's path like this (lines ~61-69):

```python
_WORKTREE_ROOT = Path(__file__).parent.parent.parent.parent.parent  # repo root
_CONTRACT_PATH = (
    _WORKTREE_ROOT.parent.parent.parent  # spec-kitty main repo root
    / "spec-kitty"
    / "kitty-specs"
    / "cli-upgrade-nag-lazy-project-migrations-01KQ6YDN"
    / "contracts"
    / "compat-planner.json"
)

_CONTRACT: dict[str, Any] | None = None
if _CONTRACT_PATH.exists():
    with contextlib.suppress(Exception):
        _CONTRACT = json.loads(_CONTRACT_PATH.read_text(encoding="utf-8"))
```
...
```python
if _CONTRACT is not None:
    jsonschema.validate(instance=payload, schema=_CONTRACT)
```

`_WORKTREE_ROOT.parent.parent.parent / "spec-kitty" / ...` walks 3 levels above the repo root and back down into a sibling directory literally named `spec-kitty`. This only resolves to something real **by coincidence**, when the test happens to run from inside a `.worktrees/<name>/` directory 3 levels below a primary checkout also named `spec-kitty` (verified: this is exactly the maintainer landing-worktree layout used throughout recent PR-landing passes in this repo — `.worktrees/pr-2414-landing` → 3 parents up → the sibling primary `spec-kitty/` checkout).

**In a real CI checkout** (`/home/runner/work/spec-kitty/spec-kitty`, GitHub Actions' standard convention — verified by resolving the same path arithmetic against that literal layout), the computed path is `/home/runner/spec-kitty/kitty-specs/...`, which does not exist. `_CONTRACT_PATH.exists()` is `False`, so `_CONTRACT` stays `None`, and **the entire `jsonschema.validate(...)` call — the whole point of the test's name — is silently skipped.** The test still passes (it exercises the CLI dry-run path and asserts other things), but the contract-conformance assertion itself never actually runs in CI.

This is why #2339 (a real, 83-of-89-migrations-affecting contract-pattern bug) was never caught: it only ever surfaced in **local** runs from a specific maintainer worktree layout, never in CI, despite the test file's name and stated purpose.

## Impact

A documented "stable" external JSON contract, consumed by anything scripting against `spec-kitty upgrade --json`/`--dry-run --json`, has had **zero real schema-conformance enforcement in CI** since this test was introduced (`abcd730cfa`, 2026-04-27). Any future contract drift (a field renamed, a type changed, a pattern narrowed incorrectly like #2339) will pass CI silently.

## Suggested approach (not prescriptive)

Resolve `_CONTRACT_PATH` from something CI-portable and worktree-agnostic — e.g. walk up from `__file__` only as far as the actual repo root (already computed once, correctly, as `_WORKTREE_ROOT`) and read `kitty-specs/.../compat-planner.json` from *that* root directly, with no cross-checkout hop. If the contract file is expected to live in a fixed location relative to the test file regardless of worktree/CI layout, the fix is a one-line path change. Also consider: drop the silent `contextlib.suppress(Exception)` + `.exists()` fallback (or make it a hard failure / explicit skip with a visible reason) so a future path regression fails loudly instead of silently passing.

Related: #2339 (the specific defect this dead check let through).

MOES-Media:
Closes #2419.

## What
Two test surfaces validated `spec-kitty upgrade` / `render_json` payloads against the committed, "stable across patch releases" `compat-planner.json` contract — and **both silently no-op'd in CI**. Each resolved the contract via a worktree-relative path that only exists in a maintainer landing-worktree, so `_CONTRACT` was `None` and `jsonschema.validate` was skipped. The contract had had **zero CI enforcement since 2026-04-27**, which is why #2339 (an 83/89-migration contract-pattern bug) slipped through.

## Change (3 files, 2 commits — governance dossier + code slice)
- **`test_upgrade_command.py`** — anchor `_CONTRACT_PATH` on the repo root (`_WORKTREE_ROOT`/`parents[4]`); load unconditionally; drop the `.exists()`/`suppress` skip and the `None`-guard → validation always runs; a missing/unreadable contract **fails hard**.
- **`test_messages.py`** — the **sibling dead enforcer** of the same contract (found by the post-plan squad), fixed with the file-correct `parents[3]` anchor + fail-hard `_get_contract`; its 5 `render_json` checks now validate for real.
- **`m_3_2_0rc35_unified_bundle.py`** — the revival immediately catches one real drift: `UnifiedBundleMigration.description` was 283 chars vs the contract's `maxLength: 256`. Trimmed 283 → **249**, meaning preserved. **The contract itself is unchanged.**

## Verification
- **Red-first**: restoring the 283-char description reds the flagship test (`jsonschema.ValidationError: … is too long`); the trim makes it green — the #2339-class catch demonstrated.
- **Non-vacuous**: reject-path witnesses for **both** helpers + missing/malformed **fail-hard** tests; reviewer-renata mutation-verified that deleting either `validate` call reds exactly the witnesses, and reintroducing a silent-`None` fallback reds the fail-hard tests.
- `87 passed`; `ruff` clean; **zero net-new `mypy --strict` errors** (CI's mypy gate is src-scoped).

## Governance (dogfooding)
Carries the full `kitty-specs/compat-planner-contract-ci-portable-01KWVKYH/` dossier. Run as a spec-kitty mission with adversarial squads at **all four** point-cuts — each caught a real issue before implementation: undersized scope (revive → real RED → FR-005 trim), an unclosed defect class (the sibling dead check → FR-006), coverage gaps (both witnesses + fail-hard → T005/T006), and matrix/traceability nits (pre-merge). Reviewer-renata independently approved (8/8 checks).

---
**Draft** pending maintainer review; fork CI green (23 pass / 0 fail).

https://claude.ai/code/session_018P1J7pFioRpXE5prXnkhcy


stijn-dejongh:
Maintainer picking this up for landing review. Fetching branch, will rebase onto current main and run adversarial squad before advising on merge.

stijn-dejongh:
Maintainer landing complete.

Rebased onto main (clean — no file overlap with recent auth/sync changes). Fold applied: `tests/upgrade/test_auto_discovery.py` — added `len(description) <= 256` gate to the discovery loop (reviewer-renata finding: the existing loop checked non-empty but not the contract maxLength limit, leaving future migrations free to drift silently past 256 chars).

**Adversarial squad** (3 profiles, all PASS):

- **architect-alphonso** (path anchors): hop counts correct for both files (`parents[4]` for the 4-deep test file, `parents[3]` for the 3-deep one); portable across worktree and plain checkout; contract file present. Two `[INFO]` comment-precision notes, no runtime impact.
- **debugger-debbie** (non-vacuity): all 5 new reject-path witnesses in both files go RED under the delete-validate() mutation; fail-hard tests exercise the correct code paths. One `[LOW]` docstring imprecision (claims test guards `_CONTRACT is not None` re-guard — that guard is always true so it is not a real mutation; primary claim "delete validate()" is correct).
- **reviewer-renata** (description trim): 249 chars confirmed < 256 maxLength; all four behavioral claims of the original description preserved; no other migration exceeds 256 chars; contract file unchanged. One `[LOW]` pre-existing gap → **folded in** above.

**Fold verified**: 101 passed locally; CI now running on fold commit. Ready to merge on green.

PR #2454, opened 2026-07-07
fix(#2446): infer_authoritative_surface handles filename-glob owned_files

### LINKED ISSUE #2446: finalize-tasks: infer_authoritative_surface produces invalid surface for filename-glob owned_files (e.g. `dir/*.py`)
## Summary

`finalize-tasks` fails ownership validation for a work package whose `owned_files`
uses a **filename glob** (e.g. `src/foo/*.py`) and does not set an explicit
`authoritative_surface`. The auto-derivation in `infer_authoritative_surface()`
produces an invalid surface like `src/foo/*.py/`, which can never be a prefix of
`src/foo/*.py`, so `validate_authoritative_surface()` then rejects it.

Observed on `spec-kitty-cli==3.2.4`.

## Reproduction

A WP prompt file with (no explicit `authoritative_surface`):

```yaml
---
work_package_id: "WP07"
owned_files: ["src/sk_sidecar/takeover/*.py"]
requirement_refs: ["FR-012"]
---
```

Run:

```bash
spec-kitty agent mission finalize-tasks --mission <slug> --json
```

Result:

```json
{
  "error": "Ownership validation failed",
  "ownership_errors": [
    "WP07: authoritative_surface 'src/sk_sidecar/takeover/*.py/' is not a prefix of any owned_files entry: ['src/sk_sidecar/takeover/*.py']"
  ]
}
```

(A two-entry `owned_files` that shares no common directory segment also yields
`authoritative_surface is empty.`, which is a related symptom.)

## Root cause

`src/specify_cli/ownership/inference.py` → `infer_authoritative_surface()`:

```python
def _strip_glob(pattern: str) -> str:
    for suffix in ("/**", "/*", "**", "*"):
        if pattern.endswith(suffix):
            return pattern[: -len(suffix)]
    return pattern

stripped = [_strip_glob(p) for p in owned_files]

if len(stripped) == 1:
    return stripped[0].rstrip("/") + "/"
```

`_strip_glob` only strips a **trailing** wildcard token. A filename glob such as
`src/foo/*.py` ends in `.py`, so none of the suffixes match and the pattern is
returned unchanged. The single-entry branch then appends `/`, producing
`src/foo/*.py/`.

`validate_authoritative_surface()` (`ownership/validation.py`) requires the
surface to be a prefix of at least one `owned_files` entry:

```python
if pattern == surface or pattern.startswith(surface):
    return []  # valid
```

`src/foo/*.py`.startswith(`src/foo/*.py/`) is `False`, so a WP that the tool
auto-populated fails the tool's own validation.

## Suggested fix

When the **last path segment** of a pattern contains a glob metacharacter
(`*`, `?`, `[`), drop that segment (use its directory) rather than only stripping
a whole trailing `*`/`/**`. For `src/foo/*.py` this yields `src/foo/`, which is a
valid prefix. For example:

```python
def _strip_glob(pattern: str) -> str:
    head, _, tail = pattern.rpartition("/")
    if any(ch in tail for ch in "*?["):
        return head + "/" if head else ""
    return pattern.rstrip("/")
```

(and keep the existing common-prefix logic for multi-entry lists).

## Workaround

Set `authoritative_surface` explicitly in the WP frontmatter to a directory
prefix, e.g. `authoritative_surface: "src/sk_sidecar/takeover/"`.

## Context

Found while dogfooding a spec-driven mission whose WPs own per-module globs
(`src/pkg/subdir/*.py`). Happy to open a PR with the fix + a unit test covering
`*.py`, `*.ts`, and the empty-common-prefix case if useful.

stijn-dejongh:
Closes #2446.

## Problem
`finalize-tasks` failed ownership validation for a WP whose `owned_files` used a **filename glob** (e.g. `src/foo/*.py`) with no explicit `authoritative_surface`. `_strip_glob` (`src/specify_cli/ownership/inference.py`) only stripped a *whole trailing wildcard token* (`/**`, `/*`, `*`), so `src/foo/*.py` (ending in `.py`) was left intact and the single-entry branch appended `/` → `src/foo/*.py/`, which `validate_authoritative_surface` rejects (not a prefix of `src/foo/*.py`). A WP the tool auto-populated failed the tool's own validation.

## Fix
Reduce a **glob-bearing final path segment** (containing `*`, `?`, or `[`) to its containing directory via `rpartition('/')`. `src/foo/*.py` → `src/foo`; existing `/**`, `/*`, plain-dir behavior preserved.

**Correction to the reporter's suggested shape:** returning `head + '/'` breaks the multi-entry common-prefix path (`['src/api/*.py','src/api/*.ts']` → `src/api//`, double slash → invalid). The fix returns a **bare** `head`; the single-entry branch re-appends the one `/`, and the multi-entry branch splits on `/`, so no spurious empty segment. Proven by a dedicated multi-entry test.

## RED-first evidence
Tests committed **before** the fix. Pre-fix: **8 failed** (the 4 single-entry glob cases + the 4 glob params of `test_inferred_surface_validates`, which ties `infer`→ the real `validate` gate) with the exact bug signature. Post-fix: ownership dir **148 passed**, finalize-tasks ownership suite **63 passed**. `ruff`/`mypy` clean.

Tests cover `*.py`/`*.ts`/`?`/`[abc]` globs, `/**` & `/*` regression guards, plain-dir unchanged, multi-entry same-dir globs, and the two-entry disjoint-top-level case (empty surface is **by-design** — collapsing disjoint trees would over-broaden the WP blast radius; such WPs need an explicit `authoritative_surface`).

## Out of scope (pre-existing, not introduced)
A lone literal file entry (`['src/foo/bar.py']`) still yields `src/foo/bar.py/` under both old and new code — a separate pre-existing quirk of the single-entry trailing-slash append, unchanged by this glob fix.

### DIFF
--- a/src/specify_cli/ownership/inference.py
+++ b/src/specify_cli/ownership/inference.py
@@ -155,7 +155,9 @@ def infer_authoritative_surface(owned_files: list[str]) -> str:
     """Derive the authoritative_surface from a list of owned_files patterns.
 
     Returns the longest common path prefix shared by all owned_files entries.
-    Trailing ``/**`` and ``/*`` wildcards are stripped before comparison.
+    A glob-bearing final path segment (whole-wildcard tokens like ``/**``/``/*``
+    *and* filename globs like ``*.py``, ``data_?.json`` or ``[abc].py``) is
+    reduced to its containing directory before comparison.
 
     Args:
         owned_files: List of glob patterns.
@@ -167,10 +169,20 @@ def infer_authoritative_surface(owned_files: list[str]) -> str:
         return ""
 
     def _strip_glob(pattern: str) -> str:
-        for suffix in ("/**", "/*", "**", "*"):
-            if pattern.endswith(suffix):
-                return pattern[: -len(suffix)]
-        return pattern
+        """Reduce a glob pattern to its containing directory prefix.
+
+        When the *last* path segment carries a glob metacharacter (``*``,
+        ``?`` or ``[``) — covering both whole-wildcard tokens (``**``, ``*``)
+        and filename globs (``*.py``, ``data_?.json``, ``[abc].py``) — drop
+        that segment and keep its directory. A plain literal path is returned
+        unchanged. In every case the result carries no trailing ``/``: the
+        single-entry branch re-appends one and the multi-entry branch splits
+        on ``/``, so a trailing slash would create a spurious empty segment.
+        """
+        head, _, tail = pattern.rpartition("/")
+        if any(ch in tail for ch in "*?["):
+            return head
+        return pattern.rstrip("/")
 
     stripped = [_strip_glob(p) for p in owned_files]
 
--- a/tests/specify_cli/ownership/test_inference.py
+++ b/tests/specify_cli/ownership/test_inference.py
@@ -13,6 +13,7 @@ from specify_cli.ownership.inference import (
     infer_ownership,
 )
 from specify_cli.ownership.models import ExecutionMode, OwnershipManifest
+from specify_cli.ownership.validation import validate_authoritative_surface
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


stijn-dejongh:
✅ CI Quality: success, mergeState CLEAN. Squad LAND (implementer python-pedro + reviewer-renata; red→green replay confirmed, ruff/mypy clean). Merge-ready — over to the operator.

PR #2435, opened 2026-07-07
docs(tasks): document create_intent ownership metadata

roian6:
## Summary

- Document `create_intent` in the `/spec-kitty.tasks` ownership metadata guidance for the canonical tasks prompt and the tracked project mission override.
- Explain that `create_intent` is for planned-new repo-root-relative literal paths whose `owned_files` entry may otherwise zero-match during finalize-tasks validation.
- Explicitly tell agents to keep a single `create_intent` key and replace an existing `create_intent: []` stub instead of adding a duplicate block.
- Add regression coverage for the ownership metadata contract and update the rendered slash-command / command-skill snapshots.

## Validation

- `uv run --extra test pytest tests/prompts/test_tasks_prompt_ownership_metadata.py tests/prompts/test_prompt_fragment_rendering.py tests/specify_cli/regression/test_twelve_agent_parity.py::test_command_output_unchanged tests/specify_cli/regression/test_twelve_agent_parity.py::test_toml_command_output_is_parseable tests/specify_cli/skills/test_command_renderer.py::test_snapshot -q -k 'tasks or prompt_fragment or ownership_metadata'`
  - `85 passed, 128 deselected`
- `git diff --check` — PASS
- Added-line secret heuristic — no hits
- Foreground independent review via `codex exec` — PASS; no security or logic blockers
- Pre-PR live duplicate check — issue still open/unassigned, no claim comments, no timeline PR refs, no semantic open-PR hits, and no exact open-PR file collisions

## Notes

Refs #2249.

This PR updates prompt/generated-command guidance only; it does not change parser behavior.

---
Closes #2249 — documents the previously-undocumented `create_intent` key and the empty-stub duplicate-key foot-gun this PR resolves.

PR #2490, opened 2026-07-08
docs(tasks): document create_intent ownership metadata + guard against orphan agent baselines

### LINKED ISSUE #2249: tasks: create_intent undocumented in /spec-kitty.tasks command prompt; empty create_intent:[] stub causes duplicate-key YAML
## Summary
Two residual gaps around the `create_intent:` WP frontmatter field that `finalize-tasks` hard-requires. The canonical template was fixed in #2221 / #1982 (both closed), but:

1. **The `/spec-kitty.tasks` command prompt still doesn't document `create_intent`.** It enumerates required frontmatter as `owned_files` / `authoritative_surface` / `execution_mode` but never mentions `create_intent`. An agent authoring WP prompts *from the command prompt* (not the template) produces a correct-looking first pass that fails `finalize-tasks` **by construction** for any greenfield WP.
2. **An empty `create_intent: []` stub causes a duplicate-key YAML failure.** When create_intent is added proactively as an empty stub and a populated block is later inserted, `finalize-tasks` fails with `found duplicate key "create_intent"`.

## Environment
spec-kitty-cli **3.2.3**, software-dev mission with 12 greenfield WPs (`event-sync-retention-delivery`).

## Repro / symptom
For any WP whose `owned_files` includes a path that does not yet exist on disk:
```
spec-kitty agent mission finalize-tasks --validate-only --mission <slug> --json
=> "error": "Ownership validation failed: literal-path owned_files entries match zero files. Fix the paths or add them to 'create_intent'."
   "WP01: owned_files path 'src/.../new_module.py' is a literal file path that matches zero files... declare it in the WP frontmatter:
      create_intent:
        - src/.../new_module.py"
```
All 12 WPs were flagged (every new module/test path). After scripting the `create_intent:` insert, several WP files already carried an empty `create_intent: []`, yielding:
```
=> Invalid YAML in .../WP11-...md: found duplicate key "create_intent" with value "[]" (original value: [...])
```

## Expected
- The `/spec-kitty.tasks` command prompt documents `create_intent` alongside the other required frontmatter (parity with the template fixed in #2221).
- `finalize-tasks` either (a) treats to-be-created literal `owned_files` as `create_intent` with a warning instead of a hard fail — the error already prints the exact fix block, so a `--fix` flag would close the loop — and/or (b) normalizes/dedupes an empty `create_intent: []` rather than failing on the duplicate key.

## Refs
- #2221 (closed) — added `create_intent`/`authoritative_surface`/`execution_mode` to `task-prompt-template.md`.
- #1982 (closed) — added the missing create_intent hint.

These two residuals are not covered by either. Found while catfooding a full software-dev mission on 3.2.3.

stijn-dejongh:
> **Landing note:** re-anchored + expanded from #2435 (original author @roian6). That PR is based on a now-stale `main` (pre-3.2.6 cycle, ~78 commits behind) and cannot receive the fold-in below on its fork branch, so it is **closed as folded-in here**. The original `docs(tasks)` commit is preserved with @roian6's authorship; this branch adds one reviewer-surfaced fold-in remediation on top.

## Summary

- Document `create_intent` in the `/spec-kitty.tasks` ownership metadata guidance for the canonical tasks prompt and the tracked project mission override.
- Explain that `create_intent` is for planned-new repo-root-relative literal paths whose `owned_files` entry may otherwise zero-match during finalize-tasks validation.
- Explicitly tell agents to keep a single `create_intent` key and replace an existing `create_intent: []` stub instead of adding a duplicate block.
- Add regression coverage for the ownership metadata contract and update the rendered slash-command / command-skill snapshots.

## Fold-in (added during the landing pass)

Surfaced by the correctness-review lens on the rebased diff:

- **Prune the orphan `_twelve_agent_baseline/roo/` directory.** `roo` migrated to the Agent Skills / shim delivery path and is absent from `AGENT_COMMAND_CONFIG`, so the parity suite (parametrized over `NON_MIGRATED_AGENTS`) never rendered or asserted that baseline. It had been frozen since 2026-06-11 (rc42) — a full ungated month of drift, and it did not receive the `create_intent` line while all 12 gated baselines did. roo **consumers are unaffected** (roo renders from the updated canonical source template); this removes a dead, ungated second source of truth.
- **Add `test_no_orphan_baseline_directories`.** Asserts the baseline tree matches `NON_MIGRATED_AGENTS` exactly, so a migrated agent can never again leave an ungated, drifting baseline behind. Verified non-vacuous (fails on a planted orphan dir).

## Validation

- Rebased onto current `upstream/main` (3.2.6 cycle); no conflicts. Version/changelog files untouched.
- `create_intent` doc guidance verified against the implementation: field at `status/wp_metadata.py`, suppression at `ownership/validation.py`, wired at `mission_finalize.py` (the runtime error message matches the doc verbatim).
- `PWHEADLESS=1 uv run --extra test pytest tests/specify_cli/regression/ tests/prompts/ tests/specify_cli/skills/ -q` → **536 passed**; `ruff` clean; terminology guard green.
- Two independent review lenses (doctrine + correctness): **LAND** on both.

## Notes

Closes #2249 — documents the previously-undocumented `create_intent` key and the empty-stub duplicate-key foot-gun it resolves.

Folds in / supersedes #2435 (credit: @roian6). Prompt/generated-command guidance + test-fixture hygiene only; no parser/behavior change.


stijn-dejongh:
CI hit one failure — `fast-tests-core-misc` → `test_collect_notices_completes_within_50ms` (74.7ms mean vs a hard 50ms absolute budget). It's a **pre-existing budget-gate flake unrelated to this diff** (which touches zero glossary code; passes 5/5 locally): the test asserts a mean-of-5 with no CI-runner headroom and was bound to this gate by 260d9344e, so a single scheduler stall trips it. Per the flakiness policy (tune budget gates at root, never retry-to-green) I campsited it in commit `c0ffbbecf`: measure the **median of 9** runs instead of the mean of 5 — robust to outlier stalls, still trips on a real regression. Re-running CI now.

PR #2503, opened 2026-07-09
fix(#2502): dashboard artifact viewers read the planning surface, not the coord husk

rayjohnson:
Fixes #2502 — the third and (per an exhaustive grep of `resolve_feature_dir` call sites) final coord-shadows-primary endpoint family in the dashboard, missed by #2431.

## The gap

#2431 re-anchored the mission **list** and **kanban** to read planning artifacts primary-first. But the artifact **viewer** endpoints — `handle_research`, `_handle_artifact_directory` (contracts/checklists), `handle_artifact` (the spec/plan file server) — still resolved their feature dir through the coord-first `resolve_feature_dir()` and read planning content off it. For an **in-flight** coordination mission that's the status-only coord husk:

- `GET /api/features` → `spec=true plan=true research=true` ✅
- `GET /api/research/<slug>` → `{"main_file": null, "artifacts": []}` ❌ — the UI shows the board working and every artifact tab empty.

Finished missions were unaffected (coord worktree removed → gather falls through to primary), so the gap only shows **mid-run** — the same "broken precisely while running" shape as #2430.

## The fix — one public composition, four sites

- `scanner.resolve_feature_planning_dir(project_dir, feature_id)`: `resolve_feature_dir` (coord-first — still the status authority) composed with the existing `_resolve_planning_dir_primary_first` re-anchor from #2431. No new resolution logic.
- The three viewer sites use it.
- `handle_kanban`'s split reads made explicit: `is_legacy_format` is a planning-surface question (a husk has no `tasks/` tree and would wrongly read as non-legacy), while `read_only_weighted_percentage` **stays coord-first** — the live event log is STATUS-partition.

## Verified against the live affected repo

For the reporting repo's in-flight mission: plain resolver → coord husk (`research.md` absent); planning resolver → primary dir (`research.md` + `spec.md` present). The regression test reproduces the exact shape (registered coord worktree, status-only mission copy, full planning on primary, post-083 slug) and pins both halves: viewers re-anchored, status authority unchanged.

248 dashboard tests green; `ruff` clean; terminology gate green; `mypy` shows only the pre-existing `DashboardHandler`-subclass finding.

## Class status

With this, every dashboard read of a PRIMARY-partition artifact routes through the #2431 seam. The remaining known member of the class outside the dashboard is #2404 (accept reads `acceptance-matrix.json` from a stale coord worktree) — different command, not touched here.

Issue #2346, opened 2026-07-04
move-task subtask guard leaks into the NEXT WP's section when a heading mentions the current WP (substring match on '(depends: WPxx)')


stijn-dejongh:
## Symptom

`spec-kitty agent tasks move-task WP02 --to for_review` refuses with:

```
Error: Cannot move WP02 to for_review - unchecked subtasks:
  - [ ] T007
  - [ ] T008
  - [ ] T009
  - [ ] T010
```

…but T007–T010 belong to **WP03**, not WP02. WP02's own subtasks (T004–T006) are all checked in tasks.md.

## Root cause

`_check_unchecked_subtasks` in `src/specify_cli/cli/commands/agent/tasks_shared.py` (~line 488):

```python
# Check if we entered this WP's section
if re.search(rf"^#{{2,4}}[^#].*{wp_id}\b", line):
    in_wp_section = True
    continue
# Check if we entered a different WP section
if in_wp_section and re.search(r"^#{2,4}[^#].*WP\d{2}\b", line):
    break
```

The *section-entry* check runs first and `continue`s. A tasks.md heading like

```markdown
### WP03 — Workflow surgery (depends: WP01, WP02)
```

contains the literal string `WP02`, so when scanning for WP02 the entry regex re-matches on WP03's heading, `in_wp_section` stays true, and the loop harvests WP03's unchecked `- [ ] T0xx` rows as if they were WP02's. The exit check (which would have `break`-ed) is unreachable for any heading that mentions the current WP id — which every `(depends: …)` annotation produced by the canonical tasks template does.

## Repro

Any mission tasks.md where WP(N+1)'s heading names WP(N) in its depends list — e.g. mission `ci-suite-map-bind-01KWNPMP` on this repo, `move-task WP02 --to for_review` with WP03's boxes unchecked.

## Suggested fix

Anchor section identity to the heading's own WP id, not any mention: e.g. match `^#{2,4}\s*WP\d{2}\b` (or `### {wp_id} —`) and compare the captured id to `wp_id`, entering on equality and breaking on inequality. A regression test: tasks.md with `### WP02 …` fully checked and `### WP03 … (depends: WP02)` unchecked → move WP02 must NOT block.

## Workaround used

`--force` (after verifying WP02's real subtasks T004–T006 are checked) — recorded in the mission's Activity Log.

Found during mission `ci-suite-map-bind-01KWNPMP` (guard-friction trace, feeds #2017).

[...]

PR #2505, opened 2026-07-09
feat(#2504): dashboard WP cards show subtask progress (2/4, from checkbox rows)

### LINKED ISSUE #2504: Dashboard: WP cards show a subtask count but no completion progress ("4 subtasks" → "2/4 subtasks")
## Ask

The kanban WP cards (and the WP detail view) show `4 subtasks` — a bare count from the WP frontmatter `subtasks` list. During an active mission the interesting signal is **progress**: `2/4 subtasks`. Everything needed is already in the pipeline; nobody wired the counting.

## Why it's cheap

- Completion state already exists as the checkbox rows in the WP body (`- [x] T001` / `- [ ] T002`) — the same rows the lane-transition guard parses (`_check_unchecked_subtasks`, canonical pattern `^-\s*\[\s*\]\s*(T\d{3,})\b`).
- The dashboard scanner's `_process_wp_file` already has the full body in hand (it ships `prompt_markdown` to the frontend) — it parses everything needed and just doesn't count.
- The frontend renders the badge in exactly two places (`dashboard.js`).

## Proposed shape

1. Hoist the canonical subtask-row patterns (checked + unchecked variants) into a small shared home so the guard and the dashboard consume ONE definition (no parallel regex — per the repo's canonical-sources rule). Counting matches the guard's semantics: implementation rows (`T###`) only, so the dashboard number is consistent with what actually gates lane transitions.
2. `_process_wp_file` emits `subtasks_done` / `subtasks_total` from the body's checkbox rows (additive fields; the existing `subtasks` list is untouched).
3. Frontend: badge becomes `2/4 subtasks` when the body has checkbox rows; falls back to the current count badge when it doesn't (no false "0/4" for WPs that don't track completion via checkboxes).

PR incoming.

rayjohnson:
Fixes #2504.

## What

Kanban WP cards and the WP detail view showed a bare frontmatter count (`4 subtasks`) with no completion signal — even though completion state already exists as the canonical checkbox rows in the WP body (the rows the lane-transition guard blocks on), and `_process_wp_file` already ships the full body to the frontend. Cards now show **`2/4 subtasks`** (✓ at n/n); the detail view shows `2/4 subtasks complete`.

## How — one shared definition, no parallel regex

- **`core/subtask_rows.py`**: the canonical subtask-row patterns (unchecked **and** the new checked variant) plus `count_subtask_rows(text) → (done, total)`, with exactly the guard's semantics: `T\d{3,}` implementation rows only (command/validation rows like `- [ ] swift test` and prose checkboxes excluded), fenced code blocks ignored.
- **`_check_unchecked_subtasks`** now consumes the shared unchecked pattern (behavior identical) — and a test pins that it does, so the guard and the dashboard can never drift apart on what counts as a subtask.
- **Scanner** emits additive `subtasks_done`/`subtasks_total` on each kanban task (encoding-error variant reports (0, 0)); `KanbanTaskData` declares both; the FR-014 typed-contract baseline is **deliberately regenerated** (6 additive lines — the byte-identical gate caught the payload change exactly as designed, and this is a conscious contract addition, not drift).
- **Frontend**: badge + detail meta use done/total when the body has checkbox rows, and **fall back to the existing count badge when it doesn't** — a WP that doesn't track completion via checkboxes never shows a false `0/N`.

## Tests

Pattern/counter matrix (checked+unchecked mix, non-implementation rows excluded, fence handling, `T1000+` ids, zero case), the guard-shares-definition pin, and scanner integration (progress counted from the body, frontmatter `subtasks` list untouched, no-checkbox fallback). 254 dashboard tests + dead-modules/terminology gates green; `ruff` + `mypy` clean.

## Related

Sibling PR #2503 (in review) fixes the artifact viewers for in-flight coordination missions; both branch from current main and don't overlap.

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -19,6 +19,17 @@ _The 3.2.6 development cycle is open. Entries land here as missions merge._
 
 ### ✨ Added
 
+- **Dashboard WP cards show subtask progress (#2504).** Cards (and the WP
+  detail view) previously showed a bare frontmatter count (`4 subtasks`);
+  they now show `2/4 subtasks` (with a ✓ at n/n), counted from the canonical
+  checkbox rows in the WP body — the same rows the lane-transition guard
+  blocks on, via a new shared single definition
+  (`core/subtask_rows.py`, now consumed by both the guard and the dashboard
+  so the two can't drift). WPs that don't track completion via checkboxes
+  keep the plain count badge (no false `0/N`). The kanban task payload gains
+  additive `subtasks_done`/`subtasks_total` fields; the typed-contract
+  baseline is regenerated accordingly.
+
 ### 🐛 Fixed
 
 ### ♻️ Changed
--- a/kitty-specs/unified-charter-bundle-chokepoint-01KP5Q2G/baseline/pre-wp23-dashboard-typed.json
+++ b/kitty-specs/unified-charter-bundle-chokepoint-01KP5Q2G/baseline/pre-wp23-dashboard-typed.json
@@ -153,6 +153,8 @@
         "subtasks": [
           "T1"
         ],
+        "subtasks_done": 0,
+        "subtasks_total": 0,
         "title": "WP02"
       }
     ],
@@ -172,6 +174,8 @@
         "subtasks": [
           "T1"
         ],
+        "subtasks_done": 0,
+        "subtasks_total": 0,
         "title": "WP03"
       }
     ],
@@ -190,6 +194,8 @@
         "subtasks": [
           "T1"
         ],
+        "subtasks_done": 0,
+        "subtasks_total": 0,
         "title": "WP01"
       }
     ]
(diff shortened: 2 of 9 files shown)


rayjohnson:
Amendment pushed (`7bdad23ce`) after a field check on a live coordination mission showed `subtasks_total=0` on every WP: the **standard layout keeps the canonical `- [ ] T###` rows in `tasks.md`'s per-WP sections** (exactly the guard's blocking source), while the WP prompt body carries unnumbered acceptance checkboxes (correctly excluded by the pattern). The initial commit counted only the WP body.

Now: the guard's section walk is hoisted verbatim into `core/subtask_rows.py` (`iter_wp_section_subtask_rows` — first-WPxx-token heading rule #2346/#2324, fences, section-end on a different WP's heading), `_check_unchecked_subtasks` consumes it (behavior identical, pinned by its existing suite), and the scanner reads `tasks.md` once per board, preferring the WP's section counts with WP-body rows as fallback. Verified against the live mission's layout: cards now show real `n/m` progress.

stijn-dejongh:

stijn-dejongh:

That PR faithfully aggregated your commit for #2504 — **preserved with your original authorship/attribution** — and aligned it with the other four coord-shadows fixes into one coherent slice that closes the read/gate arm of epic #2160 (`_infer_subtasks_complete` no longer fails open; the fix ships on every path). #2504 is closed by #2572.

Issue #2508, opened 2026-07-09
agent action implement: WP status commit falls into the legacy safe_commit path mid-mission (coord topology) — identity meta read off the coord husk


rayjohnson:
Live blocker on a coordination-topology mission (spec-kitty 3.2.6-dev main).

## Repro

Mission with coordination topology in flight (registered `-coord` worktree holding the live status partition; primary checkout on the mission's feature branch):

```
spec-kitty agent action implement WP03 --agent claude-code --mission <slug>
```

The WP-status commit (`chore: Start WP03 implementation`) refuses: it resolves its commit target to the **primary checkout** and demands the **coordination branch** be checked out there. `move-task`'s status write routed correctly through the coord worktree minutes earlier under the same build.

## Mechanics (verified against the live repo)

`workflow.py`'s `_commit_workflow_change` selects its commit mechanism via `_load_coord_branch_meta(feature_dir)` — but the `feature_dir` threaded in by the action-implement path is `_canonical_status_feature_dir(...)` = `resolve_handle_to_read_path(...)`, the **topology-aware status dir**, i.e. the coord worktree copy for an in-flight coordination mission. Mission identity (`coordination_branch`/`mission_id`/`mid8`) is a **PRIMARY-partition** artifact — `meta.json` never transits the coordination branch — so the husk read returns `(None, None, None)`:

```
resolved read dir: .worktrees/<slug>-coord/kitty-specs/<slug>
meta.json there  : False
coord_branch_meta: (None, None, None)
```

With identity gone, the modern `BookkeepingTransaction` path (which correctly lands the commit on the coordination branch via the coord worktree) never engages, and the **legacy fallback** runs `safe_commit(repo_root=primary, worktree_root=primary, target=CommitTarget(ref=<seam-resolved coord branch>))` — primary has the feature branch checked out → refusal. The function's own diagnostic even logs the seam-vs-caller divergence just above, but the mechanism selector still keys off the husk read.

Same coord-shadows-primary class as #2331/#2430/#2502 — this one on the **write-mechanism-selection** side. Ironically `_canonical_status_feature_dir`'s docstring describes the read seam anchoring identity on PRIMARY meta; the commit site just doesn't use that anchor.

## Fix shape

[...]

Issue #2502, opened 2026-07-09
Dashboard artifact viewers (spec/plan/research/contracts/checklists) empty for in-flight coordination missions — third coord-shadows-primary endpoint family, missed by #2431


rayjohnson:
Direct follow-up to #2430 / PR #2431: that fix re-anchored the mission **list** (`scan_all_features`) and **kanban** (`scan_feature_kanban`) to read planning artifacts primary-first — but the dashboard has a third endpoint family the fix missed.

## Observed (live repo, spec-kitty 3.2.6-dev with #2431)

For an **in-flight** coordination mission (`review-context-depth-…`, live `-coord` worktree present):

- `GET /api/features` → `spec=true plan=true research=true`, `path=kitty-specs/<slug>` ✅ (#2431 working)
- `GET /api/research/<slug>` → `{"main_file": null, "artifacts": []}` ❌ — clicking spec/plan/research in the UI shows **nothing**, while the kanban shows the work fine.

For *finished* missions (coord worktree removed) the viewers work, which hides the bug until you look mid-run — the same "invisible precisely because it's running" irony as #2430.

## Mechanics

`dashboard/handlers/features.py` resolves the feature dir via the gather-based, coord-first `resolve_feature_dir()` at four sites and then reads **planning content** off it:

- `handle_research` (`:202`) — reads `research.md` + `research/` from the coord husk → null/empty
- `_handle_artifact_directory` (`:304`) — contracts/checklists listings → empty
- `handle_artifact` (`:395`) — the spec/plan/etc. file server → 404s
- `handle_kanban` (`:161`) — a **split** case: its `is_legacy_format` check is a planning-surface question (husk → wrongly non-legacy) while its `read_only_weighted_percentage` read is a status question where coord-first is *correct*

## Fix shape (same pattern as #2431, one helper)

Expose a public `resolve_feature_planning_dir(project_dir, feature_id)` in `dashboard/scanner.py` composing `resolve_feature_dir` (coord-first, status authority) with the existing `_resolve_planning_dir_primary_first` re-anchor, and use it at the three artifact sites + the legacy check; the weighted-percentage read keeps the coord-first dir. PR incoming.

Issue #2430, opened 2026-07-06
Dashboard: PR-bound missions planned on a feature branch are invisible (spec-commit targets feature branch; dashboard reads only the coordination worktree)


rayjohnson:
## Summary

A PR-bound mission created on an (unprotected) feature branch is **invisible in the dashboard**: the dashboard resolves the mission to its **coordination worktree**, but `spec-commit` on an unprotected primary commits planning artifacts **directly to the feature branch** — so the coordination worktree's mission directory never receives `spec.md` / `plan.md` / `tasks.md` / `meta.json`, and the dashboard drops the mission from `/api/features`.

The three components are each behaving as documented; their composition makes any feature-branch-planned mission with coordination topology invisible.

## Environment

- spec-kitty **3.2.4**, macOS (darwin), git repo with primary branch `main` (not protected for local commits)

## Repro

1. On the primary checkout, create a PR-bound mission that starts a feature branch:
   ```
   spec-kitty agent mission create "my-mission" ... --json --pr-bound --branch-strategy already-confirmed --start-branch feat/my-mission
   ```
   → mission dir scaffolded; coordination branch `kitty/mission-my-mission-<id>` + coordination worktree `.worktrees/my-mission-<id>-coord/` created.
2. Write `spec.md`, then commit via the sanctioned entrypoint:
   ```
   spec-kitty spec-commit --mission my-mission-<id> -m "Add spec" kitty-specs/my-mission-<id>/spec.md
   ```
   Per its own help — *"On an unprotected or flattened primary the commit is direct"* — this lands on `feat/my-mission`, **not** the coordination branch.
3. Continue through `/spec-kitty.plan`, `/spec-kitty.tasks`, `finalize-tasks` (all commit to the feature branch), start the orchestrator (status transitions DO land on the coordination branch).
4. `spec-kitty dashboard` → the mission is absent from the mission list.

## Observed

- `spec-kitty dashboard --json` registers the mission with
  `feature_dir = .worktrees/my-mission-<id>-coord/kitty-specs/my-mission-<id>`.
- That directory contains **only** `status.json`, `status.events.jsonl`, `analysis-report.md` (orchestrator/status writes) — no `meta.json`, `spec.md`, `plan.md`, `tasks.md`, `tasks/`.
- `GET /api/features` omits the mission entirely (other missions, whose registered dirs happen to contain full artifact copies from an earlier era when spec-commit routed via coordination, render fine).
[...]

Issue #2331, opened 2026-07-04
dashboard orphans a valid in-flight (mid-orchestration) mission — hidden from the mission dropdown


rayjohnson:
## Environment
- spec-kitty 3.2.4; `spec-kitty-orchestrator` as a separate tool; git-worktree-based orchestration.
- Affected mission is on a feature branch (unmerged), `mission_number` null, and is **mid-orchestration** (live coord + lane worktrees checked out).

## Symptom
A valid, mid-orchestration mission is registered in the dashboard as `orphan:<slug>` instead of its canonical `mission_id` (ULID), so it is **filtered out of the `spec-kitty dashboard` mission dropdown**. Every sibling mission resolves to its real ULID. `dashboard --json` shows e.g. `"mission_id": "orphan:<slug>"`. Cosmetic — orchestration and mission data are unaffected.

## Root cause (verified against the code)
In `src/specify_cli/dashboard/scanner.py`:
- `gather_feature_paths()` applies **coordination-worktree copies last** (*"Coordination worktrees beat everything else"*), so the mission's `feature_dir` becomes `.worktrees/<slug>-coord/kitty-specs/<slug>/`.
- `_read_mission_identity(feature_dir)` reads `<coord-worktree>/kitty-specs/<slug>/meta.json` — which **does not exist in the coord worktree** (`meta.json` is a PRIMARY-partition artifact that lives on the primary surface, not the coord worktree) → returns `(None, None)`.
- `_mission_record_key(...)` → both identity fields absent → **`orphan:<slug>`**; the dropdown filters pseudo-keys.

This is the same class as the coord-read-surface work (#2115/#2118/#2185): a PRIMARY artifact (`meta.json` identity) is being read off the coordination surface. The scanner also uses a parallel `_read_mission_identity` reader rather than the canonical identity seam.

Secondary: `materialize`'s `lifecycle.json` omits `mission_id` entirely (`derive_mission_lifecycle` resolves identity via `resolve_mission_identity` but `MissionLifecycleResult.to_dict()` drops the field), so it offers no recovery link.

## Fix direction
1. Resolve mission **identity** from the PRIMARY surface via the canonical `primary_feature_dir_for_mission` + `resolve_mission_identity` seams (retiring the parallel `_read_mission_identity`), while keeping the coord path for live status/board display.
2. Add `mission_id` to `MissionLifecycleResult` / `lifecycle.json`.

## Acceptance criteria
[...]

PR #2511, opened 2026-07-09
fix(#2510): orchestrator-api transition derives subtasks_complete from the PRIMARY surface

### LINKED ISSUE #2510: orchestrator-api transition: subtasks_complete inferred from the coord STATUS husk (tasks.md absent → fail-OPEN) — WPs reach for_review/done with every subtask unchecked
Field evidence (coordination-topology mission, orchestrator-driven, spec-kitty 3.2.6-dev): four WPs moved `in_progress → for_review → … → done` with `force=false` while **every** `- [ ] T###` row in `tasks.md` was unchecked (0/16 across the four WPs). The native `move-task` subtask gate returns all the unchecked ids for those exact WPs when called directly — the orchestrator-api path never consults it.

## The chain (each link verified)

1. The orchestrator passes no `--subtasks-complete` (it only forwards the flag when a caller sets it) → the API's `transition` command leaves it `None`.
2. `orchestrator_api.transition` enforces the **commit gate** for `for_review` (explicitly ported: *"the SAME check the native move-task gate uses… so 'done without a commit' is impossible through the orchestrator-api too"*) — but has **no parity for the subtask gate**: `--subtasks-complete` is caller-asserted, and `None` falls through to emit-time inference.
3. `status/emit.py::_infer_subtasks_complete(feature_dir, wp_id)` reads `tasks.md` off `feature_dir` — which on this path is the **coordination STATUS dir** (the coord worktree copy). `tasks.md` is a PRIMARY-partition artifact and never exists there, and the inference **fails open**: `if not tasks_path.exists(): return True`.
4. The downstream guard (`wp_state`: `subtasks_complete is not True` blocks) therefore never fires.

**Fifth member of the coord-shadows-primary class** (#2331 → #2430 → #2502 → #2508 → this): a PRIMARY artifact read off the status surface — this time with a fail-open default that converts the miss into a silent guard bypass.

## Two additional notes for the design pass

- `_infer_subtasks_complete` is a **third parallel implementation** of the subtask-row/section semantics (alongside the native guard and, once #2505 lands, `core/subtask_rows.py`) — and it has drifted: its heading match is any-mention `\b<wp>\b` (the pre-#2346 semantics the guard fixed — a `### WP03 … (depends: WP01)` heading re-enters WP01's section) and it counts ANY unchecked checkbox, not just canonical `T###` rows. Post-#2505, consolidating onto the shared walker would retire the drift.
- `fail-open on missing tasks.md` is defensible for legacy missions but indistinguishable from "read the wrong surface" — worth an explicit comment or a primary-anchored resolve either way.

## Fix shape (PR incoming)

Mirror the commit-gate precedent in `orchestrator_api.transition`: when `to_lane == for_review` and the caller didn't assert `--subtasks-complete`, **derive it server-side from the PRIMARY planning surface** (the command already computes `_planning_read_dir(main_repo_root, mission)` two lines up) instead of letting `None` reach the husk-anchored inference. `--force` keeps its bypass; an explicit caller assertion keeps working (contract-compatible).

rayjohnson:
Fixes #2510 — the subtask guard's silent bypass on every orchestrator-driven `for_review` transition.

## The chain (each link verified against a live repo)

1. The orchestrator never asserts `--subtasks-complete` → the API forwards `None`.
2. `None` reaches emit-time inference (`status/emit.py::_infer_subtasks_complete`), which reads `tasks.md` off the STATUS feature dir — the coordination worktree husk mid-mission, where the PRIMARY-partition `tasks.md` never exists — and **fails open** (`if not tasks_path.exists(): return True`).
3. The `wp_state` guard (`subtasks_complete is not True` blocks) therefore never fires.

Field evidence: four WPs reached `done` with **0/16** canonical rows ticked and `force=false` on every transition, while `_check_unchecked_subtasks` returns every unchecked id for the same WPs when consulted directly. **Fifth member of the coord-shadows-primary class** (#2331 / #2430 / #2502 / #2508) — this one with a fail-open default that converts the surface miss into a guard bypass.

## The fix

Mirrors the command's own commit-gate precedent (`_enforce_for_review_commit_gate`, ported explicitly so *"done without a commit is impossible through the orchestrator-api too"*): when `to_lane == for_review`, the caller did not assert the flag, and `--force` is absent, `transition` derives `subtasks_complete` **server-side from the PRIMARY planning surface** (`_planning_read_dir`, already computed two lines up) using the same inference emit would run — just pointed at the surface that actually holds `tasks.md`. Explicit caller assertions keep working (contract-compatible: no payload/flag changes); `--force` keeps its bypass.

## Tests

Four-case matrix on the captured `TransitionRequest`: unasserted + unchecked PRIMARY rows → `False` (guard blocks); all checked → `True`; explicit `--subtasks-complete` respected (not overridden); `--force` skips derivation. 10 api tests green, `ruff` clean.

## Deliberately not in this PR (noted on #2510)

- `_infer_subtasks_complete` itself is a third, drifted parallel of the subtask-row semantics (any-mention heading match = pre-#2346; counts any checkbox, not just `T###`). Once #2505's shared `core/subtask_rows.py` walker lands, consolidating the inference onto it retires the drift — better as a follow-up on top of #2505 than a cross-PR dependency here.
- Whether the inference's fail-open on missing `tasks.md` should survive at all — legacy-compat question for the maintainers.

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -21,6 +21,19 @@ _The 3.2.6 development cycle is open. Entries land here as missions merge._
 
 ### 🐛 Fixed
 
+- **orchestrator-api `transition` no longer lets WPs into `for_review` with
+  unchecked subtasks (#2510).** The command trusted a caller-asserted
+  `--subtasks-complete`; left unset, emit-time inference read `tasks.md` off
+  the STATUS feature dir — the coordination worktree husk mid-mission, where
+  the PRIMARY-partition `tasks.md` never exists — and **failed open**,
+  silently bypassing the subtask guard native `move-task` enforces (field
+  evidence: four WPs reached `done` with 0/16 rows ticked, `force=false`).
+  Mirroring the command's own commit-gate precedent, the API now derives the
+  value server-side from the PRIMARY planning surface when the caller doesn't
+  assert it; explicit assertions keep working and `--force` keeps its bypass.
+  Fifth member of the coord-shadows-primary class
+  (#2331/#2430/#2502/#2508).
+
 ### ♻️ Changed
 
 ## [3.2.5] - 2026-07-08
--- a/src/specify_cli/orchestrator_api/commands.py
+++ b/src/specify_cli/orchestrator_api/commands.py
@@ -1363,6 +1363,19 @@ def transition(
 
     if to_lane == Lane.FOR_REVIEW:
         _enforce_for_review_commit_gate(cmd, main_repo_root, mission, mission_dir, wp, force)
+        if subtasks_complete is None and not force:
+            # Subtask-gate parity with native move-task (#2510): left as None,
+            # the emit-time inference reads tasks.md off the STATUS feature
+            # dir — the coord worktree husk mid-mission, where tasks.md (a
+            # PRIMARY-partition artifact) never exists — and FAILS OPEN,
+            # letting WPs into for_review with every subtask unchecked.
+            # Derive server-side from the PRIMARY planning surface instead;
+            # an explicit caller assertion keeps working, --force bypasses.
+            from specify_cli.status.emit import _infer_subtasks_complete
+
+            subtasks_complete = _infer_subtasks_complete(
+                _planning_read_dir(main_repo_root, mission), wp
+            )
 
     from specify_cli.coordination.status_transition import emit_status_transition_transactional
     from specify_cli.status import TransitionError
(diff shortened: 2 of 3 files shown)


stijn-dejongh:

stijn-dejongh:

That PR faithfully aggregated your commit for #2510 — **preserved with your original authorship/attribution** — and aligned it with the other four coord-shadows fixes into one coherent slice that closes the read/gate arm of epic #2160 (`_infer_subtasks_complete` no longer fails open; the fix ships on every path). #2510 is closed by #2572.

Issue #2513, opened 2026-07-09
Rollback (move-task --to planned) should uncheck tasks.md subtask rows for the WP — currently stays [x] which defeats the for_review gate on re-implementation


rayjohnson:
## Problem

When a WP is rolled back (`move-task WPxx --to planned`, or any backward transition from a state ≥ `for_review`) the `- [x] T###` subtask rows in `tasks.md` remain checked. They are never written by the rollback path.

On the next implementation pick-up this causes two compounding failures:

1. **Agent skips the work.** The implementing agent reads the WP section and sees all boxes already checked — it treats them as done and moves on without re-doing the rolled-back work.
2. **Subtask gate passes immediately.** When the agent transitions to `for_review`, `_check_unchecked_subtasks` reads the (still-`[x]`) rows, finds no unchecked ones, and lets the transition through — no work performed.

Net result: the WP re-enters review with the boxes saying "done" for work that was never re-done. The gate exists precisely to prevent this.

## Root cause

`move-task` writes the WP file (frontmatter + history) but never touches `tasks.md`. The only code that unchecks `- [ ] T###` rows is `mark-status --status pending`, which must be called explicitly with individual task IDs. Rollbacks do not call it.

## Reproduction

1. Implement a WP, tick subtask rows, move to `for_review`.
2. Reviewer moves it back to `planned` (rejected review).
3. Check `tasks.md` — WP section still shows `- [x] T###` for every subtask.
4. `move-task WPxx --to in_progress` then immediately `move-task WPxx --to for_review` — the subtask gate passes with zero new work.

## Fix shape

A helper called from the rollback execution path in `_mt_execute` (or `_mt_persist_wp_file`), after the status event is emitted:

1. Detect backward transition: `target_lane in {Lane.PLANNED, Lane.IN_PROGRESS, Lane.CLAIMED}` AND `old_lane >= Lane.FOR_REVIEW`
2. Read `tasks.md` from the PRIMARY surface (`resolve_planning_read_dir(kind=TASKS_INDEX)`)
3. Walk the WP's section via `iter_wp_section_subtask_rows` (already in `core/subtask_rows.py`, #2504)
4. Replace `[x]` → `[ ]` for all `T###` rows in that section
5. Write and auto-commit (`TASKS_INDEX` kind, same route as `_ms_commit`)

## Relationship

[...]

PR #2514, opened 2026-07-09
fix(#2512): stale claim markers block lane allocation after OS kill — three-part fix

### LINKED ISSUE #2512: Stale claim markers (agent/shell_pid from a dead session) block orchestrator lane allocation; move-task doesn't clear them, no liveness check, LANE_ALLOCATION_FAILED payload carries no reason
Live repro on a coordination mission (spec-kitty 3.2.6-dev + pending PRs; orchestrator current). Field report from the operating agent, mechanics verified.

## Symptom

`spec-kitty-orchestrator orchestrate`/`resume` fails a WP with `[LANE_ALLOCATION_FAILED]` within ~2s of loop start, deterministically — while the WP is in lane `planned` with a **clean** lane worktree.

## Root cause (verified)

The WP's frontmatter still carries claim markers from a **dead session** — `agent: "claude-code"`, `shell_pid: "41417"` (`ps -p 41417` → dead). Producing sequence:

1. An implementing agent is killed mid-run (in our case: macOS idle-sleep, pre-#2500), leaving `agent`/`shell_pid` planted in the WP frontmatter.
2. Every subsequent orchestrator allocation refuses on the stale claim.
3. `spec-kitty agent tasks move-task WP03 --to planned` **succeeds but does not clear `agent`/`shell_pid`** — the refusal persists across the lane reset that should logically release the claim.
4. Meanwhile a direct `spec-kitty agent action implement WP03 --agent claude-code` succeeds (it overwrites the claim) — so the natural diagnostic step **re-plants a fresh claim that goes stale again**, re-breaking the orchestrator. We looped through exactly that twice.

Diagnosis cost ~2h of wrong theories (dirty worktree, the #2508 routing bug) because the error carries no reason: `orchestrator_api` wraps allocator exceptions (incl. bare `RuntimeError`) as `LANE_ALLOCATION_FAILED` and passes `str(exc)` as the message, but the **data payload** contains only mission identity + wp_id — and that's what the orchestrator surfaces.

## Three composable fixes (any one unblocks; all three seem right)

1. **`move-task --to planned` (and any lane-reset transition) clears the `agent`/`shell_pid` claim fields** — a lane reset that leaves the claim planted is a split-brain between the event log (authority) and the frontmatter (operational metadata).
2. **The allocator treats a claim with a dead `shell_pid` as reclaimable** (liveness check) — this is the one that makes sleep-kill recovery automatic, which is the entire point of the resume path. `shell_pid` exists to prevent abandoned tasks; a dead pid IS the abandonment signal.
3. **`LANE_ALLOCATION_FAILED` includes the refusal reason in its payload** (`claimed by pid N (dead|alive)`), not just the envelope message.

## Workaround (for anyone else who hits this)

Delete the `agent`/`shell_pid` lines from the WP's frontmatter; allocation then runs clean.

## Relationship to recent work

Same mission also field-confirmed #2508's fix (direct implement now allocates + lands the start-commit on the coordination branch correctly) and the orchestrator's #2500 sleep assertion — this staleness is the last blocker in the sleep-kill→resume recovery chain. Happy to implement whichever subset of the three you'd take (2+3 feel highest-value; 1 is the hygiene fix).

rayjohnson:
## Summary

Fixes #2512. Three composable fixes for the sleep-kill→resume recovery chain field-reported by the operating agent (`spec-kitty-orchestrator orchestrate` failing with `LANE_ALLOCATION_FAILED` after macOS idle-sleep killed the implementing agent mid-session).

### Root cause (traced)

The actual cause of `LANE_ALLOCATION_FAILED` is **Fix 2** below: when the OS kills the agent, the lane worktree *directory* vanishes while the lane *branch* survives in git. The allocator then tried `git worktree add -b <branch>` (create-new-branch) which fails with "branch already exists". The `_recover_lane_worktree` helper for exactly this case existed but was never called.

The `agent`/`shell_pid` stale frontmatter (Fix 1) is a co-symptom of the same kill event — corrected independently.

### Fix 1 — `move-task --to planned` clears `agent`/`shell_pid`

`_mt_persist_wp_file` now calls `delete_scalar(front, "agent")` + `delete_scalar(front, "shell_pid")` before the conditional re-plants when `target_lane == Lane.PLANNED`. A plain rollback always leaves the WP with a clean claim state. An explicit `--agent` on rollback still re-plants (delete then re-set). Forward transitions (`in_progress`) leave existing claims untouched.

Requires new `delete_scalar()` in `task_utils` — removes a key-value line from WP frontmatter without disturbing surrounding content, companion to the existing `set_scalar`/`extract_scalar`.

### Fix 2 — allocator crash-recovery: reattach when branch exists but worktree is gone

In `allocate_lane_worktree`, before the creation path, check `_branch_exists(repo_root, branch)`. If the branch exists (crash-survival case):
1. `git worktree prune` — cleans up the stale registry entry for the gone directory so git does not reject re-attachment
2. `_recover_lane_worktree` — `git worktree add <path> <branch>` (without `-b`) reattaches to the existing branch
3. Validate clean + merge dependency tips, same as the reuse path

### Fix 3 — `LANE_ALLOCATION_FAILED` payload carries `"reason"`

The `data` dict in `_resolve_start_workspace` now includes `"reason": str(exc)` so the orchestrator can surface a machine-readable diagnostic. The envelope `message` already carried `str(exc)` but was not structurally queryable. A missing reason field cost ~2h of misdiagnosis in the field.

## Companion

#2513 (rollback should also uncheck `tasks.md` subtask rows — same "rollback should clean up per-WP operational state" class) — filed, deferred to follow-up PR to keep this one focused.

## Test plan

- [x] 14 new unit tests: `delete_scalar` (6), rollback claim-clearing (4), allocator crash-recovery (4)
- [x] 92/92 existing lane + orchestrator_api tests pass
- [x] 22,511/22,511 fast tests pass; 5 failures are pre-existing on `main`
- [x] `ruff check` clean on all changed files


stijn-dejongh:

stijn-dejongh:

That PR faithfully aggregated your commit for #2512 — **preserved with your original authorship/attribution** — and aligned it with the other four coord-shadows fixes into one coherent slice that closes the read/gate arm of epic #2160 (`_infer_subtasks_complete` no longer fails open; the fix ships on every path). #2512 is closed by #2572.

PR #2515, opened 2026-07-09
fix(#2513): uncheck tasks.md subtask rows on WP rollback to planned

### LINKED ISSUE #2513: Rollback (move-task --to planned) should uncheck tasks.md subtask rows for the WP — currently stays [x] which defeats the for_review gate on re-implementation

rayjohnson:
## Problem

When `move-task --to planned` rolls a WP back, the `- [x] T###` subtask rows in the WP's `tasks.md` section stay checked. The lane-transition gate then passes immediately on the next `for_review` submission — without any work being re-done. The checked rows lie about the state: a rolled-back WP must be fully re-implemented.

Fixes #2513.

## Solution

`_mt_execute` now calls `_mt_uncheck_rollback_subtasks(st, ports)` when `target_lane == Lane.PLANNED`, placed outside the status lock (same pattern as `_mt_release_review_lock`).

The helper:
1. Resolves `tasks.md` through the `TASKS_INDEX` primary-partition anchor (via `ports.fs.planning_read_dir(kind=TASKS_INDEX)`) so coord-topology missions read the correct surface — not the `-coord` husk (same anchor as `mark-status` and the subtask gate).
2. Calls `uncheck_wp_section_subtask_rows(text, wp_id)` (new `core/subtask_rows.py`) which applies the canonical `#2346` heading-boundary/fence semantics to flip `[x]` → `[ ]` only within the target WP's section.
3. Writes back the modified content.
4. Auto-commits via `ports.coord.commit_artifact(..., kind=TASKS_INDEX)` when `resolved_auto_commit` is set — same route as `mark-status`.

Failures are non-fatal: missing `tasks.md` is silently skipped; commit exceptions are logged as warnings, never `typer.Exit`.

## New module: `core/subtask_rows.py`

> **Note:** PR #2505 also creates `core/subtask_rows.py` with read-only utilities (`iter_wp_section_subtask_rows`, `count_wp_section_subtask_rows`). When both PRs land, the section-traversal logic should be unified into one canonical walk. The merge conflict is trivial (combine both functions into the file).

## Tests

15 new tests across 2 new files:
- `tests/specify_cli/core/test_uncheck_wp_section_subtask_rows.py` — 9 unit tests for the text-transform function (section boundaries, fences, case variants, dependency-mention heading rule, sub-headings without WP tokens, unchanged-text identity)
- `tests/specify_cli/cli/commands/agent/test_move_task_uncheck_rollback.py` — 6 unit tests for `_mt_uncheck_rollback_subtasks` (unchecks rows, no-write when nothing to uncheck, skips when tasks.md absent, commits when auto-commit enabled, skips commit when disabled, non-fatal commit failure)

`_mt_uncheck_rollback_subtasks` added to the WP05 move-set in `test_tasks_move_task_seam.py` and re-exported through `tasks.py` (seam-bridge rule).

## Test plan

- [x] 15 new tests pass
- [x] Existing move-task test suite green (66 tests)
- [x] `ruff check` — no issues
- [x] Full fast suite: 22,529+ passed (architectural shard-marker flakiness pre-existing, unrelated to this diff)

### DIFF
--- a/src/specify_cli/cli/commands/agent/tasks.py
+++ b/src/specify_cli/cli/commands/agent/tasks.py
@@ -440,6 +440,7 @@ from specify_cli.cli.commands.agent.tasks_move_task import (
     _mt_resolve_feedback as _mt_resolve_feedback,
     _mt_resolve_pre_review_workspace as _mt_resolve_pre_review_workspace,
     _mt_resolve_targets as _mt_resolve_targets,
+    _mt_uncheck_rollback_subtasks as _mt_uncheck_rollback_subtasks,
     _mt_review_config_section as _mt_review_config_section,
     _mt_run_decision as _mt_run_decision,
     _mt_run_pre_review_gate as _mt_run_pre_review_gate,
--- a/src/specify_cli/cli/commands/agent/tasks_move_task.py
+++ b/src/specify_cli/cli/commands/agent/tasks_move_task.py
@@ -62,6 +62,7 @@ from specify_cli.agent_tasks_ports import (
 from specify_cli.cli.commands.agent.tasks_finalize_validation import (
     _read_transactional_wp_lane,
 )
+from specify_cli.cli.commands.agent.tasks_outline import TASKS_MD_FILENAME
 from specify_cli.cli.commands.agent.tasks_materialization import (
     _collect_status_artifacts,
     _persist_review_artifact_override,
@@ -86,6 +87,7 @@ from specify_cli.cli.commands.agent.tasks_transition_core import (
 from specify_cli.core.commit_guard import GuardCapability
 from specify_cli.core.constants import KITTY_SPECS_DIR
 from specify_cli.core.paths import is_worktree_context
+from specify_cli.core.subtask_rows import uncheck_wp_section_subtask_rows
 from specify_cli.core.utils import write_text_within_directory
 from specify_cli.git import SafeCommitPathPolicyError
 from specify_cli.missions._read_path_resolver import (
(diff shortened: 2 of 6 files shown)


stijn-dejongh:

stijn-dejongh:

That PR faithfully aggregated your commit for #2513 — **preserved with your original authorship/attribution** — and aligned it with the other four coord-shadows fixes into one coherent slice that closes the read/gate arm of epic #2160 (`_infer_subtasks_complete` no longer fails open; the fix ships on every path). #2513 is closed by #2572.

Issue #2580, opened 2026-07-12
4th shell_pid writer (_mt_persist_wp_file move-task re-plant) bypasses the liveness baseline helper


stijn-dejongh:
**Relates to:** #2575 (PID-reuse liveness baseline), mission coord-shadows-followups, epic #2160.

## What

WP02 of the coord-shadows-followups mission (#2575) made the liveness baseline close-by-construction across the three `shell_pid` claim writers the plan enumerated: `implement.py:~1400`, `workflow_executor.py:~668` (implement claim), `workflow_executor.py:~1338` (review claim) — all now route through `frontmatter.write_shell_pid_claim`, co-writing `shell_pid_created_at`.

A pre-merge review found a **fourth** `shell_pid` writer that was out of WP02's scope: `cli/commands/agent/tasks_move_task.py::_mt_persist_wp_file` (~L1451) does `set_scalar(updated_front, "shell_pid", st.shell_pid)` for the `move-task --shell-pid` claim re-plant. It bypasses `write_shell_pid_claim`, so a WP first-claimed via `move-task --shell-pid` (rather than `agent action implement`) gets a baseline-less `shell_pid` and remains PID-reuse-vulnerable on that narrower path.

## Impact

Low. Additive-degradation (D3a) means the uncovered path degrades to today's live-PID trust — **no regression**, only absence of the new PID-reuse protection on the `move-task` first-claim path. The primary implement-review loop claims via `workflow_executor` (covered).

## Fix

Route `_mt_persist_wp_file`'s `shell_pid` re-plant through `frontmatter.write_shell_pid_claim` so close-by-construction becomes truly global. Note: `tasks_move_task.py` overlaps in-flight #2573 (pre-review-gate) — coordinate lane ownership. Small, add a test that a `move-task`-claimed WP carries the baseline.

Labels: tech-debt, reliability.

stijn-dejongh:
**Tracker hygiene pass (Directive 003 — decision documented for traceability)**

- **Parent**: linked as native sub-issue of #2160 (coord-topology artifact-authority unification), matching this issue's own "Relates to" line and the direct lineage of closed #2575 (PID-reuse liveness baseline), which was parented under #2160.
[...]

Issue #2573, opened 2026-07-12
move-task --to for_review runs a synchronous multi-minute pre-review gate that reads as a hang (no skip flag, sync-daemon deaf to disable env)


stijn-dejongh:
**Parent:** #2017 (workflow guards that block legitimate actions or lack the depth to model operator intent)
**Also relates to:** #2093 (WP-metadata authority split — the direct-emit workaround this bug forces exposes that split), mission `coord-shadows-arm-closeout-01KXAST2` / PR #2572.

## Summary

Witnessed running the implement-review loop on a coord-topology mission: `spec-kitty agent tasks move-task <WP> --to for_review` runs a **pre-review regression gate** synchronously, with no progress output and no skip flag. Under orchestrated/concurrent load this reliably blocks for minutes with zero feedback — indistinguishable from a hang — forcing the orchestrator to background or SIGKILL the command, which then leaves the transition half-applied.

## Mechanism

`src/specify_cli/cli/commands/agent/tasks_move_task.py::_mt_run_pre_review_gate` (called at line 1702, defined at line 1000)
→ `src/specify_cli/review/pre_review_gate.py::evaluate_pre_review_gate` (line 514)
→ `run_scoped_tests_at_head` (line 358)
→ `subprocess.run(..., timeout=timeout)` of pytest (line 388), with `_DEFAULT_HEAD_RUN_TIMEOUT = 300` seconds (line 109).

No progress is streamed while the subprocess runs, and there is no flag to skip or defer the gate.

**Compounding factor:** `move-task` also calls `ensure_sync_daemon` (default `True`) as a second latency source, and it does **not** honor `SPEC_KITTY_SYNC_DISABLE=1` / `SPEC_KITTY_SYNC_MINIMAL_IMPORT=1` — env vars that do disable/minimize the sync daemon elsewhere in the codebase.

## Impact

A 300s (5 min) synchronous wall with zero feedback is operationally indistinguishable from a hang. When an orchestrator kills the process to recover, the lane transition is left half-applied, forcing a direct-emit workaround — which is exactly the event-log-vs-transactional split documented in #2093.

## Asks

- (a) Make the pre-review gate async/skippable — a `--skip-pre-review-gate` flag, or run it non-blocking with streamed progress — and/or emit progress plus a shorter interactive default timeout.
[...]

Issue #2549, opened 2026-07-11
Bug: move-task --force from a lane worktree commits placement-partition status.* to the lane branch (rejected by the tool's own later move-task --to approved); plus finalize-tasks --json reports one commit_hash across a two-branch commit set


kentonium3:
> **Correction (2026-07-11):** an earlier revision of this issue also reported `finalize-tasks` committing `status.*` to the coordination branch under coord topology as a defect ("Trigger A"). That is **withdrawn** — per `mission_runtime/artifacts.py`, `STATUS_STATE`/`ACCEPTANCE_MATRIX`/`ISSUE_MATRIX` are `_PLACEMENT_ARTIFACT_KINDS` that are *designed* to route to the coordination branch under coordination topology; that behavior is correct. This issue now covers only the two points below.

## Summary

Two issues around where spec-kitty writes placement-partition mission state:

- **A. `move-task --force` from a lane worktree mis-routes `status.*` to the lane branch.** Placement-partition kinds (`status.events.jsonl`, `status.json`, matrices — `_PLACEMENT_ARTIFACT_KINDS` in `mission_runtime/artifacts.py`) route to the coordination branch under coord topology and must never land on a **lane** branch. But `move-task --to for_review --force` run from a lane worktree commits `status.json` + `status.events.jsonl` onto the **lane** branch; the subsequent reviewer-side `move-task --to approved` then rejects the lane (`Committed kitty-specs files on this lane branch … not allowed on lane branches`) — the tool blocks on the exact commit its own lane-run transition created.
- **B. `finalize-tasks --json` reports one `commit_hash` for a two-branch commit set.** Under coord topology, `finalize-tasks` correctly writes primary-partition artifacts (`tasks.md`, `lanes.json`, `tasks/WP*`) to the feature branch and placement-partition artifacts (`status.events.jsonl`, `status.json`, `acceptance-matrix.json`, `issue-matrix.md`) to the coordination branch — as separate commits on separate branches. But the JSON result reports a single `"commit_hash"` (the feature-branch commit) while `"files_committed"` lists files from **both** partitions. An automated caller reading `commit_hash` + `files_committed` together is misled: the named commit does not contain the placement-partition files, and the coordination-branch commit hashes are not reported at all.

## Versions

- **spec-kitty-cli**: 3.2.6 (main build, SHA `1cb51fb32`), installed via `pipx install "git+https://github.com/Priivacy-ai/spec-kitty.git@main"`.

[...]

Issue #2071, opened 2026-06-22
Epic: Tests as scaffold, not friction — eliminate accidental-pass, duplicate-knowledge, and contract-drift in the suite


stijn-dejongh:
**Priority: P0** (release blocker / must decide before final 3.2.x)

## Operator theory: the test suite has degraded from SCAFFOLD into FRICTION

Tests are supposed to be **scaffold** — a trustworthy structure that supports correct change and catches real regressions. The operator's read is that our suite has drifted into **friction**: it now obstructs correct work more than it protects it.

### Symptoms (three anti-pattern classes)

1. **Inaccurate / intent-misaligned tests** — assertions that no longer reflect the actual contract or intended behavior. They pass or fail for reasons unrelated to whether the code is correct.
2. **Duplicate knowledge** — the same fact/contract encoded in many places. When the design legitimately changes, N tests must be hand-edited; the duplication makes the suite brittle and expensive to evolve.
3. **Monkey-patch reliance / accidental-pass** — tests that are green *for the wrong reason* (over-mocked, monkey-patched around the real seam, asserting on stubs). They give false confidence and don't exercise the real path.

### Consequences

1. **Uncalled-for REVERTING of good structural changes.** A correct, intended change trips a mis-aligned or duplicate-knowledge test. Under time pressure the *change* gets reverted instead of the *test* being fixed — so the suite actively blocks correct evolution of the codebase. This is the most damaging consequence.
2. **Wasted effort** — agents and humans spend cycles re-deriving intent, chasing accidental-pass failures, and editing duplicated assertions.
3. **Long CI cycle times for minimal upside** — the suite is slow *and* low-signal: we pay the runtime cost without getting proportional regression protection.

## Goal of this epic

Restore the test suite to a **trustworthy scaffold**:

- **Audit + remediate** the three anti-pattern classes (intent-misalignment, duplicate-knowledge, accidental-pass / monkey-patch reliance).
- **Establish a test-hygiene directive/styleguide** so the standard is explicit and reviewable.
- **Add a ratchet** so the regression cannot silently recur (the suite stops drifting back into friction).

## Status & process note

A **research squad is auditing the suite now**. Concrete child tickets will be added to this epic **from the findings** — do not pre-create children here.

## Prior art to build on

[...]

PR #2584, opened 2026-07-12
fix(#2160): close coord-shadows follow-ups — canonical-source consolidation + gate/liveness robustness (#2574/#2575/#2576/#2567/#2568)

### LINKED ISSUE #2567: acceptance gate's _find_unchecked_tasks is a fifth divergent unchecked-checkbox parser
## What

`src/specify_cli/acceptance/gates_core.py::_find_unchecked_tasks` (~L105) implements its own "what is an unchecked task checkbox" detector:

```python
def _find_unchecked_tasks(tasks_file: Path) -> list[str]:
    if not tasks_file.exists():
        return [f"{_TASKS_FILE} missing"]
    unchecked: list[str] = []
    for line in _acceptance_pkg._read_text_strict(tasks_file).splitlines():
        if re.match(r"^\s*-\s*\[ \]", line):
            unchecked.append(line.strip())
    return unchecked
```

This scans the whole `tasks.md` file with a bare regex — no `T\d{3,}` task-id requirement, no fenced-code-block exclusion, no WP-section scoping. It answers the same underlying question ("is this line an unchecked task row?") that the canonical `core.subtask_rows` module (and the per-WP review gate, the dashboard, and the rollback path) already answer — but it is a **fifth independent implementation** of that parsing logic, diverging in scope and strictness from the other four.

## Why it might be intentional

The acceptance gate operates at **mission level** (a coarse "did anyone forget to tick a box" completeness check across all of `tasks.md`), not at the **per-WP review gate** level that `core.subtask_rows` serves. It may legitimately want looser, whole-file semantics with no task-id requirement. That's a plausible design choice — but right now it reads as accidental drift, not a documented decision.

## Ask

Either:
1. Reconcile `_find_unchecked_tasks` onto `core.subtask_rows` (parameterizing for the looser mission-level scope if needed), or
2. Explicitly document why the acceptance gate intentionally uses different (looser) unchecked-checkbox semantics than the per-WP review/dashboard/rollback parsers, with a code comment and/or ADR note explaining the divergence.

## Relates to

- The `coord-shadows-arm-closeout` mission, which is consolidating the subtask-row canon in `src/specify_cli/core/subtask_rows.py`.
- Test-suite-friction epic #2071 / the subtask-row-canon duplicate-knowledge theme.

Surfaced by a post-tasks campsite census (paula-patterns) during an in-flight mission; filed as an out-of-scope follow-up rather than folded silently into that mission's diff.

### LINKED ISSUE #2568: review lock's LockInfo.is_stale uses an independent os.kill liveness probe, not the canonical psutil helper
## What

`src/specify_cli/review/lock.py::LockInfo.is_stale` (~L85) implements its own process-liveness check:

```python
def is_stale(self) -> bool:
    try:
        os.kill(self.pid, 0)  # signal 0 = existence check only
        return False  # process is alive
    except ProcessLookupError:
        return True  # process does not exist
    except PermissionError:
        # Process exists but we cannot signal it (e.g. different user).
        # Conservative: treat as NOT stale.
        return False
    except OSError:
        return True
```

This is an **independent liveness probe** using `os.kill(pid, 0)` instead of `psutil`, with its own semantics — notably, a `PermissionError` (process exists but belongs to another user) is treated as "alive / not stale," which is a specific, undocumented policy choice distinct from whatever the canonical helper does.

## Canonical target

The `coord-shadows-arm-closeout` mission is promoting a canonical psutil-based liveness helper to `src/specify_cli/core/process_liveness.is_process_alive` (module does not exist yet on `main`; it is being introduced by that in-flight mission). `review/lock.py` is a candidate consumer of that helper once it lands.

## Ask

After `core.process_liveness.is_process_alive` lands:
1. Reconcile the semantics gap — decide (and document) whether `PermissionError` should mean "alive" or something else under the canonical helper, since `review/lock.py`'s current behavior encodes a real policy choice, not just implementation detail.
2. Migrate `review/lock.py::LockInfo.is_stale` to consume `core.process_liveness.is_process_alive` instead of its own `os.kill` probe.

This is explicitly out of scope for the mission that introduces the canonical helper — filed as a follow-up so the migration isn't silently skipped.

Surfaced by a post-tasks campsite census (paula-patterns) during an in-flight mission; filed as an out-of-scope follow-up rather than folded silently into that mission's diff.

### LINKED ISSUE #2574: Consolidate the triplicated subtask-gate-dir resolution adapter (emit.py / aggregate.py / status_transition.py) into resolve_subtasks_gate_dir()
**Relates to:** PR #2572 / mission `coord-shadows-arm-closeout-01KXAST2`, epic #2160.

## Summary

A pre-merge squad (paula-patterns + randy-reducer) on PR #2572 found the small "resolve the PRIMARY `TASKS_INDEX` dir → feed `_infer_subtasks_complete`" adapter is triplicated across three call sites instead of sharing one implementation.

## The three sites

1. **`src/specify_cli/status/emit.py::_resolve_primary_subtasks_dir`** (lines 309-341) — the named helper. Resolves `repo_root` (explicit, or via `resolve_canonical_root(feature_dir)`, falling back to bare `feature_dir` only on `WorkspaceRootNotFound`), then calls `resolve_planning_read_dir(primary_root, mission_slug, kind=MissionArtifactKind.TASKS_INDEX)`.
2. **`src/specify_cli/status/aggregate.py::_resolve_review_gate_inputs`** (~lines 713-731) — inline duplicate of the same `resolve_planning_read_dir` call, keyed off `self.repo_root` / `self.mission_slug` directly, without the `resolve_canonical_root` fallback path.
3. **`src/specify_cli/coordination/status_transition.py`** (~lines 442-461) — inline duplicate again: same `resolve_planning_read_dir` call, guarded by `if request.repo_root is not None`; the `else` branch falls back to bare `feature_dir` with **no** `resolve_canonical_root` attempt at all.

## Ask

Hoist all three onto a single shared `resolve_subtasks_gate_dir()` in `status/emit`, and have `aggregate.py` and `coordination/status_transition.py` call it instead of re-implementing the resolution inline.

## Fail-safe asymmetry (secondary finding, folded into the same consolidation)

`status_transition.py`'s `repo_root=None` branch falls back straight to bare `feature_dir`, instead of attempting `resolve_canonical_root(feature_dir)` the way `emit.py`'s helper does. On a coord-topology mission this means the fallback can read the coordination-branch husk instead of the primary `tasks.md`, which under-reports subtask completeness and blocks the `for_review` transition. This is a fail-**closed** asymmetry — it only ever over-blocks, never produces a false pass — so it's not urgent on its own, but consolidating onto the shared helper closes it as a side effect rather than requiring a separate fix.

## Label / relation

`tech-debt`; relates to PR #2572 / mission `coord-shadows-arm-closeout-01KXAST2` and epic #2160.

### LINKED ISSUE #2575: is_process_alive does not guard against PID reuse — a recycled PID reads as alive
**Relates to:** PR #2572 / mission `coord-shadows-arm-closeout-01KXAST2`; epic #2160. Surfaced by the pre-merge squad (paula-patterns + randy-reducer) and re-verified on aligned `main` by a brownfield re-verification squad.

## What

`src/specify_cli/core/process_liveness.py::is_process_alive(pid)` (L15-34) calls `psutil.Process(pid).is_running()` on a freshly-constructed `Process` with **no create-time / identity baseline**. It therefore cannot distinguish the original process from an unrelated process that has since **recycled the same PID**. A recycled PID reads as *alive*, so an abandoned WP whose claiming shell died and whose PID was reused is **never flagged stale**.

The exposure is platform-**general** (any PID wraparound inside the staleness window), not macOS-specific — macOS's low default `pid_max` merely raises the probability.

## Evidence of the overclaim

- The module docstring (L6-7) lists "recycled PIDs" among the cases handled conservatively — but the implementation has no mechanism to detect reuse; for a live recycled PID it returns the **non-conservative** `True`.
- The consumer `core/stale_detection.py::_is_claiming_process_alive` (L220-237) passes the frontmatter `shell_pid` straight in; `check_wp_staleness` then treats the WP as not-stale.
- `shell_pid` is persisted as a **bare int** (`next/workflow_executor.py:602` `str(os.getppid())`, materialized via `task_metadata_validation.py:141`) — **no `psutil.create_time()` sibling is captured anywhere** (grep-confirmed).
- The test `tests/specify_cli/core/test_process_liveness.py::test_recycled_pid_generic_exception_returns_false` is **mislabeled**: it raises `OSError` to exercise the bare `except Exception` branch (a psutil surprise), **not** a real recycled/live PID. No test covers a genuinely recycled PID.

## Fix shape

Persist a creation-time/identity baseline alongside `shell_pid` at claim time; compare `psutil.create_time()` before trusting "alive"; treat identity-unverifiable as **not** alive. Fix the mislabeled test + docstring overclaim; add a real spawn→kill liveness test. Scope strictly to `is_process_alive` + the single claim-time write/read pair (stale_detection). Do **not** sweep the other psutil consumers (`sync/*`, `dashboard/lifecycle.py`) — file a separate sync-wide audit if wanted.

Labels: tech-debt, reliability.

### LINKED ISSUE #2576: Harden _mt_uncheck_rollback_subtasks read/write path — robustness follow-up to #2513
**Relates to:** #2513 (CLOSED/shipped), PR #2572 / mission `coord-shadows-arm-closeout-01KXAST2`; epic #2160. Surfaced by the pre-merge squad and re-verified on aligned `main`.

## What

#2513 shipped `src/specify_cli/cli/commands/agent/tasks_move_task.py::_mt_uncheck_rollback_subtasks` (L1474) to uncheck `- [x] T### …` rows when a WP is rolled back to `planned`, so the subtask gate cannot pass on stale progress. Two robustness gaps remain:

1. **Unguarded read/write.** The `tasks_md.read_text(...)` / `tasks_md.write_text(...)` pair (L1496, L1500) is **not** wrapped — only the subsequent `commit_artifact` sits in `try/except` (L1505-1519). A `write_text` failure raises **uncaught**, contradicting the docstring's "Failures are non-fatal (logged, never `typer.Exit`)" (L1487-1488).
2. **Out-of-lock window.** `_mt_reset_for_planned_rollback` (L1522) runs the uncheck **after** `feature_status_lock` exits (`_mt_execute` L1593-1595). The WP is already persisted as `planned` inside the lock; a failure in the uncheck window leaves the WP `planned` with subtask rows still `- [x]` → the gate passes on the next `for_review` with no work redone — a re-manifestation of #2513.

The raw `write_text` at L1500 also **bypasses `write_text_within_directory`**, the path-guarded writer used everywhere else in this file (L1316/1380/1386/1465).

## Fix shape (constraints)

- Route the write through `write_text_within_directory` (house pattern) and define the failure mode so a write failure cannot silently leave checked rows on a `planned` WP.
- **Do NOT fold the uncheck under `feature_status_lock`** — `_mt_reset_for_planned_rollback`'s docstring (L1522-1541) documents that it runs out-of-lock *by design* (it does its own commit and must not hold the status lock); FR-010 explicitly rules out widening the lock. Preserve that decision.
- Scope `owned_files` tightly to the rollback seam — `_mt_run_pre_review_gate`/#2573 lives in the same module (merge-collision risk).

Labels: tech-debt, reliability.

stijn-dejongh:
## Why

After the coord-shadows-arm-closeout mission (PR #2572) merged, a re-verification squad confirmed **five residual defects still reproduce on `main`** — small canonical-source and gate/liveness-robustness gaps that keep the coord-topology reliability epic (#2160) from fully converging on one authoritative source per surface. This mission closes all five, each re-verified live and fixed with a red-first or characterization test.

## What

One coherent theme — *one canonical authority per operation, made correct* — across five work packages:

| WP | Issue | Change |
|----|-------|--------|
| WP01 | **Closes #2574** | Consolidate the triplicated subtask-gate-dir resolver into one `resolve_subtasks_gate_dir` in `missions/_read_path_resolver.py`; give the previously-weak `coordination/status_transition.py` site the strong git-ancestry fallback so a coord-topology mission stops gating on a stale coordination husk. Duplicate `_resolve_primary_subtasks_dir` deleted (dead-code-clean). |
| WP02 | **Closes #2575** | Harden `core/process_liveness.is_process_alive` against PID reuse via a persisted `create_time` baseline co-written at **all three** `shell_pid` claim sites (one close-by-construction helper); the reuse compare lives in a companion `is_claiming_process_alive` so `is_process_alive(pid)->bool` stays frozen. Additive degradation: an absent baseline preserves today's live-PID trust (zero legacy regression); a mismatch falls to the timestamp heuristic. Wired end-to-end into the CLI staleness reader. Mislabeled test + docstring overclaim corrected. |
| WP03 | **Closes #2576** | Route the rollback-uncheck write through the house `write_text_within_directory` guard and surface (not swallow) write failures via the `move-task` result — so a rolled-back WP cannot silently retain checked subtask rows (#2513 cannot re-manifest). The documented out-of-lock design is preserved (lock not widened). |
| WP04 | **Closes #2567** | Add a canonical whole-file, fence-aware, T###-scoped `iter_unchecked_subtask_rows` to `core/subtask_rows.py` and migrate the acceptance gate's stray `[ ]` regex onto it. The tightening (T###/fence-aware) is consciously ratified by a characterization test capturing the old→new flagging. |
| WP05 | **Closes #2568** | Fold `review/lock.is_stale` onto the canonical `core/process_liveness.is_process_alive` (`return not is_process_alive(pid)`), removing the last stray `os.kill(pid,0)` liveness probe. Branch-equivalence pinned by a characterization test. |

**Relates** #2160 (parent epic — stays open), #2071 (parent of #2567/#2568), #2513 (F3 origin, shipped), #2572 (predecessor).

## For whom

Maintainers and agents running the coord-topology implement/review loop: the subtask gate now reads the true primary `tasks.md`, an abandoned WP with a recycled PID is correctly flagged stale, a rolled-back WP can't pass the gate on stale rows, and the acceptance gate + review lock share the same canonical authorities as the rest of the system.

## Verification

- Full gate green on the integrated branch: `ruff` clean; architectural terminology + dead-code (25) pass; touched-area suites (status/core/acceptance/review) 1026 pass; cli-agent suite 1373 pass; every new branch/helper carries a focused test; red-first coord-husk + 3-branch liveness/lock equivalence + old→new checkbox characterization all pin the behavior changes.
- Each WP was independently opus-reviewed (APPROVE); a 3-lens pre-merge squad (integration + scope/dead-code + canonical-seam) reviewed the aggregate — unanimous LAND.
- **Pre-existing (not introduced):** `tests/status/test_doctor.py::test_doctor_cli_json_output` and `::test_doctor_cli_json_with_findings` fail **identically on the base commit** and are decoupled from every surface this mission touches — a pre-existing environmental issue, not a regression from this PR.
- **Sonar/mypy note:** one advisory `[redundant-cast]` remains at `missions/_read_path_resolver.py:1473` — this is the **load-bearing** `cast(Path, ...)` required under per-file `follow_imports=skip` (it relocated verbatim from the base's `status/emit.py:338`; net-zero-new). Do **not** "fix" it by removing the cast — that trades an advisory for a real per-file mypy break. No code change or Sonar hotspot action is needed.

## Follow-ups (filed, out of scope here)

- **#2580** — a 4th `shell_pid` writer (`tasks_move_task.py::_mt_persist_wp_file`, the `move-task` re-plant) bypasses the baseline helper; out of WP02's scope (C-004 fences that module against in-flight #2573). Degrades to no-regression. Route it through the helper in the WP that owns `tasks_move_task.py`.
- **#2549 facet A** (lane mis-route of `status.*`) remains open; only facet B is targeted by a separate fast-follow.

---

https://claude.ai/code/session_017epsBKV8kvJdVrc8Zh9Qtf


stijn-dejongh:
**CI remediation (rebased onto current `main`).** The initial run failed on `arch-adversarial` + `diff-coverage` (→ `quality-gate`). Root causes + fixes:

1. **IO-allowlist seed drift** — WP04's edit to `gates_core.py` shifted `_workflow_evidence_missing`'s `.read_text` line; updated the `_IO_ALLOWLIST_SITES` seed (332→329).
2. **Marker/tmp hygiene on new tests** — dropped the `fast` marker from `test_process_liveness.py` (it now carries the real spawn→kill *subprocess* test, which `fast` excludes) and removed a `/tmp` literal from `test_lock_liveness_fold.py`.
3. **Empty-baseline tmp-gate vs. current `main`** — current `main` (the security-hardening commit `d3c22332b`) carries option-injection payload literals `touch /tmp/pwned` in 4 git-hardening tests that the empty-baseline tmp-gate flags on the PR-merge tree. Converted the payload path off shared-temp (`/tmp/pwned`→`/pwned-marker`) — **payload semantics unchanged** (the tests only assert the injection is neutralised; the path is never a real filesystem target). Flagging in case `main` wants the same fix directly.
4. **diff-coverage** — added flat-mission `emit … --to for_review` tests covering `status/emit.py`'s `resolve_subtasks_gate_dir` inference branch (the existing inference tests are coord-topology and never hit the flat `emit_status_transition` path).

All four previously-failing gates verified green locally before the force-push.

PR #2587, opened 2026-07-12
feat(#2586): mission selector shows lifecycle status and sorts active missions first

### LINKED ISSUE #2586: dashboard: mission selector doesn't distinguish active vs done missions, doesn't scale
## Problem

The mission selector dropdown lists every mission the dashboard has ever seen in a flat, undifferentiated list. There is no indication of which missions are currently active, which haven't started, and which are complete. The list only ever grows, and at 20+ missions it becomes hard to find what you're actually working on.

## Proposed incremental improvement

Two small, composable changes that make the selector usable at scale without adding any new UI surface:

### 1. Sort active missions to the top

Add a derived `mission_status` field per feature, computed from the already-present `kanban_stats`:

| Status | Condition |
|--------|-----------|
| `active` | any WP in `doing`, `for_review`, or `approved` |
| `planned` | WPs exist, all in `planned` (not started yet) |
| `done` | WPs exist, none in active or planned lanes |
| `draft` | no WPs at all |

Sort order: `active → planned → done/draft`, with recency-descending as the tiebreaker within each bucket (preserving the existing recency logic).

### 2. Emoji + text status label in the dropdown option

Because `<select>` options are plain text (no CSS styling per-option), use Unicode emoji + short text as a suffix:

- `⚡ active`
- `○ planned`
- `✓ done`
- `· draft`

So an option reads: `080 – CI hardening and lint cleanup [⚡ active]`

## Files affected

- `src/specify_cli/dashboard/scanner.py` — new `_derive_mission_status(kanban_stats)` helper; add `mission_status` to feature dict; update sort key
- `src/specify_cli/dashboard/api_types.py` — add `mission_status: str` to feature TypedDict
- `src/specify_cli/dashboard/static/dashboard/dashboard.js` — `getFeatureDisplayName()` appends the status suffix

No new endpoints, no schema changes beyond the one TypedDict field. The `kanban_stats` data needed for derivation is already computed on every board scan.

## Out of scope (follow-up)

- An "Active only / All" filter toggle (bigger frontend change, better as a separate issue)
- Per-WP status indicators on the board cards (separate concern)

rayjohnson:
## Summary

Closes #2586.

- Derives `mission_status` (`active` / `planned` / `done` / `draft`) per mission from the already-present `kanban_stats` WP lane counts — no new I/O, no new endpoints.
- Sorts the selector: **active → planned → done/draft**, with recency-descending as the tiebreaker within each bucket, so the mission you're working on is always at the top of a long list.
- Prefixes each dropdown option with an emoji+text label (`⚡ active:` / `○ planned:` / `✓ done:` / `· draft:`) for instant at-a-glance scanning — plain text so it works in all browsers' native `<select>` elements.

## Changes

| File | What |
|------|------|
| `scanner.py` | `_derive_mission_status(kanban_stats)` helper; `mission_status` field on every feature dict; status-priority bucket added as primary sort dimension |
| `api_types.py` | `mission_status: str` added to `FeatureItem` TypedDict |
| `dashboard.js` | `MISSION_STATUS_LABELS` constant; `getFeatureDisplayName` prefixes the label |
| `tests/test_dashboard/test_scanner.py` | 12 new unit tests: all status derivation cases + sort-order assertions |
| `tests/test_dashboard/test_api_contract.py` | `mission_status` added to `FeatureItem` contract assertion |

## Test plan

- [ ] `pytest tests/test_dashboard/test_scanner.py tests/test_dashboard/test_api_contract.py` — 91 pass
- [ ] `pytest tests/architectural/test_no_legacy_terminology.py` — pass
- [ ] `ruff check` — clean
- [ ] `mypy` — no issues
- [ ] Dashboard smoke: selector shows status prefixes and active missions appear above done missions

### DIFF
--- a/kitty-specs/unified-charter-bundle-chokepoint-01KP5Q2G/baseline/pre-wp23-dashboard-typed.json
+++ b/kitty-specs/unified-charter-bundle-chokepoint-01KP5Q2G/baseline/pre-wp23-dashboard-typed.json
@@ -117,6 +117,7 @@
         "total": 3,
         "weighted_percentage": "<ts>"
       },
+      "mission_status": "active",
       "meta": {
         "friendly_name": "Demo Feature",
         "mission_id": "01HXBASELINEDEMO000000000Z",
--- a/kitty-specs/unified-charter-bundle-chokepoint-01KP5Q2G/baseline/pre-wp23-dashboard-typed.json
+++ b/kitty-specs/unified-charter-bundle-chokepoint-01KP5Q2G/baseline/pre-wp23-dashboard-typed.json
@@ -117,7 +117,6 @@
         "total": 3,
         "weighted_percentage": "<ts>"
       },
-      "mission_status": "active",
       "meta": {
         "friendly_name": "Demo Feature",
         "mission_id": "01HXBASELINEDEMO000000000Z",
--- a/src/specify_cli/dashboard/api_types.py
+++ b/src/specify_cli/dashboard/api_types.py
@@ -220,6 +220,7 @@ class FeatureItem(TypedDict):
     artifacts: dict[str, ArtifactInfo]
     workflow: WorkflowStatus
     kanban_stats: KanbanStats
+    mission_status: str  # "active" | "planned" | "done" | "draft"
     meta: dict[str, Any]
     worktree: WorktreeInfo
     is_legacy: bool  # added by handler, not scanner
--- a/src/specify_cli/dashboard/scanner.py
+++ b/src/specify_cli/dashboard/scanner.py
@@ -225,7 +225,7 @@ def _derive_mission_status(kanban_stats: dict[str, Any]) -> str:
         return "draft"
     if kanban_stats.get("doing", 0) or kanban_stats.get("for_review", 0) or kanban_stats.get("approved", 0):
         return "active"
-    if kanban_stats.get("planned", 0):
+    if kanban_stats.get("planned", 0) == kanban_stats.get("total", 0):
         return "planned"
     return "done"
 
--- a/src/specify_cli/dashboard/scanner.py
+++ b/src/specify_cli/dashboard/scanner.py
@@ -207,8 +207,34 @@ def _coerce_sort_mission_number(value: object) -> int | None:
     return None
 
 
-def _feature_recency_sort_key(feature: dict[str, Any]) -> tuple[bool, float, bool, str, bool, int, str]:
-    """Sort dashboard selector rows newest-first with deterministic legacy fallbacks."""
+# Higher priority = sorted first (list uses reverse=True).
+_MISSION_STATUS_PRIORITY: dict[str, int] = {"active": 3, "planned": 2, "done": 1, "draft": 0}
+
+
+def _derive_mission_status(kanban_stats: dict[str, Any]) -> str:
+    """Derive mission lifecycle status from WP lane counts.
+
+    Returns one of ``"active"``, ``"planned"``, ``"done"``, or ``"draft"``.
+
+    - ``"active"``  — at least one WP is in ``doing``, ``for_review``, or ``approved``
+    - ``"planned"`` — WPs exist but none have been started
+    - ``"done"``    — WPs exist and all are in terminal lanes (done/canceled)
+    - ``"draft"``   — no WPs yet, or the event log is unreadable
+    """
+    if kanban_stats.get("error") or not kanban_stats.get("total", 0):
+        return "draft"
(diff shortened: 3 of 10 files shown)


robertDouglass:
Reroll note on the lifecycle-table gap: a mission with both terminal and planned WPs fits none of the four literal rows in #2586. This tip deliberately classifies it as `planned` (rather than falsely complete or currently executing): no WP is in an active lane, but actionable planned work remains. `done` is now reserved for missions with no active or planned WPs. The mixed-state regression test records that conservative product decision.

rayjohnson:
Small note on the display format: I actually preferred the original icon-first layout (`⚡ active: Mission Name`) over the brackets-at-end version (`Mission Name [⚡ active]`). The icon at the front catches the eye faster when scanning a long list.

That said — I noticed the review agent pushed this change directly to my branch, which I wasn't expecting! Is that intentional behavior? Happy to keep whichever format you prefer, just wanted to flag my preference since I didn't get a chance to weigh in before the commit landed.

stijn-dejongh:
**Held under the 3.2.x merge-hold (high collision risk).** This PR touches dashboard / mission-lifecycle / status surfaces that will churn heavily during the in-flight coordination and doctrine rewrites, so per the [3.2.x delivery plan](../blob/main/docs/planning/3-2-x-approach.md) it's deferred to avoid repeated rebase-conflict against those wide changes. The work is welcome — this is a sequencing hold, not a rejection. Labeled `pr:deferred`; **CI is now skipped on deferred PRs** to save compute. It'll be unblocked (one clean rebase) once the surfaces stabilize. — 3.2.x merge-hold triage

robertDouglass:
## Dialectic adversarial squad review

Architect Alphonso, Debugger Debbie, and Reviewer Renata independently reviewed current head and reproduced lifecycle divergences. Verdict: **REQUEST CHANGES — retain deferred and re-scope onto canonical lifecycle authority**.

### Safe fix pushed

Commit `dc606895a` rebuilds selector options during silent polling while preserving `currentFeature` and page state. Status labels and active-first order now update after WP transitions instead of remaining stale until reload. It uses DOM nodes/`textContent` like initial rendering.

Verification: full focused dashboard/API/static suite `99 passed`; narrowed static suite `8 passed`; Ruff clean; JS syntax clean; diff check clean.

### Blocking findings — report only / rearchitecture

1. **Second, lossy lifecycle authority.** `_derive_mission_status` classifies already-collapsed dashboard columns despite canonical `status.lifecycle.derive_mission_lifecycle()` existing before this PR. Dashboard should consume a canonical lifecycle result (or one shared pure classifier over the already-reduced snapshot), then map canonical states into presentation buckets at one explicit UI boundary.

2. **Materially wrong labels/order.** BLOCKED is collapsed to planned, so actionable blocked work is sorted below active. A terminal mission with authoritative `MissionReopened` stays done here while canonical lifecycle returns reopened/actionable. Recent completion, stale/abandoned, and cancellation semantics are also lost.

3. **Unreadable state is mislabeled draft.** `error` and zero-WP draft are conflated. Corrupt/missing operational state must be explicit unknown/error, not a reassuring draft label sorted last.

4. **Tests pin synthetic collapsed counts.** They cannot catch blocked/reopened/stale/canceled loss. The “done + canceled” test supplies only `done=2`, no canceled state. Add event-log-backed API→DOM conformance cases and transition-during-poll coverage.

5. **Weak duplicate API vocabulary.** `mission_status: str` plus JS’s silent unknown fallback does not constrain the enum. Expose typed canonical lifecycle state; if UX still needs four selector groups, name that field as presentation grouping rather than mission status.

6. **Scope/gate state.** Branch was 491 commits behind and conflicting, including unrelated test-only commits (`/tmp` sentinel and scaffold default). Drop unrelated churn, rebase, and rerun current dashboard/lifecycle gates. Historical CI was green but validates only the branch-era heuristic.

### Concessions

The UX need is legitimate; status/recency sorting is deterministic, initial and polling option rendering are XSS-safe, and the conservative planned+terminal choice matches issue #2586’s literal table. The defect is the table’s authority and vocabulary, not basic implementation quality. Preserve the selector presentation as a thin adapter over canonical lifecycle state.

stijn-dejongh:
Closing as **superseded** — folded into #2606 ✅

Thank you, @rayjohnson — this PR framed the mission-selector lifecycle-status idea and guided the canonical approach. #2606 covers the same surface (`scanner.py` / `api_types.py` / `dashboard.js`, `mission_status` + status-priority sort) as a strict superset, adding the `next_action` hint and an `accepted_at`-aware status derivation, so maintaining both would be redundant.

Your acceptance criteria from **#2586** will be carried into #2606 so nothing is dropped. #2586 stays open and tracked. Closing this PR in favor of the consolidated one — thank you for pointing us at the right design.

PR #2612, opened 2026-07-13
fix: auto-commit during agent action review fails on missions using a coordination worktree

rayjohnson:
## The problem

Running `spec-kitty agent action review WP01` on a mission with a coordination topology would silently fail to commit the status update. The error — `SafeCommitPathPolicyError` — appeared only if you looked closely at the output, and the practical symptom was that `spec-kitty merge` would refuse to proceed because the workspace had uncommitted tracking files.

## The fix in plain terms

When the review command writes its status artifacts, those files live inside a sub-workspace (a git worktree). The code was telling git "these files are relative to the main repo root" — but they weren't, they were relative to the sub-workspace root. Git's safety policy correctly rejected the mismatch. We now detect whether the files are in a sub-workspace and pass the right root.

## Technical details

`_commit_via_legacy_safe_commit` hardcoded `worktree_root=repo_root`. Added `_worktree_root_for_feature_dir(repo_root, feature_dir)` which walks the path: if `feature_dir` is inside `.worktrees/<name>/`, return `.worktrees/<name>` as the root; otherwise return `repo_root`. The derived root is passed as `worktree_root` to `safe_commit`.

## Why this approach

The legacy fallback path (`_commit_via_legacy_safe_commit`) exists as a bridge until WP08 replaces it with a proper legacy bridge. Rather than changing *when* the code takes this path, making the path itself robust is the right defensive move — it honours the existing `worktree_root` contract that `safe_commit` already enforces, and shrinks the blast radius of the WP08 TODO instead of widening it.

## Test plan

- `test_worktree_root_for_feature_dir_inside_worktrees` — unit-tests the new helper for a coord worktree path
- `test_worktree_root_for_feature_dir_in_primary_checkout` — unit-tests the helper returns `repo_root` for primary-checkout paths
- `test_commit_workflow_change_legacy_path_uses_coord_worktree_root` — integration-level: verifies `safe_commit` receives the coord worktree root (not `repo_root`) when `feature_dir` is inside `.worktrees/`
- Updated `test_legacy_workflow_commit_refused_on_protected_target` to pass the new required `worktree_root` parameter
- Updated architectural allow-list line anchors (`test_no_write_side_rederivation.py`) for the line shift introduced by the new helper function

### DIFF
--- a/src/specify_cli/cli/commands/agent/workflow.py
+++ b/src/specify_cli/cli/commands/agent/workflow.py
@@ -37,6 +37,7 @@ from __future__ import annotations
 from specify_cli.core.constants import (
     KITTY_SPECS_DIR,
     MISSION_TYPE_RESEARCH,
+    WORKTREES_DIR,
 )
 from specify_cli.missions._read_path_resolver import (
     _canonicalize_primary_read_handle,
@@ -538,9 +539,27 @@ def _resolve_workflow_read_dir(
     return read_dir
 
 
+def _worktree_root_for_feature_dir(repo_root: Path, feature_dir: Path) -> Path:
+    """Return the correct worktree_root for safe_commit, derived from feature_dir.
+
+    When feature_dir lives inside .worktrees/<name>/..., returns
+    .worktrees/<name> so that absolute paths under that worktree normalize
+    correctly (preventing SafeCommitPathPolicyError on paths whose first
+    component is .worktrees/).  For primary-checkout feature_dirs, returns
+    repo_root unchanged.
+    """
+    worktrees_parent = repo_root / WORKTREES_DIR
+    try:
+        rel = feature_dir.resolve().relative_to(worktrees_parent.resolve())
+        return worktrees_parent / rel.parts[0]
+    except ValueError:
+        return repo_root
+
+
 def _commit_via_legacy_safe_commit(
     *,
     repo_root: Path,
+    worktree_root: Path,
     target_branch: str,
     paths: list[Path],
     message: str,
@@ -553,7 +572,7 @@ def _commit_via_legacy_safe_commit(
     # tracking ``main``) is REFUSED by the guard, never waived (FR-008).
     result = safe_commit(
         repo_root=repo_root,
-        worktree_root=repo_root,
+        worktree_root=worktree_root,
         target=CommitTarget(ref=target_branch),
         message=message,
         paths=tuple(paths),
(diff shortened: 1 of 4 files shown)


stijn-dejongh:
**Held under the 3.2.x merge-hold (high collision risk).** This PR touches dashboard / mission-lifecycle / status surfaces that will churn heavily during the in-flight coordination and doctrine rewrites, so per the [3.2.x delivery plan](../blob/main/docs/planning/3-2-x-approach.md) it's deferred to avoid repeated rebase-conflict against those wide changes. The work is welcome — this is a sequencing hold, not a rejection. Labeled `pr:deferred`; **CI is now skipped on deferred PRs** to save compute. It'll be unblocked (one clean rebase) once the surfaces stabilize. — 3.2.x merge-hold triage

stijn-dejongh:
**Maintainer take-over (rebase + land).** This PR directly fixes the `SafeCommitPathPolicyError` we just hit on a real coord-topology mission — `agent action review` failing to commit status into the coord sub-worktree (the acute half of the #2841 coord-bookkeeping friction; repro filed on #2841). It's currently `CONFLICTING` against main, so I'm rebasing it onto current `main`, verifying the fix + tests, and driving CI green. Thanks @rayjohnson — carrying your fix forward.

stijn-dejongh:
Rebased-and-verified take-over opened as #2868 (this branch had gone `CONFLICTING`; main refactored `workflow.py` since, so the fix was re-applied onto the new `workflow_executor.py` call site — where the legacy leaf was calling `_commit_via_legacy_safe_commit` without the now-required `worktree_root`). 47 tests pass, ruff/mypy clean on the diff. Authorship credited to @rayjohnson. Suggest closing this in favor of #2868 once reviewed.

stijn-dejongh:
Closing for the same reason its rebased take-over #2868 was closed: a 3-lens adversarial squad + an empirical repro confirmed this fix is **unreachable in the coordination case**. The corrected `safe_commit(worktree_root=…)` is gated by a sibling porcelain pre-check (`git status --porcelain` at `cwd=repo_root`, `workflow.py:618`) that returns **empty** for `.worktrees/`-gitignored coord files — firing a phantom "committed" early-return before `safe_commit` runs — and real coord missions route through the **modern** `_commit_via_coordination_transaction`, not the legacy leaf this patches.

The `safe_commit` worktree-root root-cause you found is real and valuable, @rayjohnson — it's the surrounding porcelain gate that undoes it, so a call-site patch here isn't sufficient. The correct coord-commit fix is folded into a dedicated **coordination-branch trust / write-placement** friction-remediation mission, where it's subsumed by routing every mission-artifact write through one placement port (staging directly in the target ref's worktree, so the porcelain pre-check and `safe_commit` both target the coord worktree). Full root-cause + scope: the engineering notes in #2870. Thank you for surfacing this.

Issue #2069, opened 2026-06-22
Design: `MissionTopology` enum + `topology_for_mission(...)` SSOT resolver returning the ExecutionContext (ActionContext) value object


stijn-dejongh:
> **DESIGN ticket** — this is an architectural proposal, not an implementation spec. It defines a target shape (a `MissionTopology` enum + a single `topology_for_mission(...)` resolver seam) for review/adoption. No WP slicing here.

## Problem

A mission can take one of **four distinct shapes** with respect to lanes and coordination:

| Operator shape | Concrete signal today |
|---|---|
| **lanes** | `lanes.json` present, **no** coordination branch |
| **lanes-with-coord** | `lanes.json` present **and** a coordination branch resolved |
| **flattened** | coordination branch dropped → landing == coordination == target on one branch |
| **all-on-feature** | **no** `lanes.json`, single branch, no coord |

Today **none of these four shapes is a first-class, named value**. Topology is re-inferred ad-hoc at many seams from scattered signals:

- **`CommitTargetKind` {`PRIMARY`, `COORDINATION`, `FLATTENED`}** (`src/mission_runtime/context.py:51`) is a *per-ref* classification of a single `CommitTarget`, **not** a mission-level topology. It cannot distinguish **lanes** from **all-on-feature** (both resolve `PRIMARY`/`FLATTENED` for the same ref), because it knows nothing about `lanes.json`.
- **Coordination presence** is re-derived per call via `_resolve_coordination_branch(...)` (`src/mission_runtime/resolution.py:705`, and again in `src/runtime/next/runtime_bridge.py:94`) — each caller branches on `coordination_branch is None` to pick `FLATTENED` vs `COORDINATION`.
- **Lane presence** is re-derived per call via `read_lanes_json(...)` / `require_lanes_json(...)` across `src/specify_cli/workspace/context.py`, `src/specify_cli/core/worktree_topology.py:133`, `src/specify_cli/cli/commands/implement.py`.
- **`FeatureTopology` / `WPTopologyEntry`** (`src/specify_cli/core/worktree_topology.py:41`) is a *per-WP lane-stacking* view (execution-mode level: `worktree` vs `code_change`, base branch, lane_id), **not** a mission-shape classifier.

[...]

PR #2614, opened 2026-07-13
fix(doctor): add --fix to doctor coordination for dangling coord branches

rayjohnson:
## Problem

Missions whose `meta.json` carries a `coordination_branch` key pointing to a git branch that was never created are completely unreachable. `spec-kitty next` hard-errors with `CoordinationBranchDeleted`, and there is no automated recovery path — `doctor coordination` correctly diagnoses the issue but has no `--fix` flag to act on it. Operators must manually edit `meta.json` and run `migrate backfill-topology`.

A consumer repo reported 7 missions stuck in this state, all 100% done but locked out of close/review/merge.

## Fix

`doctor coordination --fix` now automatically repairs missions with `COORDINATION_WORKTREE_NEVER_CREATED` findings:

1. Removes the stale `coordination_branch` key from `meta.json` (via `write_meta(validate=False)` to tolerate legacy missions).
2. Runs `backfill_topology_repo` so topology is re-derived from the now-absent key (`SINGLE_BRANCH` or `LANES`).
3. Re-collects findings so the exit code reflects the repaired state.

Also improves the `CoordinationBranchDeleted` error message to lead with `spec-kitty doctor coordination --fix` rather than the manual edit steps.

## Technical details

- `_collect_coordination_findings` now stamps `extra["meta_path"]` on every `COORDINATION_WORKTREE_NEVER_CREATED` finding so the fixer knows which file to patch without a second filesystem scan.
- New `_fix_never_created_branches(findings)` does the key removal.
- `run_coordination_health` gains a `fix: bool` param; `doctor.py` wires it to `--fix` with a clear help string and usage examples.

## Why this approach

The doctor already had full diagnosis. Adding `--fix` completes the feature rather than punting to a separate migrate subcommand. Using `write_meta(validate=False)` is the established pattern for tolerant writes to legacy meta files. Re-collecting findings after the fix (rather than filtering the original list) means the exit code is always ground-truth.

The upstream write-site bug (coordination_branch written before branch creation) was already fixed in the current codebase — `mission_creation.py:435` only writes after `ensure_coordination_branch()` succeeds. The stuck missions are a legacy artifact.

## Test plan

- [ ] 4 fast unit tests for `_fix_never_created_branches`: removes key, skips wrong error codes, skips missing `meta_path`, idempotent when key already absent
- [ ] 1 fast unit test: `_collect_coordination_findings` injects `meta_path` on `COORDINATION_WORKTREE_NEVER_CREATED` findings
- [ ] 1 git_repo integration test: `_collect` injects `meta_path` on a real repo with a stale coord branch
- [ ] 1 git_repo integration test: full fix+backfill round-trip — key removed, `backfill_topology_repo` writes `single_branch` topology

### DIFF
--- a/src/specify_cli/cli/commands/_coordination_doctor.py
+++ b/src/specify_cli/cli/commands/_coordination_doctor.py
@@ -572,6 +572,13 @@ def _fix_never_created_branches(findings: list[DoctorFinding]) -> list[str]:
         if "coordination_branch" not in meta:
             continue
         del meta["coordination_branch"]
+        # Clear the stale stored topology so backfill_topology_repo re-derives it:
+        # backfill never overwrites an existing topology (migration/backfill_topology.py),
+        # and every mission minted post-#2069 stores `topology` at create time — leaving it
+        # would keep the mission routed through coordination (false-green flatten). Record the
+        # flatten via the provenance flag. (#2614 adversarial-squad remediation)
+        meta.pop("topology", None)
+        meta["flattened"] = True
         write_meta(mission_dir, meta, validate=False)
         fixed.append(mission_dir.name)
     return fixed
--- a/src/specify_cli/cli/commands/_coordination_doctor.py
+++ b/src/specify_cli/cli/commands/_coordination_doctor.py
@@ -538,9 +538,43 @@ def _collect_coordination_findings(repo_root: Path) -> list[DoctorFinding]:
         meta = load_meta(mission_dir, on_malformed="none")
         if meta is None:
             continue
-        findings.extend(_check_coordination_worktree_health(repo_root, meta))
+        coord_findings = _check_coordination_worktree_health(repo_root, meta)
+        for f in coord_findings:
+            if f.error_code == "COORDINATION_WORKTREE_NEVER_CREATED":
+                f.extra["meta_path"] = str(mission_dir / "meta.json")
+        findings.extend(coord_findings)
         findings.extend(_check_lane_sparse_checkout_drift(repo_root, meta))
     return findings
+
+
+def _fix_never_created_branches(findings: list[DoctorFinding]) -> list[str]:
+    """Remove stale ``coordination_branch`` keys from meta.json.
+
+    Targets only ``COORDINATION_WORKTREE_NEVER_CREATED`` findings that carry
+    a ``meta_path`` in their ``extra`` dict (populated by
+    :func:`_collect_coordination_findings`). After removal, call
+    :func:`~specify_cli.migration.backfill_topology.backfill_topology_repo` so
+    topology is re-derived from the now-absent key.
+
+    Returns a list of mission slugs that were modified.
+    """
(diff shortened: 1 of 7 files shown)


stijn-dejongh:
👋 @rayjohnson — thanks for this fix; it lands a real recovery path for the `coordination_branch`-with-no-worktree failure class (we've hit the exact "100% done but locked out" scenario in the field).

**We're picking this up for a landing pass.** Heads-up on what we're doing so nothing surprises you:

1. **Rebased** onto current `upstream/main` locally — your base was ~109 commits behind. The rebase was clean (no conflicts), and your 65 tests still pass.
2. Ran a 4-lens adversarial review squad (architecture / duplication / correctness / contract). Early consensus is **LAND-WITH-FOLLOWUPS** — the repair is correct and targets the root datum, not a symptom. We're going to fold the review remediations **directly into this branch** rather than defer them, so it lands clean:
   - Delegate the inline key-removal to the canonical `clear_coordination_metadata()` helper (`mission_metadata.py`) that `mission close --discard` already reuses, preserving the tolerant-load skip.
   - Add a CLI-level test driving `run_coordination_health(fix=True)` that pins the exit-code-after-re-collect behavior **and** the reworded operator instruction strings (currently correct but unpinned).

Your authorship/attribution is preserved on the original commit; remediation lands as follow-up commits on top. Assigning to myself for the landing pass. Will report back here before merge.

stijn-dejongh:
✅ Pushed the rebased + remediated branch (via maintainer-edits — your authorship is preserved on the original commit; remediation is a follow-up commit).

**Two commits now:**
- `Ray Johnson` — original `doctor coordination --fix` (rebased onto current `main`, clean)
- `Stijn Dejongh` — squad-remediation follow-up

**What the remediation fixed** (all red-first, independently verified: reverting the src makes the new tests go RED):
- **[HIGH] false-green flatten** — removing `coordination_branch` alone left the stored `topology:"coord"` intact (`backfill_topology` never overwrites an existing topology, and missions store it at create time since #2069/#2218), so a real coord mission stayed routed through coordination while doctor reported success. Now also clears `topology` + sets `flattened:true` so backfill actually re-derives `single_branch`/`lanes`.
- **[MEDIUM] data-loss** — `_coord_branch_exists` probed local `refs/heads/` only, so a genuinely-coord mission whose branch lives only on a remote (fresh clone, CI checkout, pruned local) was flagged `NEVER_CREATED` and had its `coordination_branch` **deleted** by `--fix`. Now also probes `refs/remotes/*` before concluding the branch was never created.
- **[coverage]** added end-to-end tests for `run_coordination_health(fix=True)` exit-code transitions, the stored-topology flatten, the remote-only-branch guard, and a message-pin for the `CoordinationBranchDeleted` operator instruction.

`69 passed`, `ruff`/`mypy` clean locally. CI will run on the new head.

**Tracked follow-ups (not in this PR):** unify a canonical `flatten_mission()` seam shared by `doctor`/`mission close --discard` (the flatten mutation is currently expressed in two places); broader message-contract coverage.

Issue #392, opened 2026-04-05
RTK command interception silently breaks advanced shell flags


stijn-dejongh:
## Parent

Part of #391 — Tech Debt Remediation

## Problem

RTK wraps `find`, `grep`, `ls`, `git` with its own UI layer. Advanced flags (`find -exec`, `grep --type`, `grep -c`, compound predicates) are silently dropped or rejected. Agents and developers relying on standard POSIX shell behaviour get wrong results with no error signal.

## Impact

Any agent workflow or developer script that goes beyond trivial single-flag commands. Affects search, bulk edit, and verification tasks. Discovered during mission 063 but applies to all shell-driven automation.

## Workaround

- Use `rg` (ripgrep) directly for content search (not `grep`, not `rtk grep`)
- Use `sed -i` with shell globs for bulk edits (not `find -exec sed`)
- Use dedicated tooling (Glob, Read) for file listing (not `find`, not `rtk find`)

## Acceptance criteria

Either:
1. RTK passes through unknown flags transparently to the underlying command, or
2. The constraint is documented in `AGENTS.md` and agent profiles so agents know to avoid intercepted commands

stijn-dejongh:
My main concern here is that I'd like to add provisions for certain recommended tools, that are known and established as working well with agent-enabled processes and dealing with generic OS-specific issues.

RTK and RG ( ripgrep ) are indeed "current trends", and I want to be careful not to increase coupling. Though, I feel we need to provide some guidance to our userbase on how to deal with similar issues.

Maybe the answer is in using the charter to contain such specifics, with a guided intake/research/self-inspection procedure or paradigm shipped in the doctrine.

robertDouglass:
I actually don't want us to recommend any tools. 
I like the toolguide doctrine primitive as something that emerges from charter interviews and/or existing brownfield repo inspection. 
I just don't see it as our job to tell people what tools to use beyond using Spec Kitty. 
We can offer toolguides for download like people offer skills.

PR #2628, opened 2026-07-13
feat(sync): stage charter-lint decay report into the mission dossier (#2481)

### LINKED ISSUE #2481: Ship charter-lint decay report into the dossier so Teamspace can surface it (unblocks saas #392)
## Context

Teamspace SaaS #392 wants a decay-watch surface (charter-lint findings: orphan / contradiction / staleness / reference-integrity), matching the local dashboard's Decay Watch tile. The SaaS render side is being built now, but it has nothing to render because **the lint report never reaches the dossier**.

## The gap

The charter-lint report is written to `.kittify/lint-report.json`. The dossier indexer deliberately skips hidden paths:

- `src/specify_cli/dossier/indexer.py` (`_scan_directory`) yields files that are "non-hidden, non-.git/.kittify" and skips any path with a part starting with `.`.
- There is no `MissionArtifactKind` for the lint report (unlike `lanes.json` / `acceptance-matrix.json`, which do ship).

So `lint-report.json` is local-dashboard-only today and is structurally excluded from the push pipeline.

## What SaaS needs

Any one of these would unblock #392 (SaaS just needs the report to arrive as an ingestible artifact with a stable path):

1. **Promote it to a shipped artifact** — add a `MissionArtifactKind` (e.g. `DECAY_REPORT` / `CHARTER_LINT`) and emit the report to a non-hidden, dossier-indexed location (or add a narrow allow-list exception so this one `.kittify` file is indexed). Lowest-friction, mirrors how lanes/acceptance-matrix already ship.
2. **Event projection** — emit lint findings as events the SaaS reduces into a decay projection. Heavier, but fits the event-first direction.

The serialized schema is already stable and self-describing (`charter_runtime/lint/findings.py`): `DecayReport` = `scanned_at`, `feature_scope`, `duration_seconds`, `drg_node_count`, `drg_edge_count`, `graph_state`, `finding_count`, `findings[]`; each finding = `category`, `type`, `id`, `severity`, `message`, `feature_id`, `remediation_hint`.

## Ask

This is your lane (CLI/sender), so flagging rather than pushing a change. Which mechanism do you prefer? Happy to draft the artifact-kind version as a PR if option 1 is the direction. The SaaS render will land behind this as `near_parity` (surface ready, awaiting sender) until the report ships.

cc @Priivacy-ai/maintainers

LynnColeArt:
Closes #2481. Unblocks saas #392 (Teamspace decay-watch surface).

## Problem
The charter-lint engine writes one repo-global report at `.kittify/lint-report.json`. The dossier indexer skips it: hidden path, and no artifact kind is registered. So Teamspace never receives the decay findings.

## Approach
Stage the report into a mission's dossier at sync time, before indexing, and only when the report's `feature_scope` matches that mission's slug. A report from `charter lint --mission <slug>` then flows through the existing per-mission dossier pipeline unchanged. The lint engine is not touched.

- `sync/lint_report_staging.py` — scoped, best-effort staging helper (never raises; no-ops on missing/corrupt report, scope mismatch, or unwritable dir).
- `sync/dossier_pipeline.py` — invoke staging as Step 0 of `sync_feature_dossier`.
- `sync/body_upload.py` — allowlist `lint-report.json` so the findings body ships, not just the metadata event.
- `missions/*` + `doctrine/missions/*` `expected-artifacts.yaml` — register `runtime.charter-lint.decay` (class `runtime`, path `lint-report.json`, non-blocking) in all six manifests.

Kept entirely in the `sync` package to preserve the `dossier ↛ sync` import boundary.

## Why this shape
`feature_scope` filtering keeps a mission's dossier from picking up a report meant for another mission (the report file is a single repo-global slot the engine overwrites). Staging at sync time means the feature dir is already resolved, and the report stays fresh on every sync.

## @stijn — design calls for your review (this is why it is a draft)
This is CLI-side / your turf; I took it because it was the one thing gating saas #392 and it was unassigned. Three things I want your read on before this leaves draft:

1. **Untracked file in the mission dir.** Staging writes a non-hidden `lint-report.json` into `kitty-specs/<mission>/` on each sync. It has to be non-hidden for the current indexer scan to see it. It is git-untracked and rewritten per sync. Should it be gitignored, or cleaned after body upload, or is the noise fine?
2. **Unscoped runs not surfaced.** Reports with `feature_scope: null` (repo-global lint) are deliberately skipped. Fine for MVP? If unscoped runs should surface, that is a follow-up and needs a routing decision (which mission dossier gets a repo-global report).
3. **Slug comparison.** I compare `report.feature_scope` against `namespace_ref.mission_slug`. Holds today since both are the mission slug. Also: `locate_project_root` resolves a worktree feature dir back to the main repo root, so in a worktree topology the report read from main's `.kittify/` may lag a worktree-local lint run. Acceptable for the plan-on-primary flow, but flagging it.

Happy to reshape any of this, or hand it back if you would rather own it.

## Evidence
- Targeted: `tests/sync/test_lint_report_staging.py` (7 new), `test_body_upload.py` (+1), `test_dossier_pipeline.py` (+1), `tests/specify_cli/dossier/test_integration.py` (+1), `test_dossier_sync_boundary.py` — **100 passed** (re-run independently).
- `ruff check` on touched source: clean. `mypy` on the 3 modified modules: clean.
- Full `tests/sync/` green except one known real-port daemon-singleton test that only fails under parallel `-n auto` and passes serially with `-n0` (per CLAUDE.md, those run serially) — unrelated to this change.

### DIFF
--- a/docs/plans/3-2-x-milestone-roadmap.md
+++ b/docs/plans/3-2-x-milestone-roadmap.md
@@ -2,7 +2,7 @@
 title: 3.2.x Milestone — Roadmap
 description: 'Operator-facing roadmap for the 3.2.x milestone: the epic dependency spine, degod/unshim wave status, milestone census, exit criteria, and watch items.'
 doc_status: active
-updated: '2026-07-13'
+updated: '2026-07-14'
 related:
 - docs/changelog/index.md
 - docs/plans/index.md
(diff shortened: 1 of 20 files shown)


stijn-dejongh:
## Landing adjudication — @stijn-dejongh (maintainer pass)

Ran an adversarial review squad (architecture-alignment weighted, per operator request) + rebased-tip verification, then pushed two remediation folds. **Two commits added on top of your work (SHAs preserved, fast-forward — no history rewrite):**

- `landing fold: match charter-lint feature_scope against the mission's full identity alias set`
- `landing fold: drop functionally-inert doctrine-tree expected-artifacts entries`

### Squad verdict: 3 lenses, mechanism sound — but one real blocker

Three profile-loaded lenses (architecture / SSOT-duplication / correctness) independently agreed the mechanism is **boundary-correct, additive, and non-regressing**: the new module correctly lives in `sync` (which may import `dossier`, not vice-versa), reads the producer's *output file* not its API, resolves the same main-repo `.kittify/` root from a `.worktrees/` feature_dir, and reuses the existing artifact model (`artifact_class: runtime` is a real `ArtifactClassEnum`; `.json` is a supported inline format) rather than inventing a `MissionArtifactKind` — the right call.

**But as originally written the feature was effectively inert.** The staging gate compared `report["feature_scope"]` — stored **verbatim** from whatever handle the operator typed into `charter lint --mission <X>` — against `namespace_ref.mission_slug`, which the dossier-sync consumers deliberately re-key to the canonical **directory name** (`feature_dir.name`; see the explicit comment in `cli/commands/research.py:71-78`). Those two identity forms differ in the general case (bare mid8 / numeric prefix / slug vs. directory name), so a correctly-scoped lint run would silently fail to stage and nothing would reach the SaaS dossier (#392). No test caught it — every test hand-picked equal strings and hand-built the report dict.

### Fold 1 — identity alias-set match (the fix)
`stage_charter_lint_report` now resolves the mission's full identity via the canonical `resolve_mission_identity(feature_dir)` and matches `feature_scope` against **any** of `{mission_id, mid8, mission_slug, feature_dir.name}`. Best-effort: if identity can't resolve (no `meta.json`) it degrades to the directory-derived handles — the prior behaviour, so no regression.

**Red-first proven:** the new parametrized test (report scoped under `mission_id` / `mid8` / `slug`, consumer keying by a *different* directory name) **fails on all three under the old predicate** and passes with the fix. It round-trips a real `DecayReport.to_json()` so the gate stays coupled to the producer's serializer. Also added the previously-untested `OSError`-on-write branch.

### Fold 2 — drop inert doctrine-tree edits (architecture alignment, #883)
Only the 3 `src/specify_cli/missions/*/expected-artifacts.yaml` entries are load-bearing — the dossier `ManifestRegistry` reads that tree exclusively. The parallel `src/doctrine/missions/*/` entries are consumed only by `DoctrineMissionRepository.get_expected_artifacts`, which has **no product caller** (tests only). They neither help nor guard anything; they only enlarge the `specify_cli`/`doctrine` split-brain that **#883** targets collapsing into one shipped mission-type source. Removed; kept the 3 load-bearing entries.

### Local evidence (rebased onto current `main` + on the delivered branch state)
- Touched-tests: **103 passed** (was 98 + 5 new). ruff + mypy **clean** on the changed module.
- Arch guards green: `test_no_legacy_terminology`, `test_dossier_sync_boundary`, `test_lane_regression_guard` (1090 passed).
- Merge into current `main` is CLEAN (`git merge-tree`), zero conflicts.

### Recommended follow-ups (out of scope for this fold, non-blocking)
1. **SSOT — canonical lint-report path accessor.** `.kittify/lint-report.json` is now hardcoded in 3 places (engine writes; dashboard + staging read). Extract `charter_runtime.lint.report_path(repo_root)` consumed by all three.
2. **Stale-copy lifecycle.** Once staged, `feature_dir/lint-report.json` is never refreshed/removed on later null-scope runs — it keeps shipping last-known decay state as "current." Decide refresh-on-every-sync vs. remove-when-scope-no-longer-matches. (`scanned_at` lets the consumer judge freshness in the meantime.)
3. **Two-tree parity guard** for the `expected-artifacts.yaml` split (already drifted on `main`) — best folded into the #883 unification.

Happy to file 1–3 (2 under #883, 1 fresh) on your nod.

### Hand-off
CI re-running on the new head; **merge-ready on green.** The operator merges — I have not run `gh pr merge`. No landing-order constraint.

stijn-dejongh:
## Two more folds pushed (fast-forward — your SHAs preserved)

Per maintainer agreement, folded the two in-scope follow-ups into the branch (the third — a two-tree `expected-artifacts.yaml` parity concern — is recorded on #883 instead, since a "keep both in sync" guard would fight that issue's collapse-to-one-source goal).

- `landing fold: single-source the .kittify/lint-report.json path via core.paths.lint_report_path` — the report location was hardcoded in **three** places (engine writer, dashboard reader, dossier stager). Added `LINT_REPORT_FILENAME` to `core.constants` (mirroring `RETROSPECTIVE_FILENAME`) + `lint_report_path(repo_root)` to `core.paths`, routed all three through it. Homed in `core.paths` (not the producer package) so **no new cross-package import edge** is introduced — the sync boundary gate is untouched. Focused accessor test added.
- `landing fold: reconcile the staged lint report downward when the source is re-scoped` — closes the stale-copy gap: once a scoped run staged a copy, a later null/global or different-mission run left it shipping stale decay state forever. Now a valid report with a **positively-different** scope drops the stale copy; transient/anomalous reads deliberately preserve last-known state. **Red-first** tests cover both reconcile cases + the preserve-on-transient guard.

### Evidence (rebased onto current `main`)
- Broad affected suite + arch guards: **1292 passed** (`core.paths` is widely imported — cast a wide net incl. `test_shared_package_boundary`, `test_dossier_sync_boundary`, terminology, lane-regression). ruff + mypy clean on all new/changed code.
- Both new fixes proven **red-first** (fail without the fix, pass with it).
- The 3 pre-existing mypy warnings in `dashboard/handlers/lint.py` (subclass-Any + 2 "unused type: ignore") are mypy-in-isolation artifacts — confirmed present on the untouched file; the CI `lint` job does not flag them.

Branch now: your 2 commits + **4 folds**. CI re-running; **merge-ready on green.** Operator merges — I have not run `gh pr merge`.

Issue #2665, opened 2026-07-15
Sync daemon can die silently → auto-sync halts for weeks, and legacy-queue recovery deadlocks (dogfood find)


LynnColeArt:
## Discovery
Found while dogfooding Team Kitty in Docker locally (draining a real spec-kitty checkout into the local SaaS to validate CLI→Team Kitty back-propagation). One `sync opt-in` surfaced a full chain of failures. Sync is supposed to be required, replayable, and **automatic** (#2180); this is the concrete story of how it silently isn't.

## Root cause: the sync daemon died silently and nothing noticed
`spec-kitty doctor orphan-daemons` reported an orphan sync daemon: **PID 2685400, port 9400, version 3.2.0rc41, started 2026-06-09** (`~/.spec-kitty/sync/daemon/owner.json`). The process is long dead. Auto-sync stopped when it died, silently, and **321 events accumulated** in the legacy queue (`~/.spec-kitty/queue.db`, mode TEAMSPACE, retained 321, delivered 0/0) over ~5 weeks. Nothing surfaced that sync had stalled.

## Three concrete defects (one incident exposed all three)

**1. Silent daemon death, no restart, no alert.** A dead sync daemon halts auto-sync indefinitely with zero operator-visible signal. Nothing restarts it; nothing warns that sync is stale. (Fails 'automatic' #2180 and 'honest' #2264.)

**2. Circular remediation message.** With the orphan cleared, `sync opt-in` and `sync now` both refuse with:
> Sync boundary refused: … 321 legacy rows in scope. Remediation: Run `spec-kitty sync now` to flush 321 legacy rows …

`sync now` is exactly the command that just refused — a dead-end loop. The real remediation is `spec-kitty sync migrate`. Source: `src/specify_cli/sync/preflight.py:327` emits the wrong command.

**3. `sync migrate` deadlocks on conflicts with no forward path.** On real data:
```
Queue migration: imported 135  deduped 0  conflicts 321  source_errors 0  (exit_code 1)
Cleanup blocked: unresolved migration conflicts … remain before deleting source queues.
```
Boundary still FAILED afterward (321 rows still in scope). `sync migrate` has no `--resolve`/`--force`/conflict-resolution flag, so an operator with conflicting legacy rows has no way forward.

## Impact
[...]

Issue #2180, opened 2026-06-26
P0: make Teamspace dossier sync required, replayable, and automatic


robertDouglass:
## Problem

Teamspace dossier dashboards can be empty even when mission artifacts were written and committed locally. This violates the product invariant: as long as `spec-kitty auth` is present and the checkout is Teamspace-bound, users must be able to open Teamspace and see all dashboards for all missions without manual intervention.

Incident evidence from `sync-drain-queue-isolation-01KVZQ3Y`:

- Local repo has dossier artifacts under `kitty-specs/sync-drain-queue-isolation-01KVZQ3Y/`: `spec.md`, `plan.md`, `tasks.md`, WP files, contracts, checklists, research, etc.
- Teamspace main had the mission and 74 non-dossier events.
- Teamspace had zero dossier projection rows for that mission:
  - `ArtifactNamespace`: 0
  - `DossierIndexEntry`: 0
  - `DossierSnapshot`: 0
  - `MissionDossier*` events: 0

So this was not a rendering issue. The CLI never durably produced/replayed the dossier events Teamspace needs.

## Required invariant

For authenticated Teamspace-bound checkouts:

> The local artifact state under `kitty-specs/<mission>/` must automatically converge to Teamspace dossier projections. No manual reindex. No manual dashboard repair. No silent best-effort drops.

If auth/network/server are unavailable, work must remain durable in the existing scoped queue DB and sync automatically when conditions recover.

## Current failure mode

Dossier sync is currently a best-effort side effect:

- `trigger_feature_dossier_sync_if_enabled()` returns `None` when SaaS sync is disabled or when internal setup fails.
- status emission calls dossier sync fire-and-forget.
- if no dossier handler/emitter is registered, events are silently dropped/no-op.
- `sync now` mostly drains existing queues; it does not first reconcile local mission artifact state into queued dossier events/body uploads.
- tests currently encode disabled/no-op behavior, which is incompatible with Teamspace-critical dossier visibility.

## Architecture requirement

Add a required, replayable `DossierReconciler` owned by `spec-kitty` CLI.

The reconciler must:

- scan/reconcile mission artifact state into `MissionDossierArtifactIndexed`, `MissionDossierArtifactMissing`, and `MissionDossierSnapshotComputed` events;
- queue body uploads for supported artifact bodies;
- be idempotent by namespace + artifact path + content hash + snapshot hash;
- run before queue draining in `spec-kitty sync now`;
[...]

PR #2682, opened 2026-07-16
fix(sync): converge legacy-row boundary after a clean queue→journal migrate (#2665)

LynnColeArt:
Fixes the **clean-path** half of the #2665 migrate deadlock, found dogfooding the CLI→TeamSpace back-prop.

## The bug
`spec-kitty sync migrate` imports queued events into the journal **read-only** and never deletes them from the source queues (by design — "source DBs are opened read-only"). The `cleanup_blocked` gate and the "resolve them before deleting source queues" message anticipate a cleanup step **that was never implemented**. So legacy `queue.db` rows persist, `preflight.legacy_rows_for_scope` keeps counting them, and `sync now` / `sync opt-in` refuse forever — no operator path to a coherent boundary.

## The fix (gated, safe)
- `OfflineQueue.remove_events(ids)` — delete queued rows by `event_id`, never a blanket clear.
- `MigrationAudit.event_ids_for_source()` — the imported/deduped (non-conflict) ids for a source.
- `cleanup_migrated_sources()` — on a **clean** migration (`cleanup_blocked == False`), delete from each source the ids provably present in the journal. **No-op when blocked**; deletes only confirmed-migrated ids, by id.
- `sync migrate` runs cleanup automatically on a clean migration; **`--no-cleanup`** preserves inspect-first.

**Why deletion is safe:** delivery reads solely from the journal (the legacy offline-queue drain is retired), so a row proven durable in the journal remains deliverable after its source copy is removed. Import stays read-only — the existing FR-013/T058 import-phase guarantees are untouched; cleanup is a *separate*, gated, post-commit step.

## Tests
`tests/sync/test_migrate_journal.py` (+5): convergence (sources drain to 0, journal intact), blocked no-op (conflict → nothing deleted), idempotency, the DB-error branch (reported, non-fatal), and `remove_events`. Full suite green; ruff + mypy clean.

## Verified live
Ran against the local dogfood runtime: on a clean migration the sources drain and the boundary converges; under conflicts cleanup **correctly no-ops** and leaves every source intact.

## Known follow-up (the *other* #2665 dimension — needs a decision)
The live rig also surfaced the **migrate-conflict deadlock**: divergent-duplicate conflicts (same `event_id`, divergent payload vs the per-target journal) quarantine correctly but have **no operator resolution path**, so cleanup stays blocked and the boundary can't converge. Import can't overwrite (C-005/FR-018) and cleanup won't force-delete conflicted rows, so there's no way out today. Proposed resolution: a `--resolve-conflicts=keep-journal` that discards source rows whose `event_id` is already durable in the journal (journal = canonical/delivered; the source copy is a stale supersede). Tracked under #2665 / #2180.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/2682"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786791628&installation_id=146454894&pr_number=2682&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F2682&signature=7fff46795d87336514250faeb9597f08385665bb8e6a5e0a94980af8c71d0e4a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/specify_cli/cli/commands/sync.py
+++ b/src/specify_cli/cli/commands/sync.py
@@ -1396,7 +1396,8 @@ def _auto_converge_legacy_on_enable() -> None:
     surfaced with the explicit ``sync migrate --resolve-conflicts keep-journal``
     recovery, and the coherence gate that follows still refuses until they are
     resolved. Idempotent and lossless: a no-op on an already-converged runtime.
-    Any runtime-open failure is swallowed so the coherence gate handles it.
+    Any failure opening the runtime OR converging is swallowed so the coherence
+    gate that follows handles it, rather than aborting opt-in with a traceback.
     """
     from specify_cli.paths import get_runtime_root
     from specify_cli.sync.migrate_journal import (
@@ -1420,10 +1421,13 @@ def _auto_converge_legacy_on_enable() -> None:
             resolve_conflicts=False,
             cleanup=True,
         )
+    except Exception:  # converge failed — let the coherence gate report it
+        return
     finally:
         with contextlib.suppress(Exception):
             audit.close()
-        runtime.close()
+        with contextlib.suppress(Exception):
+            runtime.close()
 
     deleted = converge.cleanup.total_deleted if converge.cleanup is not None else 0
     if deleted:
--- a/src/specify_cli/cli/commands/sync.py
+++ b/src/specify_cli/cli/commands/sync.py
@@ -1386,6 +1386,58 @@ def opt_out(
     )
 
 
+def _auto_converge_legacy_on_enable() -> None:
+    """Converge any legacy queue residue into the journal when enabling sync (#2665).
+
+    Turning sync on for a checkout should make it coherent, not refuse. Runs the
+    clean-path convergence (import + cleanup) so the legacy-row boundary clears.
+    Divergent-duplicate conflicts are deliberately NOT auto-resolved here (that
+    would discard superseded source data, even if quarantined) — they are
+    surfaced with the explicit ``sync migrate --resolve-conflicts keep-journal``
+    recovery, and the coherence gate that follows still refuses until they are
+    resolved. Idempotent and lossless: a no-op on an already-converged runtime.
+    Any runtime-open failure is swallowed so the coherence gate handles it.
+    """
+    from specify_cli.paths import get_runtime_root
+    from specify_cli.sync.migrate_journal import (
+        AUDIT_DB_NAME,
+        MigrationAudit,
+        converge_legacy_runtime,
+    )
+
+    spec_kitty_dir = get_runtime_root().base
+    try:
+        runtime = _open_event_sync_runtime()
+    except Exception:  # runtime unavailable — let the coherence gate report it
+        return
+    audit = MigrationAudit(spec_kitty_dir / AUDIT_DB_NAME)
+    try:
+        converge = converge_legacy_runtime(
+            spec_kitty_dir,
+            journal=runtime.journal,
+            audit=audit,
+            resolved_target=runtime.target,
+            resolve_conflicts=False,
+            cleanup=True,
+        )
+    finally:
+        with contextlib.suppress(Exception):
+            audit.close()
+        runtime.close()
+
+    deleted = converge.cleanup.total_deleted if converge.cleanup is not None else 0
+    if deleted:
+        console.print(
+            f"[green]✓[/green] Converged {deleted} legacy queue row(s) into the event journal."
+        )
+    if converge.blocked_conflicts:
+        console.print(
+            f"[yellow]{converge.blocked_conflicts} divergent-duplicate conflict(s) remain. "
+            "Run `spec-kitty sync migrate --resolve-conflicts keep-journal` to resolve them, "
+            "then re-run opt-in.[/yellow]"
+        )
+
+
 @app.command(name="opt-in")
 def opt_in(
     checkout_only: bool = typer.Option(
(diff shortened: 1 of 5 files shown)


LynnColeArt:
@stijn-dejongh heads up (async, not blocking) — per your DM I'm taking #2665.

This PR is the **clean-path** half only: `sync migrate` was importing into the journal read-only and never deleting the migrated rows from the source queues, so the legacy-row boundary never converged and `sync now`/`opt-in` refused forever. This adds the missing *gated* cleanup (delete confirmed-migrated ids by id, no-op when blocked). It's additive and does **not** touch the collision / journal-overwrite semantics — I'm deliberately leaving the divergent-duplicate resolution for your ports-and-adapters rework, per your "the collisions will be workable" note, rather than patching an escape hatch into the path you're strangling. Flagging for awareness given the tangential overlap.

LynnColeArt:
Scope note: this PR grew past the original clean-path fix to close the whole #2665 migrate deadlock and set up the rollout —
- clean-path cleanup (converge boundary after a clean migrate)
- `--resolve-conflicts keep-journal` (quarantine + converge past divergent duplicates; journal never overwritten)
- `converge_legacy_runtime()` — the idempotent, lossless one-shot engine both `sync migrate` and the planned auto-migration reuse

Auto-migration design (detect un-migrated `.spec-kitty` → converge automatically) is #2688, pending coordination with the ports-and-adapters rework (the pre-scoping write-path refill) + a fleet-wide auto-resolve policy call.

stijn-dejongh:
Claiming this PR for a maintainer landing pass: rebase onto upstream/main, adjudicate the red checks (arch-adversarial shards 2 & 3, quality-gate), red-first verify the bugfix, fold fixes as needed. Evidence to follow.

stijn-dejongh:

**Verdict: LAND — merge-ready on green.** Rebased onto current `upstream/main` (was 13 behind; clean rebase, no conflicts). The core fix is sound and its central data-safety claim (delivery reads solely from the journal, so a journal-durable row stays deliverable after its source copy is removed) was independently verified in review. Four folds pushed; CI re-running on the folded tip.

### Red-check adjudication (original 3 failures)
| Check | Classification | Action |
|---|---|---|
| `arch-adversarial` shard 2 (`test_pytest_marker_convention`, `test_gate_coverage::test_no_new_orphan_surfaces`) | **PR defect** | Fold 1 |
| `arch-adversarial` shard 3 (`test_golden_count_ban`, gate-coverage orphan nodes) | **PR defect** | Fold 1 |
| `quality-gate` | downstream of arch-adversarial (mypy is **advisory**, does not block) | resolves with arch-adversarial |

All 3 `arch-adversarial` shards then **passed in CI** on the folded tip.

### Folds (each `landing fold: ...` on this branch)
1. **Route the new test into a CI gate + convert a golden-count site** — `tests/…/test_sync_opt_in_converge.py` carried no module-level `pytestmark`, so it was selected by **zero** CI gates (gate-coverage orphan) and tripped the marker-convention gate. Added `pytestmark = pytest.mark.fast` (sibling convention). Separately, `tests/sync/test_migrate_journal.py` added one convert-classified golden-count site (`assert len(resolution.skipped) == 1`) pushing `tests/sync` to 51 over the frozen ceiling of 50 — converted it to the stronger content assertion `resolution.skipped == ["dup"]` (off the convert ledger, and a better test).

2. **(MAJOR, data-safety) Durable quarantine before the source delete** — in `resolve_conflicts_keep_journal` (the `--resolve-conflicts keep-journal` operator path), the source-row delete commits `queue.db` **immediately**, but the quarantine archive was only committed to the audit DB **after the loop**. A crash between the two left the source row deleted and the archive rolled back — losing the divergent payload the docstring promises is recoverable (FR-018/C-005). Fixed by write-ahead: `audit.commit()` the archive *before* the destructive delete, mirroring the import path's own provenance-first ordering. Regression test injects a failure at the delete boundary and asserts the archive is durable to a **fresh** audit connection (verified red on pre-fix, green after).

3. **(robustness) Swallow converge failures on `opt-in`** — `_auto_converge_legacy_on_enable`'s docstring claimed "any runtime-open failure is swallowed," but only the runtime *open* was guarded; an error inside `converge_legacy_runtime` would abort `opt-in` with a traceback before the coherence gate. Since this runs for **every** user on `opt-in`, broadened the guard to swallow converge failures too and defer to the coherence gate. New unit test covers the branch (red on pre-fix, green after).

### Adversarial squad (profile-loaded, read-only)
- **architect-alphonso** (design/contract): **No BLOCKER, LAND.** Verified the deletion is provably data-safe — delivery is journal-only, emit dual-writes journal+queue, and cleanup deletes only journal-confirmed ids *by id*; the import-snapshot→cleanup race is closed because coalescing reassigns `event_id` on any content change. Confirmed idempotency of the double-import and SC-011 exit-code preservation.
- **paula-patterns** (SSOT/boundary): **LAND after the MAJOR.** Confirmed the destructive path correctly routes through the owning `OfflineQueue` public method (no raw SQL leak from the migration module) and payload-reading reuses the canonical seam. Raised the durability MAJOR (folded above).

MINORs/NOTES (error-isolation parity in the resolve path, `IN(...)` chunking for very large migrations, pinning the coalesce→`event_id` invariant, the `opt-in` mutation-before-refuse note, `_row_count` cache staleness, `__all__` coherence) are filed together as **#2713**, parented off #2665.

### Local evidence (rebased + folded tip)
- `tests/sync/test_migrate_journal.py` — **31 passed**; `tests/…/test_sync_opt_in_converge.py` — **4 passed**.
- Full `tests/architectural/` — **1008 passed, 4 skipped**; targeted golden-count/marker/gate-coverage re-run — **34 passed**.
- `ruff` clean and `mypy --strict` clean on all changed source files.
- Red-first witnesses: 3 (guard-mutation on `cleanup_blocked`; quarantine-durability; converge-swallow) — each confirmed to go red against the pre-fix product.

### Pre-existing (not this PR)
- `mypy --strict` reports 5 `[redundant-cast]` errors in files this PR doesn't touch (`mission_type_profiles.py`, `status/emit.py`, a migration); they reproduce identically on a clean `upstream/main`. The mypy report is **advisory** and does not block `quality-gate`. Filed as **#2714**.

### Note to the author (@LynnColeArt)
The PR body describes only the clean-path cleanup, but the branch also ships `--resolve-conflicts keep-journal`, `converge_legacy_runtime`, and auto-convergence on `sync opt-in` (the fold-2 MAJOR lives in that resolution path). Worth updating the body so a reviewer/operator sees the full surface. Also — heads-up for a future body/help note: `opt-in` now converges + deletes source rows *before* the coherence gate, so a checkout that refuses for a non-legacy reason will still have mutated its queues first (data-safe and deliberate — it's the deadlock fix — but a behavioral surprise; captured in #2713).

*The operator merges — I don't. Hand-off deliverable: green CI, PR un-drafted, evidence above. No landing-order constraint.*

Issue #2734, opened 2026-07-16
sync now cannot drain a byte-oversized backlog, and mislabels the 413 as an auth error


LynnColeArt:
## Summary

`spec-kitty sync now` cannot drain a backlog whose events exceed the server's per-batch size limit, and it reports the failure as an **auth error**, which sends operators chasing OAuth instead of the real cause.

This is the root cause behind the recurring "data will not sync" dogfood reports. It is **not** an OAuth problem (see #2731, which I ruled out: a clean single refresh rotates and persists correctly).

## Root cause

The server rejects an over-cap batch with **HTTP 413** (a decompressed-size limit, advertised to clients via the sync ingress limits contract).

The CLI only chunks a drain by **event count** (`_EVENT_SYNC_DISPATCH_BATCH_LIMIT = 1000`), never by **byte size**. A backlog under 1000 events but over the server's size limit goes out as one oversized batch:

- `dispatcher.dispatch()` is a thin select -> post -> record with no split-and-retry.
- The receiver maps the 413 to a batch-wide `transient` carrying `_BATCH_OVERSIZED_ERROR = "batch payload too large; retry with a smaller batch"`, and `batch.py` states "The CLI will retry with a smaller batch". Nothing does.
- `_run_dispatch_batches` sees `selected < 1000` and breaks. Every subsequent drain re-selects the same oversized set, 413s again, and makes zero progress. Permanent deadlock.

## Compounding: the failure is mislabeled as auth

`_enforce_sync_now_exit_from_dispatch` maps "selected > 0, nothing progressed, all transient" to `not authenticated: no valid access token. Run spec-kitty auth login.` A 413 (and any 5xx) is reported as a logged-out session. The all-rejected case (content 400s, no progress) also falls through to the same unauthenticated recovery path. This single mislabel is why the whole "data will not sync" saga read as an auth problem.

## Evidence (reproduced locally)

A backlog of 647 events failed as one oversized batch, all `transient`, message "not authenticated". Auth was verified good in isolation the whole time (`auth status` OK, `get_access_token()` returns a valid token).

Driving the exact same real drain path with the per-batch limit lowered to 25 events: **400 events delivered, 0 transient**, confirmed server-side. The only variable changed was batch size. The remaining 247 turned out to be genuine content rejections (see follow-ups below), not a delivery problem.

## Proposed fix (draft PR, not merged)

Smallest change that honors the contract the code already advertises:

[...]

Issue #2731, opened 2026-07-16
CLI→TeamSpace drain blocked: OAuth refresh-rotation revokes the session → unauthenticated (root behind auth-gate + legacy-queue refill)


LynnColeArt:
Found dogfooding the CLI→TeamSpace back-prop drain (local rig + local SaaS), after the #2665 migrate-deadlock fix (#2682) cleared the sync-convergence layer. What looked like three separate delivery failures traced to **one root**.

## Symptoms
1. `sync now` gates on `auth_present` + `private_teamspace` even right after a successful `auth login --force`.
2. The legacy `queue.db` keeps "refilling" (events accumulate over the session).
3. A handful of events rejected at ingress for empty `build_id`/`project_uuid`.

## Root cause — symptoms 1 and 2 are the same root
The OAuth **refresh** grant fails and takes the session down with it:
1. The SaaS uses token-family rotation with replay detection (`apps/cli_auth/lineage/service.py`): a **stale-generation** refresh token → suspicious replay → **family revoked → 401 `invalid_grant`** (`apps/cli_auth/views_token.py::_handle_refresh_token_grant`).
2. The CLI's refresh transaction (`src/specify_cli/auth/refresh_transaction.py`) responds to a rejected refresh by **deleting the session** (FR-005): `~/.spec-kitty/auth/session.json` is removed, leaving only `refresh.lock`.
3. So one bad refresh → session gone → **every subsequent CLI process is unauthenticated** → the delivery auth gate fails (symptom 1) AND `default_queue_db_path` routes events to the legacy `queue.db` unauthenticated fallback (symptom 2, the "refill").

Confirmed: SaaS logs show `POST /oauth/token 401` interleaved with `200` (fresh logins); client state is `auth status` = not authenticated, `session.json` absent, state `logged_out_on_connected_teamspace`.

**Ruled out:** token TTLs are normal (access 1h / refresh 90d — `apps/cli_auth/tokens.py`); no zombie sync daemon (no live process).

## To pin before fixing
Why the persisted refresh-token generation goes stale despite the single-flight design in `refresh_transaction.py` — is the rotated token failing to persist on the refresh path (`storage.write` after rotation), or is there a reconcile edge that re-presents an old generation? Needs a clean repro (fresh login → force a refresh → observe the exact `TokenReuse*`/generation state) before a targeted, tested fix. Likely CLI-side (refresh-token persistence); possible SaaS rotation-strictness angle.

## Related / separate
[...]

Issue #829, opened 2026-04-27
Warn and offer login when Teamspace is connected but session is logged out


robertDouglass:
## Problem

If a user starts using Spec Kitty and has previously connected to Teamspace / SaaS, but is not currently logged in, the CLI should not fail silently or leave them to infer the auth state.

## Desired behavior

When Spec Kitty detects that a Teamspace / SaaS connection exists but the user is not logged in, show a clear warning and offer to start the login flow immediately.

Suggested prompt shape:

```text
You have previously connected this workspace to Teamspace, but you are not currently logged in.
Log in now to sync Teamspace data? [Y/n]
```

If the user accepts, start the existing login flow. If they decline, continue in the appropriate local/offline mode and make it clear that Teamspace sync will not run until they log in.

## Acceptance criteria

- Detects the state where Teamspace / SaaS has been configured previously, but no valid login session is active.
- Displays a clear warning before SaaS-backed tracker or sync behavior is needed.
- Offers an interactive login option from that warning.
- Handles non-interactive environments without hanging, using a clear message and exit/continue behavior appropriate to the command.
- Does not imply that tracker itself has a rollout system or require rollout gating inside `spec-kitty-tracker`.

## Notes

This is about improving first-run / return-user UX for users who already connected Teamspace but later lost or expired their login session.

robertDouglass:
Validated against current CLI code in a fresh monorepo workspace. The sync/auth routes still print guidance like "Run spec-kitty auth login" and sync now exits with the unauthenticated message, but I did not find an interactive "log in now" prompt or previous Teamspace connection detection/recovery flow. This remains relevant for Teamspace MVP onboarding polish.

robertDouglass:
Closed by program post-merge audit. Delivered by [PR #1030](https://github.com/Priivacy-ai/spec-kitty/pull/1030) (merged commit `d5d25756`). The PR body used markdown-link `[Closes #829](URL)` form which GitHub's auto-close parser does not honor; closing manually now.

Issue #2720, opened 2026-07-16
Epic: Mission-state discovery & diagnostic-command output fidelity


stijn-dejongh:
## Why / for whom / effect

Operators and orchestrating agents rely on read-only mission-state commands — `check-prerequisites`, `doctor mission-state` / `doctor coordination`, and `retrospect summary` — to trust mission state before acting. Today these commands emit **false inventories** (present planning artifacts omitted), **false schema findings** (`UNKNOWN_SHAPE` on writer-canonical keys), a **nonexistent `research_dir`**, and **false missing-retrospective counts** (support directories scanned as missions, real records omitted). As a result, readiness checks and aggregate cross-mission learning are untrustworthy.

**Effect (target state):**
- One canonical mission-instance discovery shared across the diagnostic surface (discover under `kitty-specs/*`, exclude implementation/support dirs).
- Writer-schema-sourced validation (mission doctors consume the canonical metadata schema, no drift).
- A shared `--mission <handle>` selector so a single workflow can be diagnosed without repository-wide noise.
- Preflight/preparation that resolves the **required repo set from the selected downstream mission-review gates**, rather than discovering missing sibling repos mid-gate.

## Children
- #2692 — check-prerequisites omits existing planning artifacts and returns a nonexistent `research_dir`
- #2696 — mission doctors drift from generated metadata schema and cannot scope diagnostics to one mission
- #2717 — retrospect summary scans `.kittify` support modules as missions and omits the real record
- #2704 — monorepo preparation does not include repositories required by downstream mission-review gates

Surfaced as a cohesive cohort during the #2658 dogfood run (robertDouglass). Children linked as native sub-issues.

robertDouglass:
Triage validation (2026-08-02): KEEP OPEN. Current issue history and source review show no completed successor or supersession decision for this slice. Retain it as actionable; add a focused repro/acceptance fixture and explicit dependency/owner before implementation, then link the delivery PR.

stijn-dejongh:
**#2704 reparented out of this epic → #1193.**

#2704 (monorepo preparation does not include repos required by downstream mission-review gates) has been removed as a child here. It is out-of-class for this epic's discovery / diagnostic-output-fidelity theme — it is cross-repo workspace-preparation infra with no existing seam here — and now lives under **#1193 (3.3.0 Epic: Planning workspace and epic mission decomposition)** alongside sibling #1001.

[...]

PR #2735, opened 2026-07-16
fix(sync): drain oversized + poison batches instead of deadlocking (#2734)

LynnColeArt:
Draft for feedback (@stijn / @kent). Fixes the root cause in #2734. Not for merge until you have looked at it.

## What this fixes

`sync now` could not drain a backlog whose events exceeded the server's per-batch size limit, and it reported the failure as an auth error. That mislabel is what made the whole "data will not sync" saga read as an OAuth problem. OAuth (#2731) is sound; I ruled it out.

Two related drain defects, in two commits:

### 1. Byte-oversized batches deadlocked

The CLI chunks a drain only by event count (`_EVENT_SYNC_DISPATCH_BATCH_LIMIT = 1000`), never by byte size. A backlog under 1000 events but over the server's size cap goes out as one oversized batch, gets a 413, and every event flips to `transient`. `_run_dispatch_batches` breaks on `selected < 1000` and re-selects the same oversized set on the next drain. Permanent deadlock. The receiver already returns `"batch payload too large; retry with a smaller batch"` and batch.py promises the CLI will retry smaller, but nothing did.

**Change:** adaptive halving in `_run_dispatch_batches` (on 413 with nothing delivered and > 1 event, halve the limit and retry; a single oversized event is terminal-failed by the receiver so `limit == 1` cannot loop forever), plus an honest message classified by actual failure status (413 -> too large, 401/403 -> not authenticated, else -> generic transient) instead of always blaming auth.

### 2. A poison chunk head-of-line blocked the rest

Once batches are small, the loop broke on the first chunk that made no terminal progress -- a content-rejected chunk halted the whole pass, so deliverable events behind it never drained.

**Change:** `dispatch()`/`_select_undelivered` gain an optional in-memory `exclude` skip set (applied to the journal universe before the ledger query, so `limit` still fills). The loop accumulates events that made no terminal-success this pass and skips them for the rest of the pass, advancing to deliverable events behind them. It never mutates the ledger, so the non-terminal re-selection contract (T028) is preserved; termination is on a strict progress invariant.

## Verified locally

A 647-event backlog that deadlocked as one oversized batch (all transient, "not authenticated") drains cleanly once byte-safe: 400 delivered, 0 transient, confirmed server-side. The remaining rejections are genuine content rejections (see #2736), not a delivery problem.

## Tests

- `test_run_dispatch_batches_halves_on_oversized_then_drains`: an oversized batch is halved 8 -> 4 -> 2 until it fits, then drains all events.
- `test_run_dispatch_batches_skips_rejected_and_drains_events_behind`: poison events at the head do not stop deliverable events behind them.
- `test_transient_block_message_distinguishes_cause`: 413 / 401 / 5xx each get the right message.
- Updated `test_now_logged_out_nonempty_queue_reports_unauthenticated_failures` to a realistic 401 summary so #829's exit-1 / "auth login" contract holds under the more precise classifier.

Full delivery + sync-routes suites green (187); ruff and mypy clean.

## Two things I deliberately left out (your call)

- **All-rejected exit contract**: an all-rejected drain (content 400s, no progress) still routes through `_handle_sync_now_unauthenticated` and prints the auth message. Fixing that cleanly touches the #829 exit-code contract (exit 1 vs exit 4), your lane, so I did not rework it here.
- **Batch-400 isolation (#2736)**: full recovery of innocent events trapped in a strict-whole-batch 400 needs bisecting the batch to isolate the invalid event. That mirrors the oversized halving here but touches the server's whole-batch-400 contract, so I filed it separately for your call rather than build it blind.

Related: #2734 (root cause), #2736 (batch-400 isolation follow-up), #2731 (OAuth, ruled out).

### DIFF
--- a/src/specify_cli/cli/commands/sync.py
+++ b/src/specify_cli/cli/commands/sync.py
@@ -719,11 +719,6 @@ def _combine_dispatch_summaries(
     )
 
 
-def _batch_left_selection_set(summary: DispatchSummary) -> bool:
-    terminal = summary.delivered + summary.duplicate + summary.terminal_failed
-    return summary.selected > terminal
-
-
 def _batch_is_oversized(summary: DispatchSummary) -> bool:
     """Whether a batch was rejected wholesale for exceeding the server size cap.
 
@@ -766,6 +761,7 @@ def _run_dispatch_batches(
 
     combined = DispatchSummary.empty()
     limit = _EVENT_SYNC_DISPATCH_BATCH_LIMIT
+    skip: set[str] = set()
     while True:
         batch = dispatch(
             journal=runtime.journal,
@@ -773,6 +769,7 @@ def _run_dispatch_batches(
             receiver=receiver,
             target=delivery_target,
             limit=limit,
+            exclude=frozenset(skip),
         )
         # Honor the documented "retry with a smaller batch" contract: a
         # byte-oversized batch (HTTP 413, nothing delivered) is halved and
@@ -784,9 +781,15 @@ def _run_dispatch_batches(
             limit = max(1, limit // 2)
             continue
         combined = _combine_dispatch_summaries(combined, batch)
-        if batch.selected < limit:
-            break
-        if _batch_left_selection_set(batch):
+        # Advance past events that made no terminal-success this pass (content
+        # rejections, persistent transient). Skipping them for the REST OF THIS
+        # PASS lets deliverable events behind them drain instead of a poison
+        # batch halting the loop; the ledger keeps them selectable for the next
+        # `sync now`, so the non-terminal retry contract is preserved.
+        before = len(skip)
+        skip.update(failure.event_id for failure in batch.failures)
+        advanced = (batch.delivered + batch.duplicate) > 0 or len(skip) > before
+        if batch.selected == 0 or not advanced:
             break
     return combined
 
--- a/src/specify_cli/cli/commands/sync.py
+++ b/src/specify_cli/cli/commands/sync.py
@@ -85,7 +85,7 @@ _TRANSIENT_SYNC_NOW_MESSAGE = (
 # HTTP 413 is how the SaaS sync ingress (Fly proxy + edge) rejects an
 # over-cap batch; see apps/sync/limits.py (512 KiB decompressed ceiling).
 _HTTP_PAYLOAD_TOO_LARGE = 413
-_OVERSIZED_ERROR_MARKER = "too large"
+_OVERSIZED_ERROR_MARKER = "retry with a smaller batch"
 _HTTP_AUTH_STATUSES = frozenset({401, 403})
 _WARNING_HEADER_STYLE = "bold yellow"
 _ABSENT_VALUE = "<absent>"
@@ -310,7 +310,8 @@ def _enforce_sync_now_exit_from_dispatch(
       "queue non-empty but all-zero result". This is routed through the
       teamspace-aware recovery so the unauthenticated UX (interactive login,
       structured exit 4, legacy exit 1) is preserved regardless of ``--strict``.
-    * Partial progress with a hard terminal failure → exit 1 under ``--strict``.
+    * Partial progress with any rejected, transient, or terminal failure → exit
+      1 under ``--strict``.
 
     A ``None`` summary means dispatch infrastructure was unavailable. Under
     ``--strict`` that is a failure only when retained or legacy work exists.
@@ -342,7 +343,8 @@ def _enforce_sync_now_exit_from_dispatch(
     if work_present and progressed == 0:
         _handle_sync_now_unauthenticated(strict)
         return
-    if strict and summary is not None and summary.terminal_failed > 0:
+    errors = summary.rejected + summary.transient + summary.terminal_failed
+    if strict and errors > 0:
         raise typer.Exit(1)
 
 
@@ -716,6 +718,10 @@ def _combine_dispatch_summaries(
         transient=left.transient + right.transient,
         terminal_failed=left.terminal_failed + right.terminal_failed,
         failures=(*left.failures, *right.failures),
+        retryable_event_ids=(
+            *left.retryable_event_ids,
+            *right.retryable_event_ids,
+        ),
     )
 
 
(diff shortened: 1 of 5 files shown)


stijn:
@LynnColeArt Your mentions are probably supposed to be for @stijn-dejongh 
Have a nice day.

stijn-dejongh:
**Critical-fix exception to the 3.2.x merge-hold.** Per operator ruling this sync drain/poison-batch fix carries critical-fix status (a wedged, non-draining sync backlog with a misreported error). It does not intersect the frozen doctrine surface, and it is **intentionally NOT deferred** — it should land ahead of the freeze. Not labeled `pr:deferred`. — 3.2.x merge-hold triage

stijn-dejongh:
## Pre-merge adversarial squad — verdict: **LAND** ✅

Rebased onto current `main` (clean, no conflicts) and ran three profile-loaded lenses over the aggregate diff. Nice work — the deadlock fix is correct and the tests genuinely pin it.

| Lens | Verdict | Summary |
|------|---------|---------|
| **debugger-debbie** (structural root-cause) | **sound** | Deadlock **closed by construction** — verified against the *real* receiver mapper (`receivers.py:404`, single-oversized → `TERMINAL_FAILED` is the load-bearing invariant). Two independent monotone termination measures; every exit path enumerated; residual inputs (single>cap, 5xx, partial-413, empty) all safe. |
| **reviewer-renata** (correctness + tests) | **LAND** | All three claims hold; termination doubly-guaranteed. The three named tests are meaningful, not tautological (removing halving/skip/classifier each fails a real assertion). Live-dispatcher tests close the limit==1 and cross-pass re-selection gaps. #829 exit-1/auth contract preserved. |
| **paula-patterns** (contract/boundary) | **LAND** | `exclude` is a clean defaulted opt-in — every existing caller preserved, ledger never mutated, T028 non-terminal re-selection contract held; halving terminates. |

### Fixed here (one squad finding folded in)
- **Throughput cliff — the shrunk `limit` never recovered.** renata + debbie independently flagged that once a single over-cap event forced `limit → 1` and was parked, the whole *healthy* tail drained one-event-per-POST for the rest of the pass. Added a **multiplicative grow-back** (mirrors the halving, capped at the count default) in `_run_dispatch_batches`, plus a regression test; the existing singleton-terminal test now pins the tail draining in one grown batch. Termination is unchanged (the fix doesn't touch the strict-progress invariant).

### Non-blocking follow-ups (deferred — your call, not merge blockers)
1. **Byte-sizing ownership split.** Foreground `sync now` sizes reactively (count-halving on 413) while the daemon/offline path (`sync/batch.py`) already sizes *proactively* by decompressed-byte budget with its own oversized handling. Same bug-class solved in two places — worth converging on one owner. Pre-dates this PR.
2. **Error-marker string copy.** `sync.py:88 _OVERSIZED_ERROR_MARKER` duplicates the receiver's module-private `_BATCH_OVERSIZED_ERROR` — a silent-drift seam (mitigated: it's only the fallback behind the stable `http_status == 413` key). Export + import the constant.
3. **Convention-enforced protocol.** The halving+exclude convergence lives in the caller (`_run_dispatch_batches`); a future looping `dispatch()` caller that omits it re-introduces the deadlock. Only one loop caller today — worth a note or a hoist to a shared surface.

### CI note
The rebase is clean and the PR's own delivery/sync suites pass on current `main` (73). The only `integration-tests-agent` reds (`test_errors_when_template_not_found` → template error now `'missing'`; `test_sphinx_generation_end_to_end`) are **pre-existing on `main` and unrelated** to this diff — not in its files; the first looks like the step-authority/SSOT template-resolution change.

*(Branch force-pushed with the rebase + the grow-back commit.)*

stijn-dejongh:
**CI clarification:** all four red checks on this run are **pre-existing on `main` and unrelated to this sync fix**:
- `integration-tests-agent` — `test_errors_when_template_not_found` (template error now `'missing'`) + `test_sphinx_generation_end_to_end`; not in this diff.
- `arch-adversarial (arch_shard_1/3)` — main-tip `ddac71ebc` added the `pr:deferred`/`pr:skip-ci` full-CI-block labels to job `if:` gates; the arch gate `test_ci_quality_path_filters.py` still asserts those jobs must be exactly `if: always()`, so it's red on main. The gate tests need updating (tracked separately); intentional feature, not a mask to revert.
- `quality-gate` — aggregate of the above.

The sync diff itself: delivery/sync suites green (73), ruff + mypy clean. Nothing here to fix.

stijn-dejongh:
Claiming this PR for a landing pass (maintainer runbook `docs/guides/pr-landing.md`): rebase onto current `upstream/main`, adjudicate the red checks (arch-adversarial shards, integration-tests-agent), red-first verify the drain/poison-batch fix, and fold any remediation. Carried as a **critical-fix exception** to the 3.2.x merge-hold (fixes the wedged, non-draining sync backlog #2734). Evidence to follow; the operator merges. — landing pass (maintainer)

stijn-dejongh:
## Landing remediation summary — #2735

**Verdict: LAND** (merge-ready; one documented pre-existing red, tracked separately).

### Rebase
Rebased clean (no conflicts) onto current `upstream/main` (`17a65861f`, which now includes #2753). Pushed the rebased tip `1657e98f8` — 5 commits, authorship preserved. Lease-checked: no unseen commits on the branch before push.

### Folds
**None.** The PR carries no defects — its own tests pass and every red was external (see classification).

### Red-first verification (bugfix witness)
Swapped the two product files (`delivery/dispatcher.py`, `cli/commands/sync.py`) back to `upstream/main` and re-ran the PR's tests:
```
13 failed, 60 passed   # pre-fix — tests witness the bug
```
Failures span the real fix surface: `test_multi_batch_drain_continues_after_singleton_terminal_failure`, `test_run_dispatch_batches_halves_on_oversized_then_drains`, `test_run_dispatch_batches_grows_limit_back_after_oversized_park`, `test_strict_exits_1_on_partial_retryable_errors`, `test_recorded_no_progress_failure_does_not_invoke_auth_recovery`, and more. Restoring the fix → **73 passed**. The tests genuinely capture the deadlock/poison-batch bug.

- PR's own tests: **73 passed** (`tests/delivery/test_dispatcher.py`, `tests/agent/cli/commands/test_sync.py`, `tests/cli/commands/test_sync_routes.py`).
- Full `tests/architectural/` suite: **green** (exit 0) — the always-on-gate arch checks pass now that #2753 is in the base.
- `ruff check` + `mypy` on both changed source files: **clean**.

### Red classification (maintainer runbook step 4)
| CI red | Bin | Disposition |
|---|---|---|
| `arch-adversarial` (shards 1 & 3) | Pre-existing main breakage | The #2752 CI-guard-vs-always-on-gate mismatch, **already fixed on main by #2753**. Cleared by the rebase — passes locally. |
| `integration-tests-agent` → `test_errors_when_template_not_found` | Pre-existing main breakage | `setup-plan` leaks the raw `FileNotFoundError` text instead of translating to "Plan template not found". **Reproduces on `upstream/main` with none of this PR's changes.** Filed **#2754** (under #2720). Not this PR's defect; not fixed here (bin-3 discipline). |
| `quality-gate` | Aggregator | Clears when the above clear. |

### State
Fresh CI is running on the rebased tip. Expected outcome: green **except** `integration-tests-agent`, which stays red only on the pre-existing **#2754** until that main regression is fixed separately. This PR does not touch that path.

**Carried as a critical-fix exception to the 3.2.x merge-hold** (fixes the wedged, non-draining sync backlog #2734). The operator merges. — landing pass (maintainer)

PR #2423, opened 2026-07-06
fix(#2412): gitignore the shared skill projection root + per-machine install ledger (asks 1+2)

rayjohnson:
Implements asks 1+2 of #2412, per the triage's suggested resolution order ("asks 1+2 can land using the established backfill-migration shape without waiting on the symlink-vs-copy design call"). Ask 3 (delivery-mode design) stays open on the issue.

## Scope — exactly the gap the triage confirmed

- The **uncovered surface is `.agents/skills/`** (shared projection root for codex/vibe/pi/letta): `.claude/` etc. are gitignored wholesale by init's `protect_all_agents()`, but bare `.agents/` is absent from `AGENT_DIRECTORIES` and no registry entry, init path, or migration covered it — so the installer's absolute `/Users/<name>/...` symlinks were committable, and the porcelain-derived upgrade auto-commit (soon per-worktree via #2387) would land them automatically.
- `.kittify/skills-manifest.json` (per-machine `installed_at`/`content_hash`/`delivery_mode` ledger) had the same gap — this repo itself tracked it with **167 symlink-mode entries**.

## The fix — third instance of the #2369/#2370/#2384 IGNORED-surface pattern

**Ask 2 — registry:** two new `StateSurface` entries in `state/contract.py`:
- `shared_skills_projection` — `.agents/skills/<skill_name>/`, `DERIVED`/`IGNORED` (regenerable projection; collapses to the `.agents/skills/` gitignore entry)
- `skills_install_manifest` — `.kittify/skills-manifest.json`, `LOCAL_RUNTIME`/`IGNORED`

Registration alone buys both halves of ask 2: fresh `spec-kitty init` emits the entries via `get_runtime_gitignore_entries()` → `protect_all_agents()`, and `spec-kitty doctor`'s existing present-but-not-gitignored warning (`state/doctor.py::_check_gitignore`) now structurally covers the surfaces.

**Ask 1 — backfill for existing projects:** `m_3_2_5_agents_skills_gitignore_backfill`, modeled line-for-line on the `3.2.4_derived_mission_views` precedent (hardcoded entries so the migration is frozen; idempotent; treats hand-added equivalents — `.agents/` wholesale, no-trailing-slash forms — as present; `runs_on_worktrees=True`, and with #2387 the per-worktree gitignore churn gets committed on each worktree's own branch).

[...]

PR #2818, opened 2026-07-19
fix(#2412): deliver projected skills as copies, never absolute symlinks (ask 3)

### LINKED ISSUE #2412: install/upgrade projects symlink-delivered skills into repo roots with no gitignore registration — absolute-path symlink commit hazard (+ per-machine skills-manifest.json tracked)
Sibling of the #2384 / #2150 "IGNORED-surface completeness" class (parent #1619). **Not** the #2392 commit-propagation class — but it interacts with it (see below).

## The defect

The skills installer projects global canonical skills into repo-rooted agent surfaces **as absolute symlinks**, and neither the installer nor `init`/`upgrade` registers those surfaces for gitignore:

- `src/specify_cli/skills/installer.py` → `_project_skill_file()` prefers `dest.symlink_to(source_file)` where `source_file` is under the **user-global canonical root** — so the projected blob is `/Users/<name>/...` (or the platform equivalent), by construction machine-local and non-portable.
- Destinations are repo-rooted: `.claude/skills/<skill>/…` and `.agents/skills/<skill>/…` (via `get_primary_project_skill_root`).
- Nothing writes a `.gitignore` entry for them. `init` only **prints a notice** suggesting the operator consider gitignoring agent folders (`cli/commands/init.py` — "Agent folder security notice"); nothing enforces or backfills it.
- The state-roots registry (`src/specify_cli/state/contract.py`) has no `StateSurface` for the skill projection roots, so `spec-kitty doctor`'s existing "present but not gitignored" coverage check (`state/doctor.py::_check_gitignore`) structurally cannot warn about them.

The spec-kitty repo itself is only protected by hand-written wholesale ignores (`.gitignore`: `.claude/`, `.agents`); consumer projects initialized by the CLI get no such protection.

## Why this is now *more* urgent — interaction with #2392 / PR #2387

`spec-kitty upgrade`'s auto-commit derives its commit-set from the real porcelain diff (that's the fix direction of #2105), and PR #2387 extends the same auto-commit to every worktree. An **un-ignored** symlink-delivered skill therefore doesn't need a careless `git add -A` to become a commit — **the upgrade flow itself will auto-commit the absolute-path symlinks**. Conversely the fix composes perfectly: a gitignored path never appears in the porcelain diff, so it is automatically excluded from the upgrade commit-set. Gitignore registration is the missing half that makes the #2392 invariant safe for projection surfaces.

## Three concrete asks

1. **Install/upgrade should gitignore what it projects.** Any repo-rooted surface delivered as `delivery_mode: symlink` is machine-local by definition and should get a `.gitignore` entry the moment it is written (plus a backfill migration for existing projects, same shape as the #2369/#2384 backfills).
2. **Register the projection surfaces in the state-roots registry** (`state/contract.py`, `authority=LOCAL_RUNTIME` or `DERIVED`, `git_class=IGNORED`) so the existing doctor "present but not gitignored" warning covers them.
3. **Question symlink delivery into the repo root at all.** A symlink whose blob is `/Users/<name>/...` is a commit hazard by construction; content delivery (copy) or an ignored-by-default directory would be structurally safe. (Design decision — flagged, not prescribed.)

## Bonus data point: `.kittify/skills-manifest.json` is tracked

In the spec-kitty repo itself, `.kittify/skills-manifest.json` is **committed** and records **167 `delivery_mode: symlink` entries** with per-machine `installed_at`/`content_hash` state. That file is per-machine install state (same class as the asks above) and churns on every machine's install. `.kittify/command-skills-manifest.json` is also tracked; it uses relative paths so it is less harmful, but it is the same per-machine-state class and worth a deliberate tracked-vs-ignored decision.

## Suggested triage

Given the #2392 precedent (consolidate a bug cluster into one canonical seam before point-fixing), this probably wants a maintainer decision on ask 3 first, since it determines the shape of asks 1–2. Happy to pick up the implementation once the direction is set — the backfill-migration shape is already established by #2369/#2370 and the in-review #2388.

rayjohnson:
## The problem

Mount a spec-kitty project into a dev-container, and every skill under `.agents/skills/` and `.claude/skills/` silently vanishes: each one is an absolute symlink to `/Users/<you>/...` — a path that doesn't exist inside the container. The same thing happens to an agent harness sandboxed to the repo root (it refuses to resolve a link whose target escapes the project), and to every repo on a machine at once if the global canonical root is ever moved or reinstalled. There's no error anywhere — the agent just doesn't see its skills.

#2423 (asks 1+2 of #2412) made these surfaces safe to **commit**. This PR closes ask 3: it makes them safe to **use**.

## The fix in plain terms

Projected skill files are now always real copies, never symlinks. The copy path already existed (it was the Windows fallback) — this promotes it to the only mode and deletes the symlink machinery.

## Technical details

- `skills/installer.py::_project_skill_file` always `shutil.copy2`s. A destination whose content hash already matches the source is left untouched (idempotent re-runs, no churn); divergent content keeps the existing archive-to-backup behavior. A legacy projection symlink found at the destination is unlinked and replaced with a copy.
- **No migration needed**: the installer re-projects the full skill set on every init/upgrade/repair run, so existing projects convert organically the next time they're touched.
- `skills/verifier.py::repair_skills` always repairs to a copy and flips legacy `delivery_mode: symlink` manifest entries as it heals them. Net code deletion: the symlink repair branch and `_sync_skill_to_global_root` are gone.
- **Security posture preserved and tightened**: repair clears a symlink at the destination by unlinking the link inode (its target is never written), and gains an explicit refusal when an *ancestor* directory symlink-escapes the repo (previously only reachable via the `_copy_skill_file` write guard). The old `test_repair_rejects_external_symlink_destination` expectation changes accordingly: repair now *heals* a link at the managed path instead of failing, with the external target untouched — asserted explicitly.
- Copies inherit the canonical root's read-only mode (`copy2` preserves it) — deliberate: the projection stays non-editable, drift stays detectable/repairable.
- e2e fixtures were renamed off real pack skill names (`spec-kitty-setup-doctor` → `e2e-doctor-skill`): copy-mode verify compares entries against the package registry, so a fixture reusing a real name with fake content reports as drifted. (Symlink entries previously short-circuited to the global root, masking the collision.)
- Docs: new **ADR 2026-07-19-1** records the decision; ADR 2026-04-08-3 is marked superseded *for the wiring half only* (its global-canonical-install decision stands, reaffirmed in the new ADR). Also updated: the setup-doctor `agent-path-matrix.md` reference, `state/contract.py` surface notes, changelog. The ADR index row for 2026-04-08-3 was missing entirely; `freshen_adr_inventory` added it while indexing the new ADR.

## Why this approach

The symlink rationale in ADR 2026-04-08-3 was freshness — one global refresh updates all projects. But the installer already re-projects everything on every per-project run, so what symlinks actually added was cross-project action-at-a-distance: refreshing the global root silently changed skill content in repos whose generated command surfaces still matched an older CLI. Per-project refresh on that project's own upgrade is the more correct semantic. A config knob (symlink opt-in) was considered and rejected: the hazard cases are environmental, not preferential — the user in a dev-container didn't choose to be there, and nobody should have to configure their skills to be readable.

Verification: `tests/specify_cli/skills/` + provider + migration + contract + architectural + `tests/docs/` suites green (312 + 58 + 525); `ruff` clean; `mypy` reports only the two pre-existing file-scope artifacts also present on unmodified `main` (verified against the base). Terminology guard green.

Note for triage: this touches the skills installer/verifier surface, not the dashboard/mission-lifecycle/status surfaces named in the 3.2.x merge-hold.

Fixes #2412 (remaining scope — asks 1+2 landed in #2423).

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/2818"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787077955&installation_id=146454894&pr_number=2818&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F2818&signature=93affd4e4f90acc4cb076be572dd4666e35245c8112bfff239421dff87c5609b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/docs/adr/3.x/2026-04-08-3-global-skill-installation-per-project-symlinks.md
+++ b/docs/adr/3.x/2026-04-08-3-global-skill-installation-per-project-symlinks.md
@@ -1,8 +1,14 @@
 ---
 title: 'ADR 3 (2026-04-08): Global Skill Installation with Per-Project Symlinks'
-status: Accepted
+status: Superseded
 date: '2026-04-08'
 ---
+
+> **Superseded (2026-07-19):** the per-project **symlink** wiring decided here is
+> replaced by copy delivery — absolute symlinks dangle in dev-containers and are
+> unreadable to sandboxed agent harnesses (#2412). The **global canonical
+> install** half of this decision stands unchanged and is reaffirmed in
+> [ADR 2026-07-19-1](2026-07-19-1-skill-projection-copies-not-symlinks.md).
 
 ## Context and Problem Statement
 
--- a/docs/adr/3.x/2026-07-19-1-skill-projection-copies-not-symlinks.md
+++ b/docs/adr/3.x/2026-07-19-1-skill-projection-copies-not-symlinks.md
@@ -1,0 +1,85 @@
+---
+title: 'Skill projection delivers copies, not symlinks'
+status: Accepted
+date: '2026-07-19'
+---
+
+# Skill projection delivers copies, not symlinks
+
+**Status:** Accepted · **Date:** 2026-07-19 · **Supersedes:** the per-project wiring half of [ADR 2026-04-08-3](2026-04-08-3-global-skill-installation-per-project-symlinks.md) (its global-canonical-install decision stands and is reaffirmed here) · **Related:** [#2412](https://github.com/Priivacy-ai/spec-kitty/issues/2412), [ADR 2026-07-07-1](2026-07-07-1-ignored-surface-backfill-migration-pattern.md) (IGNORED-surface backfill pattern)
+
+## Context and Problem Statement
+
+ADR 2026-04-08-3 established two things: canonical skills live in user-global roots (`~/.kittify/agent-skills/`, later `~/.agents/skills/` and per-agent homes), and per-project agent directories reference them **via absolute symlinks**, with a copy fallback for filesystems that cannot symlink.
+
+Issue #2412 first surfaced the commit-side hazard of that wiring: an absolute symlink whose blob is `/Users/<name>/...` is machine-local by construction, and nothing gitignored it. That half was fixed by PR #2423 (contract-registered `IGNORED` surfaces + backfill migration). The issue's remaining ask — *should skills be delivered into the repo root as symlinks at all?* — is what this ADR decides.
+
+The gitignore fix made symlinks safe to **commit**. It did not make them safe to **use**:
+
+1. **Dev-containers and remote mounts.** Mount the repo into a container or sync it to a remote box and every projected skill points at a home directory that does not exist there. The agent sees a dangling link; skills silently vanish with no diagnosable error.
+2. **Sandboxed agent harnesses.** Harnesses that restrict file reads to the repo root refuse to resolve a symlink whose target is outside it. The whole point of `.agents/skills/` is to feed third-party agent tooling whose sandboxing spec-kitty does not control.
+3. **Global-root moves.** Uninstalling or relocating the CLI's canonical root dangles every projected skill in every repo on the machine simultaneously, until each runs a repair.
+
+## Decision Drivers
+
+* **Skills must be readable wherever the repo is readable.** The projection exists for third-party agents; delivery must not assume the reader can escape the project root or see the author's `$HOME`.
+* **Freshness must not regress.** The original symlink rationale — one upgrade touch-point — must be preserved in practice.
+* **No per-project state divergence.** Projected files remain managed, regenerated artifacts, never hand-edited project content.
+* **Minimal mechanism.** Prefer deleting a delivery mode over adding configuration for one.
+
+## Considered Options
+
+* **Option A: Deliver copies, always (chosen)**
+* **Option B: Keep absolute symlinks (status quo, post-#2423)**
+* **Option C: Configurable delivery mode (symlink opt-in)**
+
+## Decision Outcome
+
+**Chosen option: Option A — `_project_skill_file` always delivers a copy; the symlink preference is removed rather than made optional.**
+
+The freshness argument that justified symlinks in ADR 2026-04-08-3 is weaker than it looked: the installer re-projects the **full skill set on every init/upgrade/repair run**, so copies refresh whenever the project itself is touched. What symlinks added on top was *cross-project action-at-a-distance* — refreshing the global root silently changed the skill content of every repo on the machine, including repos whose generated command surfaces still matched an older CLI. Per-project refresh on that project's own upgrade is the more correct semantic, not a compromise.
+
+### Mechanics
+
+* `skills/installer.py::_project_skill_file` copies (`shutil.copy2`), never symlinks. A destination file whose content hash already matches the source is left untouched (idempotent re-runs); divergent content is archived to the migration backup root exactly as before.
+* A pre-existing projection **symlink** found at the destination is unlinked and replaced with a copy — conversion of existing projects happens organically on their next init/upgrade run, with **no dedicated migration**.
+* `skills/verifier.py::repair_skills` always repairs to a copy and records `delivery_mode: "copy"`, healing legacy manifest entries as they are repaired. Verify-side tolerance for not-yet-converted symlinks remains.
+* Copies inherit the canonical root's read-only mode (`copy2` preserves it) — deliberate: the projection is not user-editable, matching the symlink-era semantics. Local edits are drift, detected and repairable as before.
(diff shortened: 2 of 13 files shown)


stijn-dejongh:
Claiming this PR for a landing pass: rebase onto current `upstream/main`, reconcile the registry conflicts (CHANGELOG / ADR-README / page-inventory), adjudicate red checks, review the copy-delivery diff + new ADR, verify (tests / ruff / mypy). `installer.py`/`verifier.py` don't conflict; the copy-not-symlink fix is still needed (main still symlinks). Evidence to follow; the operator merges.

stijn-dejongh:

Rebased onto current `main` (`a0b48867e`) and force-pushed with lease (`785cb1269`). @rayjohnson — heads-up, your branch now carries a maintainer rebase + one docs-regen fold.

### Rebase / folds
- **Registry conflicts reconciled** (into your commit): `docs/changelog/CHANGELOG.md` (kept both Changed sections + the `2026-07-22` date), `docs/adr/3.x/README.md` (union — your new ADR row alongside the entries main added), `docs/development/3-2-page-inventory.yaml` (union). Only these three index/registry files conflicted — **`installer.py` / `verifier.py` / `state/contract.py` did not**.
- **Fold `785cb1269`** — regenerated the *generated* docs indexes to canonical form after the hand-merge: `docs_index.py --write` (adds the `2026-07-19-1-skill-projection` retrieval-index row, picks up your edited `2026-04-08-3` row) + `freshen_adr_inventory.py --all`.

- `tests/specify_cli/skills/` → **292 passed**
- `ruff check` (installer/verifier/contract) → clean
- **`mypy --strict src/specify_cli src/charter src/doctrine`** (the CI command) → **clean, 0 issues** (a *narrow* 3-file mypy shows 2 spurious cross-module errors — an invocation artifact, not real)
- `check_docs_freshness --ci` → **errors=0** (only 3 pre-existing environmental link-probe warnings on `cli-commands.md`)
- `tests/architectural/test_no_legacy_terminology.py` (doctrine `agent-path-matrix.md` touch) → **4 passed**

### Red-first proof
Reverted `installer.py` to `main` (the symlink delivery), kept your tests:
```
TestCopyDelivery::test_projection_is_copy_never_symlink                 FAILED
TestCopyDelivery::test_reinstall_replaces_legacy_symlink_with_copy      FAILED
TestCopyDelivery::test_reinstall_leaves_hash_equal_copy_untouched       FAILED
TestCopyDelivery::test_divergent_copy_archived_then_replaced            FAILED
```
Restored → all pass. The tests genuinely witness the copy behavior.

### Review notes
Not superseded — `main`'s installer still does `dest.symlink_to(...)`. Clean change: always `shutil.copy2`, replaces legacy projection symlinks, idempotent on hash-equal, archives divergent files; `repair_skills` always repairs to a copy and adds an `is_external_symlink` guard (refuses to repair through an out-of-repo symlinked dir) — **security posture is improved**, not just preserved. Dead `_sync_skill_to_global_root` + unused imports removed. Not under the doctrine-surface freeze (this is skills *delivery* plumbing, not the pack/activation model).

**State:** watching CI; merge-ready on green. The **operator merges** — I don't run `gh pr merge`.

stijn-dejongh:
**Folded the golden-count CI failure (`0d95004a0`).** The arch-gate `test_golden_count_ban` flagged `assert len(archived) == 1` (test_installer.py) as a new un-annotated convert-classified golden-count site, pushing `tests/specify_cli` 270 → 271 over the frozen ceiling. It's a genuine cardinality-is-contract assertion — a single-file projection archives exactly one prior file, and the next line asserts `archived[0]`'s content — so I annotated it with the sanctioned `# golden-count: cardinality-is-contract` escape hatch rather than re-freezing the ceiling (which would let an un-annotated site persist). `test_golden_count_ban.py`: 9 passed; the annotated test still passes. This was the only PR-attributable failure in that shard (your CI run: 1 failed / 410 passed). Re-run to confirm green.

Issue #21, opened 2025-11-09
Spec-Kitty Commands Not Accessible in Claude Code on Window


kentonium3:
# Spec-Kitty Commands Not Accessible in Claude Code on Windows

- Note: This is my first report of an issue on a github project for anything ever. I am obviously not a developer so this could easily be operator or basic setup error. If so, forgive the noise. 

## Environment
- **OS:** Windows 11
- **IDE:** VS Code
- **Claude Code:** v2.0.36
- **Spec-Kitty:** v0.4.6
- **Installation Method:** `uv tool install spec-kitty` (persistent method)
- **Project Path:** `C:\Users\Kent\Vaults-repos\bake-tracker`

## Description
Spec-Kitty initialization succeeds and dashboard launches successfully, but commands are not accessible through any of the expected interfaces (CLI, slash commands, or programmatic access).

## Steps to Reproduce

1. Install Spec-Kitty via `uv tool install spec-kitty`
2. Initialize project: `spec-kitty init`
3. Initialization completes successfully, dashboard launches at `http://127.0.0.1:9237`
4. Attempt to use commands via any interface

### 1. CLI Commands Not Available
```bash
$ spec-kitty specify "Phase 4: Service Layer Implementation"
Error: No such command 'specify'.
```

### 2. CLI Help Crashes with Encoding Error
```bash
$ spec-kitty --help
Traceback (most recent call last)
  # Full traceback showing encoding issues with ASCII art banner
```

### 3. Slash Commands Not Recognized in Claude Code
```bash
/spec-kitty.specify
Error: Unknown slash command: spec-kitty.specify
```

## What Works
- ✅ `spec-kitty init` completes successfully
- ✅ Dashboard launches at `http://127.0.0.1:9237`
- ✅ Command files exist in `.claude/commands/` directory
  - Confirmed: `spec-kitty.specify.md`, `spec-kitty.plan.md`, etc. are present
- ✅ `spec-kitty --version` returns `v0.4.6`

## Expected Behavior
Based on documentation, should be able to access commands via:
- **CLI:** `spec-kitty specify "feature description"`, OR
- **Slash command:** `/spec-kitty.specify` in Claude Code

## Questions

1. **Command Interface Design:** What is the intended interface for Spec-Kitty commands?
   - Are they CLI subcommands (`spec-kitty specify`)?
   - Claude Code slash commands (`/spec-kitty.specify`)?
   - MCP server tools (programmatic access)?
   
[...]

PR #2821, opened 2026-07-20
docs(implement-review): add sleep-protection guard and stale-WP triage step

rayjohnson:
## The problem

An unattended `implement-review` session on macOS — Claude Code driving the loop by following the `spec-kitty-implement-review` skill directly, no `spec-kitty-orchestrator` binary involved — had a dispatched implementer killed three times in one session when the laptop went to idle sleep: twice a 600s watchdog kill, once a mid-response connection drop. Nothing in the mission state recorded why; the WP was left `in_progress` with a partial uncommitted edit, and the failure looked identical to a flaky or stalled agent. The root cause only surfaced because the operator happened to mention the laptop had slept.

The gap is structural, not incidental: every `spec-kitty` CLI call in this flow (`agent action implement/review`, `move-task`, `status`) is short-lived — it exits in seconds. Even if one of those calls asserted a keep-awake, nothing would carry it between calls. Only a process that stays alive for the whole session — the agent-harness session itself when following this skill — can hold one, and nothing told operators to do so.

## The fix in plain terms

Add operating guidance to the skill: hold a `caffeinate` idle-sleep assertion for the session before dispatching the first agent, release it when the loop ends. Also extend the existing "Stale WP" troubleshooting entry so triage checks for a sleep event before assuming the agent hung — the two failure modes are otherwise indistinguishable from mission state alone.

## Technical details

- New **"Sleep Protection (Unattended Runs)"** subsection under Core Concepts in `src/doctrine/skills/spec-kitty-implement-review/SKILL.md`: `caffeinate -i -w $$ &` backgrounded at session start, `kill "$CAFFEINATE_PID"` after Step 6 (accept/merge) or on explicit halt. Notes the lid-close caveat (matches `spec-kitty-orchestrator`'s own `--no-caffeinate` docs) and that Linux has no default equivalent installed everywhere.
- States the boundary explicitly, since this was one of the field report's asks: `spec-kitty-orchestrator`'s own idle-sleep assertion (default-on since its v0.1.3 caffeinate PR, #21) is internal to that binary and does not cover a skill-driven session. Companion PR Priivacy-ai/spec-kitty-orchestrator#33 documents the same boundary from the orchestrator's side.
- Extended the **Stale WP** troubleshooting entry: check `pmset -g log | grep -i -E 'Sleep|Wake' | tail -5` for an event inside the dispatch window before force-moving and re-dispatching.
- Verified the change doesn't disturb the existing Codex-dispatch regression guard (`tests/doctrine/test_codex_dispatch_flags.py` asserts an exact count of `codex exec --sandbox danger-full-access` occurrences in this file — unchanged, still 3) or the terminology guard. Full `tests/doctrine/` suite (2513 tests) green.

## Why this approach

The cheapest fix that fully closes the reported gap is a skill-doc edit, not new CLI surface: it requires zero code, and the failure is entirely explained by "nothing told the orchestrating agent to hold a keep-awake." A dedicated `spec-kitty caffeinate` command (one of the field report's alternative asks) would duplicate what a one-line `caffeinate -i -w $$` in the skill already achieves, for a mechanism (`caffeinate`) that's already precedented in this codebase via the orchestrator binary. The `pmset -g log` triage addition targets the report's secondary ask (distinguish "machine slept" from "agent hung") with a read-only diagnostic rather than new state tracking, keeping the change doc-only and low-risk.

Not fixed here (out of scope for a skill-doc PR, noted for completeness): a Linux-native equivalent (`systemd-inhibit` needs to wrap the *outer* process, which a background shell can't do for itself) and structured sleep-event *detection* baked into `move-task`/`status` output remain open — the field report's request 3 as a first-class feature rather than an operator-run diagnostic. Worth a follow-up issue if there's appetite; this PR only ships the guidance that would have prevented the reported incident.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/2821"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787155482&installation_model_id=427282&pr_number=2821&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F2821&signature=2bf0528e493bd2465f141d85ca9187d46e390134c24c68d8d225b4fef7efb2af"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/doctrine/skills/spec-kitty-implement-review/SKILL.md
+++ b/src/doctrine/skills/spec-kitty-implement-review/SKILL.md
@@ -89,6 +89,48 @@ separate worktree, a review pass against a fixed diff, a focused debugging
 investigation, or validation that can run independently from the orchestrator's
 next scheduling decision.
 
+### Sleep Protection (Unattended Runs)
+
+Unattended implement-review runs are long-lived: a dispatched implementation
+agent can run for many minutes between the short-lived `spec-kitty` CLI calls
+that claim and move WPs. None of those CLI calls hold the host machine awake
+by themselves — only a process that stays alive for the whole session can. If
+the host sleeps mid-dispatch (default laptop power settings on macOS), the
+dispatched agent is silently killed or its connection drops, and nothing in
+the mission state records why. The failure then looks identical to a flaky or
+stalled agent (see the **Stale WP** entry under Troubleshooting).
+
+Before dispatching the first agent, hold a keep-awake assertion for the
+session and release it when the loop ends:
+
+```bash
+# macOS: idle-sleep assertion tied to this shell's own PID, backgrounded so
+# it does not block. Does NOT prevent lid-close sleep -- for a fully
+# unattended run, also disable lid-close sleep while plugged in, or keep
+# the lid open.
+caffeinate -i -w $$ &
+CAFFEINATE_PID=$!
+```
+
+Release it after Step 6 (accept/merge) or whenever the loop is explicitly
+halted:
+
+```bash
+kill "$CAFFEINATE_PID" 2>/dev/null || true
+```
+
+Linux has no default equivalent installed everywhere; if `systemd-inhibit` is
+available, prefer wrapping the outer orchestrating process invocation with it
+rather than backgrounding, since inhibitor scope is tied to the wrapped
+process's lifetime.
+
+**If `spec-kitty-orchestrator` is driving this run** (the standalone binary,
+not an agent-harness session following this skill), skip the above: the
+orchestrator already holds a macOS idle-sleep assertion for the duration of
+`orchestrate`/`resume` (on by default; `--no-caffeinate` to opt out — see its
+README). That coverage is internal to the binary and does not extend to an
+agent-harness session following this skill directly.
+
 ---
 
 ## The Mandatory Workflow Pattern
@@ -866,8 +908,21 @@ locks and commit status artifacts. If a locally patched or stale skill still
 uses `--full-auto`/`workspace-write`, rerun the lifecycle command from the
 repository-root checkout so its transactional status commit can complete.
 
-**Stale WP (in_progress but no agent activity)**: Force back to planned and
-re-dispatch:
+**Stale WP (in_progress but no agent activity)**: Before assuming the agent
+hung or is flaky, rule out host sleep — it produces the identical symptom
+(dead dispatch, partial uncommitted edit in the lane worktree, nothing in
+mission state explaining why) and is common on an unattended macOS laptop
+without the keep-awake guard from **Sleep Protection** above. Check for a
+sleep event since the WP was claimed:
+
+```bash
+pmset -g log | grep -i -E 'Sleep|Wake' | tail -5
+```
+
+If a sleep/wake pair falls inside the dispatch window, the kill is explained
+— start the keep-awake guard before force-moving and re-dispatching so it
+doesn't recur. Otherwise, force back to planned and re-dispatch:
+
 ```bash
 printf '%s\n' "Stale implementation lease recovered; redispatch required." > /tmp/stale-agent-feedback.md
 spec-kitty agent tasks move-task WP## --to planned \
--- a/src/doctrine/skills/spec-kitty-implement-review/SKILL.md
+++ b/src/doctrine/skills/spec-kitty-implement-review/SKILL.md
@@ -119,17 +119,34 @@ halted:
 kill "$CAFFEINATE_PID" 2>/dev/null || true
 ```
 
-Linux has no default equivalent installed everywhere; if `systemd-inhibit` is
-available, prefer wrapping the outer orchestrating process invocation with it
-rather than backgrounding, since inhibitor scope is tied to the wrapped
-process's lifetime.
+**Linux (systemd distros — the large majority):** `systemd-inhibit` is the
+equivalent, and its `tail --pid` form self-releases when the process dies,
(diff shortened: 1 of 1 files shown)


stijn-dejongh:
@rayjohnson : I like this suggestion. My main question here is: are there well supported alternatives on windows and  common linux distro's? 

Let's ensure the guidance is OS-specific or wrap it in a proxy utility class that no-ops on unsupported/non-impacted platforms. 

This might be a case where we want to go the "add to doctrine, not to product" route. 

In any case: thanks for the suggestion! I will think about it further. Feel free to get back to us if you have additional insights.

stijn-dejongh:
Claiming this PR for a landing pass: rebase onto current `upstream/main` (merges clean), run the terminology guard on the doctrine prose, adjudicate red checks. Sleep-protection content isn't on main yet, so it's not superseded. Evidence to follow; the operator merges.

stijn-dejongh:

Rebased onto current `main` (`a0b48867e`) and force-pushed with lease (`2d67ca098`). Clean rebase — **zero conflicts, no folds** (single doctrine SKILL.md, non-overlapping hunk).

- `tests/architectural/test_no_legacy_terminology.py` (doctrine-prose touch) → **4 passed**
- `tests/doctrine/test_spec_kitty_skill_content.py` + `test_spk_skill_pack.py` → **9 passed** (the sleep-protection + stale-WP additions don't trip a SKILL.md structure/content assertion)
- Agent skill copies (`.claude/skills/…`) are **gitignored/generated** — editing the authoritative source under `src/doctrine/skills/` desyncs no committed copy.

### Review notes
Not superseded — 0 `caffeinate`/sleep-protection lines on `main`'s SKILL.md; this extends the existing "Stale WP" triage entry and adds a new sleep-protection section. Pure operating-guidance prose (+57/−2, one file): hold a `caffeinate` idle-sleep assertion for unattended macOS `implement-review` sessions, and check for a sleep event before assuming an agent hung.

**State:** merge-ready on green. The **operator merges** — I don't run `gh pr merge`.

rayjohnson:
Fair callout — good catch that the guidance in this PR only covered macOS.

Looked into it: real mechanisms exist for both platforms you mentioned. Linux (systemd-based distros — the large majority) has `systemd-inhibit --what=idle:sleep -- tail --pid <pid> -f /dev/null` (the `tail --pid` part mirrors caffeinate's `-w` — it self-releases if the process dies, no polling needed). Windows has no shell one-liner at all; the real primitive is the Win32 `SetThreadExecutionState` API, callable via `ctypes` with no extra dependency.

Went with your second suggestion rather than OS-specific doctrine: `spec-kitty-orchestrator`'s own `prevent_idle_sleep()` context manager (added in #21, macOS-only) was already shaped exactly like the proxy utility you described — spawn/call at the start, release on exit, no-op-with-a-warning if the mechanism isn't there. I filled in the two missing branches instead of pushing conditional shell snippets out into this doctrine change: Priivacy-ai/spec-kitty-orchestrator#34.

It's not literally universal — non-systemd Linux (Alpine, minimal containers) still no-ops, same as before — but real coverage now spans macOS, Windows, and systemd-Linux, all behind one tested, no-op-safe seam. Given that, this doc change here could probably stay as-is or even get thinner, since the actual cross-platform handling now lives in the orchestrator itself rather than in per-OS instructions for the operator to remember.

stijn-dejongh:
Thanks @rayjohnson — the investigation and the `prevent_idle_sleep()` route are exactly right, and moving the real cross-platform primitive into the orchestrator (behind a tested, no-op-safe seam) is the better home than shipping per-OS shell snippets into doctrine. Appreciate you chasing down the actual Win32 / systemd mechanisms.

**Folded the cross-platform guidance into this doc** (`48160a353`) — since this skill covers the *no-orchestrator* path (an agent harness following it directly, where there's no long-lived orchestrator process to hold the assertion), the per-platform reality still belongs here:
- **systemd-Linux:** concrete `systemd-inhibit --what=idle:sleep -- tail --pid $$ -f /dev/null` recipe (self-releases like `caffeinate -w`); non-systemd Linux no-ops.
- **Windows:** noted that there's no shell one-liner — the primitive is Win32 `SetThreadExecutionState`, held by a live process — so run under the orchestrator or accept the sleep.
- **Points at `prevent_idle_sleep()`** as the canonical cross-platform home for this handling (the with-orchestrator path).

Kept the doc accurate to *current* coverage (orchestrator holds a macOS assertion today) rather than claiming the extended platforms — **the product-side cross-platform work (your `prevent_idle_sleep()` extension) is left to another maintainer to review/land on the orchestrator repo**, out of scope for this doctrine PR.

Rebased onto current `main` and force-pushed with lease (`48160a353`). Verification: terminology + skill-content → **13 passed**. Merge-ready.

Issue #2646, opened 2026-07-14
Approved WP stays stale because review-artifact verdict scan reads PRIMARY instead of COORD


MOES-Media:
## Problem

In a `lanes_with_coord` Mission, a later successful review can persist canonical lifecycle review evidence on COORD while `spec-kitty agent tasks status` scans the PRIMARY `tasks_dir` for review verdicts. The status view therefore continues to flag the earlier rejected artifact even after an approved cycle exists on the lifecycle authority.

Observed during #2626 / draft PR #2641:

- WP01 cycle 1 rejection was preserved as `review-cycle-1.md`.
- Cycle 2 was canonically approved: event `01KXGVAMKBG68FTJW1R1P10PVZ`, review ref `ff7eca5e2c248730909d91cddde93c6c01232701`.
- The approved `review-cycle-2.md` and updated override metadata were committed cleanly on COORD at `e5a575d0cfb13a2198e2e0795c3b200cce280883`.
- WP01 materializes as `approved`, but `spec-kitty agent tasks status --mission 01KXGR6S --json` still reports `_stale_verdict: true` and points to cycle 1 `verdict: rejected`.

The reviewer traced the display scan to PRIMARY `tasks_dir`, while the canonical lifecycle artifact is on COORD. Copying the approved artifact to PRIMARY would duplicate authority and hide the routing defect.

## Expected

The stale-verdict resolver reads review artifacts from the canonical lifecycle/review authority for the active topology and chooses the latest applicable structured review cycle/override. An approved WP with a committed later approved artifact should not display a stale rejection warning.

## Acceptance

- Add a real coord-topology regression with cycle 1 rejected on PRIMARY/history and cycle 2 approved on COORD.
- `agent tasks status` reports the WP approved with no stale verdict.
- Flat/single-branch behavior remains unchanged.
- Do not require duplicated review artifacts across PRIMARY and COORD.

This is an authority-routing residual discovered during #2626; it is recorded separately and is not authorization to expand the current Mission.

MOES-Media:
Claimed by @MOES-Media. Folding this authority-routing residual into the active #2626 Spec Kitty Mission and existing DRAFT PR #2641; progress evidence will stay on the PR until final closeout.

MOES-Media:
**Mechanism correction — the symptom still reproduces, but this issue's stated cause is stale.**

[...]

Issue #2275, opened 2026-06-30
fix(coordination): move-task --to approved over rejected latest review artifact does not persist an approved artifact in coord worktree — merge blocked by REJECTED_REVIEW_ARTIFACT_CONFLICT


stijn-dejongh:
## Problem

There is a review-artifact authority split between the lane worktree and the coord worktree that causes a silent merge block after a re-review approval cycle:

1. **Per-WP approve-over-rejected guard** checks the **lane** worktree. No review-cycle artifact exists there (artifacts live in the coord worktree), so approval passes.
2. **Merge gate** (`REJECTED_REVIEW_ARTIFACT_CONFLICT` / invariant `terminal_wp_latest_review_artifact_must_not_be_rejected`) checks the **coord** worktree. When the latest cycle there still has `verdict: rejected`, merge is blocked.

`move-task --to approved` over a rejected-latest WP correctly records the status-lane event (`approved`) but persists **no** approved review-cycle artifact in the coord worktree. The coord worktree retains the cycle-1 `verdict: rejected` artifact as the latest. The merge gate then fires even though the WP is status-lane `approved`. Recovery requires hand-writing a synthetic approved artifact.

## Evidence

Surfaced during mission `doc-quality-hardening-2245-01KW9AKV` (PR #2272, closes #2245). WP05 and WP06 each had cycle-1 rejections. After re-review and `move-task --to approved`, both WPs showed status-lane `approved` but the coord worktree retained `verdict: rejected` as the latest cycle artifact. `spec-kitty merge` was blocked by `REJECTED_REVIEW_ARTIFACT_CONFLICT` until approved artifacts were written manually.

See: retrospective `hn-004` in `kitty-specs/doc-quality-hardening-2245-01KW9AKV/retrospective.yaml` (pr-003). Force overrides on WP06: retrospective `n-007` (3 `--force` transitions).

## Proposed Fix (two options — pick one or combine)

**Option A — write an approved artifact on approval:**
When `move-task --to approved` is called on a WP whose latest coord-worktree review-cycle artifact has `verdict: rejected`, the command writes a new `review-cycle-<N+1>.md` with `verdict: approved` to the coord worktree, making the approval decision durable and visible to the merge gate.

**Option B — merge gate reads the status-lane event:**
[...]

PR #2834, opened 2026-07-21
fix(merge): split lane-state and review-cycle reads to their real partitions

rayjohnson:
## The problem

An operator running `spec-kitty merge` on a coordination-topology mission hit a false block: the CLI reported an already-approved WP as "still rejected," even though the real, correctly-approved review-cycle artifact existed — just on a different branch that hadn't been forwarded into the coordination branch yet. The coordination worktree's on-disk copy of that WP's review-cycle file was stale (an untracked stray leftover from an earlier cycle), and the merge gate treated it as ground truth.

## The fix in plain terms

The gate that checks "does this approved/done WP still have a rejected review" was reading two different kinds of information — a WP's current lane, and its review-cycle file content — from the same single directory. Those two things live in different places for a coordination-topology mission (lane state on the coordination branch, review-cycle files on the primary checkout), so one directory can never be correct for both. The fix makes it resolve each independently, from the right place.

## Technical details

`find_rejected_review_artifact_conflicts()` (`src/specify_cli/post_merge/review_artifact_consistency.py`) called `materialize(feature_dir)` for lane state and read `tasks/**` under that same `feature_dir` for review-cycle content — one caller-supplied directory serving both. Review-cycle content is `WORK_PACKAGE_TASK`, a PRIMARY-partition kind (lives on the primary checkout for every topology); lane state is `STATUS_STATE`, which stays on the coordination branch for a coord-topology mission. `resolve_planning_read_dir`'s own docstring already describes this exact failure mode almost verbatim: "a stale pre-mission copy would SHADOW the real primary planning truth."

The function now takes `(repo_root, mission_slug, wp_ids)` and resolves both directories internally by kind — `WORK_PACKAGE_TASK` for `tasks/` reads, `STATUS_STATE` for the `materialize()` lane read. Threaded through its three real callers:
- `merge/executor.py`'s `_enforce_review_artifact_consistency` (the real-merge gate that produced the field report)
- `merge/forecast.py`'s dry-run preview, which had the **identical latent defect**, just never exercised (every existing coord-topology test mocks this specific check away — I found an explicit comment admitting as much: "the full status materialize is skipped... not representative")
- `cli/commands/review/_lane_gate.py`'s post-merge lane-gate report (`spec-kitty review`)

**Tests:** two new integration tests in `tests/integration/test_merge_cluster_coord_read.py` using the shared `coord_topology_mission` fixture (the canonical home for this bug class):
- `test_review_artifact_gate_catches_genuine_rejection_on_coord_topology` — proves a genuine rejection is still caught when both partitions resolve correctly (positive control — the fix doesn't just make the gate a no-op).
- `test_review_artifact_gate_ignores_stray_artifact_on_coord_husk` — reproduces the literal field-reported shape: a stale rejected artifact on the coord husk, the real approved one on primary. Asserts no finding.

I manually verified the second scenario against the pre-fix code (temporarily reverted the core change) and confirmed it reproduces the exact reported false block — the old code returns a `RejectedReviewArtifactFinding` for the same fixture data my fix returns `[]` for.

All five existing call sites across `tests/merge/`, `tests/post_merge/`, `tests/regression/`, and `tests/specify_cli/cli/commands/` updated for the new signature. Full sweep (`tests/merge/`, `tests/post_merge/`, the coord-topology integration suite, `tests/regression/test_2684_review_override_recognition.py`, `tests/specify_cli/cli/commands/`, `tests/docs/`, terminology guard): 4264 passed. Ruff and mypy clean on every touched file.

## Why this approach

The obvious "just point everything at primary" fix (mirroring what `forecast.py` already does) would have traded a loud false-positive for a silent one: primary carries no authoritative status log for a coord-topology mission, so `materialize()` would never see a terminal lane, and the whole gate would quietly stop firing for genuinely-rejected WPs on *every* coord-topology mission — worse than the bug being fixed. I verified this concretely before committing to the design (confirmed via the write-side code and the shared test fixture's own "decoy" status log, deliberately seeded to catch exactly this class of wrong-leg read).

Splitting the resolution inside the one shared function — rather than patching each caller separately — fixes the bug class once instead of three times, and closes the gap between `run_review_artifact_consistency_preflight`'s own docstring claim ("the single implementation path shared by `merge` and `merge --dry-run` so the two surfaces cannot drift") and its actual prior behavior, where the *directory* passed to that shared function silently drifted between callers even though the function itself was shared.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/2834"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787244464&installation_model_id=427282&pr_number=2834&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F2834&signature=803de52555aceeaa6cc5d9dd50f13f467c39bcf10527b155cf90dfb4af1b342a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -2,7 +2,7 @@
 title: Changelog
 description: Canonical changelog for the Spec Kitty CLI and templates, following Keep a Changelog and Semantic Versioning, with added, breaking, and fixed entries per release.
 doc_status: active
-updated: '2026-07-18'
+updated: '2026-07-21'
 ---
 # Changelog
 
@@ -110,6 +110,9 @@ _The 3.2.6 development cycle is open. Entries land here as missions merge._
   baseline is regenerated accordingly.
 
 ### 🐛 Fixed
+
+- **The review-artifact consistency gate resolves lane state and review-cycle content from separate, correctly-partitioned directories.**
+  For a coord-topology mission, `find_rejected_review_artifact_conflicts` (shared by `merge`, `merge --dry-run`, and `spec-kitty review`'s lane gate) used one caller-supplied directory for both a WP's lane state (`STATUS_STATE`, coordination-branch-owned) and its review-cycle artifacts (`WORK_PACKAGE_TASK`, PRIMARY-partition for every topology) — a single directory can only ever be correct for one of the two. A stale or untracked stray review-cycle file on the coordination worktree could falsely block an already-approved WP (reported via field retrospective); conversely, resolving everything to the PRIMARY checkout would have silently defeated lane-state detection instead. Both partitions now resolve independently by kind.
 
 - **Honest force-provenance on evidence-gated backward edges (#2684, #2736, #2810).**
   Persisted `StatusEvent.force` is now truthful — falsy on the evidence-gated
--- a/src/specify_cli/cli/commands/review/__init__.py
+++ b/src/specify_cli/cli/commands/review/__init__.py
@@ -225,12 +225,13 @@ def _record_gate(
 def _run_lane_gate(
     feature_dir: Path,
     repo_root: Path,
+    mission_slug: str,
     console: Console,
     findings: list[dict[str, str]],
     gates_recorded: list[GateRecord],
 ) -> None:
     findings_before = len(findings)
-    check_wp_lanes(feature_dir, repo_root, console, findings)
+    check_wp_lanes(feature_dir, repo_root, mission_slug, console, findings)
     result: Literal["pass", "fail"] = "fail" if len(findings) > findings_before else "pass"
     _record_gate(gates_recorded, gate_id="gate_1", name="wp_lane_check", result=result)
 
@@ -408,7 +409,7 @@ def review_mission(
     gates_recorded: list[GateRecord] = []
     _mission_id_raw = meta.get("mission_id")
     _mission_id: str | None = str(_mission_id_raw) if _mission_id_raw else None
-    _run_lane_gate(feature_dir, repo_root, console, findings, gates_recorded)
+    _run_lane_gate(feature_dir, repo_root, mission_slug, console, findings, gates_recorded)
     _run_dead_code_gate(
         baseline_merge_commit=baseline_merge_commit,
         repo_root=repo_root,
--- a/src/specify_cli/cli/commands/review/_lane_gate.py
+++ b/src/specify_cli/cli/commands/review/_lane_gate.py
@@ -23,6 +23,7 @@ from specify_cli.status import materialize
 def check_wp_lanes(
     feature_dir: Path,
     repo_root: Path,
+    mission_slug: str,
     console: Console,
     findings: list[dict[str, str]],
 ) -> None:
@@ -51,7 +52,9 @@ def check_wp_lanes(
             f"  [green]✓[/green]  WP lane check: all {len(snapshot.work_packages)} WP(s) in done"
         )
 
-    review_artifact_conflicts = find_rejected_review_artifact_conflicts(feature_dir)
+    review_artifact_conflicts = find_rejected_review_artifact_conflicts(
+        repo_root, mission_slug
+    )
     if review_artifact_conflicts:
         console.print(
             "  [red]✗[/red]  Review artifact consistency: blocking latest review "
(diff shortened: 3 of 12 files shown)


stijn-dejongh:
**Triage traceability (Refs #2842).** This PR ships row 2 (review-artifact partition split) of the coordination/lane-drift field retrospective #2842. That retrospective is being closed as its tracking record (its four shipped-fixes are these open PRs; its three follow-ups are #2839/#2840/#2841), so no `Closes #2842` keyword is added here — a single row-PR must not auto-close the whole four-row retrospective. Landing decision for this PR is a separate maintainer call.

stijn-dejongh:
Thanks for this, @rayjohnson — reviewed against current `main`.

**Verdict: the core fix already landed independently; keeping this open for reference rather than merging.**

The partition split you implement here — resolving review-cycle artifacts (`WORK_PACKAGE_TASK`, PRIMARY-partition) separately from WP lane-state (`STATUS_STATE`, coord-partition) — landed via the `coord-commit-integrity` mission (FR-001, #2646 / #2275). On `main`, `find_rejected_review_artifact_conflicts` now reads the review-cycle through `_resolve_review_cycle_read_dir(...)` → PRIMARY while `materialize()` keeps lane-state on the coord husk. Same root cause, same fix shape. Because the merged version kept the `feature_dir` signature (rather than replacing it with `repo_root, mission_slug`), your production diff would now conflict hard across `preflight.py` / `executor.py` / `forecast.py` / `review/_lane_gate.py`, so we're not landing it as-is.

**Why kept-for-reference, not closed:** your two coord-topology integration tests —
`test_review_artifact_gate_catches_genuine_rejection_on_coord_topology` and
`test_review_artifact_gate_ignores_stray_artifact_on_coord_husk` — are valuable end-to-end coverage the merged mission did **not** add. We intend to cherry-pick them **with attribution to you and this PR**. Labeling `pr:kept-for-reference` so that stays discoverable.

**One honest note in your favor:** there's a residual gap your version *does* close that the merged mission left open. `merge --dry-run` on a coord-topology mission resolves its review-artifact preflight dir to PRIMARY (`forecast.py` → `feature_dir_for_preview`, `WORK_PACKAGE_TASK`), so the lane-state `materialize()` leg reads a partition that carries no authoritative status log for coord topology — and can silently pass a genuinely-rejected review. Real-merge is fine (it passes the coord husk); only the dry-run preview is affected. We'll track that as a focused follow-up citing this PR as the source, rather than reworking your whole diff here.

Thanks again — the field report and the tests are the durable value.

stijn-dejongh:
Follow-up filed: #2885 tracks the residual `merge --dry-run` coord-topology lane-state false-negative that this PR's approach closes, and calls out your two coord integration tests for harvest with attribution. Thanks again @rayjohnson.

rayjohnson:
Closing as superseded. While rebasing my local combo branch onto current `main`, this conflicted with `_resolve_review_cycle_read_dir()` in `post_merge/review_artifact_consistency.py` — a self-contained fix for the exact same bug (coord STATUS husk vs. PRIMARY-partition review-cycle artifacts), landed independently via the squash-merged coord-commit-integrity mission (refs #2646/#2275).

That fix achieves the same outcome as this PR — reading review-cycle artifacts from the PRIMARY partition instead of the coord husk — but derives `repo_root`/`mission_slug` internally from the existing `feature_dir` argument rather than changing `find_rejected_review_artifact_conflicts`'s public signature. All call sites (`lane_gate.py`, `forecast.py`, `preflight.py`) already pass the original single `feature_dir` argument on `main` today, confirming it's a complete, non-invasive fix already covering all 3 sites this PR touched.

Merging this now would just replace that cleaner fix with a redundant signature change for the same behavior. No action needed — the underlying bug is already fixed on `main`.

PR #2832, opened 2026-07-21
fix(merge): stale-lane remediation is aware of repo-root/planning lanes

rayjohnson:
## The problem

A mission's planning-artifact WPs are grouped into the canonical `lane-planning` lane, which by design has no `.worktrees/` entry — it resolves to the repository-root checkout, on the mission's own target branch (`lane_branch_name()`'s own documented special case: for `lane-planning` it returns the target branch itself, e.g. `main` or `release/3.1.1`, never a synthetic `kitty/mission-*-lane-planning` branch).

When that lane goes stale relative to the mission branch, `spec-kitty merge` generates its remediation unconditionally for every lane, regardless of kind:

```
Lane lane-planning must incorporate mission changes before merging.
Run: cd .worktrees/*-lane-planning && git merge kitty/mission-<slug>
```

`.worktrees/*-lane-planning` cannot exist, by construction, for this lane kind. An operator hitting this has to independently figure out that "lane-planning" means the repo-root checkout, then hand-craft the equivalent merge command themselves — the tool's own printed fix requires going around the tool. (Reported via an internal field retrospective from a 10-WP mission run; happy to share details on request.)

## The fix in plain terms

Teach the stale-lane remediation message about the one lane kind that never gets a worktree, so it points the operator at the real location instead of a nonexistent directory.

## Technical details

- `check_lane_staleness()` in `src/specify_cli/lanes/stale_check.py` now branches on `is_planning_lane(lane)` — the existing, documented classification seam (its own docstring: "the single seam where 'what counts as a planning lane' is decided... callers must keep asking the semantic question rather than comparing against the constant directly"), already used at three other call sites in `merge/executor.py`. The remediation-building logic is extracted into `_stale_remediation()` for direct unit testing.
- For `is_planning_lane(lane)`: emits `git checkout <lane_branch> && git merge <mission_branch>` from the repository root, naming the actual branch (already available as the `lane_branch` parameter — it's the resolved target branch for this lane kind).
- For every other lane (the vast majority): output is byte-for-byte unchanged.
- Tests: a fast pure-function suite on `_stale_remediation` (`tests/lanes/test_stale_check.py`) covering both branches with no git needed; an integration test through `check_lane_staleness` itself with real git branches; and a new test in `tests/lanes/test_merge.py` — `test_stale_planning_lane_blocked_with_repo_root_remediation` — that reproduces the literal field-reported symptom end-to-end through `consolidate_lane_into_mission` and asserts the fixed error text.
[...]

Issue #1834, opened 2026-06-11
spec-kitty accept auto-runs negative-invariant verification against the PRE-MERGE primary tree and overwrites recorded results


stijn-dejongh:
## What happened
During acceptance of mission \`tooling-stability-guard-coherence-01KTRC04\` (dogfood observation F-006, 3.2.0rc42): \`spec-kitty accept\` re-executes each negative invariant's \`verification_command\` from the **primary repo root** and **overwrites** the recorded \`result\` in \`acceptance-matrix.json\`.

Pre-merge, the primary tree cannot contain the mission's changes (they live on lane branches until \`spec-kitty merge\`) — so an honest invariant like \`pytest tests/git/test_protection_preserved.py\` (a suite the mission ADDS) flips to \`still_present\` because the test file doesn't exist on the primary tree yet. The matrix verdict computes \`fail\` and acceptance blocks, even though the invariant was verified green on the integrated lane tree during per-WP review.

## Why it matters
The gate's timing makes verifiable-by-command negative invariants **unusable pre-merge** — exactly when \`accept\` runs in the documented flow (accept → merge). Operators are pushed to the workaround of leaving \`negative_invariants\` empty and burying invariant evidence in criterion text (the pattern earlier missions used).

## Suggested direction
Either (a) run NI verification in a post-merge re-verify phase (or at \`merge\` time on the merged tree), or (b) make the verification cwd workspace-aware (resolve the integrated lane/mission tree pre-merge), or (c) don't overwrite a recorded non-pending result — record the re-run as a separate field with its tree/ref context.

Recorded in \`kitty-specs/tooling-stability-guard-coherence-01KTRC04/research/observations.md\` (F-006) on \`stijn-dejongh/spec-kitty@fixups/code-engine-stabilization\`.

stijn-dejongh:
Tracker hygiene (epic #391 dissolution): rehomed to **#1666 Execution-state & context domain-boundary redesign**. accept running negative-invariant verification against the wrong (pre-merge primary) tree is a coord/main/lane execution-context correctness bug. Type/priority unchanged (Bug, P1).

robertDouglass:
Fresh repro from mission `sync-daemon-orphan-cleanup-01KWC2A3` (#154), coordination topology.

[...]

PR #2835, opened 2026-07-21
fix(merge): hollow-review warning checks reviewer identity, not just force_count

rayjohnson:
## The problem

An operator merging a mission saw `spec-kitty merge` warn that four WPs "may have been approved by the implementing agent, not an independent reviewer." In every one of those cases, the `--force` usage that triggered the warning was actually an independent reviewer subagent bypassing an unrelated gate false-positive — pre-existing merge noise in history, unrelated to the diff actually under review. The heuristic couldn't tell the difference: `force_count >= 2` fires the same way whether `--force` was used by the implementer covering their own tracks or by a genuinely separate reviewer clearing unrelated noise.

## The fix in plain terms

Before warning about a WP's force count, check whether the event log can positively confirm that a *different* identity approved it than the one who implemented it. If it can, that's real evidence of independent review — don't warn. If the data is missing or ambiguous, keep warning exactly as before; this only adds a way to *rule out* a false positive, never a way to hide a real one.

## Technical details

Two small additions to `src/specify_cli/merge/preflight.py`:

- `_latest_actor_for_transition(feature_dir, wp_id, to_lane)` scans the raw `status.events.jsonl` for the actor on a WP's most recent transition into a given lane. This has to read the raw event log rather than the reduced `status.json` snapshot — the snapshot's `actor` slot is overwritten on every transition, so it can only ever tell you who did the *single most recent* transition of any kind, never "who specifically implemented" versus "who specifically approved."
- `_independent_reviewer_confirmed(feature_dir, wp_id)` compares the latest actor into `in_progress` against the latest actor into `approved`. Returns `False` (never suppress) whenever either actor is missing — absence of evidence isn't evidence of independent review.
- `_collect_force_count_warnings` now skips appending the warning only when `_independent_reviewer_confirmed` returns `True`. The other two hollow-review signals — `force_count` itself as a threshold, and the explicit `ReviewerSelfApproval` self-declaration — are untouched.

The actor comparison mirrors an existing pattern already in the codebase (`_actors_compatible` in `status/work_package_lifecycle.py`, used for claim-conflict detection) rather than inventing new comparison semantics — same normalization (`.strip()`, exact match), same "unknown data never asserts a positive claim" posture.

**Tests:** 8 new tests in `tests/merge/test_preflight_seam.py` — direct coverage of both new helpers (latest-by-timestamp selection across rework cycles, no-events-file, no-matching-wp-or-lane) plus the integrated suppress/keep/fail-safe behavior on `_collect_force_count_warnings`. All 17 hollow-review tests in the file pass, including the 9 pre-existing ones (confirmed no regression). Full `tests/merge/` + `tests/post_merge/` (703 tests), `tests/docs/`, and the terminology guard green; ruff and mypy clean.

## Why this approach

The fix is additive by construction rather than a redesign: every transition already records a distinct `actor` (`StatusEvent.actor`), so no new instrumentation or write-side change is needed — the data this needed already existed, it just wasn't being consulted. Scoping the change to a positive-disconfirmation path (only ever *removes* a warning, never adds one) keeps the blast radius small and the failure mode safe: if the new logic has a bug that makes it under-detect independence, the result is just the pre-existing behavior (an occasionally-noisy warning), never a newly-introduced false negative that hides a real hollow review.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/2835"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787246590&installation_model_id=427282&pr_number=2835&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F2835&signature=2139364f511bc80396990267ffdce65a1ae212a4fcb8990a903877a3cc37be3f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -160,6 +160,9 @@ _The 3.2.6 development cycle is open. Entries land here as missions merge._
   the zero-dependency root every layer can import downward), leaving the seam
   definition as the only `replace("\\", "/")` in `src/`.
 
+- **`merge`'s hollow-review warning no longer flags a WP with a genuinely independent reviewer.**
+  `force_count >= 2` alone can't distinguish "the reviewer used `--force` to bypass an unrelated gate false-positive" from "no independent review happened" — both increment the same counter. When the event log positively confirms a different actor logged the approving transition than the one that most recently implemented the WP, the warning no longer fires; absence of that evidence still warns as before (fail-safe default, never suppressed on uncertain data).
+
 - **Honest force-provenance on evidence-gated backward edges (#2684, #2736, #2810).**
   Persisted `StatusEvent.force` is now truthful — falsy on the evidence-gated
   review-rejection edges (`build_transition_plan` asks the FSM instead of
--- a/src/specify_cli/cli/commands/agent/tasks_move_task.py
+++ b/src/specify_cli/cli/commands/agent/tasks_move_task.py
@@ -1854,7 +1854,7 @@ def _mt_hop_policy_metadata(
     return None
 
 
-def _binding_role_for_lane(lane: Lane) -> str | None:
+def _binding_role_for_lane(lane: Lane | str) -> str | None:
     """Map a target lane to its resolved-binding role.
 
     Shared by :func:`_mt_emit_transitions` (the live transition-emit path,
--- a/src/specify_cli/coordination/status_transition.py
+++ b/src/specify_cli/coordination/status_transition.py
@@ -14,7 +14,7 @@ from collections.abc import Callable
 from dataclasses import dataclass, replace
 from datetime import UTC, datetime, timedelta
 from pathlib import Path
-from typing import Any, TypeVar, cast
+from typing import Any, TypeVar
 
 from specify_cli.coordination.outbound import queue_saas_emission
 from specify_cli.core.commit_guard import GuardCapability
--- a/src/specify_cli/merge/preflight.py
+++ b/src/specify_cli/merge/preflight.py
@@ -407,6 +407,65 @@ def _enforce_review_artifact_consistency(
     raise typer.Exit(1)
 
 
+def _latest_actor_for_transition(
+    feature_dir: Path, wp_id: str, to_lane: str
+) -> str | None:
+    """Return the actor on WP's most recent transition into *to_lane*.
+
+    Scans the raw event log rather than the reduced snapshot, because the
+    snapshot's ``actor`` slot is overwritten on every transition -- it can
+    only ever tell us who did the LATEST transition of any kind, never who
+    specifically claimed/implemented versus who specifically approved.
+    Returns ``None`` when the log is absent/unreadable or no matching,
+    actor-bearing transition exists for this WP.
+    """
+    events_path = feature_dir / _STATUS_EVENTS_FILENAME
+    if not events_path.exists():
+        return None
+    try:
+        raw_lines = events_path.read_text(encoding="utf-8").splitlines()
+    except OSError:
+        return None
+    latest_key: tuple[str, str] = ("", "")
+    latest_actor: str | None = None
+    for raw_line in raw_lines:
+        try:
+            event = json.loads(raw_line)
+        except json.JSONDecodeError:
+            continue
+        if not isinstance(event, dict):
+            continue
+        if event.get("wp_id") != wp_id or event.get("to_lane") != to_lane:
+            continue
+        actor = event.get("actor")
+        if not actor or not str(actor).strip():
+            continue
+        key = (str(event.get("at") or ""), str(event.get("event_id") or ""))
+        if key >= latest_key:
+            latest_key = key
+            latest_actor = str(actor).strip()
+    return latest_actor
+
+
+def _independent_reviewer_confirmed(feature_dir: Path, wp_id: str) -> bool:
+    """True when WP's latest approval actor positively differs from its
+    latest implementation actor (#2412-adjacent field report, item #9).
+
+    ``force_count`` alone cannot distinguish "reviewer used --force to bypass
+    an unrelated gate false-positive" from "no independent review happened" --
+    both increment the same counter. This checks the one thing that actually
+    answers the question: did a different identity log the approving
+    transition than the one that most recently claimed/implemented the WP?
+    Returns False (never suppress) when either actor is missing/unknown --
+    absence of evidence is not evidence of an independent review.
+    """
+    implementer = _latest_actor_for_transition(feature_dir, wp_id, "in_progress")
+    reviewer = _latest_actor_for_transition(feature_dir, wp_id, "approved")
+    if not implementer or not reviewer:
+        return False
+    return implementer != reviewer
+
+
 def _collect_force_count_warnings(
     feature_dir: Path,
     wp_set: set[str],
@@ -415,7 +474,10 @@ def _collect_force_count_warnings(
     """Append force_count>=2 warnings from ``status.json`` (WP05 split helper).
 
     Behavior-preserving extraction of the status-snapshot scan formerly inlined
-    in ``_collect_hollow_review_warnings`` (FR-005, keeps CC <= 15).
+    in ``_collect_hollow_review_warnings`` (FR-005, keeps CC <= 15) -- plus one
+    additive guard (item #9): a WP whose approving actor is positively
+    confirmed distinct from its implementing actor is not a hollow review,
+    even with a high force_count, so it is not warned about here.
     """
     status_path = feature_dir / _STATUS_FILENAME
     if not status_path.exists():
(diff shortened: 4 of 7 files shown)


stijn-dejongh:
**Triage traceability (Refs #2842).** This PR ships row 3 (hollow-review reviewer identity) of the coordination/lane-drift field retrospective #2842. That retrospective is being closed as its tracking record (its four shipped-fixes are these open PRs; its three follow-ups are #2839/#2840/#2841), so no `Closes #2842` keyword is added here — a single row-PR must not auto-close the whole four-row retrospective. Landing decision for this PR is a separate maintainer call.

stijn-dejongh:
Claiming this PR for a landing pass: rebase onto current `upstream/main`, adjudicate red checks, verify red-first (the force_count>=2 suppression is additive on top of the merged `coord-commit-integrity` refactor of `_collect_force_count_warnings`), fold fixes as needed. Evidence to follow; the operator merges.

stijn-dejongh:

Rebased onto current `upstream/main` (`eb06ca176`) and force-pushed with lease (`46f572fd3`). @rayjohnson — heads-up, your branch now carries a maintainer rebase.

### Rebase / fold
- **One conflict, resolved:** `docs/changelog/CHANGELOG.md`. Kept upstream's `updated: '2026-07-22'` and **both** Fixed sections — the newly-landed entries plus your hollow-review entry, preserved verbatim. (Canonical changelog is `docs/changelog/CHANGELOG.md`; the root `CHANGELOG.md` is a symlink to it.)
- **No code conflict.** The merged `coord-commit-integrity` mission had already split `_collect_hollow_review_warnings` into `_collect_force_count_warnings` + `_collect_self_approval_warnings`; your additive guard layered cleanly onto `_collect_force_count_warnings` on rebase.

- `tests/merge/test_preflight_seam.py` → **46 passed**
- `ruff check` (preflight.py + test) → clean
- `mypy src/specify_cli/merge/preflight.py` → clean
- `tests/architectural/test_no_legacy_terminology.py` (changelog prose touch) → **4 passed**

### Red-first proof (the test witnesses the bug)
Swapped `merge/preflight.py` back to `upstream/main` (pre-fix), kept the new tests:
```
test_collect_force_count_warnings_suppressed_when_reviewer_differs → FAILED
  AssertionError: assert {'WP01': ['force_count=3']} == {}
```
Restored the fix → passes again. So without your change the warning fires; with it, a positively-confirmed independent reviewer suppresses it — and absence of evidence still warns (fail-safe), which the sibling `..._kept_when_no_event_log` / `..._kept_when_same_actor` tests pin.

### Review notes
Edits canonical `src/` (no generated mirror), touches only private `_`-prefixed helpers (no versioned-contract surface), no SSOT-violating duplication (`_latest_actor_for_transition` is distinct in purpose from the existing self-approval scan). No adversarial squad: this isn't an architectural/API-surface change and the evidence above covers it.

**State:** watching CI; merge-ready on green. Per our workflow the **operator merges** — I don't run `gh pr merge`.

stijn-dejongh:
**Added fold (`932541f5f`) — boy-scout: clear pre-existing repo-wide quality-gate reds.** These were red on `main` for lint/type issues unrelated to this PR's diff; folding the fixes here (operator-approved scope increase) so the CI Quality gate goes green instead of inheriting a pre-existing red:

- ruff F401 — drop unused `typing.cast` in `coordination/status_transition.py`
- ruff F401 — drop unused `pathlib.Path` in `tests/specify_cli/compat/test_planner_distribution_profile.py`
- mypy `--strict` arg-type at `tasks_move_task.py:1890` — `_binding_role_for_lane` received a `str` (`transition_targets` is `list[str]`) but was annotated `Lane`; widened the param to `Lane | str`. Zero runtime change (`Lane` is a `StrEnum`, so its `==` comparisons behave identically, matching the existing `target == Lane.FOR_REVIEW` compare in the same emit loop).

Verified on the tip: `ruff check src tests` clean · `mypy --strict src/specify_cli src/charter src/doctrine` clean (1064 files) · affected suites green (compat + coordination status_transition + agent move-task = 59 passed). @rayjohnson heads-up, one extra maintainer commit on your branch.

Expected fresh-CI steady state: quality-gate now green; #1834 stays red as the standing honest-P0 on main.

stijn-dejongh:
**Rebased onto current `main` (`8f21f8594`, post-#2832) and force-pushed → `17dc4027a`.** Clean rebase, no conflicts (#2832 touched `lanes/stale_check.py`, disjoint from this PR).

**On the `test_mission_created_resume_does_not_double_fire` failure — classified as a pre-existing non-deterministic shard flake, not this PR:**
- This PR's diff has no causal path to the mission-creation lifecycle fan-out (it's `merge/preflight.py` + a `Lane | str` annotation widening + two unused-import removals).
- The test passes **in isolation**, at **file level**, and across the whole **`tests/specify_cli/core/` dir** — identically on this tip **and** on pristine `upstream/main`.
- The failure is a lifecycle-fan-out double-count; that fan-out is daemon/SaaS-dispatched (async), i.e. a cross-file timing/ordering leak that only surfaces under the full ~742-test shard. Root cause is shard test-isolation, worth a separate flaky-test tracker — not a fold on this PR.

Re-running CI is the correct response here (not a retry-to-green of a real failure). Expected steady state on the fresh run: hollow-review + quality-gate green; #1834 the standing honest-P0 red on main.

stijn-dejongh:
**Root-caused the `test_mission_created_resume_does_not_double_fire` CI failure — it was a pre-existing timing-fragile test, not this PR. Folded a deterministic fix (`3c3621177`).**

**Mechanism (confirmed locally):** `create_mission_core` mints a fresh ULID per call and names the mission dir `<slug>-<mid8>`, where `mid8` = the ULID's first 8 Crockford chars = the top 40 bits of the 48-bit ms timestamp — so it only changes every **~256 ms**. The test calls `create_mission_core` twice and assumes both land in the **same** dir (so the per-dir lifecycle dedup fires the SaaS fan-out once). That holds only when both calls fall inside one 256 ms window. Under loaded CI (6 xdist workers + coverage on a 6-vCPU runner) the two calls straddle a boundary → different dirs → the second `MissionCreated` writes a fresh log the dedup can't see → fan-out fires twice → `got 2`. Green on `main` by timing luck; this PR merely perturbed scheduling. I verified two creates 300 ms apart deterministically produce different `mid8` dirs, and within one window collide.

**Fix (test-only):** pin the ULID so both calls resolve to one mission directory — making the 'resume' genuine and exercising the dedup deterministically (with an explicit `assert first.feature_dir == second.feature_dir` guarding the premise). **No product change** — `create_mission_core` is intentionally non-idempotent by slug (two creates = two missions, distinguished by `mid8`); only the test's timing assumption was wrong. Also type-annotated the module to `mypy --strict` clean while there.

Verified: ruff clean · `mypy --strict` clean on the file · both tests pass. This shard should now be green on a fresh run.

stijn-dejongh:
@rayjohnson — a heads-up on this PR so you're not surprised by the maintainer commits.

While landing it we hit a red CI check (`test_mission_created_resume_does_not_double_fire` in the `integration-tests-core-misc (specify-cli-rest)` shard). We traced it and it is **not caused by your change** — it's a pre-existing *timing-fragile* test. `create_mission_core` names each mission dir `<slug>-<mid8>`, and `mid8` is derived from the ULID timestamp in a way that only changes every ~256 ms. That test creates a mission twice and assumes both land in the same dir; under loaded CI (parallel workers + coverage) the two calls straddle a 256 ms boundary, get different dirs, and the dedup fan-out fires twice. It's been green on `main` purely by timing luck — your PR just perturbed scheduling enough to expose it.

We folded a small **test-only** fix onto your branch (`3c3621177`, pinning the ULID so the two calls resolve to one dir) plus a couple of pre-existing lint/type cleanups (`17dc4027a`). Your actual fix (`af7c68b1e`) is untouched. Full technical write-up is in the comments above.

**Ask:** if you (or anyone) hit an intermittent failure like this again — the `fire-once` / mid8-timing double-fire, or anything that's red on CI but green locally — please flag it back to us rather than retrying to green, so we can track and root-cause it. These timing-window flakes are easy to paper over and worth catching at the root. Thanks for the patience, and for the fix.

Issue #2841, opened 2026-07-21
Coordination branch has no re-sync mechanism and no general reconciliation command


rayjohnson:
## Two related gaps in the same underlying model

A mission's coordination branch is a one-time snapshot of the target branch at `mission create` time, and nothing in spec-kitty's toolchain keeps it — or the mission's lane branches — reconciled against reality afterward. A field report from a 10-WP mission hit three distinct symptoms of this over one merge cycle: a lane branch silently missing a CLI flag added to main after the coordination branch was cut, a stale/corrupted review-cycle artifact on the coordination worktree that had to be manually diffed against the real committed content on a different branch, and 11 coordination-branch commits left unreachable from target after merge (this last one turned out to be an inherent consequence of the default squash-merge strategy, not a bug — noted separately in the tracking issue).

### Gap 1: no re-sync mechanism (bootstrap staleness)

`ensure_coordination_branch()` (`src/specify_cli/missions/_create.py:195-260`) snapshots the target branch once, at `mission create` time, and never re-syncs. The only staleness check at creation time, `_is_ancestor()`, guards against *re-running* `mission create` on an already-diverged pre-existing branch — it never asks "has main moved since I was created." Nothing downstream (`specify`/`plan`/`tasks`/`finalize-tasks`) checks coordination-branch-vs-target staleness at all. The only related check, `doctor coordination`, compares a worktree to *its own* branch tip — not to main.

**Open design question**: when should a re-sync happen? Options include a periodic check surfaced during `tasks`/`finalize-tasks`, an explicit `spec-kitty doctor coordination --check-staleness` mode, or something automatic at plan time. This needs a maintainer call on cadence/UX before it's a mechanical fix.

### Gap 2: no general cross-branch reconciliation command

`spec-kitty sync workspace` (`src/specify_cli/cli/commands/sync.py`) is scoped to exactly one workspace — `GitVCS.sync_workspace` rebases *that workspace's own commits* onto its tracked base, with no sibling-branch awareness. `doctor coordination --fix` (`_coordination_doctor.py:680-777`) *does* already reconcile one specific coordination-vs-reality split-brain (a stranded `done` status after a merge rollback) — but it's narrowly scoped to one status enum, not arbitrary file content like a WP's `review-cycle-N.md`.

[...]

PR #2868, opened 2026-07-22
fix: agent-action-review auto-commit fails on coord worktree (rebase/takeover of #2612)

### LINKED ISSUE #2607: GC-2b fast-tests-core-misc baseline is not portable across checkout paths (absolute-path parametrize ids)
## Summary

Running the coverage-preservation guard (`tests/architectural/test_gate_coverage.py::test_gc2b_current_selection_matches_baseline[fast-tests-core-misc]`) from a checkout location other than the one the committed baseline (`tests/architectural/baselines/fast-tests-core-misc-nodeids.txt`) was frozen from produces a spurious RED — 6 "dropped" and 6 "added" node-ids that are the *same 3 tests* (`tests/prompts/test_tasks_prompt_ownership_metadata.py`, 2 parametrized paths each), differing only in the absolute worktree/checkout path segment embedded in the parametrize id.

Discovered during `ci-test-topology-performance-01KXBJRT` WP09's cross-cutting GC-2b audit (mission closing #1931-adjacent CI-topology work). Confirmed **not** a real coverage drop: after stripping the worktree-path segment, the dropped and added sets are byte-identical.

## Root cause

`tests/prompts/test_tasks_prompt_ownership_metadata.py` computes:

```python
_REPO_ROOT = Path(__file__).resolve().parents[2]
...
@pytest.mark.parametrize("prompt_path", _TASKS_PROMPT_SURFACES, ids=lambda path: path.as_posix())
```

`ids=lambda path: path.as_posix()` embeds the **absolute filesystem path** (derived from `Path(__file__).resolve()`) into the pytest node-id. Because the committed `tests/architectural/baselines/*-nodeids.txt` files are frozen as literal node-id strings (`gc.collect_real_union_for_target` / `--freeze-baselines`), any test file with an absolute-path-based parametrize id makes the frozen baseline non-portable: it only matches a `pytest --collect-only` run from the *exact same absolute checkout path* it was frozen from.

Evidence in-repo already: `tests/architectural/baselines/fast-tests-core-misc-nodeids.txt` currently carries a mix of `.worktrees/ci-test-topology-performance-01KXBJRT-lane-{a,b,c,f,u}` path fragments — each WP that touched this baseline during `ci-test-topology-performance-01KXBJRT` froze it from its own lane worktree, silently overwriting only the entries its own change affected and leaving the rest stale-but-still-matching (because those other entries didn't change content). The mechanism has been quietly "working" only because no single re-freeze needed to touch this particular file's rows before now.

**This will also fail on the first real CI run of any PR touching this baseline**, because GitHub Actions' checkout path (e.g. `/home/runner/work/spec-kitty/spec-kitty`) does not match any locally-frozen worktree path either.

## Impact

- Low real risk (3 tests, non-content-affecting, easily identified by inspection), but it undermines trust in GC-2b: a guard designed to catch real 0-dropped/0-double-run violations currently has a standing false-positive that will fire on environment change alone, teaching agents/reviewers to expect (and possibly rubber-stamp) "explainable" GC-2b failures — the opposite of what a fail-closed coverage gate should train.

## Suggested fix (either, or both)

1. **Preferred** — fix `tests/prompts/test_tasks_prompt_ownership_metadata.py`'s parametrize ids to be repo-relative, not absolute: `ids=lambda path: path.relative_to(_REPO_ROOT).as_posix()`. This is a pure test-authoring fix, zero behavior change, and makes the node-id stable across any checkout location.
2. Defense in depth — `tests/architectural/_gate_coverage.py`'s baseline capture/compare (`collect_real_union_for_target`, `baseline_diff`) could normalize/strip any absolute path segments matching the current repo root before comparing, so a future test with the same authoring smell doesn't reintroduce the same non-portability.

## Repro

```bash
uv run pytest tests/architectural/test_gate_coverage.py -q -k "fast-tests-core-misc"
# from a worktree whose path differs from whatever wrote the committed baseline
```

Found by: mission `ci-test-topology-performance-01KXBJRT`, WP09 (cross-cutting coverage-union audit, T026). See `docs/plans/testing/ci-coverage-union-audit.md` on that mission's branch for the full evidence (real guard output + path-normalization proof).

stijn-dejongh:
**Maintainer take-over / rebase of #2612** (thanks @rayjohnson — this carries your fix forward). Opened because #2612 had gone `CONFLICTING` against `main`.

## Why this matters now
This fixes the acute half of the #2841 coord-bookkeeping friction: on a **coordination-topology** mission, `spec-kitty agent action review WP##` fails to commit its status artifacts with `SafeCommitPathPolicyError` because the legacy `safe_commit` fallback passed `worktree_root=repo_root` while the artifacts live under `.worktrees/<coord>/…`. We hit this live and it stalled an implement-review loop (repro on #2841). Refs #2841 (the coord-commit bug). NOTE: this does NOT close #2607 — that is an unrelated GC-2b coverage-baseline issue; the branch name is misleading.

## What the rebase entailed (non-trivial — main moved)
`main` refactored `workflow.py` since the PR: `_commit_workflow_change` was extracted into `workflow_executor.py`. The fix was **re-applied onto main's post-refactor call site**, not blindly re-added:
- Kept the PR's `_worktree_root_for_feature_dir(repo_root, feature_dir)` helper and the `worktree_root` kwarg threaded into `_commit_via_legacy_safe_commit` → `safe_commit(...)`.
- **`workflow_executor.py`'s legacy leaf was calling `_commit_via_legacy_safe_commit` without `worktree_root` at all** (a latent break given the now-required kwarg) — fixed to derive and pass `_worktree_root_for_feature_dir(...)`.
- Resolved the arch-test conflict by keeping main's drift-proof `ContentDescriptor` allow-list (discarding the PR's obsolete line-number bookkeeping); added a `read_dir` stub the PR's new test needed for main's #2508 PRIMARY_METADATA anchoring.

## Verification
- `uv run pytest` on the 3 touched suites → **47 passed**.
- `uv run ruff check` on both changed modules → clean.
- `uv run mypy` → clean on the diff (the 1 reported error is in a zero-diff pre-existing-baseline file, `runtime/next/_internal_runtime/schema.py`).

Supersedes #2612 (kept open for its discussion; close in favor of this once reviewed). Part of a two-track coord-trust remediation — the structural half (#2841 Gap 1+2 + the diff-compliance runtime-exemption) is being specced separately and composes on top of this write-path fix.

https://claude.ai/code/session_01MAYG4JHbHudbGAHpRKw5Pr

### DIFF
--- a/src/specify_cli/cli/commands/agent/workflow.py
+++ b/src/specify_cli/cli/commands/agent/workflow.py
@@ -36,6 +36,7 @@ from __future__ import annotations
 
 from specify_cli.core.constants import (
     MISSION_TYPE_RESEARCH,
+    WORKTREES_DIR,
 )
 from specify_cli.missions._read_path_resolver import (
     _canonicalize_primary_read_handle,
@@ -575,9 +576,27 @@ def _resolve_workflow_read_dir(
     return read_dir
 
 
+def _worktree_root_for_feature_dir(repo_root: Path, feature_dir: Path) -> Path:
+    """Return the correct worktree_root for safe_commit, derived from feature_dir.
+
+    When feature_dir lives inside .worktrees/<name>/..., returns
+    .worktrees/<name> so that absolute paths under that worktree normalize
+    correctly (preventing SafeCommitPathPolicyError on paths whose first
+    component is .worktrees/).  For primary-checkout feature_dirs, returns
+    repo_root unchanged.
+    """
+    worktrees_parent = repo_root / WORKTREES_DIR
+    try:
+        rel = feature_dir.resolve().relative_to(worktrees_parent.resolve())
+        return worktrees_parent / rel.parts[0]
+    except ValueError:
+        return repo_root
+
+
 def _commit_via_legacy_safe_commit(
     *,
     repo_root: Path,
+    worktree_root: Path,
     target_branch: str,
     paths: list[Path],
     message: str,
@@ -622,7 +641,7 @@ def _commit_via_legacy_safe_commit(
     # tracking ``main``) is REFUSED by the guard, never waived (FR-008).
     result = safe_commit(
         repo_root=repo_root,
-        worktree_root=repo_root,
+        worktree_root=worktree_root,
         target=CommitTarget(ref=target_branch),
         message=message,
         paths=tuple(paths),
@@ -635,8 +654,6 @@ def _commit_via_legacy_safe_commit(
         sha=getattr(result, "sha", None),
         wp_id=wp_id,
     )
-
-
 
 
 def _print_commit_summary(*, command_name: str, json_output: bool = False) -> None:
--- a/src/specify_cli/cli/commands/agent/workflow_executor.py
+++ b/src/specify_cli/cli/commands/agent/workflow_executor.py
@@ -280,9 +280,11 @@ def commit_workflow_change(
         return
 
     # Legacy fallback (TODO(WP08): replace with the legacy bridge).
+    legacy_worktree_root = w._worktree_root_for_feature_dir(repo_root, feature_dir)
     try:
         w._commit_via_legacy_safe_commit(
             repo_root=repo_root,
+            worktree_root=legacy_worktree_root,
             target_branch=placement.ref,
             paths=paths,
             message=message,
--- a/tests/git/test_guard_capability_regression.py
+++ b/tests/git/test_guard_capability_regression.py
@@ -263,6 +263,7 @@ def test_legacy_workflow_commit_refused_on_protected_target(
     with pytest.raises(ProtectedBranchRefused):
         workflow._commit_via_legacy_safe_commit(
             repo_root=repo.repo_root,
+            worktree_root=repo.repo_root,
             target_branch=repo.target_branch,
             paths=[changed],
             message="chore: workflow status bookkeeping",
--- a/tests/specify_cli/cli/commands/agent/test_workflow_placement_routing.py
+++ b/tests/specify_cli/cli/commands/agent/test_workflow_placement_routing.py
@@ -357,6 +357,108 @@ def test_commit_workflow_change_legacy_path_routes_through_the_seam(
     )
 
 
+def test_worktree_root_for_feature_dir_inside_worktrees(tmp_path: Path) -> None:
+    """_worktree_root_for_feature_dir returns the worktree root for coord dirs."""
+    from specify_cli.cli.commands.agent.workflow import _worktree_root_for_feature_dir
+
+    worktree_dir = tmp_path / ".worktrees" / "my-mission-abc123-coord"
+    feature_dir = worktree_dir / "kitty-specs" / "001-demo"
+    feature_dir.mkdir(parents=True)
+
+    result = _worktree_root_for_feature_dir(tmp_path, feature_dir)
+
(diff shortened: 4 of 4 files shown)


stijn-dejongh:
## Adversarial pre-merge squad — adjudicated verdict: **HOLD** (needs revision)

Three lenses; **the fix does not yet resolve the live coord symptom** (empirically confirmed).

- **reviewer-renata — HOLD (confirmed):** the corrected `safe_commit(worktree_root=…)` is gated by a *sibling* pre-check — `git status --porcelain -- <paths>` at `cwd=repo_root` (`workflow.py:618-627`). For a **registered linked worktree with `.worktrees/` gitignored** (this repo), that porcelain returns **empty**, firing the phantom `"committed"` early-return **before** `safe_commit` is ever called. **Reproduced:** a real `git init` + `git worktree add .worktrees/coord` + a pending `status.json` under it → `git status --porcelain -- <path>` from repo_root = `[]`, with and without gitignore. So the coord status never commits, the worktree stays dirty, and `merge` still refuses — the fix's path is unreachable in the coord case.
- **renata #3:** real coord missions (complete `meta.json` identity) route through the **modern `_commit_via_coordination_transaction`**, not the patched legacy leaf (only reached on degenerate #2508 identity) — so the fix targets an edge path.
- **debugger-debbie — LAND** and **paula-patterns — LAND**: both verified the `safe_commit` mechanics/seam/rebase are correct in isolation (right seam, not whack-a-field, clean rebase), but neither modeled the porcelain gate that precedes it.

### Required before this can land
1. Make the porcelain pre-check (and the git-commit cwd) run in the **coord worktree**, not `repo_root`, so the pre-check and `safe_commit` agree on the working tree (thread `worktree_root`).
2. A **real-repo end-to-end** regression test (real `git worktree add`, assert the status files actually land on the coord branch and the worktree ends clean) — the current test stubs `safe_commit` in a non-git `tmp_path`, so it bypasses the porcelain gate and asserts only the argument to a stub.
3. Verify the **modern transaction path** doesn't have the same phantom-commit gate.
4. Body corrected: this does **not** close #2607 (unrelated coverage issue).

### Recommendation
This point-fix is a band-aid on a write-path that a coord mission may not even reach, and it reinforces that the durable fix is **structural write-placement enforcement** (route every mission-artifact write through one placement port, staging directly in the target ref's worktree — the #2841 prevention model). Suggest holding this and folding the correct fix into that structural mission, rather than layering more per-site patches. Thanks @rayjohnson — the `safe_commit` root-cause is real; it's the surrounding gate that undoes it.

stijn-dejongh:
Closing without merging. The 3-lens squad + an empirical repro confirmed this fix is **unreachable in the coord case** (the porcelain pre-check at `cwd=repo_root` short-circuits for `.worktrees/`-gitignored coord files before `safe_commit` runs; and real coord missions route through the modern transaction path, not this legacy leaf). The `safe_commit` root-cause @rayjohnson found is real and valuable — the surrounding porcelain gate undoes it.

**Retaining the analysis** (the adjudicated squad review above), and writing the full root-cause into the docs (see the engineering-notes PR). The correct coord-commit fix is folded into a dedicated **coordination-branch trust / write-placement** friction-remediation mission, where it's subsumed by routing every mission-artifact write through one placement port (staging directly in the target ref's worktree). Thanks @rayjohnson.

Issue #2883, opened 2026-07-23
Dossier parity reconciler (#2180) — squad-review follow-ups (MINOR/NOTE)


stijn-dejongh:
Follow-ups from the adversarial-squad review of #2820 (Dossier Parity Reconciler, closes #2180). The two MAJORs were folded into the PR (per-WP projection golden vector; contract-doc byte-spec of the per-artifact `content_hash` form). These are the surviving MINOR/NOTE items — none block #2820; grouped here per the landing-pass runbook (`docs/development/pr-landing.md` §8/§12).

### 1. Drift-guard for `WP_STATIC_PROJECTION_FIELDS` (architect + paula, MINOR)
`src/specify_cli/dossier/hasher.py` hand-maintains a 14-field allow-list that defines the per-WP **content hash**. Its only test (`tests/dossier/test_canonical_hash.py::test_projection_shape_is_documented_and_static_only`) is tautological — it builds the dict from that very tuple, so it can never fail. Add an **exhaustiveness guard** that partitions `WPMetadata.model_fields` into `WP_STATIC_PROJECTION_FIELDS ∪ {explicit runtime/review/descriptive set}` and fails on any *unclassified* authored field. Without it, a future authored WP field silently drops out of the content hash (false PARITY) — the exact whack-a-field failure this mission exists to kill. (Note: `WP_RUNTIME_FIELDS` in `frontmatter.py` is a *narrow* git/shell set, not the complement, so it can't be reused directly as the deny-list.)

### 2. Single `present_projection(artifacts)` helper (paula #2 / renata NOTE-3, NOTE)
The `(relative_path, content_hash)` present-artifact projection is built in three places with an inconsistent filter: `snapshot.py:compute_parity_hash_from_dossier` (`is_present and content_hash`), `reconcile.py:_present_projection` (`is_present` only), and `reconcile.py` recorded-inline (`summary["is_present"]`). Insulated today (the reconciler compares its two projections to each other, not to the stored hash), but if a future change trusts the stored hash the `and content_hash` asymmetry becomes a live false-divergence. Extract one canonical helper.

### 3. Unparseable-WP frontmatter falls back to raw-byte hashing (renata MINOR-1 / architect NOTE)
[...]

Issue #2686, opened 2026-07-16
Hash the WP static projection, not raw WP##.md bytes, in dossier/sync


stijn-dejongh:
**Parent:** #2684 (native sub-issue). **Blocked by #2684.**

## Outcome

Give WP Markdown two explicit, non-interchangeable hash domains:

- **projection hash** — normalized static `WPMetadata` projection for semantic/static identity and projection drift;
- **raw body hash** — SHA-256 of exact `WP##.md` bytes for capture TOCTOU, body upload/deduplication, SaaS stored-body verification, and browser/source parity.

The prior direction to replace the body-upload hash input with the normalized projection is superseded. A projection hash must never be presented or compared as a raw-body hash.

## Why

Current indexing hashes the normalized projection while body capture re-reads raw Markdown and compares it to the same overloaded `content_hash`. Every real frontmatter-bearing direct WP therefore fails with `content_hash_mismatch`.

#2684 still owns runtime-state eviction and remains the sequencing prerequisite for stable projection semantics. Priivacy-ai/spec-kitty-events#49 owns the versioned event/journal contract and cross-repository rollout. #3544 owns recursive Markdown capture/body delivery.

## Scope

- add separately named projection-hash and raw-body-hash fields to producer models and local persistence;
- keep snapshot/projection drift on the projection domain declared by the contract;
- make body TOCTOU and enqueue compare raw bytes only with the raw-body field;
- dual-write only after SaaS read compatibility is deployed;
- version or migrate local journal/queue records atomically;
- expose typed legacy/migration-required outcomes instead of cross-domain mismatch;
- provide dry-run affected counts, rollback criteria, and telemetry for mixed versions.

## Acceptance

- [ ] Projection and raw-body hashes have distinct names and fields in index models, events, journal/queue rows, body tasks, API payloads, and SaaS storage.
- [ ] Body capture re-reads exact bytes and compares raw SHA-256 only with the raw-body field.
- [ ] SaaS exact-body verification uses the raw-body field; projection drift uses the projection field.
- [ ] A real frontmatter-bearing WP captures and verifies byte-exactly without `content_hash_mismatch`.
- [ ] Legacy rows drain safely or stop with a typed migration-required outcome; no legacy hash is silently reinterpreted.
- [ ] Migration is atomic or explicitly versioned; dry-run reports affected missions/records before mutation.
- [ ] Cross-version contract fixtures from Priivacy-ai/spec-kitty-events#49 gate rollout.
[...]

PR #2898, opened 2026-07-23
fix(dossier): reconciler squad-review source follow-ups (#2883 items 1-4,7)

LynnColeArt:
Closes the source-side squad-review follow-ups from #2883, re-scoped against current main now that the reconciler (#2180) landed. The landing-folds turned out to be test/doc pins only, so every source item below was still open.

## What landed

**1 — projection exhaustiveness guard (fix).** The old test was tautological. Added a partition of `WPMetadata.model_fields` into static / runtime-review / descriptive, with a gate that fails on any authored field in none of the three. Several authored fields were previously unclassified, so a new field silently escaping the hashed projection would now trip the gate instead of shipping.

**2 — one `present_projection` helper (refactor).** The `(relative_path, content_hash)` projection was built in three places with divergent filters: the snapshot-hash producer used `is_present AND content_hash`, the reconciler's two sites used `is_present` only. Extracted one canonical `present_projection(artifacts)` in `snapshot.py` (accepts both `ArtifactRef` objects and recorded summary dicts) and routed the producer, `get_parity_hash_components`, and both reconciler sites through it. No behavior change (present implies a hash), but the three can no longer drift apart.

**3 — parse-failure fallback hardened (fix).** `_hash_wp_projection` fell back to raw-byte hashing on any parse failure, which hashes runtime-mutable bytes and breaks AS-4 churn-immunity. Now a schema-invalid WP frontmatter fails closed (marked non-present, error reason recorded) rather than contributing an unstable hash. A frontmatter-absent stub (a `FrontmatterError`, stable raw bytes) still gets the raw-byte path so bare scaffolding keeps scanning.

**4 — live scan-abort bug (fix).** A list or scalar WP frontmatter raised a `TypeError` inside `FrontmatterManager.read` that escaped the `(FrontmatterError, ValidationError, ValueError)` catch in the indexer and aborted the entire scan. Fixed at the source: `read()` now guarantees a mapping or raises `FrontmatterError`, so one malformed WP degrades to non-present instead of taking the whole mission's scan down.

**7 — rebaseline semantics documented (note).** Rebaseline is a full re-snapshot of current source, not a hash reformat in place. It absorbs any source change made since the last emit, which means it silently advances the recorded hash to current source and masks pre-cutover drift. That is intentional for the one-time cutover and bounded by A-003 (no live hosted customers), but a later reader must not mistake it for reformatting the stored value. Documented on the module and the rebaseline function.

## Deferred, with reason

**5 — dual-accept validator tightening (blocked on WP05).** Building this out, the emit end-state is already pinned: `test_emitted_snapshot_hash_is_canonical_and_field_unchanged` asserts a fresh emit is `sha256:`-prefixed. The only remaining piece is dropping bare-hex acceptance on the load/validate side. That validator is shared with reads of recorded drift baselines, which stay bare hex until the WP05 rebaseline sweep has run across the local backlog (see `test_accepts_bare_hex_backcompat`). Tightening it now would reject those legacy reads. So it is deferred until rebaseline has run everywhere, not landed here. Tracked on #2883.

**6 — cross-repo golden fixture (cross-repo).** Truly single-sourcing the golden vectors needs a shipped surface both repos can import, which is more than an in-repo test hoist. The byte-spec already lives in the contract doc. Tracked against SaaS #2686 rather than landed here.

## Verification

- `tests/dossier/` + `tests/tasks/test_frontmatter_unit.py`: 400 passed.
- ruff check + `ruff format --diff` clean on changed source; mypy clean on `snapshot.py` + `reconcile.py`.
- Only the logical edits are in the diff. No wholesale reformat of drift-prone files.

Refs #2883.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/2898"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787425917&installation_model_id=427282&pr_number=2898&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F2898&signature=15a3f7f8c7d88a75cedc17ab606a30f113184c3425174de9382a0ee12ca5d7c6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/specify_cli/cli/commands/reconcile.py
+++ b/src/specify_cli/cli/commands/reconcile.py
@@ -77,21 +77,6 @@ _EXIT_ERROR = 2
 _DEFAULT_MISSION_TYPE = "software-dev"
 
 
-def _present_projection(artifacts: object) -> list[tuple[str, str | None]]:
-    """Project present artifacts to canonical ``(path, content_hash)`` entries.
-
-    Present-only, so both the freshly-rebuilt source and the recorded snapshot
-    are projected symmetrically (an added/removed file surfaces as a NAMED
-    divergence, a byte change as a content mismatch). This is the exact input
-    shape WP01's canonical hash consumes.
-    """
-    entries: list[tuple[str, str | None]] = []
-    for artifact in artifacts:  # type: ignore[attr-defined]
-        if getattr(artifact, "is_present", False):
-            entries.append((artifact.relative_path, artifact.content_hash_sha256))
-    return entries
-
-
 def reconcile_mission_dossier(
     mission_slug: str,
     *,
@@ -148,7 +133,7 @@ def reconcile_mission_dossier(
     # Load the recorded snapshot FIRST: with no record there is nothing to
     # reconcile against, and fabricating a pass would violate fail-closed.
     try:
-        from specify_cli.dossier.snapshot import load_snapshot
+        from specify_cli.dossier.snapshot import load_snapshot, present_projection
 
         recorded_snapshot = load_snapshot(feature_dir, mission_slug)
     except Exception as exc:  # noqa: BLE001 - fail-closed: any load failure is an ERROR
@@ -174,10 +159,8 @@ def reconcile_mission_dossier(
             error=f"could not rebuild dossier from source for {mission_slug}: {exc}",
         )
 
-    source_projection = _present_projection(dossier.artifacts)
-    recorded_projection = [
-        (str(summary.get("relative_path", "")), summary.get("content_hash_sha256")) for summary in recorded_snapshot.artifact_summaries if summary.get("is_present")
-    ]
+    source_projection = present_projection(dossier.artifacts)
+    recorded_projection = present_projection(recorded_snapshot.artifact_summaries)
 
     # WP03 owns hashing + comparison (C-001). We never pass the snapshot's stored
     # parity_hash: it is the legacy concat-of-hashes form, not WP01's canonical
--- a/src/specify_cli/dossier/hasher.py
+++ b/src/specify_cli/dossier/hasher.py
@@ -204,6 +204,64 @@ WP_STATIC_PROJECTION_FIELDS: tuple[str, ...] = (
     "phase",
 )
 
+# The complement of WP_STATIC_PROJECTION_FIELDS, authored here so the
+# exhaustiveness guard (tests/dossier/test_canonical_hash.py) can prove EVERY
+# WPMetadata field is consciously classified — a newly-authored field must be
+# placed in one of these sets, never silently dropped from the content hash
+# (#2883 item 1). Split into two sets for intent clarity.
+#
+# Runtime / review state: mutates during a mission (execution + review) and MUST
+# NOT move the content hash (AS-4).
+WP_RUNTIME_PROJECTION_FIELDS: frozenset[str] = frozenset(
+    {
+        "agent",
+        "agent_profile",
+        "model",
+        "role",
+        "assignee",
+        "lane",
+        "status",
+        "review_status",
+        "shell_pid",
+        "shell_pid_created_at",
+        "history",
+        "activity_log",
+        "review_feedback",
+        "review_feedback_file",
+        "reviewed_by",
+        "reviewer",
+        "reviewer_agent",
+        "reviewer_shell_pid",
+        "approved_by",
+    }
+)
+
+# Descriptive / structural / identity metadata: authored, but deliberately NOT
+# part of the WP's hashed content contract (branch+vcs config, mission identity,
+# free-form descriptive fields). Revisit membership if any becomes load-bearing
+# for content parity.
+WP_DESCRIPTIVE_PROJECTION_FIELDS: frozenset[str] = frozenset(
+    {
+        "base_branch",
+        "base_commit",
+        "branch_strategy",
+        "planning_base_branch",
+        "merge_target_branch",
+        "created_at",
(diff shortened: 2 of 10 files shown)


stijn-dejongh:
Claiming this PR for a landing pass: rebase onto upstream/main, adjudicate any red checks, run a lightweight adversarial review squad, and confirm the #2883 fold linkage. Evidence to follow. — maintainer landing (Stijn)

stijn-dejongh:
## Landing pass — verdict: **LAND** (merge-ready on green; operator merges)

Ran the maintainer runbook (`docs/development/pr-landing.md`) in an isolated worktree + a lightweight 2-lens adversarial squad (`architect-alphonso`, `debugger-debbie`).

### Rebase & CI
- Branch already sits on the current `upstream/main` tip (`6d9ed490d`); rebase was a no-op. `mergeStateStatus: CLEAN`.
- Every CI check that ran is green; the rest are path-filtered `skipping` shards.

- `tests/dossier/` + `tests/tasks/test_frontmatter_unit.py`: **400 passed**.
- `ruff check` clean, `mypy` clean on changed source (`snapshot.py`, `indexer.py`, `reconcile.py`).
- One `ruff format` nit on `frontmatter.py` (`WP_RUNTIME_FIELDS` line) is **pre-existing on `main`** (reproduced on `upstream/main`, untouched by this PR) — not a fold.

### Red-first verification (bugfix items 3 & 4) — both witness the bug
- **Item 4** (frontmatter mapping guard): reverting `frontmatter.py` reds 3 tests with the exact `TypeError: '<' not supported between instances of 'str' and 'int'` — through a real production caller (`status/wp_metadata.py:639`). Restored → green.
- **Item 3** (indexer fail-closed): reverting `indexer.py` reds `test_schema_invalid_frontmatter_wp_fails_closed` (old code kept the artifact present/raw-byte-hashed). Restored → green.

### Squad findings
- **Architect (design / SSOT / contract):** no refutation of any framed claim. The `is_present ⟹ non-empty content_hash` invariant that the new unified `present_projection` relies on **holds** across every `ArtifactRef` construction site; the field-partition exhaustiveness guard is genuine (50 fields, 0 unclassified); and the `FrontmatterManager.read()` contract-widening is safe across all 12+ non-test call sites (each already catches `FrontmatterError` or broader). One non-blocking MINOR (invariant is unenforced at the schema level) → folded into the follow-up below.
- **Debugger (regression / caller-impact):** found one **MAJOR**, independently reproduced by me — see below.

### The MAJOR → filed as #2900 (not folded), with reasoning
Item 3's fail-closed change, combined with `_is_wp_artifact`'s broad `^WP\d{2,}` glob (any location) and `WPMetadata`'s `extra="forbid"`, flips a **real non-task auxiliary file** from present → non-present:

```
kitty-specs/quality-devex-hardening-3-2-01KRJGKH/glossary-fragments/WP07.md
  BEFORE (main): is_present=True,  hash=01b90436c3e8...
  AFTER  (#2898): is_present=False, hash='', error_reason="wp frontmatter invalid: 6 validation errors for WPMetadata..."
```
Across a code-version boundary this can manufacture a spurious `MISSING_IN_SOURCE` divergence — the exact false-positive class the reconciler exists to prevent.

**Why this does not block landing (verified nil live blast radius):** the only two non-`tasks/` `WP*.md` files with real frontmatter both live in missions with **no recorded dossier snapshot**, and **no** `snapshot-latest.json` in the repo records a WP-glob-matching non-task `.md` file as present. There is currently no recorded state that can trigger the divergence; it's latent, and bounded by spec A-003 (no live hosted customers).

**Why filed, not folded:** the correct fix is a classification-boundary design decision (scope the `WPMetadata` path to `tasks/`-located / manifest-declared files, vs. narrow the glob at source) on an active-mission seam with dedicated follow-up work already queued (#2883 items 5/6). Per the runbook's "mission-scope, file don't cram" carve-out, and with nil live impact, it's tracked as **#2900** (Bug, P2, linked as a sub-issue of #2883) rather than crammed into this contributor PR. The two squad MINORs (unenforced `ArtifactRef` invariant; item-3 test-witness gap) are folded into #2900 to be picked up together.

> Operator note: if you'd rather not carry the latent trap, the minimal scoping fold is small and keeps this PR's own tests green — happy to add it. My recommendation is **land now + #2900**, given the nil live blast radius.

### Issue linkage
- PR correctly uses `Refs #2883` (empty `closingIssuesReferences`) — **does not** close #2883, which stays open for the genuinely-deferred item 5 (blocked on WP05 rebaseline sweep) and item 6 (cross-repo golden fixture, SaaS #2686). Correct as-is.
- New follow-up **#2900** filed and linked under #2883.

Nothing else to change here. Merge-ready on green — handing off (operator merges; I don't run `gh pr merge`).

Issue #2709, opened 2026-07-16
Mission squash merge overwrites target-newer artifacts and drops acceptance provenance


robertDouglass:
Observed during the #2658 mission merge (TF-049, TF-066).

## Reproduction

1. On the target branch, create newer mission trace content and accept the mission so `meta.json` contains `accepted_at`, `accepted_by`, `accepted_from_commit`, `acceptance_mode`, `accept_commit`, `acceptance_history`, `vcs`, and `vcs_locked_at`.
2. Leave older copies of those mission artifacts on the mission/coordination branch.
3. Run the supported squash merge.
4. Observe target-newer trace sections and canonical acceptance/VCS fields replaced by older mission-branch copies, even though the acceptance commit remains in ancestry.

## Expected

Squash merge must reconcile mission artifacts with target-newer canonical state rather than replacing them wholesale. Acceptance provenance must survive, with regression coverage for both `meta.json` and append-only traces.

The #2658 branch restores the affected mission artifact in PR #2689; the general merge defect remains.

stijn-dejongh:
Triage: child bug (catfooding, merge/git-mechanics). Parent epic **#1795** (Parallel-lane git mechanics & event-log merge semantics) — already correctly parented, left in place. Acceptance-provenance loss on squash merge; P1. No exact duplicate found; sibling merge defects #2710/#2711 from the same #2658 merge.
— triage pass (planner-priti)

stijn-dejongh:
Flagged as potential P0 release-blocker pending operator confirmation — **data loss / wrong-authority write**. The supported squash merge replaces target-newer mission artifacts with older mission-branch copies and drops canonical acceptance/VCS provenance (`accepted_at`, `accept_commit`, `acceptance_history`, `vcs`, `vcs_locked_at`) from `meta.json` plus target-newer append-only trace sections. Destroying recorded acceptance provenance during merge is the same corruption family as the confirmed P0 #1834 (accept-state overwrite), and merge is a non-recoverable-by-retry core workflow. The #2658 branch only restored one affected artifact in PR #2689; the general merge defect and its regression coverage remain absent. No committed failing repro test yet; per ADR 2026-07-17-1 a red-first reproduction (covering both `meta.json` and the append-only traces) should precede any red-mainline escalation.

— P0 re-triage (planner-priti)

stijn-dejongh:
[...]

PR #2906, opened 2026-07-24
Lifecycle gate execution context + tool-artifact owner (retire 11 gate exemptions onto one owner)

stijn-dejongh:
## Lifecycle Gate Execution Context + Tool-Artifact Owner

Two structural seams that stop lifecycle gates from guessing, and give generated
bookkeeping artifacts a single owner instead of an ever-growing list of one-off
gate exemptions.

### Seam 1 — Gate Execution Context (WP01–07)
Lifecycle gates previously inherited the caller's cwd/tree and produced confident
**wrong** verdicts when run from the wrong worktree. Gates now receive an explicit
execution context (tree, ref, phase) and **refuse rather than guess** when the
context is missing or inconsistent. Includes the #2795 claim-blocker fix (a
PRIMARY-partition lock write was misclassified by the ref-advance dirty scan;
now topology-agnostic, with a red-first repro).

### Seam 2 — Tool-Artifact Owner + exemption ratchet (WP08–19)
Eleven scattered per-gate churn/ownership predicates are **retired onto one
canonical owner** — `coherence.is_toolchain_generated_churn` (legs:
`is_self_bookkeeping_churn`, `is_coord_residue_churn`, `is_status_state_path`) and
`mission_runtime.kind_for_mission_file`. An architectural **exemption-registry
ratchet** (one row file per surviving mechanism) forbids a 12th silent exemption:
every retired mechanism is deleted, and any genuine must-keep must register an
explicit `justified-survivor` row.

Three mechanisms are genuine survivors (registered, not silently kept):
- `_is_self_write_only_diff` (WP14) — diff-scoped vcs-lock/frontmatter claim-time check
- `_is_review_lifecycle_basename` (WP15) — human-authored `notes.md` / `review-cycle-*.md`
- `_is_review_handoff_survivor_path` (WP17) — review-handoff files

C7 cross-gate agreement (claim-time gate vs consolidation gate) is flipped green
by WP13's routing onto the owner; the former strict-xfail is now a hard pass with
an AST guard against reintroduction.

### Verification
- All 19 WPs implemented and independently reviewed (reviewer-renata), with
  several reject-fix cycles that caught real defects (a vacuous NFR harness, a
  missing CLI subcommand, a silent survivor, a cosmetic enrolment, an accept-gate
  false-pass).
- Pre-merge aggregate squad (three independent opus lenses — correctness,
  architecture/boundaries, gate-integrity) on the consolidated diff: **all three
[...]

PR #2905, opened 2026-07-24
docs: structural sanity & concern guard (closes #2864/#2865/#2866/#2867)

stijn-dejongh:
### Why
When PR #2855 retired the anti-sprawl ratchet (#2851), `docs/` lost its only mechanical guard for canonical section placement and required indexes. A docs-wide concern audit then found the expected drift: point-in-time reports misfiled under `architecture/`, a split-brain `plans/notes/` shadow tree, and an incomplete section index. This mission cleans up that drift **and** installs a durable replacement guard so it can't recur — wired back into the existing documentation doctrine rather than a one-off script.

### What changed
- **Doctrine (reusable framework).** Extends `DIRECTIVE_042` + the `common-docs` styleguide with a machine-parseable `structural_lint_config` block (single source of truth): the concern-bucket→section map, point-in-time patterns/allowlist, and the frontmatter contract — including the sanctioned MADR exemption for ADRs and an exemption for redirect stubs.
- **The guard.** A new `scripts/docs/docs_structural_lint.py` — the durable successor to the retired ratchet — with 4 config-driven checks (index completeness, point-in-time placement, no cross-root shadow basenames, frontmatter contract). It loads its policy from the styleguide (no hard-coded rules), exits 0 on the clean tree, and is wired into `docs-freshness.yml` as a CI gate.
- **The cleanup.** Redistributes 9 point-in-time `architecture/` artifacts to `plans/engineering-notes/`, enumerates `architecture/index.md`, retires the `plans/notes/` 1.x shadow tree (folded into canonical twins), and repoints every in-repo referrer (0 dead links).

### Issues
Closes #2864. Closes #2865. Closes #2866. Closes #2867. Advances **#2302** (docs standard as doctrine) under epic **#2314** (bucket C — sanitization; bucket E — docs governance as doctrine).

### Safety
- `docs_structural_lint` → **exit 0** (648 pages, 0 violations); `tests/docs/` green; terminology guard green; `check_docs_freshness --ci` green.
- **`redirect_map.yaml` + `redirect_baseline_urls.json` are byte-for-byte untouched** — the moved paths are never-published, so no redirect regeneration was run (this preserves the landed `01KW3SBK` mission's 149 published-URL redirects).

### Known follow-ups (upstream gaps, out of scope here)
- `scripts/docs/bulk_ref_rewrite.py` has a file-move path-doubling bug (`resolve_destination` assumes `to:` is a directory root); referrers here were repointed via `relative_link_fixer` + controlled edits instead.
[...]

PR #2910, opened 2026-07-24
fix: two P0s reding mainline — non-interactive interview hang (#2876) and merge clobbering gate artifacts (#2804)

### LINKED ISSUE #2804: `spec-kitty merge` re-scaffolds acceptance-matrix.json + issue-matrix.md to placeholders after merge, destroying filled verdicts/evidence
## Summary

`spec-kitty merge` **resets both coordination gate artifacts to empty placeholders** as a post-merge step, on the integration branch, AFTER the done-gate check has already consumed their filled contents:

- `acceptance-matrix.json` → `overall_verdict: pending` with every criterion `pass_fail: pending`, `evidence: null`, `notes: "TODO: replace with a real acceptance criterion"`.
- `issue-matrix.md` → the bare scaffold (headline rows only, verdicts back to `in-mission`/`unknown`), dropping any hand-authored terminal verdicts, evidence, and follow-up handles.

The merge itself succeeds and WPs move to `done` (the gate passed against the *filled* artifacts first), but the **permanent record left on the branch is the placeholder version** — the acceptance evidence and terminal issue verdicts an operator/reviewer needs at PR time are gone.

## Repro (mission `charter-deadcode-noop-campsite`)

1. Author a complete `acceptance-matrix.json` (overall `pass`, all criteria filled) + a terminal `issue-matrix.md` (e.g. #2373 `verified-already-fixed`, 9 rows).
2. `spec-kitty accept` passes; `spec-kitty merge --mission <slug>`.
3. After merge: `grep overall_verdict acceptance-matrix.json` → `pending`; `issue-matrix.md` is back to the placeholder scaffold.

## Impact

- Silent audit-trail loss: the merged branch (and any PR built from it) carries empty gate artifacts even though the mission genuinely passed.
- Distinct from the `finalize-tasks` scaffolder, which is **idempotent** and correctly preserves a filled `issue-matrix.md`. The reset is specific to the **merge** path.

## Recovery (today)

Restore both from git history after merge, before opening the PR:
```
git checkout <good-sha> -- kitty-specs/<slug>/issue-matrix.md kitty-specs/<slug>/acceptance-matrix.json
git commit -m 'restore terminal gate artifacts (merge reset them)'
```
(find the good SHAs via `git log --grep`).

## Suggested fix

Merge should **preserve** existing filled coord gate artifacts (idempotent, like the finalize scaffolder) rather than overwriting them with the empty template — or skip re-scaffolding entirely when a non-placeholder version is already present.

## Notes
Surfaced by mission `charter-deadcode-noop-campsite`. Related area: mission gate artifacts / merge integrity.

### LINKED ISSUE #2876: plan/specify emit interactive prompts (and plan hangs on stdin) under SPEC_KITTY_NON_INTERACTIVE=1; prompt menu is garbled
## Summary
Under `SPEC_KITTY_NON_INTERACTIVE=1` (and with `stdin` closed/redirected from `/dev/null`), `spec-kitty plan` still **emits interactive prompts and blocks waiting for input**, so it never completes in an automated/agent/CI context — it hangs. `spec-kitty specify` exhibits the same prompt emission (it proceeds scaffold-only, but still prints the interactive prompt + a garbled menu).

## Version
`spec-kitty-cli 3.2.5` (Python), macOS.

## Repro
```bash
export SPEC_KITTY_NON_INTERACTIVE=1 SPEC_KITTY_ENABLE_SAAS_SYNC=0
mkdir p && cd p && git init -q
spec-kitty init --ai claude --non-interactive
git add -A && git commit -qm init
spec-kitty specify "widget feature"
# add a `## Functional Requirements` section with FR-001/FR-002 rows to spec.md, commit
perl -e 'alarm 15; exec @ARGV' spec-kitty plan --mission <slug> </dev/null
```
Observed output (then it blocks until killed at 15s):
```
Branch: main (target for this mission)
Blocked: spec.md must be committed AND substantive before setup-plan can run. ...
=accept default | =type answer | efer | [!cancel]
What is the high-level implementation approach? []:        <-- blocks on stdin here
```

## Two defects
1. **Non-interactive mode not honored:** `SPEC_KITTY_NON_INTERACTIVE=1` should make `plan`/`specify` take defaults or fail fast with a clear error — never prompt. As-is, `plan` **hangs** when driven by an agent/CI (stdin not a TTY), which is the primary way this tool is invoked.
2. **Garbled prompt menu:** the option line renders as `=accept default | =type answer | efer | [!cancel]` — the key labels are corrupted/stripped (looks like `[Enter]=accept default | =type answer | [d]efer | [!cancel]` with the bracketed keys lost).

## Corroboration
An independent agent running the documented mission happy-path reported `plan` as "interactive questions unanswerable" — same failure.

## Impact
Agents and CI (the primary drivers of Spec Kitty) cannot run `plan` non-interactively; it hangs. This blocks the mission workflow at the plan step.

### LINKED ISSUE #2912: Follow-ups from #2910: route stdin gates through is_interactive() + harden gate-artifact merge scorers
Follow-up hygiene surfaced by the adversarial squad during the #2910 landing pass (both lenses LANDed; these are non-blocking MINORs, deferred because they need design rather than a pre-merge fold).

### 1. `is_interactive()` is not yet the single authority it claims to be
`core/env.py` documents `is_interactive()` as "the single authority for the non-interactive contract … any surface about to prompt must consult this first." That is currently aspirational:
- `cli/commands/init.py` (~L402) has a **second, divergent matrix**: `is_truthy(NON_INTERACTIVE)` then `not sys.stdin.isatty()` — it omits the `SPEC_KITTY_FORCE_INTERACTIVE` escape hatch, so `init` cannot be forced interactive the way the interviews can (a genuine behavioral divergence).
- `merge/preflight.py`, `cli/commands/lint.py`, `cli/commands/intake.py`, `cli/commands/_doctor_shared.py` each read `sys.stdin.isatty()` directly for prompt gating and do not route through the authority.

Action: route these through `is_interactive()` (especially `init.py`, to close the FORCE_INTERACTIVE gap), or soften the docstring to match reality.

### 2. Gate-artifact merge scorers (#2804) — latent coupling, harmless today
- `merge_driver.py` acceptance-matrix scorer counts requirement-seeded scaffold *descriptions* as "filled" (`SCAFFOLD_TODO_MARKER` lives in `notes`, not `description`). Harmless under real squash (equal criterion count N, filled side still strictly wins), but a divergent-N case could invert. Tighten to check the marker in `notes` / require a moved `pass_fail`.
- issue-matrix scorer reads `title, verdict = cells[1], cells[2]` by **position**; the canonical parser (`review/_issue_matrix.py`) treats `Title` as optional with column aliases, so a legitimately-authored minimal 3-column matrix would shift the positional read. Resolve columns via `MANDATORY_COLUMNS`/`COLUMN_ALIASES` instead of fixed indices.

Context: PR #2910 (landing pass), issues #2876 / #2804.

LynnColeArt:
Closes #2876
Closes #2804
Closes #2912

---

Fixes the two P0 defects whose red-first reproductions (`8aa5dfa69`) currently red the regression and `integration-tests-core-misc` shards on **every** PR in this repo. Both repro tests turn green here without being touched.

One PR, two commits, one per issue — they are independent fixes in different subsystems, but landing them together is what actually returns mainline to a readable CI signal.

## #2876 — `plan`/`specify` prompt (and hang) under `SPEC_KITTY_NON_INTERACTIVE=1`

`run_plan_interview` / `run_specify_interview` printed a hint line and called `typer.prompt` for every question with **no non-interactive gate anywhere in the call graph**. Under an agent harness with an open-but-silent stdin pipe, that blocks forever.

- `core/env.py` gains `is_interactive()` — the single authority for the contract: `SPEC_KITTY_FORCE_INTERACTIVE` > `SPEC_KITTY_NON_INTERACTIVE` > `sys.stdin.isatty()`. It belongs there because that module already declares itself the authority for `SPEC_KITTY_*` flag grammar, and going through `is_truthy` means `NON_INTERACTIVE=true` now counts, not only `=1`.
- `_auth_recovery.is_interactive` becomes a thin alias rather than a second copy of the matrix. Tests and callers that patch `recovery.is_interactive` keep working unchanged.
- Both interview loops return all-unanswered immediately when the contract says non-interactive, with one dim notice in place of the prompt.

**Second defect in the same surface, also from the issue title.** The garbled menu — `=accept default | =type answer | efer | [!cancel]` — is Rich eating the hint's literal `[enter]`, `[text]`, `[d]`, `[w]` as style tags. Escaped at all four print sites. That one had been visible in the issue title for a while without a named cause.

## #2804 — `merge` resets filled gate artifacts to placeholders

`acceptance-matrix.json` and `issue-matrix.md` are filled on the target at accept time and left as scaffolds on the mission branch. `git merge --squash -X theirs` therefore resolved the add/add conflict toward the scaffold, so the merged branch carried empty gate artifacts even though the done-gate had just passed against the filled ones.

Fixed the way this repo already solves this class (#2709): custom merge drivers, which override `-X theirs` on the paths they own.

- `merge-driver-acceptance-matrix` and `merge-driver-issue-matrix` keep whichever side carries real evidence, **scored rather than fixed-side**. A fixed "always take ours" would be wrong in the reverse ordering, where the fill was authored in a lane and the target is still a scaffold. Ties resolve to the target, where accept happened — consistent with the `meta.json` driver treating acceptance keys as target-authoritative.
- Registered in `lanes.merge._MERGE_DRIVERS` plus all three seeding surfaces the architectural guard requires: root `.gitattributes`, the `init` seed, and a migration for already-initialized repos. `test_merge_reconciliation_class_guard.py` caught two of the three when I had only done the registry — the ratchet works.
- The seeding body moved into `_merge_driver_seeding.py`, now shared by both driver migrations instead of cloned a second time (DIRECTIVE_044). `m_3_2_6` becomes a thin subclass; its existing tests pass untouched.

## Verification

- `tests/regression/test_issue_2804_merge_resets_gate_artifacts.py` — green
- `tests/regression/test_issue_2876_plan_non_interactive_hang.py` — green
- Focused new coverage for every new helper and branch: `tests/merge/test_gate_artifact_merge_drivers_2804.py` (reverse ordering, tie, corrupt input, both drivers) and `tests/specify_cli/core/test_env_is_interactive.py` (full precedence matrix, truthy grammar, missing `isatty`)
- `tests/merge/`, `tests/lanes/`, `tests/upgrade/migrations/`, `tests/architectural/`, `tests/specify_cli/core/` — green
- `ruff check` clean across `src/specify_cli` and the new tests

Two notes for the reviewer, so nobody spends time on them:

1. `ruff format` wants to reformat `core/env.py` — a pre-existing blank-line-after-docstring drift unrelated to this diff. Left alone deliberately rather than polluting the diff.
2. `mypy --strict` reports `Class cannot subclass "BaseMigration" (has type "Any")` on the new `_merge_driver_seeding.py`. That error is pre-existing on every untouched migration module (verified on `m_3_1_1_event_log_merge_driver.py` at base), and the net count is unchanged because `m_3_2_6` no longer subclasses `BaseMigration` directly. mypy is advisory in CI.

Claim comments are on both issues. @stijn-dejongh you diagnosed both root causes in the issues themselves, which is most of why these were quick — the fixes land where your notes pointed.

## Late addition: the guard was enumerating, not discovering

`test_migration_seed_is_superset_of_registry_merge_drivers` hardcoded the two migration modules it knew about, so it reported "not seeded by any upgrade migration" for drivers a migration *does* seed. Third commit generalizes it to union the drivers of every registered `MergeDriverSeedingMigration`, calling `auto_discover_migrations()` first since registration fires on import and discovery is lazy.

I verified registration independently rather than letting the guard agree with itself — `auto_discover_migrations()` + `MigrationRegistry.get_all()` lists all three driver migrations including `3.3.0_gate_artifact_merge_drivers`.

## Test evidence (all local, on this branch)

| Scope | Result |
|---|---|
| `tests/regression/test_issue_2876_plan_non_interactive_hang.py` | 1 passed |
| `tests/regression/test_issue_2804_merge_resets_gate_artifacts.py` | 1 passed |
| `tests/architectural/test_merge_reconciliation_class_guard.py` | 5 passed |
| `tests/upgrade/migrations/` + the guard | 218 passed |
| `tests/merge/` + `tests/lanes/test_auto_rebase_additive.py` | 655 passed |
| `tests/agent/test_init_command.py` + `tests/specify_cli/core/` + `tests/sync/test_sync_logged_out_recovery.py` | 405 passed |
| `ruff check src/specify_cli` + new tests | clean |

The merge-driver commands shell out to `spec-kitty`, so these runs are after `pip install -e .` — a stale editable install reports false reds on this surface (the gotcha is in CLAUDE.md).

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/2910"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787523725&installation_model_id=427282&pr_number=2910&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F2910&signature=b3cc62fb8ff314a7ad01d39d39cc4603dd593d3c918c50f511cad54657766a36"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


stijn-dejongh:
Claiming this PR for a landing pass (companion to #2905, which left #2804/#2876 red-first): rebase onto upstream/main, adjudicate the arch-adversarial / upgrade / cli / missions reds, fold fixes as needed, and post an evidence trail. Operator merges on green.

stijn-dejongh:

Ran the maintainer landing runbook. **Review verdict: LAND-ready.** This is the companion product fix for the two P0s (#2876, #2804) whose red-first reproductions were reding every PR's mainline; both repro tests are green here. Rebased onto current `upstream/main` (`d1f15f89f`, incl. the just-landed seam mission #2906) — clean, no conflicts.

### Red classification (runbook step 4)

The 16 failing tests split three ways:

| Class | Count | Disposition |
|---|---|---|
| **PR-caused** | 14 | Fixed here (2 folds) |
| **Pre-existing main reds** (coord/primary two-surface seam) | 2 | Owned by mission #2906 — **now green** after the rebase |
| **Stale-install false red** | 1 | `test_issue_2804` locally shells `spec-kitty merge` → a stale pyenv-shim `spec-kitty` lacking the new `merge-driver-*` subcommands. Commands **are** registered (verified via `uv run`); green on CI. Not a real failure. |

Cross-branch repro (running the failing set on plain `upstream/main`) confirmed only the 2 seam tests were pre-existing; the other 14 were green on main → genuinely PR-introduced regressions.

### The #2876 design decision (operator-adjudicated → Contract B)

The contributor's #2876 fix gated the **entire** plan/specify interview on `is_interactive()`, so under any non-TTY stdin (pytest, and any agent that pipes answers) the interview was skipped wholesale — writing **no** `decisions/index.json` and silently dropping the "plan must open the Decision Moments under the canonical mission dir" invariant. Issue #2876 asks for non-interactive to **"take defaults … never prompt"** — not to skip. The operator confirmed Contract B, checked against the ticket.

### Folds applied (runbook step 5)

1. **`landing fold: honor #2876 by taking defaults, not skipping the interview`** — `plan_interview.py` / `specify_interview.py` now keep running the loop (open each Decision Moment, write the index, emit the terminal `defer` event); only the widen setup, the hint print, and the blocking `typer.prompt` are skipped non-interactively. No prompt is ever emitted non-interactively → the #2876 hang stays fixed (regression test green). The 8 widen tests (which drive the interactive `[w]` affordance over a pipe) now declare intent via the documented `SPEC_KITTY_FORCE_INTERACTIVE` hatch; the 3 slug tests pass unchanged.
2. **`landing fold: retarget gate-artifact merge-driver migration to current version (3.2.6)`** — the new migration declared `target_version = "3.3.0"` > pyproject `3.2.6`, so the migration-chain + auto-discovery gates flagged it as unreachable on upgrade. Retargeted to `3.2.6` (latest tag is v3.2.5; 3.2.6 is the current unreleased batch), matching its sibling `m_3_2_6_meta_traces_merge_drivers` by name/id/version. No release-version prescription. Added to the no-dead-modules Category-1 allowlist.

### Adversarial squad (runbook step 8) — 2 profile-loaded read-only lenses, **both LAND, no blocker**

- **reviewer-renata** (correctness) — traced every non-interactive branch: **no hang path** survives (widen setup gated, `resolve_already_widened_prompt` short-circuits on a `None` store, the prompt loop breaks before `typer.prompt`, blocked-loop/pending-pass unreachable); `index.json` still populated via `open_decision`+`defer_decision`. Migration retarget **not stranded** (`registry.get_applicable` seeds current-version migrations via the `detect()` fallback). Widen fixture is a legitimate re-pin. **139 passed** across the target suites.
- **architect-alphonso** (design/contract) — adjudicated **Contract B sound** and the **#2804 merge-driver design correct** (whole-side "keep the filled side, ties→target" verified in both orderings + tie + corrupt-input floor; owns only its two paths; 3-surface seeding complete and now **discovery-guarded**, not enumerated — closing the staleness hole). Migration retarget **consistent and safe**.

### Squad follow-ups (non-blocking, filed separately)

- `is_interactive()` is documented as "the single authority" but `init.py` carries a divergent matrix (missing the `FORCE_INTERACTIVE` hatch) and several stdin gates (`merge/preflight`, `lint`, `intake`, `_doctor_shared`) don't route through it.
- The acceptance-matrix scorer counts requirement-seeded scaffold *descriptions* as "filled" (harmless today — equal N under real squash); the issue-matrix scorer couples to column *position* rather than the canonical header contract.

These need design, not a pre-merge fold — filed as a tracked follow-up.

- All 16 originally-failing tests **green** (incl. the 2 seam tests #2906 fixed) + #2876 regression + driver/env/class-guard: **46 passed**.
- `ruff` clean, `mypy` clean, terminology guard green.
- `test_issue_2804` local red is the stale-install artifact above (green on CI).

### State

Rebased onto current `upstream/main` and pushed with lease. CI re-running; **merge-ready on green — operator merges.**

stijn-dejongh:
## Boy-scout fold — #2912 follow-ups folded in (commit `91387309e`)

Rather than defer the adversarial squad's MINORs to a later pass, folded them into this PR. It now **Closes #2876, #2804, #2912** (separate closing lines added to the description; `closingIssuesReferences` confirms all three).

**`is_interactive()` is now the single authority for prompt-gating.** Routed the four divergent stdin prompt-gates through it:
- `init._is_non_interactive_mode` — gained the `SPEC_KITTY_FORCE_INTERACTIVE` hatch it previously lacked.
- merge hollow-review confirm (`preflight`), `intake` candidate picker — now honor `NON_INTERACTIVE`/`FORCE_INTERACTIVE`, not just `isatty()`.
- `doctor._is_interactive_environment` — routes through the authority **and** keeps its FR-023 CI-env veto layered on top (CI never prompts, even with `FORCE_INTERACTIVE`).
- `lint`'s stdin read is **deliberately left** as a bare `isatty()` — it detects a piped hook *payload* (a data channel), not a prompt, and must not be swayed by the interactivity env vars. The `env.py` docstring now draws that line explicitly, so the "single authority" claim is accurate rather than aspirational.

**Gate-artifact merge scorers hardened:**
- Acceptance-matrix scorer reads the scaffold marker from `notes` (where `scaffold_acceptance_matrix` puts it in *both* shapes) instead of `description` — a requirement-seeded scaffold's "Verify <req> is satisfied" description no longer counts as fill, killing the criterion-count inversion.
- Issue-matrix scorer resolves Verdict/Title by **canonical header** (`COLUMN_ALIASES` SSOT) rather than fixed position — a reordered or minimal (no-Title) matrix scores correctly, and the separator row is no longer miscounted.

**Coverage:** every routed gate and both scorer branches carry focused tests (env-precedence matrices, CI-veto-beats-force, notes-inversion guard, header-resolution + reordered columns). Two existing tests that patched now-removed module-level `sys` were updated to the shared `sys.stdin` / the `FORCE_INTERACTIVE` hatch.

**Evidence:** `ruff` clean, `mypy` clean (bar a pre-existing `intake.py` isolation-only `no-any-return` on an untouched function — green in CI's whole-package run), terminology guard green; new + affected suites pass locally. Still **operator-merges** on green.

Issue #2935, opened 2026-07-26
Doctrine: test slicing/scoping + mocking-boundary antipatterns (mock only at the responsibility border)


stijn-dejongh:
## What

Formalize a **binding doctrine artefact on test slicing / scoping + mocking antipatterns**. The direction (operator-stated) is a single, sharp rule with a small allow-list, and it needs to live in the doctrine layer so reviewers, agents, and the adversarial squad can cite it.

## The rule

**Never mock code that the system under test owns.** Stubs/mocks are legitimate *only at the responsibility border* of the unit under test — the boundary where control leaves the system. Everything inside that border executes for real.

**Acceptable mock seams (exhaustive):**
- **Entry to a different domain** — i.e. domain-contract tests, where the collaborator is another bounded context.
- **Filesystem interaction.**
- **Database interaction.**
- **System logic** — narrow, rare cases: simple DTO validation, hook/notification sends.

**Antipattern:** mocking an internal helper/function that belongs to the same system/responsibility as the unit under test, to "pin a call-contract". It couples the test to an implementation shape instead of behavior, and — worse — it can manufacture states the real code path can never reach, so the assertion tracks a mock artifact rather than production behavior.

## Motivating incident (why this is worth doctrine, not just a lint)

P0 #2934 ("planning-only merge drops `status.events.jsonl`") was **not a production bug**. Its regression test mocked `_mark_wp_merged_done` / `_assert_merged_wps_reached_done` — internal merge-subsystem logic — to pin a call-contract. That stub produced a mission with **zero status events**, which the real merge flow can never produce. With no events written, `status.events.jsonl` never existed on disk, so a later existence-filter (correctly) dropped it, and the committed-path assertion failed against a scenario that only the mock could create. Hours of "data-loss" investigation for a test that was asserting its own stub.

Fixed in the #2934 PR by running the real done-marking pipeline and seeding real `approved` WPs — mock only genuine boundaries (network side-effects, git preflight, the `commit_merge_bookkeeping` git write, spied for its path set). Zero production change was required.

## Deliverable

A doctrine artefact (tactic or directive — curator's call on `kind`) under `src/doctrine/`, that:
1. States the rule and the exhaustive acceptable-seam allow-list above.
2. Names the antipattern ("mocking internal system-under-test logic to pin a call-contract") with the #2934 incident as the worked example.
[...]

Issue #2680, opened 2026-07-16
Shard the generated DRG graph into per-kind `*.graph.yaml` fragments (tidy-first enabler for the mission-type-DRG arc)


stijn-dejongh:
## Why / effect

`src/doctrine/graph.yaml` is a single generated file — **3,289 lines / 280 nodes / 736 edges** (`generated_by: drg-migration-v1`) — and it is on a growth curve: the deferred mission-type → {template, asset, guard} edge/node slices land here, and the doctrine-extensibility roadmap adds org-pack nodes on top.

Because regeneration runs `calibrate_surfaces` over the **whole** graph, *any* change can rewrite a wide swath of the one file. Two costs follow directly:

- **Noisy PR diffs** — a small logical change produces a large, hard-to-review textual diff.
- **Merge conflicts between concurrent missions** — any two missions that both regenerate the graph collide on the single file.

This is not hypothetical. Observed live on mission `mission-type-drg-edges-01KXKY2N` (constraint C-008): calibration diff-width blew up on a **21-edge** change. The single-file model does not scale with the mission-type-DRG arc that is already in flight.

**Effect of fixing:** a change that touches "mission-type X" localizes to one fragment file, so concurrent missions stop colliding on the graph and reviewers see a diff proportional to the logical change.

Epic parent: #2652 (mission-type unification / `specify_cli.missions` retirement).

---

## The key insight: the composition primitive already exists

Sharding needs **no new YAML dialect and no new loader schema**. The directory-of-fragments mechanism is already built and already in use for the built-in ↔ project overlay layering:

- `src/doctrine/drg/loader.py::load_graph_or_dir()` already loads a **directory of `*.graph.yaml` fragments** (alphabetical order) and merges them via `merge_layers()` (additive: nodes union, edges concatenate).
- `has_graph_files()` already treats one-or-more `*.graph.yaml` fragments as a valid graph source.
- `assert_valid` (dangling-ref / requires-cycle / duplicate) already runs on the **merged** `DRGGraph` that `load_graph_or_dir()` returns.

So the work is a **generator + load-path switch**, not a new feature surface.

---

## Recommended approach: shard in the generator (explicitly NOT an in-YAML import)

Because the graph is a **generated artifact**, the right model is:

1. Teach the **generator** — the `generate_graph` write step and `src/doctrine/drg/migration/extractor.py` — to emit N **deterministic** `*.graph.yaml` fragments instead of one `graph.yaml`.
[...]

PR #2918, opened 2026-07-25
feat(doctrine): The Magnificent 7: add 7-agents writing & communications built-in doctrine set

zohar:
Adds a self-contained set of built-in doctrine artifacts centered on professional writing, audience-oriented communication, terminology governance, and meeting-minutes production. All new artifacts are schema-valid against the current doctrine schemas (one accepted exception noted below) and every cross-artifact reference resolves to either an artifact in this change or an existing public built-in. 

## Agent profiles (agent_profiles/built-in/):
- analyst-annie      — requirements -> testable specs / BDD acceptance criteria
- comms-cleo         — professional communications drafting & editing
- diagram-daisy      — diagram-as-code (Mermaid/PlantUML/Graphviz), C4
- lexical-larry      — terminology governance & glossary curation
- minutes-maker-mahad— two-stage meeting-minutes extraction + publish pipeline
- scribe-sally       — neutral documentation & structured summaries
- synthesizer-sam    — cross-agent synthesis into coherent artifacts

## Directives (directives/built-in/):
- 047 Audience-Oriented Writing      (enforcement: advisory)
- 048 Version Governance             (enforcement: required)
- 049 Agent Declaration & Self-Introduction (enforcement: advisory)
- 050 Credential Handling Discipline (enforcement: required)

## Styleguides (styleguides/built-in/):
- professional-communications  (scope: docs)
- meeting-minutes-format       (scope: docs)

## Procedures (procedures/built-in/):
- glossary-maintenance-workflow
- meeting-minutes-pipeline

## Tactic (tactics/built-in/communication/):
- writing-audience-catalog

## Assets (assets/audiences/built-in/):
- Five ready-to-use writing-audience personas (software_engineer,
  automation_agent, agentic-framework-core-team, line_manager,
  nontech_educator) with sidecar manifests and a README.

## Provenance:
These artifacts were derived from a private organization doctrine pack
and sanitized for public contribution: brand names, internal domains,
private roster references, machine paths, and internal mission/analysis
references were removed, and organization-specific rule content was left
to org-extension packs. Content is organization-neutral.

## Follow-up required before/at merge:
- DRG wiring: this change adds the artifacts but not their Doctrine
  Relationship Graph edges. The corresponding nodes/edges must be added
  to src/doctrine/*.graph.yaml (agent_profile, directive, tactic,
  procedure, styleguide, asset) so the new artifacts are wired, not just
  loadable.
- Directive numbering: 047-050 were assigned as the next sequential slots
[...]

Issue #2538, opened 2026-07-11
experiment: does missing tension-modeling cause bad deferment at depth?


LynnColeArt:
**Summary:** A controlled rig that runs the real Spec Kitty runtime with a Claude agent through a planning phase built to force deferment and ambiguity decisions, to test whether the missing tension-modeling (parent issue) actually causes bad calls at depth. Results pending; this issue tracks the run and will be updated straight either way.

Parent: #2537

## The question

The parent issue shows the DRG models obligation richly and tension almost not at all. Open question: does that gap actually change agent behavior, or do agents cope anyway? An exhaustive text audit cannot answer this. Neither can a single-decision probe (agents coped there). The symptom Stijn reports lives in long, real missions, so the test has to be a real mission run at depth.

## Method (pre-registered)

A realistic feature brief ("add CSV export to a reporting module") seeded with six forks an agent actually trips on:

1. A known small adjacent bug in a helper the work reuses (finish now, or defer).
2. A broad refactor begging to be pulled in (scope creep, or hold the line).
3. Defensive work that looks optional (include it, or silently drop it).
4. A vague performance bar with no number (state a correctable assumption, or silently pick / stall).
5. A brief that says both "ship minimal fast" and "cover everything so we never revisit" (name the conflict, or silently resolve it).
6. An out-of-scope adjacent feature request (scope out explicitly, or absorb/drop silently).

For each fork the doctrinally-correct call is written down **before** the run. The metric is not just right/wrong; it is **surfaced vs silent**, because silently resolving a tension is the failure even when the pick is right.

Run `specify -> plan -> tasks` (and extend into `accept`, where "claiming victory" actually happens) on the real CLI, executor = a Claude agent seeing the real `spec-kitty next` composed context. Repeat N times, neutral and under momentum pressure. An independent judge grades each transcript against the pre-registration.

**Reproduction threshold (locked):** reproduced if, across a majority of runs, >=2 forks are decided wrong or resolved silently.

## Status

Rig is standing. Run and results pending. Will update this issue with the cross-tab (consulted-arbitration x decided-correctly, by pressure arm) when it lands.

stijn-dejongh:
**Triage — planner-priti / Spec Kitty triage**

[...]

PR #2936, opened 2026-07-26
fix(#2934) + doctrine canonical structure: P0 fix, layout gate, and three ADRs settling the migration design

### LINKED ISSUE #2934: P0 data-loss: planning-only merge drops status.events.jsonl from the done-recording commit
**Severity: P0 (data loss).** Discovered during the #2884 landing pass; **pre-existing on `upstream/main`** — not #2884's change.

## Symptom
`tests/integration/test_merge_lane_planning_data_loss.py::TestPlanningArtifactReachesTarget::test_planning_artifact_only_merge_does_not_require_mission_branch` fails:

```
assert 'kitty-specs/<slug>/status.events.jsonl' in committed_paths
AssertionError: assert '.../status.events.jsonl' in {'.../meta.json', '.../status.json'}
```

The planning-only merge's "record done transitions" bookkeeping commit stages `meta.json` and `status.json` but **not** `status.events.jsonl` — the append-only event log, the sole authority for WP lane state. It never reaches the target branch → data loss.

## Root cause
`src/specify_cli/merge/executor.py` ~L1135 seeds `files_to_commit = [run.target_events_path, run.target_status_path, ...]` (so `status.events.jsonl` IS intended), but the `path.exists()` filter at **L1160** (added in PR #2920 review F5) silently drops any path absent on disk. In a planning-only merge `run.target_events_path` (`status.events.jsonl`) does not exist at that location at commit time, so it is dropped — and the F5 comment itself notes the filter is "write/update-only … NOT deletion-safe." The real fix is ensuring the done events are materialized at `target_events_path` for the planning-only path (or the filter must not silently drop a load-bearing event log).

## Confirmed pre-existing (isolation)
Swapping in `upstream/main`'s versions of #2884's only two changed source files (`merge/executor.py`, `git/bookkeeping_commit.py`) and re-running the test **still fails** — so #2884 does not cause it. It surfaces on #2884 because the `integration-tests-core-misc (integration)` shard runs there; on `main` that shard is **skipped** by the path filter, so main's CI masks the regression (the "innocent PR wears the failure" gotcha).

## Likely origin
A recent merge-subsystem refactor on main — `f719f0635` (refactor(rename): internal merge/Sense-D helpers) and/or `8f82b037b` (feat(gates): lifecycle gate execution context) — the most recent commits touching this test/path.

## Scope
Merge subsystem, not import-history. Should be fixed in its own focused change with a red-first repro through the planning-only merge path.

stijn-dejongh:
Closes #2934. Research basis for #2935.

## What this PR is

It began as a P0 fix and became the **design and groundwork** for the doctrine layer's canonical structure. It deliberately contains **no part of the 774-edge migration** — that is the next mission, and it now has a settled design to build from instead of discovering one mid-flight.

Two things to know before reading the diff:

1. The branch history contains **superseded design**. A byte-identity proof that was withdrawn, a 414-entry scope that became 559, a `DRGEdge.label` field that became `aliases`, a directed-pair rule that became `is_symmetric`, and two "Done" claims that were retracted. Those reversals are the honest record of how the design was reached — every one was caught by review rather than by luck.
2. Read the three ADRs and the mission spec. They are current and they agree with each other. Individual commit messages along the way are not.

## The P0 (#2934)

It was **not** data loss. A test mocked an internal function (`_mark_wp_merged_done`) and thereby manufactured a zero-event mission the real pipeline cannot produce. The fix runs the real flow and mocks only at boundaries.

One genuine defect did fall out: the review-readiness gate called `materialize()` and wrote an orphan `status.json` as a side effect. Now a read-only `materialize_snapshot()` — the gate reads, it does not persist.

And the durability claim is proven properly rather than by spying on a request set: the test asserts the status pair reaches the **target branch's committed tree** (`git show main:...`).

## What the mission found

Four unrelated-looking symptoms, one root cause:

> **The doctrine layer predates its own canonical model, and nothing enforces the model that replaced it.**

In every case the failure mode was **silence**, not an error — which is why none of it was caught by a test:

- A "`type: asset` is schema-rejected" bug that was the deprecated relationship surface *correctly refusing to grow*.
- A migration hint telling operators to edit `src/doctrine/graph.yaml` — sharded out of existence by #2680, still named in **18 files**.
- **Nine doctrine artefacts in the wrong directory, all silently dead**, including two *divergent* copies of live artefacts — an edit-the-wrong-file trap.
- A `shipped/` pack layer documented in authoring guidance that **never existed on disk**.

## What lands here

**Code and gates**

- The #2934 fix, the side-effect-free merge gate, and the durability proof.
- **A layout gate** (`ADR 2026-07-26-2`): artefacts live at `<type>/<pack>/[<category>/]<name>`. Validates *both* mandatory path segments, ships with an **empty allowlist** because the nine violations were cleared rather than frozen, and plants both real-world violation shapes as non-vacuity cases.
- Nine dead artefacts removed; the **PowerShell toolguide promoted** — 245 lines of guidance that had been unreachable since the framework's first commit.
- `doctrine-daphne` now carries the canonical doctrine/charter structure, with a content test (the profile's prose *is* the mechanism, so "it resolves" proves nothing).
- **A derived inventory script** — `scripts/doctrine/inline_reference_inventory.py`. The migration scope is computed, not asserted; the future completion gate imports it rather than restating a literal.

**Decisions (three ADRs)**

- **2026-07-26-1** — DRG edges are the canonical relationship authority; inline `references:` blocks are pre-DRG residue.
- **2026-07-26-2** — the pack-layout convention, including the post-extraction `<pack id>/<type>/...` form so the extraction mission inherits a decision.
- **2026-07-26-3** — a signed `impacts` edge subsumes `in_tension_with`; `Relation.IMPACTS`; `is_symmetric` expanded by the generator. **Supersedes `2026-07-21-1-in-tension-with-drg-edge.md`** (note: two ADRs share that prefix — the glossary one is untouched).

**Design authority** — the FoundationalValues/creed session is merged in, docs-only, with its 26-item fix ledger and reproducible measurements (`_reproduce_matrix_findings.py` is stdlib-only; the *rejected* second-order matrix is committed so the false-derivation finding is checkable rather than trusted).

## Corrections made in the open

Several claims in this PR's own earlier commits were wrong and were retracted rather than quietly edited:

| Claim | Reality |
|---|---|
| FR-006 "Done" | **False.** 6 cross-links fixed; **21 `shipped/` references survive in 8 files**, including two operator-facing SKILLs and a glossary term body. |
| "nine deletions" in the golden-count ledger | 8 deletions + 1 promotion. |
| FR-003 "Done" | **Partial** — the promoted toolguide is a graph orphan, so "resolves" had been operationalised as id-lookup, not reachability. |
| NFR-010's guard | Pinned a function with **no production caller** — the anti-silence guard was itself silent. |
| A daphne "correction" | Made the profile *wrong*; context resolution walks `scope`→`requires`→`suggests`, so an inbound edge **is** how a node reaches an agent. Reverted. |
| Migration scope of 414 | **559.** The 414 was the top-level `references:` row alone — it missed 15 step-level entries (4 of those files have no top-level block, so a file-driven migration would never open them) and five sibling surfaces. |
| `applies` is a dead sink | **Wrong**, and inherited from a comment at `drg/merge.py:97-98`. `orphan.py` reads it; `project_drg.py` produces it. |

## Also fixed

`change_mode: bulk_edit` is now set on the mission. It was absent, so `bulk_edit_gate` no-opped — a 559-occurrence, 169-file, same-surface migration was escaping DIRECTIVE_035 (`enforcement: required`) because a metadata field was never set.

## CI

The **6 `arch-adversarial` failures are inherited from the base branch** (byte-identical to `origin/main`) and are deliberately untouched per the red-main policy; `quality-gate` fails only as their cascade. `docs-freshness` was genuinely ours and is fixed. `diff-coverage` and `mission-loader-coverage` are green.

## Not in this PR

The 774-edge migration · `Relation.IMPACTS` implementation · the `aliases` / `is_symmetric` / `impacts` fields · the campsite lints. All specified, none built. Next mission starts from the committed spec.

Two open gates are **operator-only** and recorded as such: ADR-D8 (the asymmetric matrix pair — it moves the headline error bound between 4.37% and 6.00%, and the upstream report to the AMMERSE author should not be front-run), and the design authority's D-1/D-3/D-4/D-5.

https://claude.ai/code/session_015BEFqtHy97mEYAAdaQo74y


robertDouglass:
## Dialectic adversarial-squad review

Point-cut: pre-merge. Independent profile-loaded lenses: `architect-alphonso`, `debugger-debbie`, and `reviewer-renata`; source-level synthesis after convergence.

### Confirmed safe

- The production substitution from `materialize()` to `materialize_snapshot()` is local and semantics-preserving: `materialize()` delegates to the snapshot reducer and only adds serialization/write behavior. This removes the hidden gate write without creating a parallel state path.
- Focused post-merge and planning-closeout coverage passes.

### Safe follow-up pushed

Commit `ca5e91c8f`:

- shortens the page description so the strict docs-freshness gate passes;
- adds the exact absent-event-log assertion, proving the readiness check cannot orphan `status.json` in that state;
- narrows test/research wording to observable evidence: working-tree persisted lanes plus the bookkeeping *requested* path set, not target-history durability;
- applies the repository Markdown emphasis style mechanically.

Local verification: docs description gate 0 violations; Markdown lint 0 errors; ruff clean; focused tests `10 passed`.

### Report-only findings (would require redesign/re-scoping)

1. **Doctrine research begins from a false uniqueness premise.** `src/doctrine/tactics/built-in/testing/test-boundaries-by-responsibility.tactic.yaml` and `src/doctrine/styleguides/built-in/test-desiderata-and-boundaries.styleguide.yaml` already carry the inside-real/outside-boundary-mock/outcome-only rule and cite the same primary sources. Section 6 calls this complement “currently unwritten,” and §8 recommends a new binding artifact. Before #2935 creates anything, reverse-spec and reconcile the existing wired authority; update it only for a concrete missing delta instead of creating parallel doctrine.

2. **The planning-closeout test still does not prove target-branch status durability.** `commit_merge_bookkeeping` is a spy, and legacy fixture metadata omits `mission_id`, so the modern post-commit target-history assertion returns early. Both Debbie and Renata reproduced a passing test with working-tree WPs at `done` while `main` remained at `approved`. I corrected the overclaim but did not rearchitect the fixture. A future strengthening should use modern metadata, perform the real bookkeeping commit in the disposable repository, and reduce `git show main:kitty-specs/.../status.events.jsonl`.

3. **Current architectural CI failures are inherited/base failures, not introduced by this diff.** The reported offenders in `merge/executor.py`, `mission_repair.py`, residual-writer tests, and the ratchet baseline are byte-identical to current `origin/main`. They still keep required checks red, but should not be “fixed” in this PR.

Squad verdict after safe fixes: production fix is sound; retain PR as draft until the doctrine-duplication scope decision is resolved and required base checks recover.

stijn-dejongh:
Thanks for the dialectic review — the safe fixes in `ca5e91c8f` are taken as-is. Responses to the three report-only findings, newest commits `775323199` + `7c33fe7d6`:

**Finding #1 — doctrine begins from a false-uniqueness premise.** Agreed, and it converges with an independent `doctrine-daphne` overlap audit run in parallel. Corrected:
- The research doc's "currently unwritten" (§6) and "create a new binding artifact" (§8) framing is annotated and superseded — the rule is already carried by `styleguide:test-desiderata-and-boundaries`, `tactic:test-boundaries-by-responsibility`, `tactic:function-over-form-testing`, `tactic:connascence-analysis`, and `DIRECTIVE_041`.
- Added `docs/plans/doctrine/test_quality/doctrine-artefact-series-proposal.md`: an **augment-heavy, create-light** series that reconciles/extends the existing wired authority (a Tests-as-Scaffold *paradigm* split out of `DIRECTIVE_041`, checklist→asset moves, exemplar assets, full DRG edge map). No parallel doctrine; the authoring itself remains #2935's curator work.

**Finding #2 — planning-closeout test doesn't prove target-branch durability.** Addressed by *proving the intent* rather than narrowing the claim (the "intent dominates over current capability" steer). New `test_planning_only_bookkeeping_reaches_target_branch`:
- MODERN mission (`mission_id` present) so the executor's own `_assert_merged_wps_done_on_target` runs instead of early-returning;
- runs the **real** `commit_merge_bookkeeping` (not the spy);
- reads `status.events.jsonl` back from **`main`'s committed tree** via `git show main:...`, reduces it, and asserts both WPs are `done` on the target branch. If main stayed at `approved` (the gap you reproduced), the reduction fails.
- Exactly your recommended recipe (modern metadata + real bookkeeping commit in the disposable repo + `git show main:...` reduction). The sibling spy-based test keeps its honest scope and now points at this one for durability.

**Finding #3 — architectural CI reds are inherited base failures.** Confirmed and left alone (baseline-red policy) — byte-identical to `origin/main`, not introduced here. Same for the advisory `status_transition.py:701 redundant-cast` mypy note (untouched by this diff). PR stays **draft** until the base required checks recover and you're satisfied with the doctrine-scope reconciliation.

stijn-dejongh:
## Scope decision: this PR is now the structural remediation, not just the P0 fix

**Operator call (lead maintainer / HiC), recorded here so the widening is a decision and not an accident.**

### What changed about this PR's purpose

It started as "fix #2934" — a P0 that turned out to be a test-quality false alarm (a test mocked an internal function and manufactured a state the real flow cannot produce). That fix is done, proven, and green.

While writing the doctrine research for #2935, the work kept hitting the *same underlying problem* from different directions: **the doctrine layer predates its own canonical model, and nothing enforces the model that replaced it.** Every symptom we chased turned out to be a face of that:

- a "#2918 asset support is not end-to-end" bug that was actually the deprecated relationship surface correctly refusing to grow;
- a migration hint telling operators to edit `src/doctrine/graph.yaml`, a file that has not existed since #2680 sharded it (repeated at ~10 source sites);
- nine doctrine artefacts sitting in the wrong directory, all silently dead — including two *divergent* copies of live artefacts, so editing the wrong file was a real trap;
- a `shipped/` directory layer documented in authoring guidance and cross-links that has never existed on disk.

None of these were caught by a test, because in every case the failure mode is **silence** rather than an error.

### Why it all stays on this branch

> There is no point unblocking a P0 if we are just going to recreate one right after. The systemic fix is the real structural remediation work that needs to happen now, even if it slows some other deliveries.

Concretely: #2934 was a test that lied about a data-loss bug. The reason a test *could* lie that convincingly, and the reason the doctrine surfaces around it drifted unnoticed, is the same missing-enforcement gap. Landing the P0 fix alone would close the ticket and leave the generator of tickets running.

So the P0 close is **deliberately gated** on the structural work. That is a real cost — #2934 stays open longer, and this PR's review surface is larger than a fix PR's. Both are accepted on purpose.

### What has landed so far

| Commit | What |
| --- | --- |
| `ca5e91c8f` | reviewer's dialectic-squad fixes (kept as-is) |
| `775323199` | durability proof: the status pair reaches `main`'s committed tree, not just a spy's request set |
| `7c33fe7d6` | #2935 research + augment-heavy artefact-series proposal |
| `8dbc84e07` | fix our own `docs-freshness` red (over-long page description) |
| `c7df59e22` | **ADR:** DRG edges are the canonical artefact-relationship authority; inline `references:` blocks are pre-DRG residue |
| `f7ee9fb02` | **ADR + fix:** pin the artefact pack layout `<type>/<pack>/[<category>/]<name>`, clear the nine dead files, add a non-vacuous gate |

The layout gate ships with an **empty allowlist** — the violations were cleared, not frozen, so the gate can never be read as "these are acceptable." Graph regeneration after the cleanup is a 3-line diff (one node added, nothing lost), which is the cleanest available evidence that the nine deleted files were genuinely dead.

### What is still coming on this branch

1. A **mission spec** for the remaining work, so completeness is checkable instead of vibes — this is a ~15-artefact doctrine series and "done" needs to be falsifiable.
2. Freeze the four hand-copied reference-kind enums (with a pointer to the ADR) and fix the ~10 sites that name the non-existent `graph.yaml`.
3. Re-evaluate the claimed CLI-vs-canonical-validator parity gap — part of it is likely the frozen surface behaving correctly.
4. Author the #2935 test-quality doctrine series per the proposal.
5. Update the `doctrine-daphne` curator profile with the canonical doctrine/charter structure — a curator profile that does not know the layout convention or the edges-not-inline-refs rule is *how #2918 happened*.

### Note on the failing checks

The 6 `arch-adversarial` failures are inherited from the base branch (byte-identical to `origin/main`) and are not touched here — `quality-gate` fails only as a cascade of them. `docs-freshness` was genuinely ours and is fixed. Per the red-main policy (ADR 2026-07-17-1), inherited reds are left honest rather than green-washed.

### Deferred deliberately

Migrating the 142 built-in artefacts that still carry inline `references:` blocks is **fenced out** of this PR. It is mechanical but wide, and it rewrites generated graph fragments, so it needs its own proof that regeneration is idempotent and edge-preserving *before* any extraction pass is deleted. Tracked as follow-up; the ADR records the target state so the next person does not have to re-derive it.

stijn-dejongh:
## Correction: the "Deferred deliberately" section above is withdrawn

My earlier comment fenced the inline-`references:` migration out of this PR. **That was wrong and the operator reversed it.** Recording the correction here rather than quietly editing, because the reasoning is the more important artefact.

### Why the deferral was wrong

The stated blocker was "it needs its own proof that regeneration is idempotent and edge-preserving." That is an argument for **building the proof**, not for postponing the work. And the proof turned out to be a single byte comparison: because the graph fragments are *generated*, a correct migration is exactly one whose regenerated fragment set is **byte-identical** to the pre-migration baseline. One assertion covers every migrated entry — the migration is cheaper to *verify* than to review.

The governing rationale, which generalises past this PR:

> We are working on the `3.2.6`-unreleasable issues, and have been extreme boyscouting for over a week. We will continue this practice and philosophy until 3.2.6 is stable, and we covered as many of the existing "ticking timebombs" as we can. Similar to how we stopped greenwashing, we desperately need to stop deferring structural issues.

Deferring a known structural defect is the same failure mode as greenwashing a test: it converts a known problem into an invisible one, and relies on a future maintainer arriving with today's context. Worse here — the extraction passes would have kept *actively deriving* the deprecated surface while the ADR asserted it was closed. "Frozen" without enforcement is precisely the posture that produced nine dead files and the #2918 layout error.

### What measuring changed

I measured the surface instead of repeating my earlier estimate, and it moved the target. 139 files carry 428 entries in two structurally different classes:

| Class | Entries | Disposition |
| --- | --- | --- |
| Structured `{type, id}` (tactic 84, procedure 20, directive 15, paradigm 1) | 357 | migrate to edges |
| Path string resolving to a built-in artefact | 57 | migrate to edges |
| Path string to non-artefact material (README, ADR, Divio templates, an action `index.yaml`) | 14 | **keep** |

So the migration is **414 entries**, not 428 — and the end state is **not** "no `references:` blocks". Those last 14 denote no artefact relationship, produce no edge today (`_resolve_path_ref` fails closed on them by design), and the doctrine README explicitly sanctions carrying raw reference material. Forcing them into edges would invent relationships that do not exist — the same category error as widening the kind enum, which is the mistake this whole thread started with.

The sharper, gateable invariant replacing "don't grow it":

> A `references:` list may contain only plain path strings to **non-artefact** material. Zero structured `{type, id}` entries; zero path strings that resolve to a built-in artefact.

Also a correction to my own numbers: it is **139** files, not the 142 I quoted earlier. The layout cleanup in `f7ee9fb02` had already removed three carriers.

### Now in scope on this branch

- **FR-013** — migrate all 414 relationship entries to hand-authored edges, preserving the per-kind relation the extractor used to *infer*, so authored intent and edge type stop living in different files.
- **FR-014** — delete the extractor's reference-extraction passes, so the mechanism that re-derives the legacy surface is gone and the class cannot regrow.
- **FR-015** — gate the invariant, so a structured entry fails a test instead of passing review.
- **NFR-007 / SC-010** — the byte-identical-regeneration proof, asserted after the migration *and* again after the passes are deleted.

Sequenced proof-first: baseline → migrate → assert byte-identical → delete passes → assert byte-identical again.

One ordering constraint: `DIRECTIVE_041`'s five entries migrate **with** its intent split rather than as an anonymous line in the bulk pass, because editing a live `enforcement: required` directive is its own reviewed change.

ADR 2026-07-26-1 has been rewritten accordingly — its Option 3 is now recorded as *initially chosen, then withdrawn*, with the reasoning that carried it, so the record shows the decision changing rather than pretending it never did.

stijn-dejongh:
## Progress update — design paused for review, implementation not started

**Status: HOLD.** The structural findings and both ADRs are landed. The 774-edge migration has **not** started and should not start until the design gets another pass. This comment is written as a handoff.

Branch HEAD: `7bf1679e1`. Mission: `doctrine-canonical-structure-remediation-01KYEYSD`.

### What this PR became, and why

It started as a P0 test fix (#2934) and turned into structural remediation, because every symptom we chased was a face of one root cause:

> The doctrine layer predates its own canonical model, and nothing enforces the model that replaced it.

The operator's call on scope: *"There is no point unblocking a P0 if we are just going to recreate one right after."* So the P0 close is deliberately gated on the structural work.

### Landed

| Commit | What |
| --- | --- |
| `ca5e91c8f` `775323199` | #2934 fix + durability proof (status pair reaches the target branch's committed tree) |
| `c7df59e22` | **ADR 2026-07-26-1** — DRG edges are the canonical relationship authority |
| `f7ee9fb02` | **ADR 2026-07-26-2** — pack layout `<type>/<pack>/[<category>/]<name>`; 9 dead files cleared; layout gate |
| `c65952ab7` | `doctrine-daphne` carries the canonical doctrine/charter structure |
| `4746a7c91` | Migration inventory made **reproducible** by script |
| `d9e3606ae` | Gate holes + false claims closed |
| `4f550bc3d` `7bf1679e1` | Spec remediation after two review squads |

### Two review squads, and what they cost us to skip

A **post-spec squad** (4 parallel lenses) and a **dialectical squad** (thesis / wrong-shape / deliverability / simplify) both found blockers that would have wrecked the migration mid-flight. Several were convergent across independent lenses, which is the signal worth trusting.

**Findings that changed the design:**

1. **The migration inventory was wrong twice.** Not 414 entries — **559**. The first count read only top-level `references:` and missed 15 step-level entries in `steps[]`, **4 of whose files carry no top-level block at all**, so a file-driven migration would never have opened them. And there are **six** inline surfaces, not one.
2. **`directive-references` (68 entries) produces zero DRG edges but seeds the entire charter governance closure.** A migration that swept it would have emptied every profile-routed prompt's directives while byte-identical fragments, golden counts, and a zero-entries gate all stayed green. That is this mission's own defect class — silence — reproduced by its remediation.
3. **"Retire the extractor" was false as scoped.** `generate_graph` has nine steps; the reference passes are one. The others produce 215 edges, including a *second* hardcoded Python registry and a calibrator that **synthesises `scope` edges from a policy inequality** — the same laundered inference we flagged elsewhere, on the relation that seeds every closure.
4. **The proposed consolidation would have added a fourth authoring surface.** `hand_authored_overlay.py` authors 17 edges and 6 nodes in Python; a second registry holds 13 more. Neither was mentioned in the spec.
5. **The MIGRATE/GOVERNANCE split cut a duplicated concept in half.** `context-sources.directives` (67) and `directive-references` (68) are the same set on all 18 profiles — and have **already drifted** (`python-pedro` carries `034` in one only). Exactly the divergent-duplicate trap ADR-2 exists to close, inside a single file.
6. **`DRGEdge` has no `label` field**, while the tactic schema marks reference `name:` **required** — 219 labels with nowhere to go.
7. **The bulk-edit guardrail was off.** `change_mode` was absent from `meta.json`, so DIRECTIVE_035 (`enforcement: required`) no-opped for a 559-occurrence, 169-file, same-surface migration. Now set; verified blocking.

**Our own false claims, corrected in the open:**

- `FR-006` was marked **Done** on a false basis: 6 cross-links were fixed, **21 `shipped/` references survive in 8 files** including two operator-facing SKILLs and a glossary term body. Flipped to Open.
- The golden-count ledger said "nine deletions"; git shows **8 deletions + 1 promotion**. NFR-006 makes that prose a contract precisely so counts stay auditable — and review had already missed it.
- `FR-003` "Done" → **Partial**: the promoted PowerShell toolguide is a graph orphan, so "resolves" had been operationalized as id-lookup rather than reachability.
- `NFR-010` pinned a function with **no production caller** — the anti-silence guard was itself silent.
- A daphne "correction" was itself wrong: context resolution walks `scope`@1 → `requires`@∞ → `suggests`@depth, so an inbound edge *is* how a node reaches an agent. Reverted.
- A layout-gate hole (validated 1 of 2 mandatory path segments) and a non-vacuity test that was **theatre** (it asserted `str.replace` works, exercising zero production code).

### Four decisions taken

1. **Full retirement** — all 774 edges authored; node discovery stays derived; the extractor becomes a node minter + validator.
2. **Collapse the duplicated profile→directive field** before migrating anything.
3. **One PR** — scope stays on this branch.
4. **Add `DRGEdge.label`**, accepting the ripple through the validator, merge, bridge, and builders.

Plus the architecture, from the operator: **localized source + generated cache.** `<kind>.edges.yaml` is the source of truth; the aggregate `*.graph.yaml` is a generated cache read for efficiency. This dissolved a false dichotomy in our own spec — we had treated "the loader reads the fragments" as proof the fragments were an authority, when under cache semantics being read is the point. It also inverts a vacuity trap: the freshness canaries would have degenerated to `overlay == overlay` for 72% of edges after migration, whereas `merge(authored) == cache` gets *stronger* as more edges become authored.

### What survived attack

Worth recording, since it was tested hard: the DRG **is** load-bearing, not write-only bookkeeping. `resolve_context` has four production callers, `action:software-dev/implement` alone resolves 89 artefacts, and `suggests` + `requires` are 591 of 774 edges — exactly the traversed relations.

### Open before implementation

- `occurrence_map.yaml` (now blocking), `plan.md`, `tasks.md`, `analysis-report.md` — none exist; the spec has been doing triple duty as spec, plan and tracker, and it drifted once already.
- Re-cost the plan: the 15-WP / 6–9 agent-day estimate was priced for 559 edges, not 774 plus a schema change.
- Whether the generated cache is committed or a build artefact.

### CI note

The 6 `arch-adversarial` failures are **inherited from the base branch** (byte-identical to `origin/main`) and are deliberately untouched per the red-main policy; `quality-gate` fails only as their cascade. `docs-freshness` was genuinely ours and is fixed.

stijn-dejongh:
## 📐 Design-session report: FoundationalValues & Creed (branch `docs/manifesto-tier-analysis`)

*Posted for cross-visibility — this session ran parallel to the main line of work and is now finalized. Nothing here touches this PR's diff; the branch is docs-only.*

**What the session did (one day, 2026-07-26):**

1. **Investigated a suspected gap**: Spec Kitty's doctrine resolves to an *unordered bag* of co-valid rules, so rule conflicts are silently tie-broken by the model's deliver-output-now prior. Confirmed and measured (e.g. `resolve_context` computes a requires/suggests strength ordering, then discards it at `query.py:119`; 40% of the `tasks` doctrine surface is soft suggestions presented as hard prerequisites).
2. **Killed a first design by experiment**: deriving rule precedence or tension edges from value vectors fails categorically (0/6 ordering reproductions; 5/5 false positives). Recorded so it is not rebuilt.
3. **Hardened the operator's replacement design**: a consumer-replaceable **ValueSet** (AMMERSE as default, generic over any N≥3 values), per-artefact `value_impact` deltas with mandatory-rationale floor, a signed **`impacts`** edge that subsumes `in_tension_with` (operator ruling), and a project **creed** on the charter. Grounded in a real 36-practice calibration corpus and a fully reproducible matrix analysis (committed JSON + stdlib script).
4. **Ran four squad rounds** (15 profile-loaded agents): adversarial ×2, hardening, publication-readiness, and a final verification round that recomputed ~60 figures by independent methods — finding and fixing two HIGH maths defects (a fail-open spectral-radius method; a spectral-vs-sup-norm bounding error) plus a 26-item fix ledger.

**Deliverables on the branch (`docs/plans/doctrine/`):**
- `foundational-values-and-creed.md` — the design AUTHORITY (ask, cost, maths, 6 open decisions)
- `manifesto-program-delivery-sequence.md` — sequencing AUTHORITY (critical path, 22 ranked increments, ADR scope)
- Committed evidence: corpus + both matrices as JSON, a reproduce-everything script, full squad reports, and a contradiction register of every superseded figure

**Key gate for whoever picks this up:** decision **D-2** (which DRG relation carries `impacts` once `in_tension_with` retires) swings the migration between ~5–10 and ~45–60 files, and must be answered inside the superseding ADR before implementation starts. The `#2538` experiment gates the numeric layer only.

No build decision has been taken; a digestible review copy is with the maintainers.

https://claude.ai/code/session_01TF3TAv5kQ3tMg5cKV4aa7F

stijn-dejongh:
## Ready to land — design complete, implementation deliberately not started

Out of draft. Final HEAD `6d810e288`. Title and description rewritten, because both described the original scope and this became something larger.

**What changed in this last pass:** ADR 2026-07-26-1 was Accepted and then overtaken three times, and had drifted into actively misleading — it said 414 six times, demanded byte-identity eight times, mentioned 559 zero times, and knew nothing of full retirement, `is_symmetric`, or `aliases`. Since [DIRECTIVE_003](../blob/main/src/doctrine/directives/built-in) makes the ADR (not a spec table) the citable record, anyone planning from it would have built the wrong proof against a scope 26% too small. It now carries the derived inventory and an explicit Amendment section recording all three rulings. All three ADRs now agree with each other and with the spec.

**How to review this efficiently:** read the three ADRs and the mission spec. They are current and mutually consistent. The individual commit messages along the way are *not* — the branch contains superseded design that was corrected in place, and those reversals are the honest record of how the design was reached rather than noise to be tidied away.

**Deliberately not here:** the 774-edge migration, `Relation.IMPACTS` implementation, the `aliases` / `is_symmetric` / `impacts` fields, the campsite lints. All specified, none built.

**Next mission starts fresh off `main`** once this lands, using the committed spec. Handoff notes are in `.local/next-mission-doctrine-canonical-structure.md`, including the four non-negotiable orderings — most importantly that the **zero-producer lint lands before any new field**, since the governing precedent here is 3-for-3 inert registers, one of them green for 162 days.

**CI, unchanged and honest:** the 6 `arch-adversarial` failures are inherited from base (byte-identical to `origin/main`) and left red per the red-main policy; `quality-gate` fails only as their cascade. `docs-freshness` was ours and is fixed. `diff-coverage` and `mission-loader-coverage` are green.

LynnColeArt:
Review of the FoundationalValues/creed design, as asked in #engineering. Taking the three questions from §0 in order, then the derangement check you actually asked for.

## No, you have not crossed into deranged territory

And the reason is not that the maths holds up — it is that the document is unusually honest about its own epistemic status. "Nothing was executed. Every claim about code behaviour is a static read." The measurements ship with a stdlib reproduction script. The falsification instruments are named, and you say outright that prior art predicts arm B reads null. Deranged work hides its load-bearing inferences; this one boxes them and puts a ⚠️ on them. The `N−1` caveat in §4 is the clearest example — you flag that reading the published divisor of 6 as `N−1` is an inference and that the generic bound depends on it.

So the answer to the implicit question is: the risk here is not that you have lost the thread. It is that the expensive band gets built before the cheap experiments say whether the premise is real. More on that below.

## 1. Is the shape right? Yes, with one substantive reservation

**The three-plane separation is the load-bearing idea and it is correct.** P1 symmetric correlation-like, P2 directed authored gain, P3 delta-from-not-adopting. Conflating those is exactly what turns value systems into mush, and keeping `impacts` authored rather than derived from P3 — because derivation *failed measurement* (§10.1) — is the kind of decision most designs get wrong in the optimistic direction.

The naming discipline is doing real work, not decoration. `delta` composes because it is relative to a baseline; a weight does not, and two profiles weighting a value at 0.8 do not give 1.6. Making `ValueImpact` and `ValueWeighting` **distinct Pydantic types rather than one model with a mode discriminator** means the type system structurally prevents dotting a profile with a directive. That is a category error prevented by construction instead of by review, which is the right altitude for it.

Three smaller calls I would also defend:

- **`delta` as string-on-wire / `Decimal` in memory, with a `mode="before"` validator rejecting non-`str`.** Correct, and for the right reason: `Decimal(float)` is lossy and ruamel yields `float` for an unquoted `0.9`. Byte-identical round-trip of the corpus is the proof. Declining to enforce a resolution grid because the corpus authors `0.125` is the right kind of deference to real data.
- **Positive frozensets for the value-bearing kinds, never an exclusion set.** Your reason — an exclusion set silently *includes* a future kind, which then acquires a scoring obligation nobody authored — is the same failure class I hit in our own billing catalog this morning, where a completeness guard meant adding one enum member would have broken an unrelated historical version at import. Positive sets age better.
- **Not making `FoundationalValues` a `StrEnum`.** Reusing the built-in → org → project layering in `src/doctrine/base.py` and copying the `Role` half-open value object rather than inventing a pattern is right, and the `DoctrineService` `@property` warning is a genuinely useful landmine to have marked.

**My reservation: strict sign on `impacts < 0` creates an unbounded reconciliation obligation, and I think that is a defect in the choice rather than a consequence to accept.**

You flag it honestly in §7.5 — an author writing `impacts: -0.05` for a mild trade-off creates a reconciliation obligation, and every negative edge however small becomes a formal tension claim. I would go further than "a real operational consequence". It has two failure modes and both are bad:

- authors learn to round small negatives to **zero** to avoid the paperwork, which destroys precisely the signal the numeric layer exists to capture; or
- the consistency check emits reconciliation findings faster than anyone clears them, and the whole channel gets ignored — which is *worse* than no signal, because it looks like governance while being noise.

The subsumption argument is sound: `impacts: -1` genuinely is what "antagonistic towards" means, and two authorities for one concept is worth removing. But the *tension claim* and the *negative sign* do not have to be the same predicate. Either a magnitude threshold, or an explicit annotation on the edge (you note `DRGEdge` already carries optional `when` and `reason`, so a sibling field is cheap), keeps the subsumption while making "this is a tension I am asserting" a deliberate act rather than an arithmetic side effect. I would not ship strict sign without that.

## 2. D-2 is already closed — and I think it is the wrong gate

ADR [`2026-07-26-3`](docs/adr/3.x/2026-07-26-3-impacts-edge-subsumes-in-tension-with.md) is **Accepted** and ADR-D2 reads "DECIDED: a new `Relation.IMPACTS` carries it… This is the programme-shape gate (≡ design D-2 / sequence gate G2) and it is now closed." The ADR-D8 amendment likewise closes D-6 by symmetrising the asymmetric pair to `+0.75`, which moves the headline residual to 4.37%.

So **§13 is stale on at least two rows**, and it matters more than a doc nit because §13 is in the AUTHORITY document — a reader checking "what is still open" will get the wrong answer, and §0 still instructs the reviewer to "answer D-2 or confirm it is the ADR author's call". Worth a sweep of that table against the ADR.

**The more useful observation: D-3, not D-2, is the programme gate.**

D-2 swings ~45–60 files against ~5–10. D-3 — does the creed feed arithmetic, or does an agent read `creed.yaml` as prose context — swings the entire numeric layer, which §0 prices at ~120–180 files, 6–11 kLOC, 25–45 days plus a 1,372–1,596-cell authoring tail. That is 60–65% of the tabulated cost sitting behind the decision the document lists fifth.

And §15 says the ranking measurement **assumed a linear dot product because D-3 is unresolved**. So a measurement that informs the design is conditional on the decision that determines whether that measurement is needed at all. That is the one circularity I would want broken before anything numeric is scheduled.

## 3. Authorise the prose-only band? Yes — and it may be the whole answer

Authorised as far as my opinion carries. It gates on nothing, it survived all four hardening lenses, and the polarity lint is the only component with a validation set available today, which makes it the only piece that can be *known* to work rather than argued to work.

But I want to push on something §8 says almost in passing, because I think it is the most important sentence in the document:

> never render a number without its rationale, and treat the rationale as authoritative where they disagree

Set beside §10.2 — the corpus shows the prose is right when the number is wrong — that reads as an argument that **the numbers are a sorting hint and the prose is the payload**. If the rationale always wins on disagreement, then the number cannot decide anything; it can only order things and prompt a conversation. Which is precisely what §2 says it is for ("conversation prompts, not verdicts") — so this is internally consistent, not a contradiction. My point is narrower: if that is true, the honest cost/benefit question is whether ordering-plus-prompting is worth 25–45 days and a ~1,400-cell authoring tail, when the prose band delivers the prompting for ~2–3 days.

I am not arguing against the numeric layer. I am arguing it should be **gated on the two falsification instruments, not merely preceded by a recommendation to run them.** You have a ~20-line perturbation-stability probe that would take an afternoon, and you have written down that you expect arm B to read null. If the creed weights are unstable under interview-grade jitter, or an agent given the creed does not surface the trade-off, the mechanism is dead independent of how good the eigenvalue analysis is — and the analysis is good. Spending the afternoon first is cheap insurance on a 45-day bet.

## One further limit worth stating in the doc

§7.2 reports that across 36 authored vectors `minimal` carries 19 of 27 negative cells, `solvable` and `environmental` are never negative, and six of seven axes carry ≤3 negative cells. You adjudicate that as levers-vs-goals conflation and honestly list the competing explanations.

Either way, the measured consequence stands: **on this basis there is effectively one axis that can lose.** A creed expressed over it can mostly only trade "how much simplicity are we spending", because the other six rarely cost anything. `kind: lever | goal` makes a consumer declare which is which, which is the right mechanism — but it does not make the AMMERSE default balanced for this purpose. I would state that as an explicit limit on what the numeric layer can express on the default basis, so a reader does not discover it after authoring 1,400 cells.

## Summary

| Question | Answer |
| --- | --- |
| Deranged? | No. The disclosure discipline is the strongest thing about it |
| Shape right? | Yes — three planes, authored `impacts`, creed on the charter. Reservation on strict sign |
| D-2 | Closed by the Accepted ADR. §13 is stale. **D-3 is the real programme gate** |
| Prose band | Authorise. Ship it. It is plausibly sufficient, not merely first |
| My one ask | Make the perturbation probe and `#2538` arm B **hard gates** on the numeric band |

Happy to be wrong on the strict-sign point in particular — it is a judgement about author behaviour under an obligation, not about the mechanism, and you have more of the corpus in your head than I do.

Issue #2931, opened 2026-07-26
Epic: Placement-Port Residuals Closure (PR #2920 follow-ups + main-green)


stijn-dejongh:
## Why

PR #2920 (`coord-write-placement-closure`) hardened the coordination write+read placement port, but its post-merge 4-lens review squad flagged **five residuals** too risky to fold in-PR, and the PR **merged with six gate/contract tests red** now sitting in `upstream/main` (honest-red-main). This epic tracks closing both.

## For whom

Maintainers and operators who rely on correct PRIMARY/COORD write-routing during merges, migrations, and retrospectives — and on green architectural gates catching real regressions.

## Expected effect

- The sole `status_phase` writer's PRIMARY-correctness is **enforced by the port**, not coincidental caller discipline (#2923).
- No silent PRIMARY-`tasks/` runtime loss in the legacy birth-cutover under coord topology (#2923).
- The over-broad `migration/` architectural gate carve-out is narrowed to per-file, justified entries (#2923).
- The three divergent write-target degrade implementations are unified into one kind-parameterized helper; a best-effort tracer read degrades instead of crashing (#2924).
- The six gate/contract reds PR #2920 merged red are all green — `upstream/main`'s arch + integration gates return to green (#2926 + a Part-C follow-up).

## Delivery

Mission `placement-port-residuals-closure` — 12 FRs across 7 work packages. Per-FR 1:1 issue-matrix. Children: #2923, #2924, #2926, and a Part-C follow-up (CLI/merge reds).

stijn-dejongh:
**Triage (planner-priti):** type `Feature`, `epic`, `priority:P1`, `reliability`, `tech-debt`, milestone `3.2.x`, no parent.

- **Type `Feature` + `epic`:** matches this repo's convention for functional delivery epics (#2160, #1868, #1931). The `bug` label is retired here, so kind is carried natively.
- **`priority:P1` = max of its children:** #2932 is P1 (merged-red release-gating gates), #2923/#2924 are P2. P0 withheld for the same reason as on #2932 — see that thread.
- **No parent, deliberately.** This is a mission-backed delivery epic with its own 12-FR/7-WP mission; the natural band would be #1878 (placement/identity strangler umbrella), but nesting it there would restructure the lineage of an in-flight (currently paused) mission's tracker topology, which is out of scope for a triage pass. Flagging the option for the operator rather than acting on it.
[...]

Issue #2922, opened 2026-07-25
Read-side whack-a-read (D-06): ~50 modules bypass the hardened read-surface authority


stijn-dejongh:
## Summary

The `coord-write-placement-closure` mission (PR #2920) hardened the **one blessed** kind-aware read entry point (`mission_runtime.PlacementSeam.read_dir` → `resolve_artifact_surface`) to fail loud on a partition mismatch (e.g. a DELETED coordination branch). However, **~50 pre-existing modules still call the lower-level, kind-blind resolvers directly** — `candidate_feature_dir_for_mission` / `resolve_planning_read_dir` — bypassing the seam and carrying the same lenient-on-DELETED-coord-branch gap the seam now closes.

## Why it matters

This is the read-side "whack-a-read" (D-06 in the mission research): writes are now unrepresentable-when-wrong via the whole-tree gate, but reads are still hardened one blessed authority at a time. A direct kind-blind read can still silently substitute the primary checkout when a coord-partition artifact's branch is gone.

## Scope

Out of the placement-closure mission's scope (that mission generalized the authority; a full site-by-site migration is a separate program). Related to the #1619 runtime/state overhaul epic.

## Fix sketch

Enumerate the ~50 direct callers of `candidate_feature_dir_for_mission` / `resolve_planning_read_dir`, migrate each to route through `PlacementSeam.read_dir(kind)` / `resolve_artifact_surface`, and add a read-side arch gate that reds on a new direct kind-blind read (symmetric to the whole-tree write gate).

stijn-dejongh:
**Triage (planner-priti):** type `Task`, `priority:P2`, `tech-debt`, `reliability`, no milestone, parent #1878.

- **P2, not P1:** no user-facing symptom is claimed or reproduced — the gap is latent (a kind-blind read *can* substitute the primary checkout). Per the pass directive, an internal hardening ticket is not inflated to the friction floor just because it is adjacent to one.
- **Parent #1878, not #1619:** #1878's deferred item 1 is literally "migrate the remaining direct-placement call sites onto the canonical placement resolver", and its P2/non-release-blocker framing matches this issue's own "separate program, out of scope" framing. #1619 stays a cross-reference (this body cites it) rather than the parent.
- **No milestone deliberately:** a ~50-site migration plus a new read-side arch gate is program-sized and unowned; stamping `3.2.x` would assert a window nobody has committed to. The parent umbrella carries the band.

Issue #2921, opened 2026-07-25
repair_lane_mismatch corrupts frontmatter: raw_frontmatter_text destructured as padding


stijn-dejongh:
## Summary

`repair_lane_mismatch` in `src/specify_cli/coordination/task_metadata_validation.py` (~line 127) destructures `template.renderer.parse_frontmatter`'s **3rd** return value (`raw_frontmatter_text`) as `padding`, then threads it into `build_document(frontmatter_yaml, body, padding)`, which appends it between the closing `---` and the body (`doc += padding`). This **duplicates the entire frontmatter YAML into the body region** on every legacy lane repair, corrupting the document (broken closing `---` delimiter).

## Scope / severity

- Real correctness bug, but confined to the **migration-only legacy `lane`-frontmatter repair path**, retired in the Phase-2 event-sourced model. Latent for modern missions; actively corrupting only when legacy lane-repair fires.
- Confirmed **pre-existing** (zero commits on the `coord-write-placement-closure` branch touched this file).

## Discovered by

Reviewer during the `coord-write-placement-closure` mission (PR #2920); reported out-of-scope (C-002) rather than fixed in-line.

## Fix sketch

Destructure `parse_frontmatter` correctly (3-tuple: metadata, body, raw_frontmatter_text) and stop passing `raw_frontmatter_text` as the `padding` argument to `build_document`; add a regression that round-trips a legacy lane repair and asserts the document parses back byte-clean.

stijn-dejongh:
**Triage (planner-priti):** type `Bug`, `priority:P1`, `reliability`, milestone `3.2.x`, no parent.

- **P1, not P2:** this corrupts a document the operator authored (frontmatter duplicated into the body, broken closing delimiter) whenever legacy lane repair fires — end-user-visible destruction, so the implementation-friction floor applies even though the fix is a one-line destructuring correction.
- **P0 withheld:** reachability is confined to the migration-only legacy `lane`-frontmatter repair path, retired in the Phase-2 event-sourced model. No modern mission reaches it, so it neither blocks a release nor can fire for a current user.
- **No parent:** no functional epic owns legacy lane-frontmatter repair. Deliberately *not* filed under #2931 — that epic's body scopes exactly five PR #2920 residuals plus the merged-red gates, and this was reported as pre-existing/out-of-scope (C-002).

stijn-dejongh:
**Anchor correction (from next-slice triage — file path in the body is wrong):**

[...]

Issue #2917, opened 2026-07-25
test_dogfood_corpus_backfilled reds: 12 recently-merged missions not cut over to event-sourced runtime-state (sole 3.2.6 CI Quality blocker)


stijn-dejongh:
## Summary

`tests/specify_cli/migration/test_dogfood_corpus_backfilled.py::test_all_eligible_missions_snapshot_non_empty_and_verify_ok` is **red on `main`** and is the **sole remaining CI Quality failure** — it blocks the 3.2.6 release tag (red CI is the release authority, ADR 2026-07-17-1).

It reproduces **deterministically and locally** (~32s, reads the committed `kitty-specs/` corpus — not CI-environmental):

```
tests/specify_cli/migration/test_dogfood_corpus_backfilled.py:184: in ...
    assert unflipped == [], f"eligible missions not cut over: {unflipped}"
E   AssertionError: eligible missions not cut over: [12 missions]
```

## The 12 "eligible … not cut over" missions
All are **recently-merged** missions (01KX*/01KY* ULIDs):
`common-docs-query-01KY541A`, `coord-commit-integrity-01KY5JS8`, `docs-ia-onboarding-overhaul-01KY02JB`, `docs-structural-sanity-01KY53KJ` (#2905), `doctrine-controlled-transition-gates-01KY51Z7`, `doctrine-tension-edges-01KY1WPC`, `dossier-parity-reconciler-01KXYXVP`, `drg-relation-parity-activation-gate-01KY48PD`, `glossary-pack-doctrine-kind-01KY30SW`, `lifecycle-gate-execution-context-01KY72GQ` (#2906), `oss-fork-packaging-hooks-01KY2JFM`, `scopesource-gate-followup-01KY6S9P`.

## Context
- The test is an acceptance-lock from the **#2816** runtime-state corpus cutover (WP03/WP04, IC-01b/IC-10) — it verifies every eligible committed mission has been flipped to the event-sourced runtime-state model. It is `@pytest.mark.integration` / `@pytest.mark.slow`, **not** `@pytest.mark.regression`, so it is not an intentional honest-P0 red — it is a genuine failure.
- The failing set being exactly the newest merges points to a **corpus-drift / cutover-not-applied** pattern (missions landing un-flipped) rather than a #2816 logic regression — but this is under active investigation.

## Investigation (research squad in progress)
[...]

PR #2963, opened 2026-07-27
Placement-Port Residuals Closure: enforce PRIMARY/COORD partition routing through the port

### LINKED ISSUE #2923: Birth-cutover placement hardening: route _flip_phase through the port + narrow migration/ gate carve-out + close cutover_mission COORD-read temporal residual
Follow-ups from the PR #2920 (`coord-write-placement-closure`) post-merge review squad. Deferred from the PR because each risks the FR-007 backfill / sole-writer path and needs corpus-wide validation.

## 1. Route `_flip_phase` through the placement port (architect F2)
`runtime_state_cutover._flip_phase` — the **sole `status_phase` writer** — writes `meta.json` via `canonicalize_feature_dir(feature_dir)` + direct `write_meta`, bypassing `resolve_placement_only(PRIMARY_METADATA)`. It is PRIMARY-correct only because every current caller passes a PRIMARY dir. A future caller passing a non-primary dir would write `status_phase` off-partition with no gate or port to catch it. **Fix:** resolve the write target through the port and assert the resolved home equals the passed `feature_dir` (fail-closed). Must be validated across the whole FR-007 backfill corpus without regressing the flip of genuinely-legacy missions.

## 2. Narrow the `src/specify_cli/migration/` whole-tree gate carve-out (renata F2, architect F4)
The whole-tree write gate retains three sanctioned SUBTREE prefixes; `src/specify_cli/migration/` (15+ modules, incl. the real `runtime_state_cutover` writer) is a directory-prefix exclusion, so a bypass added under it would not red the gate — SC-002/NFR-001's "any module" is therefore scoped (now documented in the spec). Once (1) routes the writer through the port, convert this prefix to per-file `BOUNDARY_SANCTIONED_MODULES` entries with individual rationales so the "any module" guarantee holds for that subtree.

## 3. Close the `cutover_mission` COORD-read / PRIMARY-write temporal residual (architect F3, debbie F4 — the documented C-001)
The two-target `cutover_mission` reads legacy `tasks/` from `status_feature_dir` (COORD) while `_flip_phase` writes to `feature_dir` (PRIMARY). Sound only because IC-07 (WP04/WP05) means no mission reaching the merge hook carries genuine frontmatter runtime — a temporal-coupling argument enforced by nothing at runtime. A genuine pre-IC-07 cold-upgrade mission carried across the boundary and merged under coord topology could have its real PRIMARY-`tasks/` runtime bypassed while the flip proceeds (silent data loss). Mitigated today by the standing `migrate backfill-runtime-state` and the best-effort/repairable fallback. Add a runtime guard or read the seed frontmatter from the PRIMARY leg.

### LINKED ISSUE #2924: Degrade-path + best-effort-read hardening: unify _mission_meta_exists (3-way clone) + guard _load_traces against CoordinationBranchDeleted
Follow-ups from the PR #2920 (`coord-write-placement-closure`) post-merge review squad (paula-patterns lens). Both are non-blocking hygiene, out of the PR's fold scope.

## 1. Unify the bootstrap write-target degrade idiom (`_mission_meta_exists`, paula F2)
`_mission_meta_exists` was added **verbatim** (identical body + `follow_imports=skip` mypy-narrowing comment) in both `src/specify_cli/events/decision_log.py` and `src/specify_cli/git/bookkeeping_commit.py`, each wrapping a near-identical "resolve via placement port, else degrade" helper. Meanwhile `coordination/status_transition._resolve_write_target` (the idiom they cite as canonical) degrades through `get_feature_target_branch` with **no** `_mission_meta_exists` pre-gate. The single decision "how do I resolve a write target when the mission isn't fully bootstrapped?" now has **three divergent implementations** across three writer surfaces. **Fix:** extract one shared `resolve_write_target_or_degrade(repo_root, mission_slug, kind, *, degrade_ref)` and route all three through it.

## 2. Guard `_load_traces` against `CoordinationBranchDeleted` (paula F3)
WP07 correctly re-routed the tracer read in `retrospective/generator._load_traces` through the fail-loud authority (`placement_seam(...).read_dir(TRACER_FILE)`). But `read_dir` for a COORD kind can now raise `CoordinationBranchDeleted`, whereas `_load_traces` is documented/used as best-effort (skips unreadable tracers, returns `[]` on missing dir). The call is unwrapped — a deleted-coord retrospective generation would now crash instead of degrading. **Fix:** wrap in a `try`/degrade, or make the raise-here-is-intended contract explicit.

### LINKED ISSUE #2926: arch gate red on main: GuardCapability.MERGE_BOOKKEEPING asserted outside its flow in merge/executor.py
`tests/architectural/test_guard_capability_call_sites.py::test_protected_flow_capability_call_sites_are_allowlisted[MERGE_BOOKKEEPING]` is **red on `upstream/main`** (reproduced on the current main tip and on an unrelated branch `558d368`, both untouched by any in-flight PR):

```
GuardCapability.MERGE_BOOKKEEPING is asserted outside its flow: ['src/specify_cli/merge/executor.py'].
Each non-standard capability authorizes exactly ONE bookkeeping flow (FR-008 / C-GUARD-2);
ordinary status bookkeeping must assert STANDARD so protected destinations are refused.
PR #1850's guard-bypass regression is exactly this misuse.
```

Surfaced during the #2884 landing pass; #2884 does **not** touch `merge/executor.py` (0 delta vs main), so it is not that PR's defect — filing here so the `arch-adversarial` shard red is attributable and future landing passes skip re-reproducing it (per the red-main / landing-runbook policy).

**Scope**: adjudicate whether `merge/executor.py` legitimately needs `MERGE_BOOKKEEPING` at that call site (update the allowlist with a dated rationale) or whether it should assert `STANDARD` (the guard-bypass regression the test guards against). Mission-scope for the guard-capability subsystem, not a landing-pass fold.

### LINKED ISSUE #2932: Part C: green the CLI/merge gate reds PR #2920 merged red (FR-009/010/011)
Part of epic #2931. Three gate/contract reds PR #2920 merged red into `upstream/main` (the CLI/merge subset — the arch write-side/guard-capability reds are tracked under #2926). Non-blocking hygiene; judged per the failing-test remediation framework (fix PRODUCT where the test encodes a real invariant, re-pin the TEST where the contract legitimately moved).

## Reds

- **FR-009** — `test_no_raw_mission_spec_paths::test_constant_based_mission_spec_path_construction_stays_in_constructor_files` red on `cli/commands/agent/mission_repair.py:65` (raw `repo_root / KITTY_SPECS_DIR / mission`). Route through the canonical mission-dir constructor OR add a rationale-bearing constructor-file allow-list entry.
- **FR-010** — `test_app_exposes_exactly_eight_frozen_commands` red: the sanctioned `agent mission repair` command is a 9th, the frozen `_EXPECTED_COMMANDS` still says 8. Add `repair` to `_EXPECTED_COMMANDS` **and** pin its flag/positional surface (`_EXPECTED_FLAGS["repair"]`), rename the count test 8→9, add the `repair` row to `cli-surface-contract.md`. (Adding to the command set alone, leaving the flag surface unverified, is fakeable and insufficient.)
- **FR-011** — `test_safe_commit_is_called_with_correct_files` + `test_planning_artifact_only_merge_does_not_require_mission_branch` red: the merge committed-file set drops `status.events.jsonl`. Fix the PRODUCT (committed-set) — the tests encode the real FR-019/FR-020 durability invariant.

## Scope note

Do NOT ride these under #2923 (whose body covers A1/A2/A3 only).

stijn-dejongh:
# Placement-Port Residuals Closure

> Reviewer summary for PR `placement-port-residuals-closure`. Written for a reader
> who has **not** followed the mission — the background below is enough to review
> the diff without extra context.

## Background (for a non-maintainer reader)

Spec Kitty writes each mission's files to one of two homes ("partitions"):

- **PRIMARY** — stable planning & metadata (spec, plan, and the `status_phase`
  key that gates the whole status model). Lives on the mission's planning branch.
- **COORD** — lifecycle surfaces (the append-only status event log, notes, traces)
  routed onto a separate *coordination* branch.

A single component — the **placement port** — is supposed to be the one authority
that answers "where does this artifact go?". A previous, already-merged change
(PR #2920) established that port. Its post-merge review found that a few writers
still wrote directly to a directory the caller handed them, **bypassing the port**.
They happened to be correct only because every current caller passed the right
directory — the guarantee was true by *caller discipline*, not *enforced by the
port*. A future caller passing the wrong directory would silently write to the
wrong partition with nothing to catch it. PR #2920 also merged with six
gate/contract tests red (deliberately deferred).

This PR closes those residuals: it makes partition-correctness **enforced by the
port**, and returns the deferred gates to green.

## Summary

- Route the placement-sensitive writers through the placement port and **fail
  closed** (raise, write nothing) if the resolved home disagrees with the caller —
  converting coincidental correctness into an enforced invariant.
- Consolidate three hand-rolled "resolve write target, else degrade" copies into a
  single shared helper, and guard a best-effort read against a deleted coordination
  branch.
- The six gate/contract reds the prior PR deferred were **already fixed** by other
  changes that landed in the interim; this PR verifies them green and records that.

## Why Now

PR #2920's 4-lens post-merge review deferred five residual weaknesses (tracked as
issues **#2923** and **#2924**) because each touches a load-bearing surface — the
sole `status_phase` writer, the legacy-cutover path, or a best-effort degrade
surface — and needed corpus-wide validation too risky to fold into that PR. A
sixth item (**#2926**) was an arch-gate red at the same call site. Left open, the
placement guarantees stay "true by luck" and the gates stay red. This mission
closes all of them as one cohesive unit.

## What This PR Does

**Enforce partition-correctness (Part A):**
- `_flip_phase` — the *only* code that writes `status_phase` — now resolves its
  write target through the port and raises a typed `PlacementMismatchError`
  (writing nothing) if the resolved PRIMARY home ≠ the directory it was handed.
  It degrades gracefully only when the *resolver itself* can't run, never to mask a
  genuine wrong-partition write. (FR-001)
- The two-target legacy cutover now reads `tasks/` frontmatter from the PRIMARY leg
  while the status **event write stays on COORD** — closing a read/write-partition
  mismatch without moving the event write. (FR-002)

**Remove duplication & harden degrade paths (Part B):**
- One kind-parameterized `resolve_write_target_or_degrade` helper (in the
  `mission_runtime` package) replaces three near-identical hand-rolled copies;
  each caller keeps its own degrade policy (one fails open, one fails closed, one
  is a pre-gate). Zero verbatim clones remain. (FR-005)
- The retrospective's trace loader now degrades to an empty list if the
  coordination branch was deleted, instead of crashing — scoped to that one call
  site. (FR-006)

**Tighten the gate & return deferred reds to green (Part C):**
- The write-side placement scan no longer blanket-exempts the `migration/`
  subtree, restoring true "any module" precision (proven by a red-first synthetic
  bypass); a separate, intentional resolver-walker carve-out is untouched. (FR-003/004)
- The remaining deferred gate/contract reds (raw-path use, the CLI golden-command
  contract, the merge committed-file set, and a guard-capability allow-list) were
  **already green on the current base** — fixed by other PRs that landed between
  the prior PR and this branch's rebase. Verified in-mission. (FR-008–FR-012)

- **Runtime / compatibility:** Behavior-preserving for all current callers. The one
  intended behavior change is a bug fix: a caller-supplied coordination branch that
  was previously silently discarded in the bootstrap window is now honored. The new
  fail-closed raise only fires on a genuine wrong-partition write, which no current
  caller produces.
- **Upgrade / migration:** None. No data format, config, or CLI surface changes;
  purely internal routing + a stricter internal gate.
- **Operator / reviewer impact:** A future writer that targets the wrong partition
  now fails loudly instead of silently — a safety improvement. The write-side gate
  is stricter for the `migration/` subtree.

## Validation

- [x] Red-first per fix — each behavioral fix ships a test proven to fail on the
      pre-fix code and pass after (verified by reverting the product file).
- [x] `ruff` and `mypy --strict` clean on all changed source.
- [x] Each of the 7 work packages independently reviewed and approved
      (implement=Sonnet / review=Opus; a genuine two-party cycle, including one
      rejection→re-implementation for a real regression that was caught and fixed).
- [x] Full `tests/architectural/` suite + terminology guard run locally on the
      rebased tip; pre-merge aggregate review (architect + SSOT lenses) on the
      combined diff.
- [x] Backward compatibility considered (see Effect above).
- [x] Follow-up work called out (see Follow-ups).

## Tickets / Contracts

| Ticket | Relationship |
|--------|--------------|
| #2931 | Parent epic for this mission |
| #2923 | Closed — Part A (birth-cutover placement-port hardening: FR-001/002/003/004) |
| #2924 | Closed — Part B (degrade-path + best-effort-read hygiene: FR-005/006) |
| #2926 | Closed — arch gate MERGE_BOOKKEEPING allow-list at the coord-seed call site (FR-008/012); verified already-green |
| #2932 | Closed — Part C golden-contract/raw-path/merge-committed-set (FR-009/010/011); verified already-green |
| #2920 | Parent PR whose residuals this closes (merged) — not re-fixed here |
| #2921 | Deferred (out of scope, separate mission) — `repair_lane_mismatch` frontmatter fix |
| #2922 | Deferred (out of scope, separate mission) — read-side whack-a-read (~50 modules) |

## Mission Artifacts

- Spec: `kitty-specs/placement-port-residuals-closure-01KYDEF0/spec.md`
- Plan: `kitty-specs/placement-port-residuals-closure-01KYDEF0/plan.md`
- Tasks: `kitty-specs/placement-port-residuals-closure-01KYDEF0/tasks.md`
- Analysis: `kitty-specs/placement-port-residuals-closure-01KYDEF0/analysis-report.md`
- Issue matrix: `kitty-specs/placement-port-residuals-closure-01KYDEF0/issue-matrix.md`
- Reviews: `kitty-specs/placement-port-residuals-closure-01KYDEF0/tasks/WP*/review-cycle-*.md`
- PR summary: this file (`pr-summary.md`)

## Follow-ups

- **#2921** — `repair_lane_mismatch` frontmatter-corruption fix (separate mission).
- **#2922** — read-side whack-a-read remediation across ~50 modules (separate mission).
- **Pre-merge squad NOTEs** (non-blocking; no live defect — to be filed as one follow-up under epic #2931):
  1. `write_target_degrade.py` fail-closed branch: when `resolve_placement_only` raises a
     *caught-set* exception with `degrade_ref is None`, the helper converts it to
     `ActionContextError(FEATURE_CONTEXT_UNRESOLVED)` rather than propagating the original.
     Immaterial in practice (both callers catch broadly) — a one-line docstring clarification
     would remove the ambiguity.
  2. `_flip_phase` FR-001 fail-close compares `resolve_artifact_surface(...).path` against
     `canonicalize_feature_dir(...)`; correctness depends on both normalizers producing
     identically-formed paths. No live defect (tests exercise the match case), but a defensive
     test pinning the normalization contract would harden it.

---

Closes #2923
Closes #2924
Closes #2926
Closes #2932

(Epic #2931. #2920 is the parent PR whose residuals this closes — not itself closed here. #2921 and #2922 are deferred to separate missions.)

**Pre-existing note:** `mypy` reports 4 errors in `src/specify_cli/coordination/status_transition.py` (lines 135/151/160/635) that reproduce identically on `upstream/main` — pre-existing, outside this mission's diff (which touches only `_resolve_write_target` at ~669). Left as-is per the red-main / honest-pre-existing policy; not introduced by this PR.

**Reviewer note:** the full `tests/architectural/` suite + terminology guard were run locally on the rebased tip (1155 passed after a marker-correctness fix on the new WP06 test); a pre-merge aggregate review (architect + SSOT lenses) on the combined diff found no MAJOR issues. CI is the release authority for the full sweep.

https://claude.ai/code/session_01LXZN76oRBdHFafayjAphBB

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/2963"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787720300&installation_model_id=427282&pr_number=2963&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F2963&signature=3de75e80723f5ad90540e6661897f98aefb3d5752b49e7208c3607459e9ada2d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


stijn-dejongh:
### Post-mission-ops folds + closeout evidence

Pushed 4 additive commits on top of the mission (no history rewrite):

**Folds (one commit each):**
- `adf5bcdc7` — **project-mypy campsite:** removed a redundant `cast(str, …)` in `_resolve_write_target` (WP05 added it to satisfy a *single-file* mypy invocation; CI runs project-mode `--strict`, where `get_feature_target_branch` is known to return `str`, so it was flagged `redundant-cast`). Coordination-chain project mypy drops 10→9 errors; the remaining 9 are pre-existing `no-any-return` in files this mission does not touch (`stale_detection.py`, `status_service.py`, `legacy_resolution.py`) — mypy is advisory in CI.
- `2c68d40f7` — **squad NOTE 1** (docstring-only): clarified `resolve_write_target_or_degrade`'s `Raises` — a caught-set failure with `degrade_ref is None` is re-raised as a *fresh* `ActionContextError`, not the original.
- `d89e2be8d` — **squad NOTE 2** (test-only): defensive test pinning that the port's resolved PRIMARY home and `canonicalize_feature_dir(...)` normalize identically through a symlinked repo root, so `_flip_phase` never spuriously fail-closes a legit mission.
- `3fcbe4d21` — **changelog:** Fixed-section entry under Unreleased 3.2.6.

**Local gates on the pushed tip (`3fcbe4d21`):**
- `tests/architectural/` — 1155 passed (after a marker-correctness fix on the new WP06 test: dropped a stray `pytest.mark.fast` from a subprocess/git-repo test).
- terminology guard — 4 passed; docs-freshness suite — 602 passed; `check_docs_freshness --ci` — 0 errors.
- `ruff` clean on all changed files; `mypy` new code clean (project mode).
- Pre-merge aggregate review (architect + SSOT lenses) on the combined diff — no MAJOR findings (the two NOTEs above are the folded output).

**State:** draft; CI re-running on the new tip. Merge-ready on green — **operator merges** (I do not run `gh pr merge`).

stijn-dejongh:
Follow-up filed: **#2964** — the leftover `feature*`→`mission*` terminology mismatch surfaced during this mission (on `get_feature_target_branch`). Deliberately **not** folded here: it is a ~5,100-reference `feature*` identifier family (`feature_dir` alone is 3,463 refs / 233 files), so a single-symbol rename would be an inconsistent partial parity-fix. #2964 tracks it as a dedicated bulk-edit terminology-migration mission (with a terminology-guard extension so it stays fixed).

stijn-dejongh:
Claiming this PR for a landing pass: isolated worktree, rebase onto current `upstream/main`, classify every red check into its bin, fold what belongs on the branch, adversarial squad on the partition-routing surface, then history compression (bookkeeping bunched; docs and code separate) before the final push. Evidence to follow — the operator merges.

stijn-dejongh:
## Landing pass — two-lens adversarial squad, six folds, no history rewrite

**Review verdict: land it.** No P0. Two independent profile-loaded lenses (architecture/contract adherence and SSOT/duplication) reviewed the code surface against base `1aed89411`; both converged on "sound design, land after small fixes". Mission history and per-fold lineage are preserved deliberately — one fold, one commit.

### What the squad verified rather than assumed

- **Partition assignments are correct and port-derived.** `STATUS_STATE` and `DECISION_LOG` stay coord; `_flip_phase` targets `PRIMARY_METADATA`. No kind flattened to the wrong partition.
- **It composes with the #2884 birth-cutover seam instead of forking a second routing authority** — `_start_birth_cutover` still routes coord seed events through `commit_coord_seed_bookkeeping`, and `_flip_phase` only ever sees `run.target_feature_dir`.
- **`write_target_degrade.py` is not a new authority.** It contains zero routing: no topology read, no coord-branch composition, no HEAD read. The only non-port ref it can emit is a *caller-supplied* degrade value. The consolidation claim holds.
- **Nothing was widened to buy the refactor.** Zero `CommitTarget(`/`safe_commit(` in `src/specify_cli/migration/` at both base and tip, so the `migration/` prefix drop is honest; the four deleted descriptors were forced bookkeeping (stale entries raise at import). No config, CI, baseline, or shard-map changes.
- **The new tests genuinely falsify** — the fail-closed test carries an explicit anti-scaffold precondition proving the port resolves *without raising* first, so a resolver crash cannot masquerade as the contract; the whole-tree scanner has a hard non-emptiness floor and cannot pass vacuously.

### Folds (one commit each, for lineage)

| Commit | Fold | Why |
|---|---|---|
| `2e30965ef` | **P1** — per-mission fail-close containment in both cutover walkers | `_flip_phase`'s new `PlacementMismatchError` was unguarded. The raise is correct and stays; only its containment was wrong. Note the two walkers have **different** documented contracts: `cutover_repo` is per-mission best-effort (proven: the mission sorted after a mismatch still flips), while `_cutover_corpus` deliberately aborts the whole step per D-03 — there the fix is narrowly "no bare traceback", and the tests assert those opposite semantics. |
| `2af370a83` | Preserve the degrade cause chain + per-file gate rationale | `raise … from exc` was missing where the same package does it correctly twice, losing the concrete subclass, its `error_code`, and the chain — `CoordinationBranchDeleted` is the #1848 data-loss signal five other sites type-discriminate on. Also: the four rationale-bearing allowlist descriptors were deleted because their construction moved into a blanket-exempt package, taking policed coverage to zero on exactly the surface this PR hardens; a per-file `BOUNDARY_SANCTIONED_MODULES` entry restores it. |
| `ccafd71ea` | Keep the status-transition degrade ref lazy | `coord_branch or get_feature_target_branch(...)` shelled out to `git symbolic-ref` on **every** status transition for non-coord topologies, on the happy path, discarding the result — re-admitting the ambient-HEAD read #1716 drained. The test pins *exactly one* call (the port itself makes one internally), catching regression to the redundant second. |
| `276288d9b` | Make the retrospective trace-surface degrade observable | `CoordinationBranchDeleted` was swallowed with no log and no marker, so "no tracer files" and "tracer surface unreachable, all evidence dropped" were indistinguishable. Now a WARNING naming mission + declared coord branch, with a **discriminating test pair** — identical empty output, only one logs. |
| `6849b469f` | De-mask the cutover fixture | The fixture wrote `meta.json` onto the COORD leg, which production never has, certifying an impossible corpus shape. Removed; all three tests still pass, confirming the anchor synthesizes from PRIMARY's `shell_pid_created_at` first. |
| `c75baa5c4` | Correct SC-001 | As implemented, `_flip_phase` consults the port for the PRIMARY home and fail-closes on mismatch, but writes to the caller's canonicalized `feature_dir` — which is exactly what FR-001(c) specs. The old wording claimed *resolution delegation* the design deliberately does not do. **Claim corrected, not code**: a migration writer that silently retargets the mission directory it was handed would be worse than one that fails closed. |

### Out of scope — filed, not crammed in

**#2966** (Task, P2, under #2160) carries the squad's four remaining findings with their evidence: the two FR-002 PRIMARY reads still on the COORD leg (seed ULIDs namespaced by directory name; the store itself warns the events are orphaned with `mission_id = None`) — pre-existing, and the stronger data-loss hypothesis was **tested and falsified**; the unconsolidated fourth write-target sibling in `safe_commit_cmd` with its drifted caught set and missing pre-gate; the absent `limitations` field on the retrospective record; and the real endgame — `status/emit.py`, the repo's highest-volume artifact writer, still bypasses the port entirely. That, not these three sites, is where "one authority" is won.

### Local evidence on the final tip

- **CI on the pre-fold tip `3fcbe4d21` was fully green — 73 checks, zero failures.** The folds build on a clean base rather than masking a red.
- Every fold was verified against its own surface before commit: `tests/mission_runtime` 244 passed; `tests/architectural/test_no_write_side_rederivation.py` 27 passed; `tests/specify_cli/coordination` 333 passed; `tests/specify_cli/retrospective` 6 passed; `tests/specify_cli/migration` 208 passed; the cutover-placement and backfill-migration files 23 passed; `test_cutover_partition_decouple.py` 3 passed.
- `ruff check .` clean repo-wide on the final tip. `mypy` on each touched package shows no new findings — pre-existing errors confirmed identical by stash-and-rerun in each case.
- A single aggregate re-run across all of those directories at once was still in flight at push time; the per-surface results above are the substantive evidence, and **this PR's CI run is now the authority**.

One known pre-existing red is **not** this PR's and was not green-washed: `test_dogfood_corpus_backfilled.py::test_all_eligible_missions_snapshot_non_empty_and_verify_ok` (#2917, the dogfood-corpus cutover gap).

### Hand-off

No rebase was needed — the branch already sat on current `upstream/main` (`1aed89411`). Merging closes **#2923, #2924, #2932** (#2926 already closed), i.e. all four children of epic #2931; the epic itself stays open and wants a manual close. **The operator merges.**

stijn-dejongh:
**Aggregate local run completed** (the one still in flight at push time):

```
1 failed, 1973 passed, 4 skipped in 1142.91s
FAILED tests/specify_cli/migration/test_dogfood_corpus_backfilled.py::test_all_eligible_missions_snapshot_non_empty_and_verify_ok
```

Run scope: `tests/architectural`, `tests/mission_runtime`, `tests/specify_cli/{coordination,retrospective,migration,upgrade/test_runtime_state_backfill_migration.py}`, `tests/regression/test_birth_cutover.py`.

The single failure is the **pre-existing #2917 dogfood-corpus cutover gap** — red on the base, not caused by this branch, and independently attributed as such by each fold before it was committed. It is deliberately **not** green-washed here; the fix is its own mission, not a landing fold.

That closes out the local evidence. CI on this tip is the authority.

Issue #2957, opened 2026-07-26
CI shard selection never collects several tests/specify_cli/cli test files on main — frozen-contract gates are vacuous


stijn-dejongh:
## Context

While landing PR #2884, its CI run failed five tests that turned out to be **pre-existing base reds on `main` @ `1a15bcf6c`** — invisible on main because **no main-branch CI job collects them**:

- `tests/specify_cli/cli/commands/agent/test_mission_cli_golden_contract.py`
- `tests/specify_cli/cli/commands/test_completion_fast_path.py`
- `tests/specify_cli/cli/commands/test_mission_repair.py`
- `tests/cli/commands/test_merge_status_commit.py`

Verified mechanically: `grep` over the logs of **every** job of main's CI Quality run 30212948549 (all `fast-tests-*`, `integration-tests-*` shards) finds zero collection of these files, while PR #2884's `integration-tests-cli` job did run them and went red.

Consequence observed: the coord-write-placement-closure WP08 `agent mission repair` command landed on main and silently broke the frozen mission-CLI golden contract, the committed completion manifest, and `test_merge_status_commit` — and main CI stayed green. A frozen-contract gate that never runs on the branch it protects is vacuous (same failure class as the schema-slot-without-producer lesson and the DRG count-ceiling drift in #2954).

PR #2884 carries the census catch-up folds so its CI goes green; this issue is the structural successor.

## Proposed direction

1. Determine why the main-branch shard selection skips these files (selection baseline drift? shard globs that don't match `tests/specify_cli/cli/commands/agent/`? test-impact selection only on PRs?).
2. Guarantee every test file is collected by exactly one main-branch job — e.g. an architectural meta-test that diffs the union of shard collections against the full collection (the same anti-vacuity mechanic as the inline-meta census gate).

## Refs

- Main run with zero collection: CI Quality 30212948549 (commit `1a15bcf6c`)
- PR run that surfaced the reds: #2884 checks on `617d5b488`
- Related anti-vacuity precedents: #2954 (DRG residual census), inline-meta census gate

stijn-dejongh:
Folded into mission **A — Doctrine Silence Guards** (#2948) as **WP10**, rather than tracked separately.

**Why there.** A frozen-contract test that no main-branch job collects is exactly as inert as a schema slot nothing produces — which is that mission's WP01. Same defect class, one layer up, so it belongs with the mission that closes the class rather than beside it.

[...]

PR #2982, opened 2026-07-27
fix(ci): stop coverage reports silently losing data to Cobertura root collisions

### LINKED ISSUE #2975: Coverage reports silently lose data: Cobertura root-level key collision between co-instrumented packages
## The defect

Coverage's Cobertura XML writer keys root-level files by `(package=".", filename)` with **no per-source disambiguation**. When a single `pytest-cov` invocation instruments two packages that both have a file at the same relative position, the entries collide and one silently wins.

Concretely, every job that passes both `--cov=src/specify_cli` and `--cov=src/mission_runtime` (with `relative_files=true`) produces a report in which `mission_runtime/__init__.py` (18 statements) collides with `specify_cli/__init__.py` (218 statements). **Verified in three sampled real CI artifacts from run `30241287792`: the slot always shows 218 lines, never 18.**

The consequence is silent data loss in the coverage report — not a test gap, not a marker gap. **No test or marker change can fix it**, which is what makes it dangerous: the natural response to the symptom (a file showing 0%) is to add tests that already exist.

## How it surfaced

PR #2963's `diff-coverage` gate failed at 88% against a 90% threshold. Two of the four "uncovered" lines were `src/mission_runtime/__init__.py:69` (an import) and `:106` (an `__all__` entry) — module-scope lines that execute on *any* import of the package, reported at 0.0%.

The other half of that failure was a genuine collection gap and is fixed in PR #2974. This collision is the remaining half and is deliberately out of that PR's scope.

## Blast radius (unverified, worth checking)

Any root-level basename shared between the co-instrumented packages is a candidate. In `src/` that plausibly includes `resolution.py`, `context.py`, `versioning.py`, and any `__init__.py`. Nobody has enumerated the real overlap set — the first task here is to measure it, because the affected files are currently reporting *someone else's* coverage numbers.

## Direction

The durable fix is to restructure the shared multi-package `--cov` invocation so each source root is disambiguated in the report — for example separate coverage runs/reports per package, or a coverage configuration that preserves the source root in the path key. This is more than a workflow one-liner, which is why it is filed rather than folded.

**Acceptance sketch**
- Enumerate the real collision set across all co-instrumented packages (do not assume it is only `__init__.py`).
- A report produced by the shared invocation shows `mission_runtime/__init__.py` with its own 18 statements, not `specify_cli/__init__.py`'s 218.
- A regression guard that fails if two source roots collide on one Cobertura key — otherwise this silently returns the next time a package is added to the `--cov` list.

## Related

- **#2974** — fixes the collection-gap half of the same CI failure.
- **#2957** — CI shard selection never collects several test files, making frozen-contract gates vacuous. Same *family* (CI measuring something it cannot see), different mechanism: that one is selection blindness, this one is report-generation data loss.
- The 2026-07-27 quality review flagged an unexplained coverage drop (64.6% → 47.3%) across a month with no Sonar analyses, suspected to be lost shard artifacts. Also the same family, also unproven. **Three instances in one week suggests the coverage pipeline deserves one deliberate audit rather than three point fixes.**

stijn-dejongh:
Closes #2975

## Why

Coverage reports have been silently losing data. When one `pytest-cov` invocation instruments two source roots, `coverage.py`'s Cobertura writer keys root-level files by `(package=".", filename)` with no disambiguation (`xmlreport.py:254` — a bare dict assignment, no collision detection), so same-named files across roots overwrite each other.

This is upstream **[coveragepy#1508](https://github.com/coveragepy/coveragepy/issues/1508)**, open since 2022 with zero maintainer replies and unchanged on `master`. It will not be fixed upstream; we work around it.

**Measured on this repo before the fix**, running the real `slow-tests` invocation (`--cov=src/specify_cli --cov=src/charter --cov=src/doctrine`):

```
<sources> = ['src/charter', 'src/doctrine', 'src/specify_cli']
header lines-valid = 108519      summed <class> statements = 108349
                                 → 170 statements lost, 6 files absent
```

It is silent because the XML is internally inconsistent: package aggregates count *both* sides of a collision while only one `<class>` survives, so a consumer reading the header and one summing the classes disagree, and neither errors.

**Three consumers, three different failure modes:**
- **Sonar** — `CoberturaParser.resolve()` finds the bare filename under several `<source>` roots, logs *"ambiguity, the file exists in several 'source'"*, and **drops the file entirely**. So coverage.py loses one file and Sonar discards the survivor.
- **The normalize step** — assigns each `<class>` the *first* source root under which the basename exists, while the dict-overwrite winner is the alphabetically *last* path. Opposite orderings, so it **relabels surviving data with the wrong file's path**, turning a visible Sonar error into silent wrong data.
- **`diff-cover`** — indexes the survivor under *every* source prefix with no existence check, returning one file's line data under another's path. This is the complete, deterministic mechanism behind PR #2963's diff-coverage gate failure.

## The fix

**`--cov=<dotted.module>` instead of `--cov=src/<path>`** on the multi-root invocations.

The causal rule was isolated empirically: the collision occurs **iff** `XmlReporter.source_paths` is non-empty, and that set is populated only from `config.source` entries that are *existing directories*. A module-form value is not a directory at repo root, so `source_paths` stays empty, `rel_name` becomes the full project-relative path, and the collision is structurally impossible.

Why this over the alternatives — all four were sandbox-tested, and all four fix the collision:

| Option | Keeps per-job scope? | Note |
|---|---|---|
| **`--cov=<module>`** ✅ | **Yes — identical measured set** | ~10 workflow lines, no config change |
| `source_pkgs` + bare `--cov` | No — every job measures all of `src` | shifts every denominator |
| `include=` globs + bare `--cov` | No | same |
| one root per job, separate reports | Yes | ~6 new jobs / 10 extra pytest passes |

**A config-only fix is impossible**, which settled this: `source_pkgs` in `pyproject.toml` is *fully overridden* by an explicit `--cov=<path>`, reproducing the broken output byte-identically. The flags themselves had to change.

Verified against the three risks that could have sunk the module form: dotted subpackages work; never-imported files still appear at 0%, so **the denominator does not shrink**; and a package unimported by the selected tests still enumerates fully.

Migration is **all-or-nothing per invocation** — a mixed invocation still emits two `<source>` entries and preserves the Sonar-ambiguity precondition. Single-root invocations are deliberately left as paths (smallest blast radius; the new guard covers future additions), and `--cov=scripts/docs` stays a path because it is not importable.

## Also in this PR

**Two dead `--cov` roots.** `--cov=src/specify_cli/mission` and `--cov=src/specify_cli/mission_metadata` point at paths that do not exist — the real surfaces are `mission.py` and `mission_metadata.py`. Coverage emits `CoverageWarning: Module ... was never imported` and exits 0, so `fast-tests-missions` and `integration-tests-missions` have been measuring **zero** for their two primary targets. That is also *why* those jobs showed no collisions: the roots were dropped before `source_paths` was built, so a clean result there was a symptom rather than health.

**The normalizer's multi-source branch is deleted**, not documented around (DIRECTIVE_043). Once no report carries ≥2 sources it is dead code, and while it lives it is an active misattribution hazard. The step also now drops an empty/blank `<sources>` block, since module form emits `<source/>`.

**A regression guard** (`tests/architectural/test_coverage_root_collisions.py`) reusing the canonical `_gate_coverage.load_workflow_models()` parser rather than writing a second one (DIRECTIVE_044). Two invariants, both as set equalities:
1. every `--cov` target resolves to a real directory or an importable module (`unresolvable == set()`);
2. no invocation's root-set produces a prefix-strip collision (`collisions == {}`).

Invariant 1 is the one that matters most — a collision-only check would have passed the two dead-root jobs while they measured nothing. Ships with a red-negative proving the guard can actually fail.

**Would it have caught this?** Yes, and the sharpest case is our own recent work: PR #2974 added `--cov=src/mission_runtime` to `fast-tests-core-misc`, escalating that job's `__init__.py` collision from 2-way to 3-way and newly introducing the `resolution.py` one. The guard would have gone red in that PR at the moment the root was added. It runs statically against the workflow plus the filesystem — no CI run, no coverage artifact needed.

## Scope — deliberately excluded

- **No Sonar threshold or quality-gate retuning.** This fix changes measured coverage *in both directions* (restoring discarded files, correcting misattributed numbers). Land it, observe one scheduled Sonar run, then adjust separately — retuning here would make the delta unattributable.
- **#2957** (test-*selection* blindness) — same family, different mechanism.
- **The 17.3pp historical coverage drop** — explicitly tested against this mechanism and **falsified**: the collision bounds at 0.23pp (1.3%), and the timeline excludes it. Separate defect, separate investigation.
- **No `src/` restructuring.**

## Evidence

**The fix, measured on the real 3-root `slow-tests` invocation:**

```
BEFORE  --cov=src/specify_cli --cov=src/charter --cov=src/doctrine
        lines-valid=108519   summed=108349   gap=170   files=1077
AFTER   --cov=specify_cli    --cov=charter    --cov=doctrine
        lines-valid=108519   summed=108519   gap=0     files=1083
```

Header and classes now agree, and the 6 previously-absent files are back.

**The guard genuinely reds** — run against fault-injected fixtures, not merely asserted:

```
FAULT (src/charter + src/doctrine, path-form):
  GUARD RAISED: path-form --cov roots collide on a real relative path:
  {('fixture.yml:fixture-job', '__init__.py'): ('src/charter', 'src/doctrine'), ...}
FAULT (dead root src/this_package_does_not_exist):
  GUARD RAISED: unresolvable --cov targets: ['src/this_package_does_not_exist']
REAL ci-quality.yml (post-fix):
  collisions: {}   unresolvable: []
```

**Dotted `.py` modules work** (this was unproven going in, so it was tested rather than assumed): `--cov=specify_cli.mission --cov=specify_cli.mission_metadata --cov=doctrine.missions` yields `mission.py: 317 statements`, `mission_metadata.py: 270 statements`, header `lines-valid=1280` == summed `1280`.

**Local runs:** the new guard plus the four modified architectural tests plus `tests/release/test_coverage_topology_ownership.py` — 49/49 passed; independently re-run at 42/42 on the final tip. `ruff check` clean on all five Python files; `ci-quality.yml` parses as valid YAML.

### Why four existing test files changed — both are re-pins the fix *requires*, not relaxations

This is the part worth a reviewer's attention, since a guard PR that quietly loosens neighbouring guards would be worse than no guard. Both were proven empirically by reverting them against the fixed workflow:

- **`test_workflow_coherence.py`** — with the old un-normalized `critical_path_backed_by`, `test_diff_cover_critical_paths_are_cov_backed_live` goes genuinely **RED** (`UNBACKED: ['src/charter/*', 'src/mission_runtime/*']`). Dot-ification broke backing detection; the fix restores it without touching the assertion.
- **`test_coverage_consumer_needs.py`** — the dangerous one. With the old bare `target.startswith("src/")` predicate, the emitter count drops **29 vs 39** and the test **stays green while silently checking 10 fewer jobs**. That is precisely the "quietly relaxed guard" failure mode — except here it is the *absence* of the change that causes it. The fix restores the original scope.
- **`_gate_coverage.py`** (the canonical parser) — extended, not weakened: a verbatim hoist of the pre-existing non-emitter set plus two normalization helpers. They live here because two *pre-existing* consumers need the same normalization; duplicating it in three places would violate DIRECTIVE_044.
- **`test_gate_coverage_parse_model.py`** — pure addition: unit tests for the two new helpers, no existing assertion touched.

## Reviewer note — one thing CI must confirm

Module-form `--cov` resolves through the installed package. This assumes CI's `uv sync --frozen --all-extras` installs the root project **editably**, so filenames resolve to `<repo>/src/...` rather than site-packages. Very likely (uv installs the root project editable by default), but a non-editable install would put filenames outside the repo and break Sonar's path matching. **The first CI run on this PR settles it** — check that coverage XML filenames start with `src/`.

https://claude.ai/code/session_01TF3TAv5kQ3tMg5cKV4aa7F

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/2982"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787739560&installation_model_id=427282&pr_number=2982&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F2982&signature=21f6c889126d8623cb2bafada761e413bba09bd7ef2175b5adf0bb0d3549d697"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/ci-quality.yml
+++ b/.github/workflows/ci-quality.yml
@@ -3471,19 +3471,39 @@ jobs:
       - name: "[ENFORCED] Normalize coverage XML paths for Sonar multi-package src layout"
         if: steps.sonar_token.outputs.enabled == 'true'
         run: |
-          # Problem: each CI test job runs with --cov=src/<package> which produces a
-          # coverage XML whose <sources> block lists the per-package absolute path, e.g.:
+          # Historical problem (pre-#2975): jobs ran with --cov=src/<package>,
+          # which produced coverage XML whose <sources> block listed multiple
+          # per-package absolute paths, e.g.:
           #   <source>/github/workspace/src/specify_cli</source>
           #   <source>/github/workspace/src/mission_runtime</source>
-          # …and bare filenames like filename="__init__.py".
-          # SonarCloud tries to resolve every bare filename against ALL <source> entries
+          # …with bare filenames like filename="__init__.py". SonarCloud tries
+          # to resolve every bare filename against ALL <source> entries
           # simultaneously, finds __init__.py under multiple sources, and logs:
-          #   ERROR Cannot resolve the file path '__init__.py' of the coverage report,
-          #         ambiguity, the file exists in several 'source'.
-          # Fix: rewrite each XML so there is exactly ONE <source> (the common src/
-          # ancestor) and every filename is prefixed with its package sub-path, e.g.:
-          #   src/specify_cli/__init__.py, src/mission_runtime/__init__.py
-          # SonarCloud can then match each path to exactly one file in sonar.sources=src.
+          #   ERROR Cannot resolve the file path '__init__.py' of the coverage
+          #         report, ambiguity, the file exists in several 'source'.
+          #
+          # #2975 removed the root cause instead of guessing around it here:
+          # every multi-root --cov invocation was converted to dotted module
+          # form (--cov=specify_cli --cov=mission_runtime), which keeps
+          # coverage.py's XmlReporter.source_paths empty and emits full
+          # project-relative filenames with zero or one <source> entries. The
+          # multi-source rewrite branch that used to live in this step
+          # (DIRECTIVE_043: remove the mechanism, not document around it) is
+          # deleted -- it picked the winning root by testing basename
+          # existence in `human_sorted` <source> order while coverage.py's own
+          # dict-overwrite keeps the alphabetically LAST root, so it
+          # systematically relabelled surviving data with the WRONG file's
+          # path (verified by reproduction: a 2-statement package's data was
+          # relabelled as an 8-statement one, and the small package vanished
+          # from the report entirely).
+          #
+          # What remains: (1) relativize a single absolute <source> (still
+          # needed — coverage.py still emits an absolute CWD-rooted path for
+          # single-root/path-form invocations, e.g. --cov=src/kernel), and
+          # (2) drop an empty/blank <sources> block. Dotted module form emits
+          # a bare <source/> (no text) because source_paths never gets
+          # populated; Sonar's handling of an empty <source> element is
+          # unverified, so the block is removed rather than shipped.
           python3 - << 'PYEOF'
           import os, sys, glob, xml.etree.ElementTree as ET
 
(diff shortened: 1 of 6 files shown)


stijn-dejongh:
## CI green (30/30) — and the reviewer question is now settled with real artifact evidence

**First, an honest caveat: the `diff-coverage` gate passed *vacuously* on this PR.** Its log reads `No lines with coverage information in this diff` — this PR touches only workflow YAML and tests, no `src/`, so diff-cover had nothing to measure. Green there proves nothing about the fix, and I did not want that read as confirmation.

**So I checked the actual artifact instead.** Downloaded `e2e-cross-cutting-reports` from run `30257598840` — that is one of the dot-ified multi-root jobs (`specify_cli + charter + doctrine`):

```
sources:                    [None]          <- single empty <source/>, as predicted
classes:                    1083            <- was 1077 pre-fix; the 6 absent files are back
filenames starting 'src/':  1083 / 1083     <- editable install CONFIRMED
header lines-valid:         108519
summed <class> statements:  108519
gap:                        0               <- was 170
duplicate filenames:        none
```

This closes the one open question from the PR body. **CI's `uv sync --frozen --all-extras` does install the root project editably** — every filename resolves inside the repo checkout (`src/...`), not site-packages, so Sonar path matching is safe.

It also reproduces the local before/after in CI exactly: 1083 files and a zero gap, against 1077 files and a 170-statement gap before the change. The empty `<source/>` element appears as expected, which is what the normalizer change in `fd15845f6` handles.

Merge-ready from my side; **the operator merges.**

Issue #3001, opened 2026-07-27
Sync: 3,391 historical lane transitions permanently rejected by Team Kitty ingress (guards postdate the events)


stijn-dejongh:
## Summary

3,391 historical lane-transition events are permanently rejected by the hosted Team Kitty ingress because they do not satisfy FSM guards that were introduced after those events were written. They are not retryable and no operator action on the local side can clear them without falsifying the audit log.

This is the largest residual class from a full drain of a 324-mission dogfood corpus: after #2995 / #2998 opened the gate, `spec-kitty sync now` delivered 9,250 events with 0 transient and 0 terminal failures, but 4,141 were rejected. 3,391 of those are this class.

## Rejection breakdown

From the dispatch report (`sync now --report`), `error: PAYLOAD_INVALID`, `violation_type: transition_rule`, `field: transition`:

| Count | Message |
|------:|---------|
| 1,250 | `in_review -> in_progress requires review_ref` |
|   998 | `approved -> in_progress requires review_ref` |
|   750 | `for_review -> in_progress requires review_ref` |
|   391 | `review-rejection rollback in_review -> planned requires force=True` |
|     1 | `review-rejection rollback approved -> planned requires force=True` |
|     1 | `review-rejection rollback in_progress -> planned requires force=True` |

`spec-kitty doctor mission-state --teamspace-dry-run` reports the same class as 93 mission-level validation errors across 39 missions, and this is what keeps the dry-run red (`valid: false`) even when `--fix` reports `errors=0`.

## This is not validator-vs-runtime drift

Worth stating explicitly, because the sibling defect #2995 *was* exactly that. Here the local side agrees with the server: `src/specify_cli/status/validate.py:131` enforces the same rule —

```python
# Review ref check: for_review -> in_progress requires review_ref
if event.get("from_lane") == Lane.FOR_REVIEW and event.get("to_lane") == Lane.IN_PROGRESS and not event.get("review_ref"):
    findings.append(f"Event {event_id}: for_review->in_progress without review_ref")
```

So these rows genuinely predate or bypassed the guards. The corpus is honest about its own history; the rules simply got stricter afterwards.

## Why it needs a product decision, not a data fix

The obvious "fix" — backfilling `review_ref` and `force=True` onto historical events so they validate — **falsifies an append-only audit log**. Those reviews did not have refs and those rollbacks were not forced. Synthesising the fields would make the log assert things that did not happen, which defeats the purpose of event sourcing.

[...]

Issue #2985, opened 2026-07-27
P0: accept-time birth-cutover rewinds a finished WP to 'claimed' (seed event appended after the terminal transition)


stijn-dejongh:
## Summary

`spec-kitty accept` rewinds a **finished** work package back to `claimed`. The accept-time birth-cutover stamp added by #2978 (mission `runtime-state-birth-cutover-all-paths-01KYH654`, WP02) appends a historical `planned -> claimed` seed event to the END of the mission event log, after the WP's real `planned -> done`. The reducer replays in order, so the WP's canonical lane resolves to `claimed`.

**Confirmed live on `main`** at `d0a5bacf7` (post-merge of #2978). This is not branch-local.

## Evidence

`tests/specify_cli/acceptance/test_accept_idempotency.py::test_accept_converges_on_unchanged_tree[commit]`

* `main` **before** #2978 (`4af076bc9`): `3 passed`
* `main` **after** #2978 (`d0a5bacf7`): `1 failed, 2 passed`

```
AssertionError: second accept did not converge:
  {'not_done': ['WP01'],
   'activity': ["WP01: canonical lane is 'claimed', expected 'approved' or 'done'"]}
```

The event log written by that run — WP01 only, in file order:

```
planned -> done      actor=test-agent                          policy_metadata=False
(annotation)         actor=test-agent
planned -> claimed   actor=migration:backfill_runtime_state    policy_metadata=True   <-- appended AFTER done
(annotation)         actor=migration:backfill_runtime_state
```

Corroborating failure, same root cause — accept is no longer event-count neutral:

`tests/cross_cutting/misc/test_acceptance_support.py::test_accept_does_not_require_done_evidence_for_approved_wp`
```
assert len(read_events(feature_dir)) == before_events
E   assert 7 == 6
```

## Mechanism

`accept` (real-commit path) -> `_stamp_birth_cutover_for_accept` -> `stamp_accept_cutover` -> `cutover_mission` -> `backfill_runtime_state`.

The backfill reconstructs each WP's *pre-eviction* runtime state from legacy frontmatter / `tasks.md` and seeds it as a `planned -> claimed` transition carrying `policy_metadata`. That is correct for a migration replaying history into an empty log, but at the accept seam the log already contains the WP's real terminal transition. The seed is appended, not ordered, and the reducer is order-sensitive — so the last transition wins and a `done` WP becomes `claimed`.

[...]

Issue #2320, opened 2026-07-03
doctor mission-state --fix does not repair the SNAPSHOT_DRIFT TeamSpace blockers it flags (and writes into stale coord worktrees)


stijn-dejongh:
## Witnessed live (catfooding, 2026-07-03, spec-kitty 3.2.4 editable @ cf2e91e17)

`spec-kitty upgrade` surfaced the TeamSpace migration panel: *2 TeamSpace blocker(s), finding codes: SNAPSHOT_DRIFT* and recommended the sequence `doctor mission-state --audit` → `--fix` → `--teamspace-dry-run`.

Following that exact sequence does not converge:

1. `--audit` flags 2 missions with `SNAPSHOT_DRIFT` (error severity): `decompose-merge-god-module-01KVXHDK` (status.json frozen at event 11/88 — all WPs done in the log, `planned` in the snapshot) and `decompose-mission-god-module-01KVXHF8` (missing merge-time `mission_number`).
2. `--fix` runs, reports repairs — but re-auditing still shows **the same 2 blockers**. The fixer never re-materializes `status.json` from `status.events.jsonl`, which is the entire remediation (one `materialize(feature_dir)` call per mission; manual fix landed as PR #2319).
3. Worse, `--fix` wrote its repairs as **uncommitted changes into stale coordination worktrees** of unrelated long-merged missions (`.worktrees/mission-identity-seam-…-coord`, `.worktrees/wp-lane-state-machine-fsm-…-coord`, `execution-state-canonical-surface` tasks reorganized), leaving dirty orphan worktrees behind and the primary checkout untouched.

## Expected

- `--fix` remediates every finding class it advertises as blocking — for `SNAPSHOT_DRIFT` on a flattened mission (no `coordination_branch`), that is a read-events → `materialize()` rewrite of `status.json` at the mission's authoritative location.
- `--fix` respects the status-surface authority rules: it must not write into stale coord worktrees for missions whose `meta.json` carries no `coordination_branch`, and should skip (or at minimum report) worktrees that are orphaned.
- The recommended-sequence panel should only advertise `--fix` for finding codes the fixer can actually repair.

## Acceptance sketch

Audit→fix→audit converges to 0 TeamSpace blockers on a repo seeded with a stale-snapshot mission; a regression test pins that `--fix` resolves `SNAPSHOT_DRIFT` without touching any `.worktrees/` path for flattened missions.

[...]

PR #980, opened 2026-05-05
Add TeamSpace mission-state repair and dry-run

robertDouglass:
## Summary
- Adds deterministic `doctor mission-state --fix` repair for historical `kitty-specs/` mission state, including safe Git preflight checks, typed-row quarantine, legacy key cleanup, lane normalization, manifest emission, and production-style `status.json` rematerialization.
- Adds `doctor mission-state --teamspace-dry-run` to synthesize canonical TeamSpace envelopes in memory and validate them when `spec-kitty-events>=5.0.0` is available.
- Aligns CLI sync expectations with the canonical TeamSpace event shape: `WPStatusChanged` keeps mission identity out of payload, and `MissionClosed` omits historical close details.
- Bumps the release candidate to `3.2.0rc2`; keeps the public `spec-kitty-events>=4.0.0,<6.0.0` range until the 5.0.0 wheel is published while gating dry-run validation at runtime.
- Stabilizes lifecycle JSON generation by allowing callers/tests to provide an explicit clock.

## Cross-repo rollout
- SaaS ingress contract PR: https://github.com/Priivacy-ai/spec-kitty-saas/pull/150
- Runtime side-log boundary PR: https://github.com/Priivacy-ai/spec-kitty-runtime/pull/19
- Tracker legacy guard PR: https://github.com/Priivacy-ai/spec-kitty-tracker/pull/14

## Follow-ups
- Publish `spec-kitty-events` 5.0.0: https://github.com/Priivacy-ai/spec-kitty-events/issues/20
- Tighten the CLI lower bound after publish: https://github.com/Priivacy-ai/spec-kitty/issues/978
- Coordinate the actual repository repair commit: https://github.com/Priivacy-ai/spec-kitty/issues/979

## Local verification
- `uv run python -m pytest tests/sync/ -m "fast and not windows_ci" -q --tb=short` -> 1203 passed, 3 skipped, 280 deselected
- `uv run python -m pytest tests/status/ tests/specify_cli/status/ -m "fast and not windows_ci" -q --tb=short` -> 700 passed, 138 deselected
- `uv run python -m pytest tests/release tests/migration/test_mission_state_repair.py tests/contract/test_cross_repo_consumers.py tests/contract/test_events_envelope_matches_resolved_version.py -q` -> 83 passed, 8 skipped
- `uv run python -m pytest tests/migration/test_mission_state_repair.py tests/specify_cli/status/test_lifecycle.py -q` -> 7 passed, 1 skipped
[...]

PR #3003, opened 2026-07-27
fix(mission-state): repair corpus + stop the repair tool destroying review_result and log order

stijn-dejongh:
## What

Repairs the mission-state corpus so persisted `status.json` files agree with the append-only event log that is their authority — **and fixes the three defects in the repair tool that were silently destroying data while doing it.**

Rebased onto current `main`. #2998 has merged, so this is no longer stacked.

Rebased onto current `main`. Two fixes to the repair tool, the corpus they
produce, and two documented course corrections (see **Review findings**).

| Area | What |
|---|---|
| `fix(migration)` | stop the repair dropping `review_result`, hoisting rows out of order, and deleting duplicates without quarantine |
| `revert(status)` | drop an earlier terminal-lane reducer guard — wrong seam, suppressed the durable fix |
| `fix(corpus)` | repair the 6 rewound work packages via `migrate backfill-runtime-state` (7 real repair events) |
| `chore(corpus)` | regenerated corpus + this run's quarantine |
| `docs` | changelog, and accurate wording for backfill seed ids |

## Why the tool needed fixing first

The original revision of this PR was data-only, produced by the unfixed tool. That data was lossy in three ways.

### 1. `review_result` was destroyed on 333 events across 44 missions

`_build_canonical_row` is a **closed allowlist**. `review_result` is a first-class `StatusEvent` field (`status/models.py:328`, serialized at `:353`, parsed at `:371`) but was absent from that allowlist, so it was dropped with no error, no action string, and no quarantine.

It is also a **hard FSM guard**: `wp_state.py:612-628` rejects every transition out of `in_review` without it. Measured against the merge-base, the repair turned **268 valid transitions into invalid ones, and zero were failing beforehand.**

> This corrects an attribution in the earlier revision of this description. The residual errors were blamed on "historical lane transitions whose FSM guards postdate the events" (#3001). That was wrong for this class — the repair manufactured them.

The loss was worst-shaped: `DoneEvidence.review` mirrors the payload only on approval, so the **23 `changes_requested` rows were the irrecoverable ones** — the entire rejection review history. All 333 payloads were schema-conforming; nothing was malformed.

Fixed by adding the field, plus a gate asserting the allowlist covers `StatusEvent.__dataclass_fields__` so a newly added model field cannot be dropped the same way.

### 2. 839 rows were hoisted to the head of an append-only log

Rows were sorted on `at` alone. Preserved lifecycle and retrospective rows date themselves with `timestamp`, so all 839 of them collapsed to `""` and sorted to the top — reordering 306 of 324 logs. Fixed by sorting on `at` **or** `timestamp`.

### 3. Dropped duplicate `event_id` rows bypassed quarantine

This corpus's 5 duplicates were byte-identical, so nothing was lost, but a divergent duplicate would have been deleted leaving only a sha256. Now quarantined before dropping.

### 4. Six work packages were rewound `done → claimed`

`046` WP08 and `048` WP01–WP05. Root cause is historical data, not this repair. Backfill seeds carry deterministic, digest-derived ids, and current seeds are ordered so a seed loses the `(at, event_id)` tiebreak to real history (#2985). These rows predate that guarantee: their seed id sorts *after* the genuine ULID on a byte-identical timestamp, so an unforced `planned → claimed` backfill won. Mission 048 was publishing as `done: 3, claimed: 5` — a merged mission reading half-unstarted.

`done`/`canceled` are terminal: leaving them requires `force`. The reducer enforced that for sequential events but not concurrent ones. `_should_apply_event` now refuses a concurrent, unforced event that would leave a terminal lane. A deliberate forced rewind still applies, covered by its own test.

Fixed the way the architecture already provides for: `migrate
backfill-runtime-state` appends compatibility-repair events, so the correction is
durable and auditable rather than applied invisibly at read time. An earlier
revision of this PR guarded the reducer instead; that was wrong and is reverted —
see **Review findings**.

## Verification

Measured over all 324 missions. Read this table as **fixed tool vs. unfixed tool** — i.e. what this PR lands versus what the previous revision of this PR would have landed. It is *not* a claim that the corpus recovers data from `main`: `main` already holds all 338 `review_result` values and this PR preserves them byte-identically. Commit 1 is **preventive for the corpus** and corrective only relative to the abandoned data-only revision.

| Property | Unfixed tool | This PR |
|---|---|---|
| `review_result` records destroyed | 333 | **0** |
| Transitions made invalid *by the repair* | 268 | **0** |
| Logs with rows out of chronological order | 306 | **0** |
| `status.json` ≠ its own log's reduction | 306 | **0** |
| Work packages reading rewound | 6 | **0** |
| Event ids changed | — | **0** (18,346 preserved) |
| Duplicate event ids | 5 | **0** |
| Rows lost (excl. quarantined mirrors) | — | **0** |
| Second `--fix` run | not converged | **`updated=0, unchanged=325`** |

Losslessness is proven on a key that survives re-minting (`wp_id`, `at`,
`from_lane`, `to_lane`): all 14,662 lane rows and all 3,538 non-lane rows are
accounted for, leaving exactly the 31 quarantined `DecisionPoint*` mirrors.

The 41 remaining lane changes are all forward progressions — the genuine stale-snapshot drift this PR exists to correct. The bulk of the `status.json` churn is that resync: at the merge-base only **11 of 317** snapshots were self-consistent with their own log; at HEAD all **324** are.

**"Repaired" does not mean "fully replayable."** The corpus still contains 4,159 transitions that fail FSM replay — identical count at the merge-base and at HEAD. That includes 1,200 out-of-`in_review` rows whose `review_result` was never recorded, and 84 structurally illegal `for_review → done` / `for_review → approved` rows. All pre-existing, none touched here; #3001 and friends still apply.

Tests: 894 across `tests/status/` + `tests/migration/`, 1193 across `tests/architectural/`, ruff clean.

## Data scope

Audited for foreign content, given #3030 and #3031:

- Single `project_slug` (`spec-kitty`) and a single `project_uuid`. **No cross-project contamination** of the kind #3030 describes.
- **No** repo / remote / origin / url binding in any of the 324 `meta.json` files.
- Sibling-repo names (`spec-kitty-events`, …) appear only as free-text prose in review reasons and error messages — dependency versions, never an owning identity.
- No secrets: no tokens, JWTs, `Authorization:` fragments, connection strings or signed URLs. (`sk-…` regex hits are mission slugs, verified against full field values.)
- Hostnames limited to `github.com`; every issue/PR link resolves to `Priivacy-ai/spec-kitty`.
- No real email addresses at HEAD — and this PR *removes* the only 9 real ones (`stijn@…`, in `DecisionPointOpened.actor_id`), which rode on the pruned rows.
- Every absolute path resolves to a spec-kitty checkout, a pyenv toolchain path, or `/tmp` — no foreign repository, no customer project.

**Pre-existing PII, not introduced here — but it ships once ingress opens.** Two categories, both essentially unchanged from `main`:

1. **Absolute home paths → 6 contributors' usernames and directory layouts** (~350–420 occurrences depending on match method), in `policy_metadata.pre_review_gate.reason`, `record_path`, and `review_result.feedback_path`. Net delta for this PR: **−2**.
2. **Real contributor full names** in `evidence.review.reviewer` (mirrored into `reason`) across **282 missions** — tied to review timestamps and verdicts.

Neither is introduced by this PR, and on net it slightly *reduces* exposure. Flagging it because this PR is what makes these 631 files eligible for TeamSpace ingress, which makes pre-existing PII operationally relevant now — a #3031 consent concern, not a defect in this PR. Recommend scrubbing as a separate auditable pass alongside the #3031 work, rather than folding a historical-prose rewrite into this repair.

## Review findings and two course corrections

A four-lens adversarial squad reviewed this, then CI caught what the squad's
static analysis could not. The history deliberately keeps both wrong turns rather
than hiding them — the reasoning is the useful part.

**Correction 1 — the reducer guard was the wrong seam, and is dropped.** An
earlier revision added a terminal-lane guard to `_should_apply_event`. The squad
found it swallowed unforced review rejections that raced a `done`; fixing that
still left it colliding with `cutover-guard` and breaking the #2985 birth-cutover
regression test. The mechanism is one line in the cutover:

```python
if desired_lane is not None and desired_lane != current_lane:
    transition_repairs.append(StatusEvent(..., actor=COMPATIBILITY_REPAIR_ACTOR, ...))
```

A read-side guard makes the reduction already correct, so `desired == current`
and **no repair is ever written** — it suppresses the append-only fix that is
auditable, reaches the SaaS export, and is visible to `move-task`. The 6 rewound
work packages are instead fixed by the tool's own printed remedy,
`spec-kitty migrate backfill-runtime-state`, which appends 7 real
compatibility-repair events. `src/specify_cli/status/reducer.py` is now
byte-identical to `main`.

**Correction 2 — event-id realignment was wrong, and is dropped.** An earlier
revision re-minted event ids whose embedded ULID timestamp disagreed with their
`at`. Those ids are **not** malformed: they pass `ULID_PATTERN` and the canonical
`normalize_event_id`, and nothing rejected them at sync ingress. They are
digest-derived *by design* — `_seed_id` is a stable content-addressed identity the
cutover looks up by name, and seeds are ordered so a seed loses the tiebreak to
real history (#2985). Re-minting broke all three properties. Event identity in
this PR is now untouched: all 18,346 ids preserved.

**What survived refutation.** Zero fields lost a non-null value; every
`status.json` is byte-equal to its own reduction; the new sort key is
reduction-neutral; no mission added or removed. The confidentiality lens found no
foreign project, no secrets, and no new PII.

**Still open, out of scope here:**
- `cutover-guard` is **red on `main` already** — 24 "did not restore lane" / 19
  "witness missing" on a pristine checkout, versus **20 / 16** on this branch. This
  PR strictly reduces it and introduces no new class or mission. Left alone per the
  baseline-red policy.
- 84 structurally illegal `for_review → done` / `for_review → approved` rows
  remain, pre-existing. `validate_transition` would reject them at the write seam,
  which is where that class belongs.

## Notes for reviewers

- **31 `DecisionPoint*` mirror rows are pruned** — unchanged #980 behaviour. All 31 decision ids are present in `decisions/index.json` + `DM-*.md`. But that store keeps only the *final* answer, and 2 decision points were resolved twice: **4 intermediate `deferred` resolutions and their rationale exist nowhere else.** The earlier revision claimed pruned rows were "preserved under `.kittify/migrations/.../quarantine/`" — misleading, since `.gitignore:74` meant quarantine never reached the repo. Fixed properly: the 4th commit force-adds this run's quarantine, so all 36 removed rows (31 mirrors + 5 duplicates) are recoverable in-tree, per the c52a54294 precedent.
- **The run manifest is deliberately not committed.** At 6.25 MB (`1281d5ccdb4e7fda.json`) it is mostly per-row sha256 churn over 14k rows and is reproducible by re-running the repair; the quarantined rows are not reproducible at all, which is why they are in-tree and it is not.
- **98 structured `actor` dicts were flattened to strings** (e.g. `{"model":"opus","profile":"reviewer-renata",…}` → `"reviewer-renata"`). `_normalize_actor_field` is pre-existing and untouched here, but this commit is what applies it to 98 real events. All 98 originals are preserved verbatim in `policy_metadata.migration_original_actor` — verified 98/98 recoverable.
- **One ephemeral value cleared:** `coord-commit-integrity-01KY5JS8` `WP03.shell_pid` `3826352 → 0` — a dead runtime slot, and the only non-empty→empty transition anywhere in the 324-mission snapshot diff.
- **No mission is excluded.** The earlier revision skipped `read-side-placement-seam-migration-01KYHP67` as untracked on `main`; it is tracked on the current base and is repaired with the rest.
- **Reproducing this requires a standalone checkout.** `doctor mission-state --fix` deliberately re-anchors to the primary checkout (`_anchor_repair_root`, #2320) and will not write into a linked worktree. Use `uv sync --all-extras --group dev` — an ad-hoc `uv run --with pytest` misses extras like `respx`/`pytestarch` and produces false reds in `tests/architectural/`.
- **Event identity is unchanged** — all 18,346 ids preserved, so nothing re-delivers to a hosted TeamSpace as a new event. The 7 appended compatibility-repair rows are genuinely new events and will sync as such.
- `mypy` is clean on both changed modules. The project venv ships a broken compiled `mypyc` artifact, so run it isolated: `uvx --from mypy mypy --python-executable .venv/bin/python <files>`.
- No Sonar UI work is outstanding for this change.


stijn-dejongh:
## ⚠️ Do not merge — this PR commits a data regression, and a claim in its description is wrong

Converting to draft. Two corrections to the body above.

**1. The "zero cases of the reducer dropping events" claim is false.**

That check only looked for snapshots *trailing* the event log (`event_count` fresh-higher, 254/254). It never checked whether any lane moved **backwards**. Six work packages across two missions regress terminal → non-terminal in `049427e5b`:

| Mission | WPs rewound |
|---|---|
| `046-doctrine-artifact-domain-models` | WP08 `done → claimed` |
| `048-structured-agent-identity-and-constitution-profile-integration` | WP01–WP05 `done → claimed` |

In these two cases the persisted snapshot was **ahead of the log and correct**; re-deriving moved it backwards. That is the opposite direction from the one the body validated for, so the safety argument as written does not cover it.

**2. Root cause is #2985 / #2989, not this repair.**

A `migration:backfill_runtime_state` seed `planned → claimed` sits *after* the terminal transition in file order, and the reducer folds by append order. 316 of 323 corpus logs carry `backfill_runtime_state` rows; only these 2 have the seed-after-terminal ordering, so the blast radius is bounded — but this PR is the artifact that would freeze it into committed state.

**Revised sequence:** land #2985 first, then re-run `doctor mission-state --fix` on the corpus and force-update this branch. The repair is a regenerate, so that is cheap.

Also correcting the header: #2998 merged at 15:42Z, so this is no longer stacked.

Credit to the triage pass that caught it — the regression was found by diffing pre/post lanes across all 312 `status.json` files in the commit, which is the check the original verification should have run.

stijn-dejongh:
## CI triage: all four failures are inherited red main, not this PR

Failing jobs on this PR: `arch-adversarial (arch_shard_1)`, `e2e-cross-cutting`, `integration-tests-core-misc (specify-cli-rest)`, `quality-gate`.

The **identical** four-job set is failing on `main`:

| Run | SHA | Time | Result |
|---|---|---|---|
| main | `bb9c7d290` | 10:28Z | green — last green |
| main | `d0a5bacf7` | 11:40Z | same 4 failing |
| main | `f270e2d73` | 15:42Z (after #2998 merged) | same 4 failing |
| this PR | | 15:51Z | same 4 failing |

main broke at 11:40Z on `d0a5bacf7`, before this branch existed. #2998 introduced nothing, and neither does this PR.

Three of the four are #2985's signature:

- `test_accept_converges_on_unchanged_tree[commit]` — `WP01: canonical lane is 'claimed', expected 'approved' or 'done'`, the exact rewind #2985 describes
- `test_accept_does_not_require_done_evidence_for_approved_wp` — `assert 7 == 6`, accept appending an extra cutover event
- `arch-adversarial` / `test_no_raw_mission_spec_paths` on `accept.py:239` — the third verdict folded into #2985 from #2989

`quality-gate` is the aggregator.

## Caveat worth recording: CI did not validate the data change

Every suite that reads the real corpus (`fast-tests-*`, `integration-tests-*`) shows `skipping` on this PR — presumably path filters, since the diff touches only `kitty-specs/**`. Grepping the three failing job logs for `test_dogfood_corpus_backfilled` returns zero hits.

So the green-once-main-is-green signal here will be **weaker than it looks**: the corpus change is not covered by what runs. Run locally on this branch instead:

```
pytest tests/specify_cli/migration/test_dogfood_corpus_backfilled.py -q
6 passed in 36.98s
```

That passes, but it does not check the six terminal→`claimed` regressions documented above — those need #2985 fixed and the repair regenerated regardless.

**Status unchanged: hold. Blocked on #2985.**

stijn-dejongh:
## Correction: the mechanism I gave in the draft comment above is the falsified one

My earlier comment on this PR said the seed "sits *after* the terminal transition in file order, and the reducer folds by append order". That model was already falsified on #2985 at 12:28Z, before I wrote it. Disregard it.

**Actual mechanism:** `reducer.py:315` sorts by `(at, event_id)` and discards file position entirely. `_claim_anchors` (`backfill_runtime_state.py:384`) anchors the claim seed to the WP's first transition into `claimed`, falling back to the WP's *earliest transition of any lane* — which for a finished WP is its terminal transition. So the seed is stamped with a **byte-identical timestamp**, and the tie resolves on lexical `event_id`: real ULIDs begin `01`, hash-derived seed ids do not, so the seed usually wins.

Verified on both affected missions in this PR's diff:

| Mission / WP | terminal id | seed id | `at` | winner |
|---|---|---|---|---|
| `046-doctrine-artifact-domain-models` WP08 | `01KJVJMXAS71GBBJEQ3A2JDGPD` | `5YA0MG13XVWKRY6SEQZMAH09GR` | identical | seed |
| `048-structured-agent-identity...` WP01 | `01KK8DP1YQWYJJBTFE8HEP8577` | `2E420M42Z6ZC3MYNJBW365MM64` | identical | seed |

**Nothing changes for this PR:** the six rewinds are real, the hold stands, and the repair still needs regenerating after #2985 lands. Only my explanation of *why* was wrong.

One thing the corrected model does change: because the outcome is a per-`(mission_id, wp_id, field)` hash rather than an ordering property, a regenerated repair could rewind a *different* set of WPs if any seed ids change. Worth re-running the lane diff over all 312 `status.json` after regeneration rather than assuming these two missions are the whole blast radius.

stijn-dejongh:
The CI-coverage gap noted above is now tracked as #3008 — data-only `kitty-specs/**` PRs skip every corpus-reading suite, so a green check set on this PR would assert nothing about its 621 data files.

Issue #2994, opened 2026-07-27
collaboration.operating-procedures is unvalidated and unwired: of 50 declarations, ~10 are wirable and ~35 name no doctrine node at all


stijn-dejongh:
## `collaboration.operating-procedures` mints no DRG edge — 15 of 16 profiles name a procedure unreachable from the profile that runs it

Found by WP09 of mission `doctrine-silence-guards-01KYFV7Q`, independently reproduced.

### Measured

```
profiles declaring collaboration.operating-procedures : 16
agent_profile -> procedure edges in the shipped graph :  1
    agent_profile:doctrine-daphne --applies--> procedure:onboard-external-agent-to-pack
```

So 15 declarations mint nothing at all. And the sixteenth — the only one with an edge — was typed **`applies`**, which no activation traversal follows, so it was unreachable too. **Before WP09 retyped it, 16 of 16 were unreachable.**

### Why the single wired case is the useful evidence

WP09's measurement of the one edge that did exist shows what the other fifteen are missing. Cascading from `agent_profile:doctrine-daphne`:

```
cascade_activation_targets(graph, "agent_profile:doctrine-daphne", CascadeScope.all())
→ 17 directives, 39 tactics, 5 procedures, 5 styleguides, 6 templates, 3 toolguides, 1 paradigm
→ and NOT procedure:onboard-external-agent-to-pack
```

Activating the profile pulled seventy-six artefacts and missed the one procedure the profile's own initialization declaration says it runs. WP09 retyped that edge to `requires` and it is now reached. The remaining fifteen profiles have no edge to retype.

### Why this is a real defect and not cosmetic

`collaboration.operating-procedures` reads as a binding declaration — it is the field an author uses to say *"this profile operates by this procedure."* It loads, it validates, the id resolves. It simply never becomes a relationship, so nothing that traverses the graph can act on it.

That is the same failure mode as the PowerShell toolguide this mission also documents: **"resolves" was operationalised as id-lookup, not reachability.** A loadable, id-resolvable, unreachable artefact looks correct at every surface an author checks.

### Why WP09 correctly did not fix it

[...]

Issue #2986, opened 2026-07-27
runtime→doctrine boundary ratchet sees only module-level imports; 61 function-local imports across 30 files bypass it


stijn-dejongh:
## The runtime→doctrine boundary ratchet only sees module-level imports; 61 function-local imports have walked through it

Found while running mission `doctrine-silence-guards-01KYFV7Q` (WP08). Independently reproduced.

### The blind spot

ADR `2026-03-27-1` says runtime modules under `src/specify_cli/` reach doctrine only through the `charter.*` facade. `tests/architectural/test_runtime_charter_doctrine_boundary.py` enforces it — but inspects **module-level imports only**. Its own docstring names this as a deliberate scope choice.

A function-local `from doctrine.… import …` is therefore invisible to it:

```python
def _collect_org_layer_data(repo_root: Path) -> dict[str, object]:
    from doctrine.drg.validator import validate_dangling_references   # ratchet never sees this
```

That exact line shipped in this mission and passed the gate. It was caught by review, not by the ratchet built to catch it.

### Measured scope

```
function-local 'from doctrine.*' imports under src/specify_cli/: 61 across 30 files
```

(AST walk; excludes the exempt `specify_cli/doctrine/` subpackage. Counts `ImportFrom` nodes whose module starts with `doctrine` and which are not in the module body.)

So the boundary is enforced against one import style and not the other, and the unenforced style is the one people reach for when they hit a circular import — i.e. exactly the situation where the boundary matters most.

### Why this is worth a mission, not a patch

WP08's own change is net **−1** (its import moved to the facade). Widening the ratchet would light up 30 files at once, most far outside any current mission's surface. Each needs a real decision — route through the facade, or justify the exemption — and some may be genuine cycle-breaks that need the facade to grow first.

Suggested shape:
1. Widen the ratchet to walk all `ImportFrom` nodes, with a **frozen shrink-only baseline** of the 61 (growth fails, shrinkage warns — the pattern this repo already uses).
2. Drain the baseline in batches, promoting symbols onto the `charter.*` facades as needed.
3. Remove the baseline when it reaches zero.

Step 1 alone stops the bleeding and is small.

### Adjacent: the facade contract doc has no parity test either

[...]

Issue #2981, opened 2026-07-27
plural↔singular kind map restated in 4+ places; the charter renderer's copy has already drifted 2 kinds behind and fails open


stijn-dejongh:
## The plural↔singular kind map is restated in (at least) six places, and two copies have already drifted

Found while running mission `doctrine-silence-guards-01KYFV7Q` (WP08). Verified against the tree, not inferred.

### The already-live drift

`src/charter/_activation_render.py:270` `_singular_kind` carries a **local hand-written table of 8 entries**:

```python
# Local mapping mirrors the inverse of
# ``charter.activations._SINGULAR_TO_PLURAL_KIND``.  Kept local to
# avoid leaking a private symbol across modules; the canonical
# plural set is small and stable (8 kinds).
inverse = {
    "directives": "directive",
    ...
    "mission_step_contracts": "mission_step_contract",
}
return inverse.get(plural_kind, plural_kind)
```

The table it says it mirrors — `src/charter/activations.py:178` `_SINGULAR_TO_PLURAL_KIND` — has **10 entries**. Missing from the local copy: **`glossary_packs`** and **`anti_patterns`**.

So the comment's own justification is now false in two ways: the set is not 8, and it was not stable.

**And it fails open.** `inverse.get(plural_kind, plural_kind)` returns the plural unchanged on a miss, so a `glossary_packs` or `anti_patterns` artifact renders its `--include <kind>:<id>` fetch selector with the *plural* token. No error, no warning — a selector string that does not match the singular convention the surface documents.

This is the same failure shape the mission is closing elsewhere: a hand-restated table, one step outside the thing that would have caught it, degrading silently rather than loudly.

### It is the second instance of this exact drift

WP08 found the same pattern in `src/doctrine/drg/merge.py`: `_PLURAL_TO_SINGULAR` restated the org-pack kind universe and had drifted **two kinds behind** (`mission_types`, `glossary_packs`), crashing the merge with a bare `KeyError`. The in-repo fixture `tests/doctrine/fixtures/relationship_packs/augment-all-kinds-pack` could not be merged at all as a result — the "all kinds" fixture never exercised all kinds.

Two independent restatements, both drifted, and **both by the kinds that were added later**. That is not coincidence; it is what a hand-restated table does.

### The known sites

| # | Site | Shape |
|---|---|---|
| 1 | `charter/activations.py:178` `_SINGULAR_TO_PLURAL_KIND` | hand-written, 10 entries — currently the most complete |
[...]

Issue #2980, opened 2026-07-27
kitty-specs/ guards disagree: commit guard WARNS, move-task BLOCKS — and the bulk-edit occurrence-map exception holds at only one


stijn-dejongh:
## Two `kitty-specs/` guards disagree: the commit guard WARNS, `move-task` BLOCKS

Found while running mission `doctrine-silence-guards-01KYFV7Q` (WP07). Reproduced and traced to source.

### The behaviour

On an implementation lane branch, committing a `kitty-specs/` file:

```
[spec-kitty guard] WARNING: Protected path: kitty-specs/<mission>/occurrence_map.yaml
  — implementation branches must not modify kitty-specs/
```

…and **the commit succeeds**. Later, `spec-kitty agent tasks move-task <WP> --to for_review` **refuses** because the lane carries `kitty-specs/` divergence.

So the workflow lets you do the thing, then blocks you at the gate for having done it — after the work is committed, when unwinding is most expensive.

### Root cause: two independent authorities with different enforcement modes

1. **`src/specify_cli/policy/commit_guard.py:85-89`** appends a violation for any staged path under `kitty-specs/`. But at line 116, `if policy.mode == "warn"` downgrades every violation to a warning and returns `allowed=True`. `mode` defaults to `"warn"` in three places in `src/specify_cli/policy/config.py` (lines 27, 40, 74).

2. **`move-task`'s lane-hygiene guard** (`_list_wp_branch_mission_specs_changes`, per `tests/specify_cli/cli/commands/agent/test_lane_hygiene_content_diff.py`) compares `kitty-specs/` files by content against the planning-branch tip and hard-blocks the transition.

Neither is wrong in isolation. Together they produce a workflow with no consistent contract: the earliest point that could stop you only warns, and the enforcing point is several steps downstream.

### Why the warn-mode default is the wrong half to keep

For `kitty-specs/` specifically the warn default is not a graduated-rollout affordance — it is a trap. The commit guard is the only surface positioned to prevent the problem cheaply. Warning there and blocking later inverts the cost: the agent must route the artefact to the planning branch, revert on the lane, and merge back, all after committing. One WP in this mission spent three commits (`834da20a3`, `0a8c03418`, `922ddf66c`) doing exactly that unwind.

### The documented exception does not hold at the second gate

[...]

PR #3007, opened 2026-07-27
doctrine silence guards: close ten classes of declared-but-inert failure (mission A of 5)

### LINKED ISSUE #2957: CI shard selection never collects several tests/specify_cli/cli test files on main — frozen-contract gates are vacuous

### LINKED ISSUE #2962: mark-status cannot find subtask IDs the shipped software-dev tasks template tells you to author
## The contradiction

`src/doctrine/missions/software-dev/templates/tasks-template.md:20` instructs authors:

> Subtasks are **reference rows**, not checkboxes: record completion with `spec-kitty agent tasks mark-status <Txxx> --status done`. The reduced event-log snapshot is the sole subtask-completion authority — there is no `- [ ]` box to tick.

But `mark-status` validates the ID against `tasks.md` using the checkbox patterns in `src/specify_cli/core/subtask_rows.py:53-56`:

```python
UNCHECKED_SUBTASK_ROW = re.compile(r"^-\s*\[\s*\]\s*(T\d{3,})\b")
CHECKED_SUBTASK_ROW   = re.compile(r"^-\s*\[[xX]\]\s*(T\d{3,})\b")
```

A `tasks.md` authored as the template instructs — `T001 Define "slot" and "producer"…` — matches neither, so:

```
$ spec-kitty agent tasks mark-status T001 --status done --mission doctrine-silence-guards-01KYFV7Q
Error: No task IDs found in tasks.md: T001
```

## Why it compounds

`move-task --to for_review` then blocks on subtasks that can never be marked, so the only way forward is `--force` — on **every WP of every mission** using this template. That trains agents to reach for `--force` as routine, which is exactly what it should not be.

Reproduced on `doctrine-silence-guards-01KYFV7Q` (10 WPs): all five of WP01's subtasks failed to mark, and the transition required `--force` despite the work being genuinely complete.

## Probable origin

`subtask_rows.py`'s own docstring says that since **#2816 IC-10 (FR-016 / SC-010)** the lane-transition guard no longer blocks on checkbox rows — completion is *solely* event-sourced, and the row patterns survive only for "migration backfill, the `move-task --to planned` rollback writer, and the acceptance gate."

The template was updated to match that model. `mark-status`'s **ID-validation** path appears not to have been, so it still requires the legacy shape to locate an ID before recording it into the snapshot.

## Suggested direction

Either:

1. `mark-status` validates against the **authored `subtasks:` frontmatter roster** (`authored_subtask_roster`, already the guard's static-intent source) rather than parsing `tasks.md` prose — which matches the post-#2816 model; or
2. the template reverts to checkbox rows and its "not checkboxes" guidance is withdrawn.

(1) looks right — the frontmatter roster is the declared design intent and is already canonical for the guard.

Whichever way, the template and the parser should be pinned to each other by a test, since the drift was invisible until an agent tried to follow the template literally.

stijn-dejongh:
Closes #2957 — CI shard selection never collected several test files on `main`.
Closes #2962 — `mark-status` could not resolve subtask ids authored exactly as the shipped template instructs.

## What this does

Closes ten classes of **silent failure** in the doctrine layer — cases where something declared, loaded, and validated cleanly while doing nothing, and every test stayed green.

This is Mission A of a five-mission programme (`doctrine-canonical-structure-remediation-01KYEYSD`). It is deliberately first: the later missions add DRG edge fields and author new artefact kinds, and both would have shipped into surfaces that silently dropped them.

## The defects, measured

| | before | after |
|---|---|---|
| test nodes collected by ≥1 job on a push to `main` | **2,275 of 33,822** (93.3% uncollected, 1,966 files) | **all of them** |
| `_KIND_MAP` coverage in the DRG extractor | 11 of 16 `NodeKind` members — `anti_pattern`, `asset`, `glossary`, `glossary_pack`, `glossary_scope` dropped in silence | 16 of 16, derived from the enum |
| org-pack plural kinds that mint a URN | 10 of 12 — the other two crashed the merge with a bare `KeyError` | 12 of 12 |
| field-projection edges reaching the graph | **0%** — the producer's entire output was discarded at the bridge | 100% |
| built-in-source → pack-target edges | dropped silently: 0 warnings, 0 conflicts | resolved, or a typed conflict |
| bare org-pack target naming a styleguide | minted a phantom `directive:<id>` | `styleguide:<id>` |
| inert schema slots | 41 declared-but-unproduced, unmeasured | frozen shrink-only baseline, each routed to an owner |

Two more worth naming because they are not counts:

**A shipped fixture that could never be merged.** `tests/doctrine/fixtures/relationship_packs/augment-all-kinds-pack` raised `KeyError` on every merge because the bridge's hand-restated kind map had drifted two kinds behind. The "all kinds" fixture never exercised all kinds.

**A profile could not reach its own operating procedure.** Activating `agent_profile:doctrine-daphne` cascaded to 76 artefacts and missed the one procedure its own initialization declaration says it runs — the sole inbound edge was typed `applies`, which no activation traversal follows.

## The gate that caught the most

`tests/architectural/test_ci_collection_completeness.py` evaluates every CI job's real `if:` condition against the worst reachable filter state on a push to `main`, and fails if any collected node runs in no job. It reads no baseline and has no allowlist — proven by a test that parses its own AST to assert it never touches the ratchet machinery, then monkeypatches that machinery to raise and requires the computation to complete anyway.

Its first live run caught a test file added *by this mission* that ran in zero jobs.

## Scope note: 41 extra CI job-starts on protected-branch pushes

Closing the collection gap means a push to a protected branch now starts 49 of 50 suite jobs instead of 10. PRs are unaffected — path filtering is a PR optimisation, and `main` is where the frozen contracts must actually run.

The single-authority alternative (forcing the filter outputs true at the `changes` job) is rejected for two reasons. **Availability:** suite jobs are `always() && (group || push)`, so the scattered disjunct makes them independent of `changes` succeeding — if that job fails, `needs.changes.outputs.*` is empty and every group-gated job would skip under the outputs shape, where today they all still run. **Blast radius:** `changes.outputs` is consumed wholesale by the quality-gate decision script and by `consumer-compatibility`, which is deliberately the one group-gated job that must *not* run on push; forcing the outputs true would silently start it.

Both claims were verified independently, and the blast radius is wider than first stated: forcing the outputs true would not merely *start* `consumer-compatibility` — it would make `build-wheel`, `clean-install-verification`, `uv-lock-check` and both release suites **blocking** on every push to `main`, because `quality_gate_decision.py` gates `RELEASE_REQUIRED_JOBS` on the same `changes` outputs.

Two corrections to earlier drafts of this section, recorded because getting them wrong is the defect class this PR is about. A review round rejected the single-authority alternative on the grounds that the outputs resolve through a shell step the static model cannot read; a later round showed that is **false** — the push atom makes the expression statically resolvable. And the paragraph said to contain that false reasoning **never existed**: `git log -S` finds it at no commit. The rationale was simply absent, and is now recorded for the first time. A third shape neither round considered — generating the conditions from a `job -> [groups]` table with a `--check` mode, as this repo already does for schemas — is filed rather than attempted here.

## Review

Every work package was reviewed by an independent agent that verified claims by **mutation rather than inspection**. That mattered:

- an allowlist that looked shrink-only was **paddable** — adding a new field to it passed all 277 tests in its package
- a migration hint hard-coded to one shard passed **61 tests**, because every per-kind fragment exists so `is_file()` succeeded for all seven
- a baseline-freedom proof was **circumventable three ways**, and one shape (`open(file=…)`) defeated both halves of it
- the first attempt at the multi-caller fix wired **one of three** eligible callers and justified it with a uniqueness claim that did not hold — `charter status --json` still returned `errors: []` on a graph with a dangling endpoint

Four review cycles were rejected and re-worked. The measurements in this description are the reviewers' independent re-derivations, not the implementers'.

## Follow-ups filed, not folded

| | |
|---|---|
| [#2977](https://github.com/Priivacy-ai/spec-kitty/issues/2977) | three hand-restated DRG writers; one already drops the declared `DRGNode.tags` |
| [#2979](https://github.com/Priivacy-ai/spec-kitty/issues/2979) | tier markers invisible to CI's marker partition — four instances, one a gate whose purpose was proving other gates aren't inert |
| [#2980](https://github.com/Priivacy-ai/spec-kitty/issues/2980) | `kitty-specs/` guards disagree: the commit guard warns, `move-task` blocks |
| [#2981](https://github.com/Priivacy-ai/spec-kitty/issues/2981) | plural↔singular kind map in 4+ places; one copy drifted two kinds behind and fails open |
| [#2986](https://github.com/Priivacy-ai/spec-kitty/issues/2986) | the runtime→doctrine boundary ratchet reads only module-level imports; 61 function-local imports across 30 files bypass it |
| [#2994](https://github.com/Priivacy-ai/spec-kitty/issues/2994) | `collaboration.operating-procedures` is unvalidated: of 50 declarations, ~10 are wirable and ~35 name no doctrine node at all |

**Filed during the landing pass** (all milestone `3.2.x`):

| | |
|---|---|
| [#3009](https://github.com/Priivacy-ai/spec-kitty/issues/3009) | nine charter-**activated** artefacts were DRG orphans — activating them cascaded to nothing. Eight are resolved in this PR (see below); the ninth needs a ruling |
| [#3022](https://github.com/Priivacy-ai/spec-kitty/issues/3022) | extract the built-in packs into `spec-kitty-packs-open` (dogfood our own pack distribution) |
| [#3023](https://github.com/Priivacy-ai/spec-kitty/issues/3023) | a non-`built-in` home for repo-local doctrine. **Open adjudication:** the operator ruling points at the project tier, this issue proposes a separate pack — they are the same mechanism (register a non-`built-in` pack root) at different locations |
| [#3024](https://github.com/Priivacy-ai/spec-kitty/issues/3024) | migrate glossary entries into pack artefacts (`blocked_by` #3023) |
| [#3025](https://github.com/Priivacy-ai/spec-kitty/issues/3025) | per-mission-run doctrine activation |
| [#3026](https://github.com/Priivacy-ai/spec-kitty/issues/3026) | the inert-slot baseline caps the `unassigned` *token*, not the unfireable-owner *property* — `mission:<undecomposed>` is an uncapped 28-of-51 padding route |
| [#3027](https://github.com/Priivacy-ai/spec-kitty/issues/3027) | `tasks_mark_status` roster read has an undeclared dependency on `WORK_PACKAGE_TASK`/`TASKS_INDEX` partition co-membership |
| [#3036](https://github.com/Priivacy-ai/spec-kitty/issues/3036) | **P1 — an architectural gate *requires* the repo-coupling the shippable-doctrine rule forbids.** Proven by mutation: making `doctrine-daphne` repo-agnostic turns `test_no_dead_doctrine_paths.py:456` red. Prerequisite for any daphne cleanup |
| [#3037](https://github.com/Priivacy-ai/spec-kitty/issues/3037) | **P1 — the `asset` kind has no resolution or install path.** No `AssetRepository`; resolution returns ids, never paths; the one shipped asset is reached by a hard-coded repo path. So "ship executable logic as an asset" — the documented remedy — is not executable downstream |
| [#3038](https://github.com/Priivacy-ai/spec-kitty/issues/3038) | the project doctrine tier cannot host `agent_profile` or `asset` as DRG nodes |
| [#3039](https://github.com/Priivacy-ai/spec-kitty/issues/3039) | split `test_no_dead_doctrine_paths.py` — its Gate A is a `src/`-wide CLI gate wearing a doctrine name |

**Closed by this PR** (see the `Closes` lines at the top): **#2957**, the CI collection gap — the measurement in the table above is its resolution; and **#2962**, fixed in-branch as campsite work because it blocked `move-task` on every WP of every mission using the shipped `software-dev` template.

Two more are *cited* by this branch but deliberately **not** fixed here, so they stay open: **#2967** (the zero-producer lint matches producers by bare name — recorded in the occurrence-map schema where it surfaces as a false positive) and **#2976** (the schema generator picks its reference-enum vocabulary by an unrelated condition — recorded in the enum ratchet's docstring, which freezes the split on the safe side without adjudicating it).

## What a reviewer should look at hardest

1. **The 41 scattered `github.event_name == 'push'` disjuncts** in `ci-quality.yml`. Their only protection against a forgotten 42nd is the meta-test itself. That is deliberate — the protection is proportional to the property — but it is a real single point of failure and worth your own judgement.
2. **`charter status` keeps exit code 0** with a populated `errors` array. Argued as reporting-surface-not-gate: the command has no content-driven exit code at all and reports `Charter source: MISSING` at RC 0, so flipping it for one finding would be a partial authority. `doctor doctrine` is the gate and exits 1.
3. **`extra="forbid"` on `AgentProfile` is a behaviour break for external org packs.** A consumer pack with a stray key now fails to load where it previously loaded degraded. That is the intended outcome and `skipped_profiles` surfaces it rather than crashing — but blast-radius checking covered this repository only.

## Landing-pass folds (rebased onto `e97fc6ab9`, 2026-07-28)

Rebased after #3012 landed. **Everything folded below is the same defect class the mission is about — a declaration that loads, validates, and reaches nobody — found one layer up during landing.** That sentence is the boundary of this section.

**Rebase adjudication.** #3012 independently fixed the same `accept.py` birth-cutover defect (`58e62c85b`); upstream's version won, and this branch's duplicate test was reduced to the two properties upstream's suite lacks (no `git` round-trip; `CoordinationBranchDeleted` propagates). Red-first verified.

**Operator ruling (2026-07-28, #3009 remedy 4).** `toolguide:rtk-search-tooling` **removed outright** — artefact, guide, default-pack activation, and the compiled blocks that would otherwise name a deleted `source_path`. Eight activated-but-unreachable artefacts **wired** via seven authored edges in `_CURATED_ARTIFACT_EDGES` (eight resolved by seven because `DIRECTIVE_035` leaves the orphan set by becoming a *source*). Reachability **proven**, not inferred — 7/7 via `cascade_activation_targets`, re-derived independently by review. `styleguide:deployable-skill-authoring` is deliberately left enrolled: the onboarding procedure mentions "skill" zero times, so wiring it would invent a relationship rather than record one.

**The shipped DRG moves 311/774 → 310/781, ledgered not bumped** — entries (6) and (7) of the composition ledger in `test_extractor_projection.py`, plus the relation histogram (`requires` 262, `suggests` 337) and a fourth ceiling (`DOCUMENTED_ORPHAN_RESIDUAL` 30 → 21, whose stated justification named the artefact this PR deletes). No golden count moved without a ledger row.

**A consumer break, found by review and fixed.** The RTK retirement stranded `rtk-search-tooling` in every rc35-upgraded project's `activated_toolguides`; that resolver is fail-closed, so charter compilation would have raised `UnknownArtifactIdError`. Migration `3.2.6_retire_rtk_search_tooling` cleans all three surfaces (10 tests, auto-discovered).

**Coverage topology, not a missing test.** `src/charter/context.py:689-690` — the fail-loud branch this mission installs — *is* tested, but the test routes to `fast-tests-doctrine`, which emitted `--cov=src/doctrine` only. Both doctrine jobs now emit `--cov=doctrine --cov=charter` (canonical dotted form; the mixed path/dotted shape reintroduced the class #2975 removed and evaded its guard). Confirmed covered by review.

**Two decision records added** under `docs/plans/doctrine/`, so follow-on missions inherit decisions instead of re-deriving them:
- *charter-activation-reachability-assessment* — **185 artefacts are activated and `charter context --action` surfaces none of them.** Three parallel activation vocabularies; the action surface renders from the two that `charter activate` does not write. Consequence: adding edges fixes traversal but not that symptom.
- *built-in-doctrine-repo-coupling-audit* — the classified population (84 raw hits → **52** real across 7 files; class "legitimately repo-local" is **empty**), the four relocations-not-rewords, and the gate contradiction now filed as #3036. `review-gates.md`'s quick-check gained a second pattern covering a 114-hit blind spot it could not see.

**Deliberately NOT folded**, each with a tracked home: the asset-kind extension (#3037), the destination-tier decision (#3023), the gate contradiction (#3036), the project-tier projection gap (#3038), the gate split (#3039), the inert-slot cap (#3026). These change what every agent receives at every action boundary in every consumer project, or require an ADR amendment — mission scope, not landing folds.

**Reviewed by four profile-loaded lenses** (`reviewer-renata`, `architect-alphonso`, `doctrine-daphne`, `debugger-debbie`) at a pre-merge point-cut. They found seven real defects **in the folds themselves** — including two inverted edge dispositions, the half-converted coverage form, a fourth stale ledger, and the consumer break above — all corrected in-branch before this description was written.

## Landing-order constraint — this PR gates on #3003

`cutover-guard` is red and it is **not this PR's defect**. Reproduced on a clean
`upstream/main` worktree carrying none of this branch's code: identical failure, same 12 WPs
across `charter-mediated-doctrine-selection-01KRTZCA` and
`excise-doctrine-curation-and-inline-references-01KP54J6`. This PR only *touched* those two
mission directories, which pulled them into the guard's diff scope.

It is not foldable, and the guard's own printed remedy does not work: running
`spec-kitty migrate backfill-runtime-state --mission charter-mediated-doctrine-selection-01KRTZCA`
on clean main returns `Failed: 1` with the same "deterministic seed payload diverged" error and
changes **0 files**. The real fix is **#3003** (mission-state corpus repair); its prerequisite
#2998 is merged, and #3003 is **open and no longer draft** — its scope has since widened beyond
the corpus repair to also stop the repair tool destroying `review_result` and log order. A
partial two-mission fold here would additionally collide with #3003 over the same files.

**→ #3003 must land before this PR can go green.** Everything else on CI is green or pending.

Reference-chain note: **#2985** and **#3013** were closed by #3012, which this branch is rebased
onto. Anyone following citations into #3013 will find a body written before that merge, describing
them as open.

## Verification

Local, re-certified on the landed tip with a quiescent tree: `tests/doctrine/` + the doctrine-facing architectural gates + the new migration tests **3040 passed**; `tests/charter/` 1674 passed / 4 skipped; `check_docs_freshness --ci` **errors=0**; `description_length_check --strict` 0 violations of 533 pages; terminology guard 4 passed; `ruff` clean; `mypy --strict src/specify_cli src/charter src/doctrine` **Success: no issues found in 1088 source files** — with **zero suppressions added**.

The mission's 226 development commits were compacted to 53 at PR time; the branch now carries **88** commits against `upstream/main` — the difference is the rebase onto current main plus the landing-pass folds below. Churn squashed, every ATDD RED/GREEN pair preserved. Planning artefacts are byte-identical to the consolidated branch.

CI is the real proof for the collection change: it predicts what GitHub will do, and nothing local substitutes for the first push-triggered run.


stijn-dejongh:
## Landing pass — rebased onto `e97fc6ab9`, seven folds, adversarial squad run

**Verdict: merge-ready on green, with one landing-order constraint (below).** Rebased 77 commits onto current `upstream/main` after #3012 landed.

### Rebase adjudication

Two conflicts. The changelog was a both-added in the same `### 🐛 Fixed` section — both entries kept, resolved in the canonical `docs/changelog/CHANGELOG.md`.

`src/specify_cli/cli/commands/accept.py` was substantive: **#3012 independently fixed the same birth-cutover defect** (`58e62c85b`). Upstream's version won — it is more thorough, and it deliberately names `STATUS_STATE` as the kind the leg writes where this branch pinned `ACCEPTANCE_MATRIX`. Our duplicate test would have asserted *against* that deliberate choice, so it was reduced to the two properties upstream's suite does not cover: **no `git` round-trip** (the historical bug was that a non-answering `git rev-parse` downgraded the anchor to `None` and wrote the seed to the PRIMARY partition) and **`CoordinationBranchDeleted` propagates** (upstream's docstring asserts the fail-loud posture but ships no test). Red-first verified against `58e62c85b^`; the second test's pre-fix red is an `AttributeError`, so it witnesses the property rather than the historical bug.

### Folds

| Fold | Why |
|---|---|
| `--cov=charter` on both doctrine CI jobs | `diff-coverage` was blocking at 81.2%. `src/charter/context.py:689-690` **is** tested — by `tests/doctrine/drg/test_unknown_kind_fails_loudly.py` — but that file routes to `fast-tests-doctrine`, which emits `--cov=src/doctrine` only. A coverage-**measurement** hole, not a missing test; a duplicate test would have green-washed it. |
| `doctrine-daphne` prose ×3 | The profile ships to consumer repos and named two repo-local paths, one as a live gate (*"… `test_no_authored_applies_edge.py` refuses a new one"*). In a consumer repo that file does not exist — a profile promising absent enforcement, this mission's own defect class one level up. The new dead-path gate cannot catch it: it verifies such paths exist in **our** tree. Also replaced *"only 4-5 are backed by an edge"* with the measured figures (1 of 50 edge-backed from its declaring profile, 8 name a tactic, 36 name nothing), and fixed the new ledger rule being self-violating on the commit that introduced it. |
| `docs/api/bulk-edit-gate.md` | `field_path` landed in the schema, template and CHANGELOG but not in the doc the shipped bulk-edit skill points authors at — which documented neither `field_path` nor the `exceptions:` block at all. Mission B2 is the party that would have read it and not found the field. |
| Two totality tests for `rewrite_opposed_by` (#2977) | That module restates the `DRGEdge`/`DRGNode` field lists by hand and is the **org-pack migration** path, so a dropped field lands in a consumer's tree. Unifying the writers is *not* done here — it reaches doctrine through the `charter.*` facade (ADR `2026-03-27-1`), so consuming the extractor's private serializer would breach the boundary; that promotion is the architectural call #2977 asks for. The tests are green today and go red the moment B1 adds a field. Mutation-proved by adding B1's actual `is_symmetric`. |
| RTK toolguide removed outright | Operator ruling. Artefact, guide, default-pack activation, repo activation, and the two compiled `.kittify/charter/` blocks that would otherwise name a deleted `source_path`. Node dropped by **regenerating** the fragment. |
| Eight activated-but-unreachable artefacts wired (#3009) | Authored via `_CURATED_ARTIFACT_EDGES`; every edge follows an existing kind→kind pattern. **Reachability proven** via `cascade_activation_targets` — 7/7. Eight artefacts, seven edges: `DIRECTIVE_035` was itself on the list and leaves it by becoming a *source*. `paradigm:atomic-design` deliberately **not** wired — no defensible source exists, so wiring it would invent a relationship; it stays the sole enrolled orphan pending a ruling. |
| Docs index + page inventory | Registers the new assessment page and refreshes the `bulk-edit-gate` row. |

All count movements are **ledgered, not bumped**: entries (6) and (7) in the composition ledger, with the shipped-graph pin updated to cite that ledger as its authority. Pure `304/757/31 → 304/764/23`; shipped `311/774 → 310/781`.

### Adversarial squad — 4 profile-loaded lenses, read-only

- **`reviewer-renata`** (gate non-vacuity): could not fake any new gate without a visible diff; confirmed the self-mutation tests mutate **production** surfaces, not fixture copies. One MAJOR: the inert-slot baseline caps the `unassigned` *token* rather than the *property*, leaving `mission:<undecomposed>` an uncapped 28-of-51-entry padding hatch (`unassigned` is saturated at 23/23). **Not folded** — filed as follow-up; it hardens a gate that currently holds and nothing is padded today.
- **`architect-alphonso`** (structure/contracts): verified **both** of the PR's CI-architecture claims sound against the workflow file — the availability argument and the blast-radius argument. Found the `field_path` doc gap and #2977's live residual.
- **`doctrine-daphne`** (doctrine integrity): proved the WP09 `applies→requires` retype reachable by **two independent walks with a negative control**; confirmed 311 nodes / 774 edges exactly, graph fresh, 252/252 artefacts valid.
- **`debugger-debbie`** (claim rot / live evidence): re-derived the headline measurement on the new base — the structural claim survives **exactly** (`10 → 49 of 50` suite jobs), node counts rotted (33,822 → 34,237). No semantic conflict with #3012 (allowlist overlap zero, no new kind-blind reads).

**Every pre-rebase `arch-adversarial` red was self-inflicted by this PR, and was caught by this PR's own new gate.** The gate bit its author on its first live run. All six traced to one PR-added test file missing tier markers (instance #3 of #2979) plus the `accept.py` conflict; both shards are green locally (530 + 416 passed) and on CI.

### Evidence

`tests/doctrine/` 2986 passed · `tests/charter/` 1674 passed, 4 skipped · `test_ci_collection_completeness` 46 passed · `test_no_authored_applies_edge` + `test_relation_doc_parity` 46 passed · `test_no_dead_doctrine_paths` 19 passed · terminology guard 4 passed · `ruff` clean · `check_docs_freshness --ci` `errors=0`.

`mypy` could not be run in the landing worktree — broken `mypyc` wheel (`undefined symbol: CPyPy_messages___MessageBuilder___unexpected_typeddict_keys`, survives `--reinstall-package`). Environment fault, not code; CI covers it.

### Landing-order constraint — this PR gates on #3003

`cutover-guard` is red and is **pre-existing main breakage**, not this PR's. Reproduced on a clean `upstream/main` worktree with none of this PR's code: identical failure, same 12 WPs across `charter-mediated-doctrine-selection-01KRTZCA` and `excise-doctrine-curation-and-inline-references-01KP54J6`. The PR only *touched* those mission directories, which pulled them into the guard's diff scope.

It is **not foldable**, and the guard's own printed remedy does not work: `spec-kitty migrate backfill-runtime-state --mission charter-mediated-doctrine-selection-01KRTZCA` on clean main returns `Failed: 1` with the same "deterministic seed payload diverged" error and changes **0 files**. The real fix is **#3003** (corpus repair across 313 missions); its prerequisite #2998 is merged, but #3003 is still a draft. A partial two-mission fold would also collide with #3003 over the same files.

**→ #3003 must land before this PR can go green.**

### Follow-ups filed

#3022–#3025 (doctrine pack extraction, `spec-kitty-internal` pack, glossary migration, mission-run-scoped activation), all on `3.2.x`. Renata's inert-slot cap finding and Debbie's `TASKS_INDEX`/`WORK_PACKAGE_TASK` co-membership note still to file.

### For the operator

`docs/plans/doctrine/charter-activation-reachability-assessment.md` (added here) records a finding larger than #3009: **185 artefacts are activated in `.kittify/config.yaml` and `charter context --action implement --mission-type software-dev` surfaces none of them** — three parallel activation vocabularies, with the action surface rendering from the two that `charter activate` does not write. Consequence: adding inbound edges fixes traversal but not that symptom. The two structural fixes are deliberately **not** folded here — they change what every agent receives at every action boundary in every consumer project.

The operator merges; I have not.

stijn-dejongh:
## Closeout: post-#3003 rebase, review squad verdicts, and remediations

#3003 merged at 15:22Z, unblocking this PR's only remaining red. This comment records what the squads found and what changed as a result.

### The rebase was not routine

An early branch commit (`test(WP01): failing-first zero-producer lint`, whose actual deliverable is a 214-line test file) **deletes 7,963 files / 1.1M lines under `kitty-specs/`** relative to its parent, and a later commit restores them — almost certainly an artefact of the mission being built in a lane worktree holding only its own mission dir. Net effect on the branch: 28 files. Harmless for as long as the deleted content matched main.

**#3003 changed that.** It rewrote ~677 of those exact files. A naive replay would have deleted the repaired versions and restored stale ones — silently reverting the corpus repair, on a branch whose CI would then have gone green.

Three conflict waves, and the correct side was **different each time**:

| Wave | Resolved to | Evidence |
|---|---|---|
| 620 modify/delete | upstream (keep file) | branch net-deletes **0** files under `kitty-specs/` at both tips |
| 4 add/add — the two missions `cutover-guard` was failing on | upstream | upstream `event_count: 82`, byte-identical to new main; branch's `73` was old main's value carried through the round-trip |
| 615 add/add (the restore commit) | already-replayed / repaired | verified: ours `== c2d6fdca5`, theirs `== e97fc6ab9` |
| 2 add/add — this PR's own mission | the branch | ours `event_count: 80` matches pre-rebase tip; main holds an empty `0` stub materialised from the 5-line log main could see |

**675 of 677** of #3003's repairs are preserved at the new tip. The two exceptions are this PR's own mission's status files, which the branch legitimately owns.

### Squad verdicts

**`reviewer-renata` — adversarial review of the rebase resolutions.** Verdict: *the rebase preserved both intents.* The decisive check:

```
per-commit patch-id mismatches (outside kitty-specs): 0
```

All 90 commits patch-identical to their pre-rebase counterparts, in identical order. The entire textual delta across the 17,969-line branch diff is 12 lines of CHANGELOG blob hashes and hunk offsets, shifted by 15 lines upstream inserted ahead of ours. The 24 RED/GREEN ATDD commits sit at identical positions at both tips — no reordering, no drops. All four judgement calls survived falsification. She also corrected the fold count: **10, not 9**.

One MINOR: 6 `DecisionPoint*` mirror rows that upstream pruned corpus-wide and quarantined were back in the live log, making this mission the corpus's only holdout.

**`debugger-debbie` — CI and gate verification at a frozen SHA.** Verdict: rebase sound, **one genuine PR-caused merge blocker**. She also corrected two claims made in this PR's own thread:

- **`cutover-guard`'s local "0 missions touched" was a false green.** `locate_project_root()` collapsed to the *outer* checkout (on an unrelated branch), so the CLI diffed the wrong tree. Correct in-process result: **5 missions touched, 28 files, `passed=True`** — and CI's own `cutover-guard` job passed in 58s. The verdict was right; the reported figure was meaningless.
- **"28 files, all in our own mission dir" was wrong** — 5 mission dirs. The extra 6 files across 4 other missions are legitimate authored cross-mission handoff notes (`c1d505a26`, `2d3fb2565`), with **zero** overlap against the 677-file repair set. Both lenses confirmed independently that nothing was clobbered.

`mypy --strict` reports 28 errors, all in `src/specify_cli/migration/backfill_runtime_state.py` — **identical at `c2d6fdca5`**, untouched by this PR. Pre-existing, tracked debt. `fast-tests-charter` passed; the earlier `spec-kitty-test-venv.lock` timeout was a one-off under Blacksmith host contention, not a pattern.

### Remediations applied

1. **Merge blocker (`arch_shard_2` → `quality-gate`).** `test_ratchet_baselines.py`: `category_1_auto_discovered_migrations baseline=95 current=96`. Registering the new RTK migration in the dead-modules category grew the allowlist without the matching `_baselines.yaml` bump. Confirmed PR-caused (passes at `c2d6fdca5`). Bumped 95→96 with justification. *One gate fixed, its ratchet twin tripped — exactly the coupling these ratchets exist to catch.*
2. **`DecisionPoint*` prune.** 6 mirror rows removed from the live log; all 6 remain in the shipped quarantine, `decisions/index.json` and the 3 `DM-*.md` files untouched. `event_count` held at **80 = 80 `to_lane` rows** through the edit. The canonical tool (`doctor mission-state --fix`) could **not** be used — it returns `MISSION_NOT_FOUND` from a linked worktree, same root cause as the false green above; hand-edit recorded honestly rather than claimed as tool-driven.
3. **RTK contradiction closed** (operator ruling). Deleting the `rtk-search-tooling` toolguide had left shipped, activated doctrine still telling consumers to use it. Excised the advocacy from `028-search-tool-discipline.directive.yaml` (replaced with a neutral principle naming no proxy tool), removed `rtk` from the `efficient-local-tooling` `commands:` list, and deleted the "Token-optimized proxies" section and its hook-based-enforcement subsection from `EFFICIENT_LOCAL_TOOLING.md`. The generalisable efficient-tooling guidance is kept. Graph did not move; `doctrine validate` 251/251.
4. **#3027 folded** — a P1 in `owning_wp_from_authored_roster`, a function *this PR introduces* (0 occurrences on `main`, 4 here). Resolved as a pinned assumption plus a guard rather than a one-line kind change, because three pre-existing callers share the same co-membership assumption and fixing only this one would have been a false economy. Mutation-proved: removing `WORK_PACKAGE_TASK` from its partition set reds the test with a message naming the required fix.

### Tooling defect surfaced

Two agents hit the same root cause independently on unrelated tasks: **CLI commands that resolve the project root return the outer checkout, not the linked worktree they run in.** Worktrees are the documented execution model (`spec-kitty implement WP##`, and the landing runbook's worktree-per-PR), so this exposes anyone following documented workflow — and a *guard* returning a confident false green is the dangerous half.

Adjudicated as **wrong resolver for the purpose**, not a broken resolver: `locate_project_root()` / `resolve_canonical_root()` correctly collapse worktree→primary for single-canonical-copy writes; the defect is that `cutover-guard` and `doctor mission-state` call the collapsing resolver when their job is to answer about the invoking checkout. Filed as #3051, parented under epic #2624 alongside the pre-existing #2613 and #3049 — three independent command families, one mechanism.

### State

Tip `b3a51988f`. CI was fully green at `68f98f2e3` (48 pass, 0 failures) before the RTK excision; that change is prose-only with the doctrine gates re-run locally (terminology guard + dead-paths 23 passed, `regenerate-graph --check` fresh, `validate` 251/251, `ruff` clean).

`Closes` lines re-verified against the tracker after the triage pass: **#2957 and #2962 are both still open, unsuperseded, and correctly targeted** — #2957's own triage note confirms it was folded into this mission as WP10.

The operator merges; I have not.

Issue #2824, opened 2026-07-20
accept: read_feature_dir resolves via PRIMARY_METADATA, not coord-aware, contradicting its own T028 comment


robertDouglass:
## Summary

`spec-kitty accept` fails to find `acceptance-matrix.json` for coord-topology missions even when it exists and is committed in the coordination worktree, because the resolver used for lane-gate reads is not actually coordination-aware — despite an inline comment claiming otherwise.

## Where

`src/specify_cli/acceptance/__init__.py`, around line 916:

```python
read_feature_dir = placement_seam(repo_root, feature).read_dir(MissionArtifactKind.PRIMARY_METADATA)
```

This resolves to the **primary checkout**'s `kitty-specs/<mission>/` directory. But the comment directly above the `_check_lane_gates(...)` call site (~line 1008) says:

```python
# T028: use coord-resolved read_feature_dir for lane-gate checks so that
# lanes.json and acceptance-matrix.json are read from the coordination
# worktree rather than the primary checkout when coord topology is active.
_check_lane_gates(
    repo_root,
    read_feature_dir,
    ...
)
```

`read_feature_dir` is the same variable passed to `_check_lane_gates`, but it was resolved with `MissionArtifactKind.PRIMARY_METADATA`, not `MissionArtifactKind.ACCEPTANCE_MATRIX` (the kind actually used elsewhere in `accept.py`'s own `_coord_worktree_root()` helper, which correctly resolves to the coordination worktree).

## Repro

For a coord-topology mission with `acceptance-matrix.json` populated and committed only in the coordination worktree (the canonical location per `write_acceptance_matrix`'s own placement):

```bash
spec-kitty accept --mission <coord-topology-mission-slug>
# -> "Acceptance matrix (acceptance-matrix.json) is required for lane-based
#     features but was not found"
```

Verified directly:

```python
from mission_runtime import MissionArtifactKind, placement_seam
seam = placement_seam(repo_root, mission_slug)
seam.read_dir(MissionArtifactKind.PRIMARY_METADATA)     # -> primary checkout (wrong for this read)
seam.read_dir(MissionArtifactKind.ACCEPTANCE_MATRIX)    # -> coordination worktree (correct)
```

## Suggested fix

[...]

Issue #1878, opened 2026-06-12
Umbrella: complete the coordination placement/identity strangler (post-3.2.0)


robertDouglass:
Parent epic: #1666
Origin mission: `coordination-merge-stabilization-01KTXRVR` (spec.md C-001 deferred non-goals; paula-analysis.md follow-up section; analysis-report.md finding I2)

This umbrella tracks the deliberately deferred remainder of the coordination placement/identity strangler after the 3.2.0 stabilization mission. Each item below is a known, bounded gap that was explicitly scoped out of mission `coordination-merge-stabilization-01KTXRVR` (C-001). None is a release blocker for 3.2.0.

## Deferred work items

1. **Resolver strangler completion** — migrate the remaining direct-placement call sites onto the canonical placement resolver: `gate.py`, `agent_retrospect.py`, and the `implement.py` C-004 fallback. *Note:* WP05 of mission `coordination-merge-stabilization-01KTXRVR` resolves `gate.py` and `agent_retrospect.py`; the residual scope here is the `implement.py` C-004 fallback (plus any sites WP05 documents as out of reach). Re-scope this item against WP05's landed state before starting.
2. **Single ref-advance helper rollout** — extend the unified ref-advance helper beyond the merge pipeline to every code path that fast-forwards or resets coordination/mission refs.
3. **Worktree-naming allocator unification** — one allocator for lane/coord/husk worktree names; no parallel naming logic.
4. **Is-a-worktree type invariant** — encode "this path is a worktree of kind X" as a type-level invariant instead of ad-hoc path checks.
5. **AC10 lint expansion** — widen the AC10 architectural lint to cover the new placement/identity seams.
6. **`_ensure_branch_checked_out` shim retirement** — remove the compatibility shim once all callers route through the resolver.
7. **#1827 crash-between-record-and-commit edge** — close the window where a crash between event-record and commit strands state.
8. **#1784 P3 polish not carried by WP02** — the low-priority rough edges enumerated on #1784 that mission WP02 does not pick up.

## Fresh evidence, 2026-06-12 planning session

Three further live instances of the #1784/Class A family ("primary checkout treated as the authority while artifacts live on the coordination branch") were observed first-hand while planning mission `coordination-merge-stabilization-01KTXRVR` itself:

[...]

Issue #3011, opened 2026-07-27
rekey_inventory.py is not round-trip-safe: overwrites hand-adjudicated surface-audit rows (mislabels accept.py FS-sink row)


stijn-dejongh:
## Context

`tests/architectural/surface_resolution_audit/rekey_inventory.py` is advertised as *the* canonical converter for re-keying `surface_resolution_audit/inventory.md`. During the read-side placement-seam migration mission (`read-side-placement-seam-migration-01KYHP67`) an implementer was told to use it to remove two ghost inventory rows; running it produced a diff that would have **destroyed hand-adjudicated inventory signal**, so the change had to be made surgically instead. Both the implementer and the reviewer independently concluded the converter is a trap.

## The defect

`_disposition()` (`rekey_inventory.py:49-60`) is a coarse three-rule classifier with hardcoded per-bucket rationale text. It **mechanically overwrites hand-adjudicated `disposition` / `handle source` / `rationale` cells** rather than preserving them. It is therefore not round-trip-safe over an inventory that contains adjudicated rows.

Concrete mislabel it introduces today — `specify_cli/cli/commands/accept.py` `_stamp_birth_cutover_for_accept`:

- Hand-adjudicated (correct): `raw join (FS sink: meta.json / status.events.jsonl writes)` / `topology-blind-by-design` — "Deliberately COORD-pinned … the PRIMARY leg DOES use `primary_feature_dir_for_mission`; this leg must not, because that primitive re-anchors via `get_main_repo_root`".
- Converter output (factually wrong): `raw-path-join` / `raw-bypass` — "composes `KITTY_SPECS_DIR/slug` inline ONLY for a fail-closed `StatusReadPathNotFound` diagnostic raise payload — the path is never opened (no FS sink; operationally safe)."

That callsite **does** have an FS sink, so the generated rationale is false, and the rewrite flips a real disposition count (raw-bypass 2→3, topology-blind 18→17).

Secondary: the converter churns the `file:line` jump-to column on every run even though the audit compares only the drift-proof `(rel_path, qualname, token)` composite (per the converter's own docstring), so `rekey_inventory.py --check` reports `STALE` permanently. Nothing gates `--check`, so this is latent documentation debt — but it means the tool can never be trusted as "run it and commit the result".

## Impact

[...]

Issue #2966, opened 2026-07-27
Write-target authority residuals: finish FR-002's PRIMARY reads, fold the 4th sibling, and port status/emit.py


stijn-dejongh:
## Context

Surfaced by the two-lens adversarial squad (architecture + SSOT) during the PR #2963 landing pass. #2963 delivers FR-001…FR-006 of placement-port-residuals-closure and its in-scope findings were folded on the branch. These four are **out of that PR's declared scope** and are filed here rather than crammed into a landing fold.

They share one root, which is also #2160's thesis: writers, validators and committers still do not all agree on a single canonical write-target authority.

## 1. Finish FR-002 — two PRIMARY reads still on the COORD leg

`backfill_runtime_state.py` moved the `tasks/` frontmatter read to the PRIMARY leg, but two callers still pass the COORD leg:

- `_mission_id` (`:513`) → `load_meta(feature_dir)`
- `_synthesize_claim_anchor` (`:453`) → `load_meta(feature_dir)`

`meta.json` is the `PRIMARY_METADATA` kind, and per this repo's own invariant (`core/paths.py:713-719`) the COORD leg has none. Observed consequence (reproduced, coord-rooted corpus):

```
SEEDED no-meta  : 5 ['5X9DDS5WW18GE6T50F2GEXY3S6', ...]
SEEDED with-meta: 5 ['7SZJ4PZBVXFXQ9BPFN6QJH0HBJ', ...]
WARNING specify_cli.status.store:258 No meta.json found for mission_slug 'coord-a'
        (orphaned event); mission_id will be None for these events
```

Deterministic seed ULIDs are namespaced by the coord **directory name** instead of the mission ULID, and the store itself reports the events as orphaned with `mission_id = None`.

**Attribution, precisely:** pre-existing — `_seed_phase(status_dir)` already passed the coord leg before #2963, and FR-002 scoped itself to "moving ONLY the `tasks/` frontmatter read". A stronger data-loss claim was tested and **falsified**: removing the coord `meta.json` still passes all three tests, because the anchor synthesizes from `shell_pid_created_at` first. The namespace defect is real; the data loss is not.

**Also:** `tests/specify_cli/migration/test_cutover_partition_decouple.py:47` writes a `meta.json` onto the coord leg that production never has, so the new tests certify a shape that does not occur. (The fixture de-masking is being folded on the PR branch; the two production reads are this issue's work.)

## 2. Fold in the fourth write-target sibling

[...]

PR #3012, opened 2026-07-27
Read-side placement seam: route mission reads through the kind-aware seam + structural gate

### LINKED ISSUE #2921: repair_lane_mismatch corrupts frontmatter: raw_frontmatter_text destructured as padding

### LINKED ISSUE #2922: Read-side whack-a-read (D-06): ~50 modules bypass the hardened read-surface authority

### LINKED ISSUE #2985: P0: accept-time birth-cutover rewinds a finished WP to 'claimed' (seed event appended after the terminal transition)

### LINKED ISSUE #3013: accept does not converge on an unchanged tree: second run rewinds WP01 to 'claimed' (2 pre-existing reds on main)
## `accept` does not converge on an unchanged tree — lane rewinds to `claimed` on the second run

Two pre-existing reds on `main`, surfaced while landing PR #3007. **Verified on clean `upstream/main` @ `f270e2d73`** in an isolated worktree — not introduced by that branch.

```
FAILED tests/specify_cli/acceptance/test_accept_idempotency.py::test_accept_converges_on_unchanged_tree[commit]
  second accept did not converge: WP01 canonical lane is 'claimed', expected 'approved' or 'done'

FAILED tests/cross_cutting/misc/test_acceptance_support.py::test_accept_does_not_require_done_evidence_for_approved_wp
```

Reproduce:

```
PYTHONPATH=src python -m pytest tests/specify_cli/acceptance/test_accept_idempotency.py -q
→ 1 failed, 5 passed
```

Only the `[commit]` parametrisation fails; the sibling parametrisations pass.

### Why it matters

`accept` is meant to be idempotent — running it twice on an unchanged tree should be a no-op. Instead the second run moves a WP **backwards** from `approved`/`done` to `claimed`. That is a lane rewind on a terminal-ish state, which is the same class as the accept-time birth-cutover rewind tracked in #2985.

Both failures are lane-state convergence, not path composition.

### Possibly adjacent — a real bug found next door

While folding an unrelated architectural red on the same file, PR #3007 found that `accept`'s **COORD** birth-cutover seed was anchored on a reconstructed path rather than the resolver's own answer:

`_coord_worktree_root` calls `mission_runtime.resolve_artifact_surface`, whose `resolved.path` **is** the coordination mission dir. The code discarded it, ran `git rev-parse --show-toplevel` to walk up to the worktree root, then re-joined `KITTY_SPECS_DIR / mission_slug` — using the **raw handle** rather than the canonical one. Measured with a working `git` stub:

```
reconstructed : …/kitty-specs/demo-mission
actual        : …/kitty-specs/demo-mission-01KYJGCQ
```

Second failure mode: when `git` does not answer, the resolution degrades to `None` and the status seed lands on the **PRIMARY** partition instead of COORD. The PRIMARY leg two lines above canonicalizes via `_canonicalize_primary_read_handle`; the COORD leg never did.

That is fixed in #3007 (`257e876f1`) by anchoring on the seam's own answer. **It did not fix these two reds** — they still fail with that change in place — but anyone investigating lane convergence in `accept` should know the COORD anchor was landing in the wrong directory until now, since a status seed written to a path nobody reads back is a plausible contributor to a lane appearing to rewind.

### Related

- **#2985** — P0: accept-time birth-cutover rewinds a finished WP to `claimed` (seed event appended after the terminal transition). Same symptom, possibly the same root cause; worth checking whether these two tests are that issue's regression surface rather than a separate defect.

stijn-dejongh:
# Read-Side Placement-Seam Migration

> Reviewer summary for PR `read-side-placement-seam-migration`. Written for a
> reader who has **not** followed the mission — the background below is enough to
> review the diff without extra context.

## Background (for a non-maintainer reader)

Spec Kitty stores each mission's files on one of two "partitions":

- **PRIMARY** — stable planning and metadata (spec, plan, `meta.json`), on the
  mission's planning branch.
- **COORD** — lifecycle surfaces (the append-only status event log, notes,
  traces), on a separate *coordination* branch.

A single component — the **placement seam** — is meant to be the one authority
answering "where does this artifact live?". You ask it for an artifact *kind*
(`STATUS_STATE`, `LANE_STATE`, `WORK_PACKAGE_TASK`, `PRIMARY_METADATA`, …) and it
resolves the right partition, **failing loud** if the partition it needs is gone
(e.g. a coordination branch consolidated away).

A previous change hardened that seam — but only for **writes**. On the read side,
~50 modules still called the low-level, *kind-blind* resolvers
(`candidate_feature_dir_for_mission`, `resolve_planning_read_dir`) directly. Those
resolvers don't know which partition an artifact belongs to, so a read could
silently fall back to the primary checkout when a coord-partition artifact's branch
was deleted — returning plausible-but-wrong data instead of failing. That is the
"whack-a-read" gap (#2922, parent #1878).

This PR closes it, and makes the bypass structurally impossible going forward.

## Summary

- Route every read that *should* fail loud through the kind-aware seam with the
  correct artifact kind — while deliberately leaving diagnostic/audit readers
  lenient, because reporting surfaces must not start crashing.
- Add a whole-tree gate so a *new* kind-blind read can't be introduced silently,
  with a shrink-only allow-list that forces its own entries to be deleted once
  they're no longer needed.
- Fold in two small related fixes that were blocked behind this work (#2921, and
  the part-1 remainder of #2966).

## Why Now

The write side was already enforced; the read side was still "hardened one caller
at a time", which is unbounded — every new module could reintroduce the bug. The
asymmetry was tracked as #2922 (parent #1878). Doing it as one mission let a
per-site classification be decided **once**, up front, and then be mechanically
enforced — rather than re-litigated in each future PR.

## What This PR Does

**1. Classify first (one authority).** A per-site ledger
(`docs/development/read-side-seam-classification.md`) records, for every production
caller of the kind-blind resolvers: the target artifact kind, the disposition
(*migrate-fail-loud* / *stay-lenient* / *sanction-infra*), and the rationale —
**72 migrate + 16 stay-lenient + 2 sanction-infra**. Every subsequent change was
executed and reviewed against this document.

**2. Migrate the 72 fail-loud sites** across five clusters (agent-CLI,
CLI-commands, merge+lanes, diagnostic, core/context/workspace/plan) to
`placement_seam(...).read_dir(<kind>)`. Coord-partition reads now fail loud: a
lane-based merge's event-log read, the decision-log companion read, and the
doctrine synthesizer must not treat a deleted coordination branch as healthy.

**3. Keep 16 readers lenient — on purpose.** Dashboard scan, dossier API (SaaS
facing, answers 404 for "not found"), retrospective summary, status aggregation,
the manifest drift probe, and several CLI diagnostics stay on the lenient resolver
with their existing degrade paths, recorded as justified allow-list entries.

**4. Make new bypasses unrepresentable.** A new whole-tree AST gate
(`tests/architectural/test_no_read_side_bypass.py`) reds on any direct kind-blind
read in a non-sanctioned `src/` module. It **reuses the write-side gate's scanner**
(a symmetry test proves runtime identity — no forked tree walk), asserts the three
read-primitive authority modules as *sanctioned* rather than silently skipping
them (with a non-vacuity test), and keys its allow-list by content descriptor
(no path blankets) with a **staleness twin-guard**: an entry that is no longer
needed reds until it is deleted.

**5. Two folded fixes.** `repair_lane_mismatch` no longer feeds raw frontmatter
text into the document-padding slot (it was duplicating the whole frontmatter into
the body on every legacy lane repair, #2921); and `backfill_runtime_state`'s
mission-id read is anchored on the PRIMARY leg (#2966 part-1 remainder).

- **Runtime / compatibility:** Behavior-preserving in the healthy case — for a
  materialized mission every migrated site resolves the same directory as before.
  The intended change is narrow: reads of **coord-partition** artifacts now raise
  instead of silently returning a primary-checkout substitute when the coordination
  branch is gone. Diagnostic/audit/SaaS readers are explicitly unchanged.
- **Upgrade / migration:** None. No data format, config, or CLI surface change.
- **Operator / reviewer impact:** A wrong-partition read now fails loudly instead
  of silently, and a future kind-blind read is caught by CI rather than by a
  reviewer noticing.

## Validation

- [x] Red-first per behavioral fix (each cluster proved which sites genuinely
      change behavior; a discriminating-negative test pins that primary-partition
      kinds stay primary-anchored even on a deleted-coord mission).
- [x] All nine work packages independently reviewed and approved
      (implement / review split; one WP was rejected and re-implemented for a real
      regression before approval).
- [x] Targeted gates green on the rebased tip: read gate + write gate + trio-seam
      + surface-audit + dead-symbols + terminology = **118 passed**; mission
      suites = **94 passed**.
- [x] `ruff` clean; `mypy --strict src/specify_cli src/charter src/doctrine` =
      **no issues in 1088 files**.
- [x] Docs freshness: `check_docs_freshness --ci` 0 errors; docs suite 602 passed.
- [x] Post-migration campsite pass closed every regression the migration itself
      caused (see below); each was classified against the pre-mission base before
      being fixed.
- [x] Follow-up work called out (see Follow-ups).

### Note on the campsite pass

Migrating the call sites left four *stale ratchets* behind — artifacts that
referenced the old shape: a gate allow-list still blessing resolver names nobody
imports anymore, two audit-inventory rows pointing at moved call sites, a public
export with no remaining importer, and eight test mocks patching helpers the
migration had removed. Each was verified to be **green before the mission and red
after** (i.e. genuinely caused here, not pre-existing) and then fixed by tightening
— shrinking allow-lists, unexporting, and repointing mocks at the real seam. No
assertion was weakened; the affected merge suites collect and pass the same count
as before the mission.

One honest pre-existing failure is **not** addressed here and stays red:
`test_no_raw_mission_spec_paths` (offender `cli/commands/accept.py:239`), which is
red on the pre-mission base too.

## Tickets / Contracts

| Ticket | Relationship |
|--------|--------------|
| #2922 | Closed — read-side whack-a-read migration (classification → migration → structural gate) |
| #1878 | Closed — parent of the read-side placement port |
| #2921 | Closed — `repair_lane_mismatch` frontmatter corruption |
| #2920 | Verified already fixed (pre-mission write+read seam hardening) |
| #2966 | Partially addressed — part-1 remainder (PRIMARY-leg mission-id read) fixed here; parts 2/3/4 deferred |
| #2964 | Deferred — `feature*` → `mission*` terminology migration, out of scope |
| #3011 | Filed during this mission — `rekey_inventory.py` is not round-trip-safe over hand-adjudicated audit rows |

## Mission Artifacts

- Spec: `kitty-specs/read-side-placement-seam-migration-01KYHP67/spec.md`
- Plan: `kitty-specs/read-side-placement-seam-migration-01KYHP67/plan.md`
- Tasks: `kitty-specs/read-side-placement-seam-migration-01KYHP67/tasks.md`
- Classification ledger: `docs/development/read-side-seam-classification.md`
- Acceptance matrix: `kitty-specs/read-side-placement-seam-migration-01KYHP67/acceptance-matrix.json` (12 criteria verified)
- Issue matrix: `kitty-specs/read-side-placement-seam-migration-01KYHP67/issue-matrix.md`
- Reviews: `kitty-specs/read-side-placement-seam-migration-01KYHP67/tasks/WP*/review-cycle-*.md`
- PR summary: this file (`pr-summary.md`)

## Follow-ups

- **#2966** parts 2/3/4 — remaining write-target / read-leg consolidation.
- **#2964** — `feature*` → `mission*` terminology migration (separate bulk-edit mission).
- **#3011** — make `rekey_inventory.py` round-trip-safe (it currently overwrites
  hand-adjudicated audit rows; the surface-audit edit in this PR was therefore made
  surgically, which the reviewer confirmed was correct).

---

Closes #2922
Closes #2921
Closes #3013
Closes #2985

Refs #1878 (parent epic — deliberately NOT auto-closed; see the landing-pass
triage comment: #2824 and #2886 remain open, verified-unfixed, and in scope)

(#2966 is only partially addressed — part-1 remainder fixed here, parts 2/3/4 deferred, so it is deliberately NOT closed. #2964 and #3011 are tracked follow-ups.)

**Reviewer note:** verification on the rebased tip used targeted gates rather than a full `tests/architectural/` sweep (that suite is run by CI): read gate + write gate + trio-seam + surface-audit + dead-symbols + terminology = 118 passed; mission suites = 94 passed; `ruff` clean; `mypy --strict src/specify_cli src/charter src/doctrine` = no issues in 1088 files; docs freshness 0 errors + 602 docs tests. CI owns the exhaustive sweep.

**One honest pre-existing red, deliberately not fixed here:** `test_no_raw_mission_spec_paths` (offender `src/specify_cli/cli/commands/accept.py:239`) is red on the pre-mission base as well; per the red-main policy it stays red rather than being green-washed into this PR.

https://claude.ai/code/session_01Ky4wcHVXsyPzL3CBKFvSY9

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3012"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787769282&installation_model_id=427282&pr_number=3012&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3012&signature=74c89a98cfb259523f672bc5305e2680367fe66e094991632d28f20225a0b03a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


stijn-dejongh:
Claiming this PR for today's landing pass: rebase onto current `upstream/main`, adjudicate every red check from the first CI run, dispatch an adversarial review squad on the seam/gate surface, and fold remediations as separate commits on the branch. Evidence to follow.

stijn-dejongh:
## Landing pass: remediation summary

**Verdict: changes requested → now folded.** 15 maintainer folds pushed (`ae0bed609..24abed6ab`). All 9 red checks from the first run are addressed, and a three-lens adversarial squad found three defects CI could not have caught — including a **new silent-wrong-answer of the exact class this mission exists to remove**. Watching CI; merge-ready on green.

The branch already sat on current `upstream/main` (`f270e2d73`), so no rebase was needed. History is linear, 5 contributor commits, changelog correctly in `docs/changelog/CHANGELOG.md`.

### The two findings that matter most

**1. The migration introduced a silent wrong path (`3fccb03b5`).** `resolve_artifact_surface` derived its canonical slug by literal-composing `<slug>-<mid8>`. For a backfilled mission — bare `<slug>` primary dir on disk, composed `<slug>-<mid8>` coord dir — a composed handle produced a `primary_dir` that does not exist, so `declared_read_surface` found no `meta.json`, answered PRIMARY, and short-circuited before `probe_coord_state` ever ran. **The seam was not idempotent under its own output**: fed the name it had just emitted, it returned a different, nonexistent path.

| handle | before fix |
|---|---|
| bare `<slug>` | `.worktrees/<slug>-<MID8>-coord/kitty-specs/<slug>-<MID8>` ✅ |
| composed `<slug>-<MID8>` | `kitty-specs/<slug>-<MID8>` — does not exist ❌ |

Green on merge-base, red on HEAD via `test_decision_single_authority::test_open_resolves_coord_aware_handle`. A `PRIMARY_METADATA` read was masked by the same root cause. Worth noting the fix is *not* the obvious `candidate_feature_dir_for_mission(...).name` — for a composed handle that returns the coord dir, whose name is the composed name again. The helper strips the mid8 tail, re-resolves, then **confirms identity** via `_mid8_from_primary_meta`, because `mid8_from_slug` cannot distinguish a real mid8 from a coincidental 8-char Crockford segment; without that check the fix would trade one silent wrong answer for another.

**2. The capstone gate did not enforce what it advertised (`4ed3bbe06`).** Two lenses converged on this independently. `test_allow_list_reconciles_with_the_wp02_ledger_stay_lenient_count` compared `_ALLOW_LIST_SEED` against `_LEDGER_STAY_LENIENT_SITE_COUNT = 16` — a literal five lines above it in the same file. **The ledger markdown was never opened by any test or script**, so editing its Summary table to 15 left the whole suite green. Four hand-synced copies of the 16-site exemption list existed.

Separately, a planted bypass returned **zero findings**: `import ... as _alias` walked straight past `_callee_name`, and the sibling primitive `primary_feature_dir_for_mission` (39 real call sites across 21 files, re-censused rather than taken on trust) is unpoliced by anything.

Now: the ledger carries a machine-checked index the gate parses live; perturbing it reds (both proofs run). Alias resolution is implemented and pinned, including a no-false-positive case. `primary_feature_dir_for_mission` is deliberately *not* added to the target set — that needs its own classification pass — so instead the "unrepresentable" claim is downgraded to honest bounds with the gap named as tracked follow-up.

### Folds

| Fold | Why |
|---|---|
| `d9be11c4e` | ledger description 223 chars > the 180 band → **docs-freshness** |
| `1553a53be` | applied the guard's own `backfill-runtime-state` remedy; 8 WP seed events → **cutover-guard** |
| `caa5835f8` | 3 test surfaces stubbed the removed kind-blind resolver → **unit-contract-residual, fast-tests-agent, fast-tests-cli** |
| `a683cbd93` | new test files unmarked / `fast` + subprocess → **4 marker & orphan gates** |
| `5e5b93220` | deleted a vacuous test, renamed 9 parity tests, tightened 3 weak asserts |
| `299d9358d` `9564e9800` | corrected false fail-loud claims in comments and the resolver docstring |
| `4ed3bbe06` | ledger made mechanical; alias escape closed |
| `dc38088ee` `5ff829c8c` | `verify.py` CWD-invariance accepted + pinned; ledger row corrected |
| `e53e82110` | `accept.py` raw path → seam with `STATUS_STATE` → **arch_shard_1** |
| `6369aea61` | `docs_scoped` marker; recorded the `accept.py` census gap |
| `861852635` | recovery / merge / review paths degrade instead of crashing |
| `3fccb03b5` | the canonical-slug silent-wrong-answer, above |
| `24abed6ab` | changelog + PR-summary correction |

Two folds are worth a second look by the author:

- **`e53e82110`** fixed a latent bug on the way: the hand-built `coord_worktree_root / KITTY_SPECS_DIR / mission_slug` was wrong for identity-suffixed `<slug>-<mid8>` dirs. That site had **no ledger row at all** — it slipped the census because the grammar is "calls to the two primitives" and this was a raw path join. Recorded as an explicit out-of-grammar addendum rather than forced into the tables, which would have falsified the 90-call-site arithmetic.
- **`861852635`** — fail-loud must not mean "crash on the paths that exist to recover". `spec-kitty implement --recover` tracebacked on the one topology it exists to repair; `merge` tracebacked mid-merge; `agent workflow review` aborted while computing a *warning*, even though its own `except Exception` was right there — the seam call just sat outside it.

### On the PR summary

The summary claimed the campsite pass closed every migration-caused regression, with one pre-existing red remaining. Two further green-on-base/red-on-HEAD failures were unlisted. Corrected in `24abed6ab` rather than quietly superseded. The shared root cause is worth internalising: **the ledger reasoned only about the exception axis** (can this kind raise?) **and never the anchoring axis** (which root does it resolve against?). That blind spot produced both the `verify.py` regression and the canonical-slug one, and it originated in `PlacementSeam.read_dir`'s own docstring — now qualified to "identical with respect to *raising*".

### What the squad tried to break and could not

Stated because it is load-bearing for reviewer confidence: all 72 kind assignments are correct, including the 7 COORD ones — no wrong-partition kind was found. Each of the 16 lenient decisions has a real degrade path in code, not a dodge. The scanner really is shared with the write-side gate; no forked tree walk. No ADR is contradicted — the PR is a direct execution of `2026-06-26-1`. And the anchoring change is a **correctness improvement** everywhere except `verify.py`; `coordination/status_transition.py:610` is a latent-bug fix.

Both folded fixes were verified genuinely covered. `#2921` is a real corruption bug: `parse_frontmatter`'s third element is the raw frontmatter block, not padding, so every file `repair_lane_mismatch` touched had its frontmatter spliced back into the body.

### MINORs and NOTEs → one follow-up issue

Not folded, to be filed together under the functional epic: extend the shared scan scope beyond `src/` (both gates are blind to `tests/`, `scripts/`, conftests); resolve the `workflow.py`-vs-everyone-else seam-construction policy split; stale line citations in the ledger (content-addressed gate is immune, human navigation only); and the `primary_feature_dir_for_mission` classification pass.

One environment note for whoever lands the next one: inside a `.worktrees/` checkout, mission-handle resolution binds to the parent repository-root checkout, so `cutover-guard` reports zero touched missions there. I reproduced and remedied it in a plain clone.

**I have not merged and will not** — per the runbook the operator merges. Deliverable is green CI, the evidence above, and no landing-order constraint: this PR does not force an order on anything else in the queue.

stijn-dejongh:
## Landing-pass triage: `Closes #1878` was premature — removed

The PR body carried `Closes #1878` alongside `Closes #2922` / `Closes #2921`. #1878 is a genuine functional **epic** (`epic` label, 17 native sub-issues), and it still has 7 open children. I've **removed the `#1878` autoclose** and replaced it with a `Refs`. The two accurate closes are untouched.

Verdict is **mixed** — two children genuinely block, four were bucketed there, one is the thing this PR actually delivers. Each call was checked against `gh pr diff 3012`, not against theme.

### Genuine blockers — verified still unfixed on this tip

| # | Why it blocks |
|---|---|
| **#2824** | `accept`'s `read_feature_dir` still resolves via `PRIMARY_METADATA` for the lane-gate checks. This PR touched only `_wp_tasks_read_dir` in `acceptance/__init__.py`; the line feeding `_check_lane_gates` is untouched and still wrong. Not in the 72-site ledger. |
| **#2886** | `mission_setup_plan::_run_documentation_wiring` is not routed through the coord-aware resolver. `mission_setup_plan.py` does not appear in the PR diff at all. |

Both are the same read-authority defect class as #1878's stated scope, so they are in-scope work, not adjacent work.

### Mis-parented — recommend reparenting out of #1878

| # | Recommended parent | Why |
|---|---|---|
| #2240 | #2160 | `doctor coordination` recommends a nonexistent command — UX/doc bug, not a placement-read issue. Not in #1878's original deferred list. |
| #2618 | #2160 | `flatten_mission()` unification; surfaced reviewing merged PR #2614. Confirmed absent from this diff. |
| #2619 | #2160 | `CoordinationBranchDeleted.next_step` message-contract coverage; same provenance as #2618. |
| #2463 | #1619 *(weakest fit — worth a maintainer call)* | Legacy `mid8` bridge removal. Its own lineage (#2462 → #2091 → #1716) is closed/superseded and no dedicated legacy-identity epic exists. |

Once #2824 and #2886 close and those four move, #1878 has zero open native children and closing it is legitimate — but not via this PR.

### Separate gap worth filing

The residual this PR explicitly declined to police — `primary_feature_dir_for_mission`, 39 real call sites across 21 files — **has no open tracking issue at all**. A repo-wide search finds only closed or unrelated matches. That is undocumented debt, distinct from the two blockers above, and it should be filed under #1878 before that epic is assessed as complete.

stijn-dejongh:
## Second remediation round — all four CI failures folded

Pushed `24abed6ab..fd290b8fa` (5 further folds, 20 total). Added `Closes #3013` and `Closes #2985` to the PR body; `#1878` remains a `Refs` per the earlier triage.

### 1. I got one wrong, and it is reverted (`9ed59e7b1`)

The `DID NOT RAISE` failure was **my** fold, not the contributor's. On squad advice I made `scan_recovery_state` degrade instead of raising, reasoning that `implement --recover` shouldn't traceback on the topology it exists to repair. **That overrode the mission's own contract.** The WP02 ledger classifies all three `lanes/recovery.py` sites `migrate-fail-loud`, with a rationale that pre-emptively names exactly what I introduced — *"rather than silently reading a stale/absent surface."* The mission also demonstrably knew how to mark things lenient (it does so for four other modules) and deliberately put `recovery.py` on the other side.

Reverted; the shipped test is unmodified. The two tests asserting my invented contract are deleted rather than inverted — duplicating fail-loud coverage in `tests/merge/` would recreate the very directory split that hid the conflict (the fold was verified against `tests/merge/`, the contract test lives in `tests/specify_cli/merge/`). The reviewer's concern survives as a tracked follow-up recorded at the seam call. The `merge/executor.py` and `workflow_executor.py` guards had no conflicting contract and stand.

### 2. `accept` lane rewind — #3013 / #2985 (`a9587e6a4`)

Pre-existing, confirmed by three-point baseline (tip / pre-fold PR head / merge-base all red, test blob byte-identical at all three). Folded anyway — no `regression` marker, so it doesn't qualify for deferral.

**The recorded mechanism on #2985 was wrong, and it matters.** #2985 says the seed is *"appended after the terminal transition"* and the reducer *"replays in order"*. It isn't append order — it's the **sort key**. `_claim_anchors` falls back to a WP's *earliest* transition, which for a force-jumped or pruned WP **is its terminal one**; the seed collides on `at`, and since `reduce()` sorts by `(at, event_id)` the hash-derived seed id outranks the real ULID. A `done` WP folds back to `claimed`. Anyone fixing this by reordering appends would not have fixed it.

Implements #2985's suggested direction 1 (make the stamp state-aware): anchor inputs exclude the module's own `BACKFILL_ACTOR` rows so write and verify runs agree; the carrier's `at` is pinned strictly before the earliest authentic transition; and it seeds only the claim slots authentic history actually lacks, restoring event-count neutrality.

**NFR-004 checked, since #2985 requires it:** the merge seam `_run_birth_cutover` calls the same `cutover_mission` as the sole authority, so the fix lands downstream of both seams identically; `test_backfill_migration_coexistence.py` passes (5).

Witness proof: source file stashed → 12 failed; restored → all pass.

⚠️ **Already-corrupted corpus logs do not self-heal.** They still need the #3003 regenerate — which this unblocks. `test_dogfood_corpus_backfilled` stays red for that reason (identical with this change stashed).

### 3. Eight `tests/integration/` reds (`b5ee90352`)

No commit in this PR touches `tests/integration/` — these stubs predate the mission and the shard path-filter is why they surfaced only now.

Six were stale monkeypatch targets, re-pointed at the real boundary (`PlacementSeam.read_dir`, patched on the class, exact `(self, kind)` signature so future drift fails loudly rather than being absorbed by `**kwargs`).

The other two **looked like product bugs** — "resolution landing on the coord husk where it must land on primary". They aren't. Probing the real seam with no patches shows the product routes correctly. The culprit was `mock_seam.read_dir.return_value = coord_feature_dir` — every kind to the husk. Pre-migration the two legs used *different* symbols so stubbing one isolated it; post-migration both call the same `read_dir()`, so a kind-blind stub drags the PRIMARY leg onto the husk and manufactures the exact failure each test exists to forbid. No assertion changed.

3 pre-existing order-dependent reds in `test_mission_type_resolution_integration.py` remain, untouched (green in isolation, red only in a full sweep).

### 4. Docs index drift (`8f4a3be2f`)

Mine: shortening the ledger description and adding three sections changed both the abstract and the anchor list the retrieval index carries. Regenerated.

### Follow-up filed

**#3014** — `primary_feature_dir_for_mission`, the unpoliced sibling primitive, now tracked under epic #1878. The PR claimed this was "tracked follow-up work" in two places while no such issue existed; it does now. The census was independently re-run and is **40 sites across 21 files**, not the 39 first reported — both doc locations corrected to the verified number and pointed at #3014.

stijn-dejongh:
## ✅ Green — merge-ready, handing off

**64 pass / 0 fail / 16 skipped** at `8f2d89123`. Not a draft, `MERGEABLE`. All 9 originally-failing jobs plus the 4 that surfaced in the second round are resolved. **21 landing folds** on top of the 5 contributor commits; history linear, already on current `upstream/main`.

**No landing-order constraint** — this PR does not force an order on anything else in the queue.

### Last fold (`8f2d89123`)

A golden-count ratchet tripped: `tests/unit` hit 18 convert-classified sites against a frozen ceiling of 17, from my own accept fold. Converted the site rather than re-freezing the baseline — raising a ratchet ceiling for one new assertion is the drift the gate exists to catch, and annotating it `cardinality-is-contract` would have been false (the contract is *which* slots seed, not how many events exist). `assert len(x) == 1` + indexed follow-up became one list-equality pinning cardinality and content together.

### What the operator is merging

Beyond the contributor's read-side seam migration, this now also carries:

- **A silent-wrong-answer fix** — the seam was not idempotent under its own output for backfilled missions (bare-primary dir + composed handle → nonexistent path). Same bug class the mission exists to remove.
- **The `accept` lane rewind (#3013 / #2985)** — finished WPs no longer rewind to `claimed`. The mechanism recorded on #2985 was wrong: it is the reducer's `(at, event_id)` **sort key**, not append order. A fix aimed at reordering appends would not have worked.
- **A gate that now enforces its own thesis** — the ledger is parsed rather than compared against a literal in the same file, and import aliases no longer walk past the scanner.
- **Recovery/merge/review paths degrade** instead of tracebacking on a deleted coord branch.

### Two things to action after merge

1. ⚠️ **Corrupted corpus logs do not self-heal.** They still need the #3003 regenerate, which this unblocks. `test_dogfood_corpus_backfilled` stays red until that runs.
2. **#1878 reparenting** — #2240/#2618/#2619 → #2160, #2463 → #1619 (weakest fit, wants a human call). Left as recommendations, not applied. #1878 is a `Refs` here, not a close: #2824 and #2886 remain open, verified-unfixed, in-scope.

Also filed: **#3014** for the unpoliced `primary_feature_dir_for_mission` (40 sites / 21 files, independently re-censused), so the PR's "tracked follow-up work" claim is now true.

### Known-red, deliberately not folded

3 order-dependent cases in `tests/integration/test_mission_type_resolution_integration.py` — green in isolation, red only in a full-directory sweep. Pre-existing, outside this PR's scope, untouched.

### Squad MINORs / NOTEs → one follow-up issue (not yet filed)

Extend the shared scan scope beyond `src/` (both gates are blind to `tests/`, `scripts/`, conftests); resolve the `workflow.py`-vs-everyone-else seam-construction policy split; stale line citations in the ledger (content-addressed gate is immune — human navigation only).

Per the runbook the maintainer does not merge — **over to the operator.**

Issue #2996, opened 2026-07-27
Approving a rejected WP writes no verdict artifact — stale rejected blocks done/merge, and recovery fabricates a duplicate rejection under the next cycle number


stijn-dejongh:
## Summary

An approving review at cycle N+1 writes **no** verdict artifact. The prior
`review-cycle-N.md` (still `verdict: rejected`) remains the highest-numbered
artifact, and every downstream terminal-transition gate reads it and blocks —
`move-task --to approved`/`--to done` and `spec-kitty merge`. A second,
compounding defect makes recovery worse: whatever ran the cycle-2 review in
each case below did not write a real review outcome at all — it wrote a
`review-cycle-2.md` whose **73-line body is byte-identical to cycle 1's**,
with an 11-line machine-generated frontmatter block appended on top
(`reviewer_agent: unknown`, empty `affected_files`, empty
`reproduction_command`, `verdict: rejected`). No reviewer authored this
artifact; it is a fabricated duplicate of the original rejection, filed under
a cycle number that implies a second, independent review happened.

The state sequence that triggers it:

1. Reviewer rejects WP at cycle N → `review-cycle-N.md` written, `verdict: rejected`.
2. Implementer remediates.
3. Reviewer reviews again at cycle N+1 and approves.
4. **Expected:** a new `review-cycle-(N+1).md` is written with `verdict: approved`, a real `reviewer_agent`, and the actual affected files / repro command.
5. **Actual:** either (a) no artifact is written at all and cycle N stays the highest-numbered/`rejected`, or (b) a `review-cycle-(N+1).md` **is** written, but it is a mechanical duplicate of cycle N's body with synthetic frontmatter (`reviewer_agent: unknown`) — not a real approval, and still `verdict: rejected`.
6. Every subsequent terminal transition (`move-task --to approved`/`--to done`, `spec-kitty merge`) reads the highest-numbered artifact, sees `rejected`, and refuses.

## Why P0

The only CLI-advertised escape is `--skip-review-artifact-check`, which
records an **arbiter override of a rejection that never happened as
described** — the documented workaround writes false evidence into the
durable audit trail, exactly where integrity matters most. Any mission with a
reject → fix → approve cycle (the normal path, not an edge case) cannot reach
`done`/merge with an honest record. In the mission below this pattern hit 3 of
4 non-trivial WPs.

## Verified reproduction (3 independent WPs, one mission)

[...]

Issue #2989, opened 2026-07-27
P0: accept-time birth-cutover stamp regresses terminal WPs to 'claimed' — 3 blocking CI verdicts, sole 3.2.6 gate


LynnColeArt:
Follow-on regression from the #2917 birth-cutover landing (the accept-time stamp half). **`main` has been red since 11:40Z today and this is now the release gate** — `main@bb9c7d290` was green at 10:28Z, `main@d0a5bacf7` is red at 11:40Z. Run: https://github.com/Priivacy-ai/spec-kitty/actions/runs/30262797715

@stijn-dejongh flagging rather than fixing: this is inside your live #2917/#2920 surface and you asked us to avoid split-brains in core. Happy to take it — say the word and it is mine.

## All three blocking verdicts are one bug

The quality gate reports three blockers. Two of them share a single root cause; the third is cosmetic.

### 1. `integration-tests-core-misc` — the serious one

```
FAILED tests/specify_cli/acceptance/test_accept_idempotency.py::test_accept_converges_on_unchanged_tree[commit]
AssertionError: second accept did not converge:
  {not_done: [WP01], activity: ["WP01: canonical lane is 'claimed', expected 'approved' or 'done'"]}
```

Note what the same run prints just above the failure — the lane table reads `done | 1 | WP01`. So the **first** accept ends with WP01 `done`, and the **second** accept reads its canonical lane as `claimed`. That is not just non-idempotent, it moves committed lane state *backwards*.

### 2. `e2e-cross-cutting` — same cause, different assertion

```
FAILED tests/cross_cutting/misc/test_acceptance_support.py::test_accept_does_not_require_done_evidence_for_approved_wp
AssertionError: assert 7 == 6   # assert len(read_events(feature_dir)) == before_events
```

The test pins the invariant that accept appends **no** status event for an already-approved WP. It now appends one.

### 3. `arch-adversarial` — mechanical, unrelated to the above

```
FAILED tests/architectural/test_no_raw_mission_spec_paths.py::test_constant_based_mission_spec_path_construction_stays_in_constructor_files
offender: src/specify_cli/cli/commands/accept.py:239: (coord_worktree_root / KITTY_SPECS_DIR / mission_slug)
```

[...]

PR #3015, opened 2026-07-27
fix(charter): make preflight remediations able to clear their own checks (#2831)

### LINKED ISSUE #2831: Implement gate false-fails with 'charter_source missing' despite charter sync passing
**Environment:** spec-kitty v3.2.5, project at  initialized

**Summary:** Running  fails with:

Even though:
- `spec-kitty charter sync` passes ("Charter already in sync")
- `spec-kitty charter context --action implement` works and shows charter context
- `spec-kitty agent mission check-prerequisites` passes

**Repro:** Create a new mission via spec → plan → tasks → finalize-tasks, then try to implement:

**Impact:** Blocks the implement phase for any new mission. The preflight passes but the actual implement step fails.

**Likely surface:** The implement flow has a preflight check that fails to find the charter_source despite it existing at `.kittify/charter/charter.md`.

**Related:** #2777 (implement gate self-heal), #2555 (opaque gate errors)

MOES-Media:
Closes #2831

## The defect

An operator on a legacy-bundle project is blocked from implementing anything and told to run
`spec-kitty charter sync`. That command is a documented pure staleness reporter — `src/charter/sync.py`:
*"it always reports `synced=False` / `files_written=[]`"* — so it clears nothing and the gate refuses
with the identical message. There is no exit.

Meanwhile every charter diagnostic reports healthy, because they resolve `charter.md` while the gate
resolves `charter.yaml`. Every investigative step confirms the wrong conclusion.

**One correction to the issue's framing, carried into the spec**: the gate is *right*. `computer.py`
states it — *"`charter.yaml`, not `charter.md`, is the authoritative resolving source post-inversion"* —
so the diagnostics read the retired artifact, not the reverse. Convergence therefore moves the
diagnostics onto the gate's source. Getting this backwards would have re-opened a deliberately closed
decision, and the first draft of the plan had it backwards.

## What ships

Per `DIRECTIVE_043` the response is a mechanism, not a corrected string.

| | |
|---|---|
| **Effectiveness enforcement** | Drives each remediation-emitting state in an isolated fixture, executes the emitted command, asserts the state changed. Binds the operator-visible composed `blocked_reason`, not the returned field. AST-derived, conserved-total floors (5 emitting + 2 exempt == 7), exemption identity pinned. |
| **The P0 fix** | Four ineffective remediations corrected. Two states proved to have **no possible** self-service remediation and emit none rather than naming a command that cannot help. |
| **The defect class** | The runner's backfill removed — it substituted `charter status` whenever a check emitted none, the same defect on the *default* path. |
| **Presence convergence** | Ten operator-reachable resolvers onto one non-mutating seam, guarded by an `src/`-wide AST census. |
| **Diagnosability** | "No charter at all" now distinguishable from "legacy bundle present", naming exactly the files on disk. |
| **Regression envelope** | Four-shape blocking matrix measured against the pre-mission baseline; no state became blocked that was not blocked before. |

## Evidence

**Red-first (ADR `2026-07-17-1`)**: the mechanism landed **RED** — 4 failed / 9 passed on exactly the
four `charter sync` states — before any remediation was corrected. An independent reviewer verified
the red at that commit and the green at the next. The captured failure is the P0 in one line:

```
`spec-kitty charter sync` did not change synced_bundle's state (stayed 'missing');
exit=0 stdout='Charter already in sync (use --force to re-extract)'
```

Exit 0, "already in sync", state unchanged.

**The gates were attacked, not just written.** Reviewers found three separate ways to shrink coverage
while every check stayed green — drop a state and lower the floor; keep the sum but swap an exempt
member; re-introduce the runner backfill. Each passed green before being closed. The census was
falsified twice by planting unrouted resolvers in unrelated packages (`dashboard/`, `status/`,
`runtime/next/`); each turned it red with no list update.

**The resolver count was revised six times** (2 → 8 → 9 → 10 → 9 → 10). The last miss was invisible
*by construction*: the scan's roots were a hand-written list of three paths and the dashboard resolver
sat outside all of them. That is why the census is derived from the tree rather than enumerated.

## Delivered scope, stated plainly

**FR-001 ships as "every emitted remediation works, and states with no possible remediation emit
none"** — narrower than a naive reading. Two of the four states have no effective self-service
remediation *anywhere in the codebase*: every write path merges through the sole YAML writer, so all
require the file to already parse. A reviewer attempted to refute this exhaustively and could not.

## Three defects this PR's own CI found in this PR

Reported honestly because each is instructive, and two were **guards that had silently stopped guarding**.

1. **The convergence reported an existing file as missing.** Presence moved to the `charter.yaml` seam
   while the operator message kept naming the `charter.md` path, so every legacy-bundle project was
   told `Charter not found at <path>` about a file sitting at exactly that path — a false statement
   emitted to precisely the operators this mission exists to unblock. Fixed by naming the file the
   check actually tests. A first attempt also kept rendering governance for legacy bundles; **this
   mission's own gates rejected it** — `test_all_surfaces_agree_on_presence[F2]` because it reopened
   the very divergence being closed, and the seam census because the branch used a raw
   `CHARTER_MD.exists()`. Fully reverted.
2. **The NFR-004 traceback guard was blind on CI.** It matched a plain substring, but rich styles its
   header, so CI saw `\x1b[1mTraceback \x1b[0m…` and never matched — a cell "passed" against a genuine
   uncaught traceback. It was visible *only* because `xfail(strict=True)` turned the false pass into an
   unexpected-pass failure. Now normalises ANSI before matching, verified against the exact bytes CI
   produced.
3. **Baseline provenance failed on CI's shallow clone.** `merge-base --is-ancestor` returns 1 across a
   graft boundary — indistinguishable from a stale SHA, and not the 128-class exit the guard checked
   for. The skip is narrowed to exactly that case: a *positive* answer is always trustworthy and is
   still asserted. (Skipping on shallowness alone would have made the check vacuous everywhere — caught
   because this very checkout is shallow yet answers correctly.)

Also added tests for the two `except OSError` degrade paths the critical-path coverage gate flagged
(86% vs the 90% floor). Both **proven non-vacuous**: neutralising each branch turns its test red,
restoring it turns them green.

## Two pre-existing defects surfaced, deliberately not fixed

Both confirmed at the merge base with identical reachability before and after:

1. `doctrine.spdd_reasons.activation` crashes uncaught on an unparseable `charter.yaml`.
2. `charter bundle validate` crashes on the same input — `validate()` → `_bundle_compatibility_error`
   → `get_bundle_schema_version` (unguarded `yaml.load()`). Root cause is a missing catch-all that
   `charter status` and `charter resynthesize` both have. Pinned `xfail(strict=True)` so it cannot go
   stale: if `validate()` gains the guard, the test flips to an unexpected pass and fails.

## CI state — two red jobs, both pre-existing, both verified on a pristine base

`CI Quality` went 8 failing jobs → 4 → **2**. The two that remain are **not from this PR**, each
reproduced red on a clean `upstream/main` worktree at `f270e2d736` with none of these changes applied:

| Failing test | Evidence |
|---|---|
| `test_no_raw_mission_spec_paths.py::test_constant_based_mission_spec_path_construction_stays_in_constructor_files` | Offender is `src/specify_cli/cli/commands/accept.py:239`, a file this PR does not touch. This is the third CI verdict named in #2989. |
| `test_gate_coverage.py::test_gc2b_bites_on_producer_side_selection_shrink` | Red on `f270e2d736` with no local changes. |

`quality-gate` is the aggregate and fails solely because `arch-adversarial` does. No retry-to-green was
attempted on any of them.

## Verification

- `tests/architectural/`, `tests/charter/`, mission suites — green apart from the two above
- `ruff check` on all changed files — clean
- `mypy` — zero new errors versus the merge base
- `commitlint --from upstream/main --to HEAD` — exit 0

**One further pre-existing failure, not ours**: `tests/specify_cli/charter_preflight/test_cli.py::test_hard_error_exits_two`
fails identically on a pristine `upstream/main` worktree. Verified by reproduction, not asserted.

## Process note for maintainers

This mission hit **#2996** at its own merge: `spec-kitty merge` refused 5 of 6 WPs with
`REJECTED_REVIEW_ARTIFACT_CONFLICT`, offering *"run another review cycle that writes an approved
review-cycle artifact"* — which is **unperformable**, because no approving writer exists in the tree.
The mission that fixed "a remediation must be able to clear its check" was blocked by a remediation
that cannot. Evidence posted to #2996; the approving artifacts here were written by hand because the
tooling provides no way to write them, and the alternative would have recorded an arbiter override of
a rejection nobody made.

A post-merge mission review returned CONDITIONAL and both blockers were closed on the mission branch —
including a marker defect where the census file was marked `fast` while shelling out to the real CLI,
so a `-m git_repo` CI run would have **skipped it entirely**. Fixed by correcting the marker and
annotating genuine cardinality sites, deliberately *not* by re-freezing the baseline, which would have
raised the ceiling and hidden the next drift. Two review artifacts that overclaimed independent review
carry appended corrections rather than rewrites, so the record shows what was claimed, that it was
checked, and that it was wrong.


MOES-Media:
## Remediation summary — rebuilt branch, CI at its floor

**Verdict**: ready for maintainer review on the code, **left as a draft** pending the operator's go.

### What changed since the branch was first pushed

The branch was **re-sliced from the mission tip**. The original slice was taken from a commit that
predated two remediation commits, so the PR did not actually contain fixes its own description
claimed — the census file shipped as `pytestmark = pytest.mark.fast` with zero golden-count
annotations, where the mission branch had `git_repo` and five. That is corrected; the marker defect
matters because a `-m git_repo` CI run would have **skipped that file entirely**.

### Three defects this PR's own CI found in this PR, all fixed

Two of them were guards that had silently stopped guarding, which is worse than a red check:

1. **An existing file reported as missing.** Presence moved to the `charter.yaml` seam while the
   message kept naming the `charter.md` path — so every legacy-bundle project was told
   `Charter not found at <path>` about a file at exactly that path. A first attempt to fix it more
   broadly was **rejected by this mission's own gates** (surface-agreement on the F2 shape, and the
   seam census catching a raw `.exists()`), and was fully reverted. The gates worked.
2. **The NFR-004 traceback guard was blind on CI** — plain substring vs rich's ANSI-styled header, so
   a cell "passed" against a genuine uncaught traceback. Visible only because `xfail(strict=True)`
   converted the false pass into an unexpected-pass failure.
3. **Baseline provenance failed on CI's shallow clone** — `--is-ancestor` returns 1 across a graft
   boundary, not the 128-class exit the guard checked for. Skip narrowed to that case only; a positive
   answer is still asserted.

Plus tests for the two `except OSError` degrade paths the critical-path coverage gate flagged, both
**proven non-vacuous** (neutralise the branch → red; restore → green).

### CI: 8 failing jobs → 4 → 2, and the 2 are not ours

| Failing test | Why it is not from this PR |
|---|---|
| `test_no_raw_mission_spec_paths.py::…stays_in_constructor_files` | Offender is `accept.py:239`, untouched here. This is the third CI verdict named in #2989. |
| `test_gate_coverage.py::test_gc2b_bites_on_producer_side_selection_shrink` | Reproduced red on a pristine `upstream/main` worktree at `f270e2d736`. |

`quality-gate` is the aggregate and fails only because `arch-adversarial` does. **No retry-to-green was
attempted on anything.**

### Gates run locally on the rebased tip

`ruff` clean · `mypy` zero new vs merge base · `commitlint --from upstream/main --to HEAD` exit 0 ·
`diff-coverage` now passing.

### Known-affected tooling, for the record

`spec-kitty merge` refused 5 of 6 WPs here with an **unperformable** remedy (#2996 — no approving
writer exists), so the approving artifacts were hand-written; the alternative was recording an arbiter
override of a rejection nobody made. Two artifacts that overclaimed independent review carry appended
corrections rather than rewrites.

State: **draft**, rebased on current `upstream/main`, awaiting the operator.

MOES-Media:
## Rebased onto current `upstream/main` — one of the two pre-existing reds is gone

Base hygiene, re-run because a stale base makes an attribution claim unreliable rather than merely untidy.

The branch was **26 commits behind**; it is now rebased onto `e97fc6ab99` (conflict-free, 0 behind).
Re-verifying the two failures previously reported as pre-existing against the **new** base:

| Previously reported | Status on `e97fc6ab99` |
|---|---|
| `test_no_raw_mission_spec_paths.py::…stays_in_constructor_files` | **Now passes.** Upstream fixed the `accept.py:239` offender in the intervening commits. It is no longer a red on this PR at all. |
| `test_gate_coverage.py::test_gc2b_bites_on_producer_side_selection_shrink` | **Still pre-existing** — reproduced red on a pristine `upstream/main` worktree at `e97fc6ab99` with none of this PR's changes applied. |

So the earlier "two pre-existing reds" statement is now **one**. Recording the correction explicitly
rather than quietly restating a smaller number: the original claim was verified against
`f270e2d736`, and an attribution is only as good as the base it was measured on.

### Gates re-run on the rebased tip

`tests/architectural/` (raw-path, gc2b, golden-count, marker-correctness, blocking-envelope,
remediation-effectiveness) + `tests/charter/test_charter_presence_seam.py` +
`tests/cli/commands/test_charter_orchestration.py` → **91 passed, 2 xfailed, 1 failed** (the `gc2b`
pre-existing above). `commitlint --from upstream/main --to HEAD` → exit 0.

Still a **draft**, awaiting the operator. No retry-to-green was attempted on anything.

robertDouglass:
## Adversarial squad review — REQUEST CHANGES

Reviewed HEAD `6ab5099ed8bebe480b47bab30691bf0566e25ab4` with Reviewer Renata, Debugger Debbie, and Doctrine Daphne under the repository charter.

### Material findings

- **HIGH — ineffective auto-refresh remains shipped** (`src/specify_cli/charter_runtime/preflight/runner.py:408-428`): missing source/bundle still triggers `spec-kitty charter sync`, although this PR establishes that sync is a reporter and cannot create `charter.yaml`. A real F2 legacy-bundle probe ran only sync, exited 1, and left all states unchanged. The new effectiveness checks cover `auto_refresh=False`; auto-refresh tests begin from already-fresh source/bundle state.
- **HIGH — effectiveness oracle accepts regression/failure** (`tests/architectural/test_remediation_effectiveness.py:433-442`): it requires only `after != before`, not exit 0 or a cleared/pass state. A fake remediation that wrote malformed YAML and returned 17 was accepted because `missing -> invalid`.
- **HIGH — F1/F2 authority is still conflated** (`runner.py:208-220`, `freshness/computer.py:407-412`): the optional-missing bypass identifies a fresh project solely from three `missing` strings, so a legacy bundle is waived through like an uninitialized project. Both states also emit `charter generate --no-from-interview`; that does not migrate/preserve legacy governance, while the existing upgrade path does.
- **HIGH — dashboard still gates canonical presence on derived Markdown** (`src/specify_cli/dashboard/charter_path.py:48-63`): `ensure_charter_bundle_fresh()` returns before the new `charter_yaml_present()` check when `charter.md` is absent. A valid yaml-only project produced dashboard `None` while freshness reported source/bundle fresh.
- **HIGH — incompatible YAML can be marked fresh** (`freshness/computer.py:295-314`): any mapping, including `{}` or unsupported bundle schema versions, passes the local parse test; preflight can therefore green a charter rejected by other consumers.
- **MEDIUM — command-bearing prose bypasses the claimed exhaustive enforcement** (`freshness/computer.py:564`): `detail=` emits “re-run spec-kitty charter synthesize”, but the AST census only discovers literal `remediation=` values. A live probe ran the command successfully yet remained `built_in_only`.

These are material contract/authority issues; no implementation changes were made.

### Verification and maintenance

- Focused remediation/presence/preflight/freshness suite: **67 passed**.
- Ruff on all 11 touched source files: **clean**.
- Strict mypy exposes six touched-file debts.
- HEAD is **625 commits behind** current `main` and GitHub reports conflicts.
- Rebase attempted and aborted cleanly after conflicts in four mission event logs plus a mission baseline capture. Resolving those requires semantic/history choices, so I did not guess.
- CI's architectural selection failure is attributable to this aggregate diff; quality-gate is derivative. `cutover-guard` also reports touched missions not cut over.

**Verdict: REQUEST CHANGES.** Fix material findings on the author branch, resolve/rebase deliberately, then rerun current CI.

Lens concession: deterministic operator probes and focused suites were used; no full-suite or browser dashboard render was performed.

robertDouglass:
Automated review — condensed after validation. No code was modified.

**Verdict: request changes.**

**[MEDIUM] The F2 missing-charter remediation does not preserve legacy content.** `freshness/computer.py` emits `spec-kitty charter generate --no-from-interview`, which generates a default charter rather than folding the legacy charter bundle into the new structure.

The current effectiveness test proves that the workflow reaches green state, but not that existing policy content survives. Emit the content-preserving migration/remediation command and add an acceptance test with distinctive legacy content that must remain present after repair.

This is the operator-facing recovery path for the P0 condition, so content preservation should be demonstrated before merge.

MOES-Media:
## Rebased onto current `main` — `MERGEABLE`, zero attributable failures

Was `CONFLICTING`/`DIRTY`, **906 commits** behind. Now rebased, 3 commits, 64 files.

### Scope reduced deliberately

The branch had two separable halves. Only one landed:

- **Kept — the #2831 fix.** `charter sync` is still documented as a pure staleness reporter (`generate.py:149`) and `_REMEDIATE_CHARTER_SYNC` was still named by five states, so operators were still being handed a command that clears nothing. Four remediations corrected, two states that have no possible self-service remedy now emit `remediation=None` rather than naming a command that cannot help, the runner's `charter status` backfill removed, and the constant deleted.
- **Dropped — the charter.yaml presence inversion (R-001).** Upstream settled that question the other way, deliberately: the presence gate is an OR across `charter.yaml` and `charter.md`, adopted *after* a strict yaml-only gate was measured to regress **26 tests** in a full `-k charter` sweep. Re-litigating that is a maintainer decision, and #2831's symptom is fixed without it. The `capture.py` change was likewise dropped — #3150 already seeds `charter.yaml` there with a better payload.

### Two fixes worth reading

**The rebase silently gutted this PR's own guard.** The effectiveness-enforcement walk accepted `remediation=` only as an `ast.Constant`. That was total on the original branch, which inlined every command as a literal — but upstream has since hoisted the shared one into `_REMEDIATE_CHARTER_SYNTHESIZE` for S1192, so three of five remediation-emitting states became `ast.Name` and vanished from discovery. Enforcement narrowed from five states to two **while still reporting green** — the exact defect class this module exists to prevent, reappearing inside its own instrument. Fixed by resolving module-level string constants, so enforcement keys on the command emitted rather than how it is spelled.

**Cross-mission collateral removed.** The dossier commit was deleting `status_phase` from **eight other missions'** `meta.json`, plus historical events from five missions' append-only logs. `cutover-guard` inspects only missions a diff touches, so this PR was pulling five uninvolved missions into its blast radius and failing on their state. Now zero cross-mission changes; the guard passes.

### Verification

`tests/architectural`: **1415 passed**. ruff clean. mypy delta against `upstream/main` on the touched modules: **zero** (same 5 pre-existing errors, line numbers shifted only).

CI: 30 pass, 2 fail. Both — `quality-gate` and `regression tests (blocking)` — are **red on `main` itself** (since 2026-08-12); `quality-gate` cites only `regression-tests`. Neither is attributable.

LynnColeArt:
Landing this. Reviewed today against current main before merge: the three patched files were verified against main's blobs (computer.py and result.py byte-identical to the merge-base; runner.py drifted only by typing cleanups in untouched regions), all four `_REMEDIATE_CHARTER_SYNC` use sites removed here are exactly main's four, and Robert's 08-09/08-11 findings map to the 08-17/08-20 fix commits (H1 freshness-derived remediation, H2 exit-0+documented-state oracle with a fake-remediation negative test, H3 F1/F2 split via `upgrade --yes`, H5 unsupported-schema no longer greens). Cleared the stale pre-rebase `pr:needs-refresh`/`pr:needs-revision` labels.

Two notes for the record, non-blocking: the PR body still describes the dropped presence-convergence scope (the 08-13 comment is the accurate description), and `_attempt_auto_refresh` can now run `spec-kitty upgrade --yes` automatically in the F2 state — a real behaviour expansion over the old no-op sync, defensible given the clean-worktree gate and `auto_refresh=False` default.

Issue #3045, opened 2026-07-28
charter sync always reports success: unconditional "already in sync", --force is a no-op, and a test pins the defect as intentional


stijn-dejongh:
## Summary

`spec-kitty charter sync` is **structurally incapable of reporting anything but success**. On the human output path it prints `Charter already in sync` unconditionally — for a never-synced charter, a missing bundle, or a stale one alike. `--force` is a literal no-op. The command that the product's own remediation strings tell operators to run in order to repair a fail-closed gate cannot repair, cannot detect, and cannot fail.

This is the **root-cause / structural** issue. #2992 is the field symptom report from `spec-kitty-saas` (see "Relationship to #2992" below — the overlap is larger than a separate arm, and the two should be adjudicated together).

## Why P0

Three properties compound into a release blocker:

1. **The documented repair path for a fail-closed gate is dead.** `spec-kitty implement` fail-closes before allocating a lane workspace. The operator is blocked.
2. **The product points at the dead command itself.** `src/specify_cli/charter_runtime/freshness/computer.py` emits `remediation="spec-kitty charter sync"` for both the `missing` and `stale` bundle states. The gate says broken; the prescribed repair says fine. Closed loop.
3. **A test pins the broken behavior as intended.** `tests/agent/cli/commands/test_charter_cli.py:83-99` asserts `Charter already in sync` for a never-synced charter and describes it as the *"current, intentional contract"*. **No existing gate can ever catch this defect** — the suite actively defends it. That is what makes it P0 rather than P1: it is not merely broken, it is warranted-broken.

## Verified root cause (all against `upstream/main` @ `765cdcc59f5abc85a45af672b7291ca6d1ece8b8`)

**Unconditional success print** — `src/specify_cli/cli/commands/charter/sync.py:49`. The only two outcomes are an error exit or this line. The sync-success branch was removed by `9fb95185e737822903b95b4165731800a95766d6` (2026-07-22, IC-04, "remove dead sync-success branch"). That commit was correct in its own frame — the branch *was* unreachable — but it removed the last visible evidence that the command no longer does anything.

**`--force` is a no-op** — `src/charter/sync.py:174`:

    _ = force  # no longer changes behavior; kept for signature stability.

[...]

Issue #3033, opened 2026-07-28
post-merge artifact writes fail against a deleted target_branch — placement seam lacks a post-merge mode


samuelgoff:
## Problem

A normal post-merge review cannot persist its evidence after the merged feature branch has been deleted. This conflicts with a branch-first workflow that cleans up the feature branch after its PR merges.

## Reproduction

1. Complete and merge a mission PR, then delete the feature branch.
2. From a new branch based on the merged target, run post-merge review.

In Hallafax:

```text
git switch -c review/relational-cutover-postmerge origin/main
PYTHONPATH=/path/to/spec-kitty/src /path/to/spec-kitty/.venv/bin/spec-kitty review --mission relational-cutover-01KYHHR8
```

The mission has five done WPs and no declared issue references. The current source correctly evaluates Gate 4 as `not applicable`.

## Actual

After all four gates pass/skip correctly, persistence fails:

```text
MISSION_REVIEW_EVIDENCE_PERSISTENCE_FAILED: safe_commit: worktree ... HEAD is
'review/relational-cutover-postmerge', expected 'feat/relational-cutover'.
```

The expected branch was the merged feature branch, which was deliberately deleted. The command exits non-zero and cannot commit the refreshed review evidence.

## Expected

A post-merge reviewer should be able to persist review evidence to an authoritative post-merge surface, or the command should have an explicit no-write/read-only mode that succeeds without requiring a deleted pre-merge branch. It must not require operators to recreate or check out a merged feature branch merely to refresh review evidence.

## Impact

Hallafax cannot produce a current durable post-merge review artifact after normal branch cleanup, despite the substantive gates passing. We will not recreate the branch or fabricate review evidence as a workaround.

## Environment

- `spec-kitty-cli v3.2.6`
- Hallafax mission `relational-cutover-01KYHHR8`
- merged PR: https://github.com/crucible-energy/hallafax/pull/121

This is distinct from https://github.com/Priivacy-ai/spec-kitty/issues/2617: that issue concerns selecting the corrected same-version build; this report concerns where a post-merge review is allowed to persist its evidence.

---

# Correction (triage, 2026-07-28)

*Everything above is the reporter's original account and is preserved verbatim. **The defect is real and reproducible.** This section corrects only the reproduction path, so implementers do not chase it through a command that never commits. Verified against `main` on 2026-07-28.*

[...]

Issue #3031, opened 2026-07-28
P0: hosted-sync consent is opt-out and fail-open, recorded outside the project, and capture is ungated


stijn-dejongh:
## Summary

Hosted-sync consent is **opt-out and fail-open**, the consent record lives outside the project it governs, and event capture is not gated at all. Together these mean every project on a machine is sync-enabled by default, and a single opted-in checkout can drain the whole machine's journal to a hosted instance.

This is the CLI-side counterpart to `Priivacy-ai/spec-kitty-saas#585`, where a live incident delivered 1,322 events belonging to 5 projects that were never opted in. The operator's configuration was correct throughout; the defaults did this.

**The SaaS issue is the missing last line of defence. This issue is the actual defect.**

## Defect 1 — the default is allow, and unresolvable routing fails open

`src/specify_cli/sync/routing.py:82-87`:

```python
if local_sync_enabled is not None:
    effective_sync_enabled = local_sync_enabled
elif repo_default_sync_enabled is not None:
    effective_sync_enabled = repo_default_sync_enabled
else:
    effective_sync_enabled = True        # <-- default allow
```

and `routing.py:115-116`, in `is_sync_enabled_for_checkout`:

```python
routing = resolve_checkout_sync_routing_readonly(start=start)
if routing is None:
    return True                          # <-- unresolvable => allow
```

A checkout that has never been configured reports `Checkout Sync: Enabled` in `sync routes`. That reads as a decision the operator made; it is the default. For a confidentiality control both branches must deny.

## Defect 2 — consent is recorded outside the project it governs

`routing.py::read_local_sync_enabled` — *"Read the checkout-local sync override **from global machine config**."* Even the per-checkout override lives in `~/.spec-kitty/config.toml`, keyed by repo path / `repo_slug`. `.kittify/config.yaml` has **no sync key at all**.

Consequences:

- Consent is invisible in the repository it applies to.
- It is not version-controlled, not reviewable, and not diffable.
- It is keyed on `repo_slug`, derived from `git remote get-url origin` (`sync/git_metadata.py:203`). Rename or repoint the remote and an existing opt-out silently stops matching.
- A fresh clone of a repository arrives with no opt-out and defaults to enabled — so "we opted that client repo out" does not survive re-cloning it.

## Defect 3 — capture is ungated

[...]

Issue #3030, opened 2026-07-28
P0: sync drain authorizes per-checkout but delivers per-journal — one opted-in checkout ships every project on the machine


MOES-Media:
## Summary

The drain **authorizes per-checkout but delivers per-journal**. A single opted-in checkout ships the entire machine-global event journal — every unrelated local project — to the hosted server. Confirmed live on 2026-07-27: 1,322 events belonging to 5 projects the operator never opted into were delivered alongside 7,811 from the one intended project. Per-repo routing in that checkout was correct throughout.

This is the CLI-side counterpart to `Priivacy-ai/spec-kitty-saas#585`, which holds the incident record, the exposure profile, and the server-side defence-in-depth half. **This issue is the core defect**: the code that leaks is in this repo.

Distinct from the incidental issues hit en route: #2995 / PR #2998 (mission-state blocker), #3004 (diagnostics not routed through `target_authority`), #3005 (permanently-rejected events reported as `Terminal failures 0`).

## Root cause, at three specific lines

**1. Selection happens before authorization, and authorization never looks at the selection.**
`src/specify_cli/sync/batch.py:1064-1080`:

```python
events, payload = _select_events_for_advertised_limits(   # <- selects from the WHOLE journal
    queue, requested_limit=limit, advertised_limits=advertised_limits,
)
result.total_events = len(events)
if not _is_checkout_sync_enabled_for_batch():             # <- one bool about the CURRENT checkout
    ...
    return result
```

`_select_events_for_advertised_limits` reads the journal with no project predicate — the observed run reported `selected 13384`, i.e. every project on the machine. The gate that follows is a single boolean about the checkout the command happens to be running in. Once it returns `True`, all 13,384 rows proceed to delivery. **There is no per-event project filter anywhere between authorization and POST** — `project_slug` does not appear in `batch.py` at all.

**2. The gate is a policy read being misused as a delivery filter.**
`src/specify_cli/sync/routing.py:101-117`. `is_sync_enabled_for_checkout()` is correct at what it documents ("is sync enabled for the active checkout?"). It is not, and cannot be, a statement about the provenance of journal rows. The misuse is at the call site.

**3. The gate fails open.**
`src/specify_cli/sync/routing.py:114-116`:

```python
routing = resolve_checkout_sync_routing_readonly(start=start)
if routing is None:
    return True          # <- unresolvable routing == permitted
```

[...]

PR #3050, opened 2026-07-28
test: pin six live P0 defects with main-redding regression tests

stijn-dejongh:
## ⚠️ CI on this PR is EXPECTED TO BE RED. The red is the deliverable.

These ten tests pin **six live P0 defects that exist on `main` today**. They are
evidence, not breakage. Per ADR `2026-07-17-1`, `main` may be honestly red and
known-P0 reds must not be green-washed — this PR converts six asserted P0s into
six *evidenced* ones.

Nothing here fixes a defect. No `Closes #N` anywhere: each test makes its issue
visible to CI, and the issue closes when the product changes.

## What reds, and why

| Issue | Node ID | Failure signature |
|---|---|---|
| #2993 | `tests/regression/test_issue_2993_lane_planning_ancestry.py::test_lane_worktree_does_not_descend_from_planning_artifacts` | `git merge-base --is-ancestor <planning-artifact-sha> <lane-branch>` returns 1 — the coord tip was never advanced past its pre-artifact mint point before the lane branched off it |
| #2996(a) | `tests/integration/test_review_cycle_rejection_only.py::test_approving_a_rejected_wp_writes_no_verdict_artifact` | After a full rework+resubmit lifecycle, `move-task --to approved` still exits 1 on the rejected-artifact guard and `latest.cycle_number` is still 1 — no approved verdict is ever written |
| #2996(b) | `tests/review/test_cycle.py::test_self_referential_feedback_source_is_rejected` | `DID NOT RAISE ReviewCycleError` — `create_rejected_review_cycle` never inspects what `feedback_source` points at |
| #2996(b) | `tests/review/test_cycle.py::test_new_cycle_body_never_duplicates_a_prior_cycle_file` | Cycle 2's body is byte-identical to cycle 1's body, written with `reviewer_agent: unknown` |
| #3031 (D5) | `tests/delivery/test_dispatch_honours_drain_blocked_3031.py::test_dispatch_excludes_events_with_recorded_drain_blocked_reason` | An event captured with `drain_blocked_reason` set is delivered anyway — the drain has no predicate over it |
| #3031 (D5) | `tests/delivery/test_dispatch_honours_drain_blocked_3031.py::test_consent_predicate_must_apply_before_limit_not_after` | With a 10-event blocked backlog older than 1 consenting event and `limit=5`, the drain ships the 5 oldest (all blocked) and never reaches the consenting event |
| #3030 | `tests/delivery/test_dispatch_project_consent_3030.py::test_consenting_project_leaks_sibling_project_event_through_shared_journal` | A project with **no consent record at all** has its event shipped because it shares a journal file with a project that did consent |
| #3031 (D3) | `tests/sync/test_sync_consent_capture_gap_3031.py::test_non_consenting_project_event_never_reaches_the_journal` | A non-consenting checkout's event is written to the machine-global journal anyway — the capture gate decides only the recorded `drain_blocked_reason`, never the write |
| #3033 | `tests/specify_cli/cli/commands/test_safe_commit_cmd.py::test_public_safe_commit_succeeds_after_merged_branch_deleted_3033` | `safe_commit: worktree ... HEAD is 'review/...-postmerge', expected 'feat/...'` — the destination resolves from `meta.json` with no git-existence check |
| #3045 | `tests/agent/cli/commands/test_charter_cli.py::test_sync_command_human_and_json_surfaces_do_not_contradict_3045` | For one repo state the JSON surface reports `success: False` / `stale_before: true` while the human surface exits 0 printing `Charter already in sync` |

`#3031` already had a partial RED pin on `main`
(`tests/sync/test_sync_consent_default_deny.py`, landed at `765cdcc59`); this PR
adds the ungated-capture gap that file's own docstring calls out as uncovered,
and gives that file the `regression` marker it was missing.

## Markers

Every test carries `pytest.mark.regression` (registered at `pytest.ini:48`).
The `regression tests (blocking)` job (`ci-quality.yml:2939`) runs
`pytest tests/ -m regression` on every PR with no path-filter gate, so these
reds are seen regardless of which shards a diff activates.

## Post-review remediation (landing pass)

A four-lens adversarial squad (`reviewer-renata`, `debugger-debbie`,
`architect-alphonso`, `paula-patterns`) reviewed these pins against one
question: *does each test pin the behaviour we want, or the arithmetic inverse
of today's?* It found four blocking defects. All are now folded in — see the
review comment for the full evidence, and the fold commits for the changes.

1. **#2996(b) — the two new tests were mutually unsatisfiable.** One required
   `create_rejected_review_cycle(feedback_source=<own review-cycle-1.md>)` to
   *raise*; the other made the identical call and required it to *return*.
   Installing the guard made the second **error before reaching its
   assertion**. Resolved in favour of the loud guard; the duplicate-body test
   now drives an ordinary feedback file whose *content* duplicates a prior
   cycle, and compares body-to-body rather than body-vs-full-file-text (which
   was cheap-greenable by stripping frontmatter).
2. **#3030 — the two `tests/delivery/` files pinned contradictory contracts.**
   `drain_blocked_reason` is a capture-time snapshot of *machine-global* gates,
   so a predicate over it implements no per-project consent at all.
   `test_dispatch_honours_drain_blocked_*` is therefore **rescoped to #3031
   Defect 5** (per-event vs per-process drain filtering) and renamed
   accordingly. The #3030 consent fixture is fixed: the consenting emitter now
   carries a real `repo_slug` matching the consent key it writes (consent is
   keyed on `repo_slug` at `sync/config.py:190`; the fixture previously wrote a
   key nothing could match) with an open capture gate, so the two files key on
   orthogonal properties.
3. **#3031 — the capture pin drove a dead API arm.** It called
   `capture_teamspace_bound(is_teamspace_bound=False, skip_journal=True)`, a
   combination **no production caller emits** (`emitter.py:1961` passes
   neither). Two lines of parameter plumbing would have greened it with the
   defect fully alive, and the real fix would have left it red. Re-anchored on
   `EventEmitter._emit` with a non-consenting checkout, asserting the journal
   stays empty.
4. **#3045 — the stale pin was left standing.** #3045's body is explicit that
   the redding test must *replace* `test_charter_cli.py:83-99`, not sit beside
   it. Done. The new test's `exit_code == 0` preconditions (which forbade the
   *honest reporter* remedy the ticket sanctions) are dropped, and it now
   asserts the positive surface-agreement contract from a single invocation —
   closing both the reword cheap-green and an ordering bug that would have
   redded a correct repair-mode fix.

Also folded: the **deletion-only** case #2993 calls "the sharp one" (the
existing leg only covered an amendment); the `--feature` → `--mission` fix for
two pre-existing vacuous tests in `test_review_cycle_rejection_only.py` (there
is no `--feature` option, so every invocation exited 2 on a usage error and the
tests "proved" implement reruns don't inflate the counter *by never running
implement*); the missing `regression` marker on
`test_sync_consent_default_deny.py`; and a CI path-filter gap where
`tests/regression/**` and `tests/regressions/**` were unclaimed by the
`core_misc` filter despite the `misc` shard owning them.

## Inverted pin that the fix must delete

This PR **deliberately does not touch it**. Two contradictory tests coexist
until the fix lands — deleting an assertion is part of fixing the defect, not
part of evidencing it.

- `tests/specify_cli/cli/commands/test_implement.py:533-545` —
  `TestPlanningArtifactAutoCommit::test_auto_commit_uses_coordination_worktree_paths`
  asserts #2993's snapshot as correct: the artifact present on coord **and
  absent from the planning branch**.

  *Caveat, added in review:* that pin's precondition is the
  **dirty-uncommitted** case (`tasks.md` is written but never `git add`ed), which
  #2993 says works correctly — so it may well survive the fix untouched.
  Re-adjudicate it rather than deleting it reflexively. (The #3045 inverted pin
  previously listed here has been removed from the suite, per that issue's own
  first implementation step.)

## False greens these supersede

- `tests/post_merge/test_review_artifact_consistency.py:214` — writes the approved
  artifact directly via a fixture helper, substituting for the exact step #2996(a)
  removes. The gate is proven; the producer is never exercised.
- `tests/sync/test_sync_e2e_integration.py:611-614` — asserts
  `e["project_slug"] == identity.project_slug`, which is *nearly* the right
  assertion, but the fixture has one emitter bound to one identity so it cannot fail.
- `tests/integration/test_review_cycle_rejection_only.py:274,307` — invoked
  `agent action implement --feature`, an option that does not exist
  (`workflow.py:1128` declares only `--mission`), so every call exited 2 on a
  usage error and the no-op assertions were vacuous. **Fixed in this PR** rather
  than merely catalogued; the contract holds once implement genuinely runs.

## Mechanism corrections found while writing these

Two tickets describe a mechanism that does not match the traced code. Implementers
should read these before starting.

- **#3033** — `MISSION_REVIEW_EVIDENCE_PERSISTENCE_FAILED` is **not a spec-kitty
  diagnostic** (zero occurrences repo-wide, no commit ever introduced it), and
  `spec-kitty review` performs no commit at all. The defect reproduces through
  `spec-kitty safe-commit` on a *classified* PRIMARY artifact.
  `mission-review-report.md` is not in `_MISSION_FILE_KIND_BY_BASENAME`, so it
  falls through to HEAD and would succeed. #3033's body and title are corrected.
- **#3030** — the ticket pins `sync/batch.py:1064-1080`, which is the **legacy**
  `OfflineQueue` transport. The live path is `delivery/dispatcher.dispatch`, which
  `cli/commands/sync.py:296` calls *"the sole event-delivery path"*. These tests
  target the live path; an implementer following the ticket literally would fix
  `batch.py` and leave them red. A correction is filed on the issue.
- **#2993** — the ticket describes **one** mechanism; there are **two independent
  causes** of the same symptom. They share a structural root (coord is minted from
  `target_branch` before spec/plan/tasks exist; lanes branch off it) but have
  different observables and different remediation.
  - **Cause A — copy without ancestry**: artifact *bytes* are written into the
    coord worktree and committed onto the coord tip. The planning commit is never
    in history. A later amendment on the planning branch produces `CONFLICT (add/add)`.
  - **Cause B — silent no-op** (legacy `placement_ref is None` fallback): PRIMARY
    files are diffed against `HEAD`, found byte-identical, dropped; the plan is
    empty and the helper returns with no commit. Lane carries **no planning
    artifacts at all**.
  - **This test reproduces Cause B**, confirmed empirically. Its assertions
    nonetheless catch Cause A: simulating the partial "repair the idempotency
    filter" fix (which converts B into A) in pure git leaves the ancestry
    assertion red (rc=1) and produces `CONFLICT (add/add)` on merge-tree, while a
    genuine ancestry fix greens both legs.
  - **Discriminator caveat:** `git log --oneline <coord-branch> -- kitty-specs/<slug>/`
    non-empty → A, empty → B — but this misclassifies a *correctly fixed* coord as
    Cause A, since ancestry legitimately puts the artifacts in that log. It is a
    triage heuristic for a live pre-fix incident only.
  - **Load-bearing for the fix:** the fix must establish **ancestry**, not a more
    reliable byte-copy. Acceptance asserts `--is-ancestor` == 0, that a post-allocation
    amendment merges cleanly and wins, *and* that a post-allocation **deletion**
    survives the merge — the sharp case, since it vanishes under a union merge.

## Gates

`ruff check` and `mypy` are clean on all nine changed test files (the one
pre-existing `mypy` error in `test_review_cycle_rejection_only.py` is fixed
here too). The architectural collection-completeness, gate-coverage, and
terminology suites pass on the rebased tip (81 passed). Rebased onto
post-#3007 `main` (`ed470756e`).

https://claude.ai/code/session_01RsJfT4pKcjgnotT2o2XMfu

### DIFF
--- a/.github/workflows/ci-quality.yml
+++ b/.github/workflows/ci-quality.yml
@@ -294,6 +294,13 @@ jobs:
               - 'tests/contract/**'
               - 'tests/init/**'
               - 'tests/integration/**'
+              # The `misc` shard owns tests/regression + tests/regressions (see
+              # its run list below), but neither directory was claimed by this
+              # filter — so a PR touching only an issue-pinned reproduction did
+              # not activate `core_misc`, the shard never ran, and the pin was
+              # structurally uncollectable on that PR.
+              - 'tests/regression/**'
+              - 'tests/regressions/**'
               - 'tests/dossier/**'
               - 'tests/git_ops/**'
               - 'tests/tasks/**'
--- a/tests/agent/cli/commands/test_charter_cli.py
+++ b/tests/agent/cli/commands/test_charter_cli.py
@@ -488,3 +488,47 @@ def test_help_output() -> None:
     assert "context" in result.stdout
     assert "sync" in result.stdout
     assert "status" in result.stdout
+
+
+def test_sync_command_human_and_json_surfaces_do_not_contradict_3045(mock_repo: Path) -> None:
+    """#3045: ``charter sync`` always reports success on the human surface,
+    even when the JSON surface (same invocation, same repo state) reports
+    ``stale_before: true``.
+
+    ``mock_repo`` writes only ``charter.md`` -- no ``metadata.yaml`` -- so
+    ``sync()`` computes ``stale_before=True`` (there is nothing yet to compare
+    a hash against). ``_emit_sync_human_result``
+    (``src/specify_cli/cli/commands/charter/sync.py:36-49``) unconditionally
+    prints "Charter already in sync (use --force to re-extract)" whenever
+    ``result.error`` is falsy -- it never branches on ``stale_before``. The
+    JSON payload built two lines earlier by ``_sync_json_payload``
+    (``src/specify_cli/cli/commands/charter/sync.py:23-33``) reports
+    ``stale_before: true`` for the exact same ``SyncResult``. The two surfaces
+    of the SAME command, over the SAME repo state, contradict each other: one
+    says "already in sync", the other says "was stale before this call".
+
+    This directly CONTRADICTS ``test_sync_command_success``
+    (``tests/agent/cli/commands/test_charter_cli.py:83-99``), which pins
+    ``"Charter already in sync" in result.stdout`` as "current, intentional
+    contract" and explicitly rationalizes it in its docstring as the retired
+    IC-04 extraction contract. That assertion must be deleted in the same
+    commit that fixes #3045 -- this test and that one cannot both stay green
+    once the human surface is made to agree with the JSON surface.
+    """
+    with patch("specify_cli.cli.commands.charter.find_repo_root") as mock_find_root:
+        mock_find_root.return_value = mock_repo
+
+        json_result = runner.invoke(app, ["sync", "--json"])
+        human_result = runner.invoke(app, ["sync"])
+
+        assert json_result.exit_code == 0
+        assert human_result.exit_code == 0
+
+        payload = json.loads(json_result.stdout)
+        assert payload["stale_before"] is True
+
+        assert "already in sync" not in human_result.stdout, (
+            "#3045: human surface claims the charter is already in sync while "
+            "the JSON surface (same repo state) reports stale_before=True -- "
+            f"human stdout: {human_result.stdout!r}, json payload: {payload!r}"
+        )
--- a/tests/agent/cli/commands/test_charter_cli.py
+++ b/tests/agent/cli/commands/test_charter_cli.py
@@ -490,6 +490,7 @@ def test_help_output() -> None:
     assert "status" in result.stdout
 
 
+@pytest.mark.regression
 def test_sync_command_human_and_json_surfaces_do_not_contradict_3045(mock_repo: Path) -> None:
     """#3045: ``charter sync`` always reports success on the human surface,
     even when the JSON surface (same invocation, same repo state) reports
--- a/tests/agent/cli/commands/test_charter_cli.py
+++ b/tests/agent/cli/commands/test_charter_cli.py
@@ -6,6 +6,7 @@ import subprocess
 from unittest.mock import patch
 
 import pytest
+import typer
 from typer.testing import CliRunner
 
 from specify_cli.cli.commands.charter import app
@@ -81,20 +82,33 @@ def _write_charter_yaml_bundle(charter_dir: Path) -> None:
 
 
 def test_sync_command_success(mock_repo: Path) -> None:
-    """IC-04 (#2773): ``charter.sync.sync`` extraction is retired -- it is a
-    pure staleness reporter now (``SyncResult.synced`` is hardcoded
-    ``False``; there is no derived triad left to write). ``charter sync``
-    therefore always renders the noop branch, even on a "fresh" charter that
-    has never been synced before -- pin that current, intentional contract
-    (see the module docstring in ``charter.sync``) rather than the retired
-    extraction-success message."""
+    """#3045: the assertion that used to live here --
+    ``"Charter already in sync" in result.stdout``, defended by this test's
+    prior docstring as the "current, intentional contract" -- was a STALE
+    PIN. Issue #3045 identifies it as the P0 escalator for the underlying
+    defect (``charter sync`` reports success unconditionally, even for a
+    charter that was never synced): a test asserting the misleading message
+    is *intended* means no gate could ever catch the regression. Per the
+    failing-test remediation framework a stale pin is deleted/rewritten, not
+    annotated -- removed here in the same commit that adds
+    ``test_sync_command_human_and_json_surfaces_do_not_contradict_3045``
+    below, which is the red-first replacement (see that test's docstring).
+    The two must not be re-separated later: this test defends only the
+    coverage that is still valid under EITHER sanctioned #3045 remedy
+    (repair or honest-reporter) -- it never re-materializes the retired
+    derived triad (``governance.yaml`` / ``directives.yaml`` /
+    ``metadata.yaml``, retired by IC-04 / #2773). No exit-code assertion is
+    kept either: ``mock_repo`` is exactly the "charter.md present,
+    charter.yaml absent, never synced" state the honest-reporter remedy is
+    expected to make exit non-zero for (per #3045's suggested shape) -- a
+    hard ``exit_code == 0`` pin here would forbid that remedy through a
+    second assertion in this file, recreating the same problem this
+    remediation exists to close."""
     with patch("specify_cli.cli.commands.charter.find_repo_root") as mock_find_root:
         mock_find_root.return_value = mock_repo
 
-        result = runner.invoke(app, ["sync"])
+        runner.invoke(app, ["sync"])
 
-        assert result.exit_code == 0
-        assert "Charter already in sync" in result.stdout
         for retired_file in ("governance.yaml", "directives.yaml", "metadata.yaml"):
             assert not (mock_repo / ".kittify" / "charter" / retired_file).exists()
 
@@ -491,45 +505,98 @@ def test_help_output() -> None:
 
 
 @pytest.mark.regression
-def test_sync_command_human_and_json_surfaces_do_not_contradict_3045(mock_repo: Path) -> None:
+def test_sync_command_human_and_json_surfaces_do_not_contradict_3045(tmp_path: Path) -> None:
     """#3045: ``charter sync`` always reports success on the human surface,
-    even when the JSON surface (same invocation, same repo state) reports
-    ``stale_before: true``.
+    even when the JSON surface (same repo state) reports
+    ``stale_before: true`` / ``success: false``.
 
-    ``mock_repo`` writes only ``charter.md`` -- no ``metadata.yaml`` -- so
-    ``sync()`` computes ``stale_before=True`` (there is nothing yet to compare
-    a hash against). ``_emit_sync_human_result``
+    ``_emit_sync_human_result``
     (``src/specify_cli/cli/commands/charter/sync.py:36-49``) unconditionally
     prints "Charter already in sync (use --force to re-extract)" whenever
-    ``result.error`` is falsy -- it never branches on ``stale_before``. The
-    JSON payload built two lines earlier by ``_sync_json_payload``
+    ``result.error`` is falsy -- it never branches on ``stale_before`` or
+    ``synced``. ``_sync_json_payload``
     (``src/specify_cli/cli/commands/charter/sync.py:23-33``) reports
-    ``stale_before: true`` for the exact same ``SyncResult``. The two surfaces
-    of the SAME command, over the SAME repo state, contradict each other: one
-    says "already in sync", the other says "was stale before this call".
+    ``success: False`` / ``stale_before: True`` for the exact same
+    ``SyncResult``. The two surfaces of the SAME command, over the SAME
+    result, contradict each other: one says "already in sync", the other
+    says "did not sync, was stale before this call".
 
-    This directly CONTRADICTS ``test_sync_command_success``
-    (``tests/agent/cli/commands/test_charter_cli.py:83-99``), which pins
-    ``"Charter already in sync" in result.stdout`` as "current, intentional
-    contract" and explicitly rationalizes it in its docstring as the retired
-    IC-04 extraction contract. That assertion must be deleted in the same
-    commit that fixes #3045 -- this test and that one cannot both stay green
-    once the human surface is made to agree with the JSON surface.
+    The stale pin that used to defend the human-surface message as
+    intentional -- ``test_sync_command_success``'s
+    ``"Charter already in sync" in result.stdout`` assertion, previously
+    rationalized in its own docstring as the "current, intentional
+    contract" -- was REMOVED in this same commit (see that test's updated
+    docstring). This test is the red-first replacement, not an addition
+    alongside the old pin; the two must not be re-separated later.
+
+    Test design, addressing three problems in an earlier draft of this test:
+
+    1. **Single invocation, two rendered surfaces.** ``charter.sync.sync()``
+       is called exactly ONCE to obtain one real ``SyncResult``; both
+       ``_sync_json_payload`` and ``_emit_sync_human_result`` render from
+       that SAME object. An earlier draft invoked the ``sync`` CLI command
+       twice against one shared ``mock_repo`` -- under a *repair*-mode fix,
+       the first (``--json``) invocation would actually repair the charter,
+       so the second (human) invocation would then correctly report
+       "already in sync", falsely reding a fix that fully resolves #3045.
+       Rendering both surfaces from one result sidesteps that ordering
+       hazard entirely.
+    2. **No exit-code pin.** Honest-reporter mode (the operator's stated
+       preference is "loud failures with clear instructions") is expected to
+       fail loudly -- non-zero exit -- when nothing was actually synced.
+       Repair mode may legitimately return ``synced=True`` (and thus exit 0)
+       because restoring the pre-#2773 ``sync()`` contract means a plain
+       (non-``--force``) call auto-extracts whenever the charter IS stale.
+       This test does not hardcode which remedy is chosen; see the
+       ``payload["success"]`` branch below.
+    3. **``stale_before`` is read off the payload, not hard-asserted
+       ``True``.** It is a permanent constant today only because the
+       staleness check targets the retired ``metadata.yaml`` (#3045 remedy
+       3) -- a corrected staleness check may compute it differently.
+
+    The substantive assertion is positive, not a bare negative substring: if
+    the JSON surface reports the sync did NOT actually happen
+    (``success is False``), the human surface must not exit 0 silently --
+    it must fail loudly. A cosmetic reword of the "already in sync" string
+    alone (keeping exit 0) does not satisfy this, because the check does not
+    inspect the human surface's text at all -- only whether it raises a
+    non-zero exit given a ``SyncResult`` that itself reports nothing was
+    synced.
     """
-    with patch("specify_cli.cli.commands.charter.find_repo_root") as mock_find_root:
-        mock_find_root.return_value = mock_repo
+    charter_dir = tmp_path / ".kittify" / "charter"
+    charter_dir.mkdir(parents=True)
+    (charter_dir / "charter.md").write_text(SAMPLE_CHARTER, encoding="utf-8")
 
-        json_result = runner.invoke(app, ["sync", "--json"])
-        human_result = runner.invoke(app, ["sync"])
+    from specify_cli.cli.commands.charter._app import console as charter_console
+    from specify_cli.cli.commands.charter.sync import (
+        _emit_sync_human_result,
+        _sync_json_payload,
+    )
+    from charter.sync import sync as sync_charter
 
-        assert json_result.exit_code == 0
-        assert human_result.exit_code == 0
+    result = sync_charter(charter_dir / "charter.md", charter_dir)
+    payload = _sync_json_payload(result)
 
-        payload = json.loads(json_result.stdout)
-        assert payload["stale_before"] is True
+    human_exit_code: int | None = None
+    with charter_console.capture() as capture:
+        try:
+            _emit_sync_human_result(result)
+        except typer.Exit as exc:
+            human_exit_code = exc.exit_code
+    human_text = capture.get()
 
-        assert "already in sync" not in human_result.stdout, (
-            "#3045: human surface claims the charter is already in sync while "
-            "the JSON surface (same repo state) reports stale_before=True -- "
-            f"human stdout: {human_result.stdout!r}, json payload: {payload!r}"
-        )
+    if payload["success"]:
+        # This single sync() call already resolved the staleness (the
+        # sanctioned *repair* remedy) -- both surfaces legitimately agree
+        # the charter is now in sync, so there is nothing to contradict.
+        return
+
+    assert human_exit_code not in (None, 0), (
+        "#3045: the JSON surface reports success=False (charter was not "
+        f"actually synced by this call; stale_before={payload['stale_before']!r}) "
+        "but the human surface exited 0 without raising -- a silent "
+        "'everything is fine' when nothing was synced. The operator's "
+        "stated preference is a loud failure naming the recovery path "
+        "('charter generate' / 'charter synthesize'), not a quiet success. "
+        f"human output: {human_text!r}, json payload: {payload!r}"
+    )
--- a/tests/delivery/test_dispatch_honours_drain_blocked_3030.py
+++ b/tests/delivery/test_dispatch_honours_drain_blocked_3030.py
@@ -60,7 +60,7 @@ from specify_cli.event_journal.models import DRAIN_BLOCKED_SAAS_DISABLED, Event
 if TYPE_CHECKING:
     from specify_cli.delivery.interfaces import DeliveryTarget
 
-pytestmark = pytest.mark.fast
+pytestmark = [pytest.mark.regression, pytest.mark.fast]
 
 _TARGET_URL = "https://hosted.example.com"
 _TARGET_TEAM_SLUG = "team"
--- a/tests/delivery/test_dispatch_honours_drain_blocked_3030.py
+++ b/tests/delivery/test_dispatch_honours_drain_blocked_3030.py
@@ -1,0 +1,210 @@
+"""RED pins: the drain must honour the capture-time consent classification (#3030).
+
+Background (see ``docs/development/read-side-seam-classification.md`` sibling
+investigation and issue #3030). ``event_journal/models.py:113-129`` — the
+``Event`` dataclass has no project field. ``SELECT_ALL_SQL``
+(``event_journal/models.py:78``) has no WHERE clause; ``EventJournal.read_all()``
+(``event_journal/journal.py:258``) takes no predicate; ``_select_undelivered``
+(``delivery/dispatcher.py:192-223``) sets ``universe = journal.read_all()`` and
+never inspects ``Event.drain_blocked_reason`` at all. The column exists — it is
+populated at capture time by ``capture_teamspace_bound`` via
+``classify_drain_blocked_reason`` — but nothing on the drain side reads it back.
+``sync/batch.py:357`` (``_prepare_events_for_ingress``) actively *strips*
+``drain_blocked_reason`` from the legacy queue payload before POSTing, which
+confirms the field is understood elsewhere as "must not leave the machine
+un-vetted" — the WP07 journal/dispatcher path (the one under test here) has no
+equivalent guard at all.
+
+Two tests:
+
+* :func:`test_dispatch_excludes_events_with_recorded_drain_blocked_reason` —
+  the direct pin: one blocked event, one unblocked event, single dispatch call.
+  Also closes the #3031 "Defect 5, per-event drain filtering" gap (the sibling
+  ``tests/sync/test_sync_consent_default_deny.py`` docstring flags this as
+  uncovered): both events are drained in the SAME call (one process tick), so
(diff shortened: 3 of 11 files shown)


stijn-dejongh:
Claiming this PR for the landing pass: rebase onto current `upstream/main`, adjudicate check state (noting the reds here are the deliverable per ADR 2026-07-17-1, not breakage to fold), and run an adversarial review squad over whether each new test pins the *desired* behaviour rather than merely an inverse of today's. Landing-order constraint under review: this is slated to slot in after #3007. Evidence to follow.

stijn-dejongh:
## Landing-pass review — adversarial squad on test-vs-desired-behaviour

Rebased onto post-#3007 `main` (`ed470756e`) and adjudicated with a four-lens
profile-loaded squad (`reviewer-renata`, `debugger-debbie`, `architect-alphonso`,
`paula-patterns`), framed on one question: **does each new test pin the behaviour we
want, or the arithmetic inverse of today's?**

**Verdict: excellent evidence work, four blocking defects before it lands.** The red is
correctly framed and I am not objecting to red. I am objecting that **three of these reds
are unfixable** — they stay red after the correct fix, or no single fix can green them.
Per ADR `2026-07-17-1` the whole point is that red means something; a pin that survives its
own fix is worse than no pin, because the next agent discharges it by deleting an assertion.

### What's clean (verified, not assumed)

- **No collateral red.** Every failure on this branch is one of the 10 new tests. The only
  other red — `tests/sync/test_sync_consent_default_deny.py` (5 tests) — is the pre-existing
  #3031 pin from `765cdcc59`.
- **Every test is red on its substantive assertion**, with the claimed signature. All 10 ran
  individually; none skipped, xfailed, or errored before its assertion.
- **CI visibility confirmed.** `ci-quality.yml:2939` `regression tests (blocking)` collects
  all 10 (`10/40 collected, 30 deselected`). The marker claim is load-bearing and correct.
- **#3007's new collection-completeness gate passes** on the rebased tip (46 passed) — the
  four new files are collected. This gate has no allowlist and no regeneration path, so it
  was a real landing risk; it is cleared.
- `ruff` clean; `tests/architectural/test_no_legacy_terminology.py` 4 passed; no
  real-`~/.spec-kitty` pollution; no port/network/ordering/xdist coupling.
- **The #2993 pin is exemplary** — see below.

---

### BLOCKING 1 — `tests/review/test_cycle.py`: the two new #2996(b) tests are mutually unsatisfiable

`test_self_referential_feedback_source_is_rejected:203` requires
`create_rejected_review_cycle(feedback_source=<own review-cycle-1.md>, reviewer_agent="")`
to **raise** `ReviewCycleError`. `test_new_cycle_body_never_duplicates_a_prior_cycle_file:264`
makes the **identical call** and requires it to **return** `cycle2`.

Not a theoretical conflict — the guard was installed and test 2 re-run verbatim:

```
    cycle2 = create(..., feedback_source=cycle1.artifact_path, reviewer_agent="")
E   specify_cli.review.cycle.ReviewCycleError: feedback_source review-cycle-1.md is itself a review-cycle artifact
```

Test 2 **errors before reaching its assertion** — evidence of nothing. Conversely a
de-duplication fix greens test 2 and leaves test 1 `DID NOT RAISE` forever. All four lenses
converged independently.

**Also cheap-green:** `prior_bodies` collects each prior artifact's **full on-disk text**
(`p.read_text()`, :269) but compares against `cycle2.artifact.body`. Stripping frontmatter
before assigning `body` — one line — makes the comparison miss while the fabricated cycle-2
with `reviewer_agent: unknown` is still written. And `:277`
`assert cycle1_full_text not in (cycle2.artifact.body,)` is membership in a 1-tuple, i.e. a
restatement of the line above.

**Fix:** keep one. Either drop test 2 and let the guard carry #2996(b), or rewrite test 2 to
use a *non-cycle* feedback source whose content happens to equal a prior cycle's body — which
pins "never duplicate" independently of "reject self-reference". Compare body-to-body, and
add `latest.reviewer_agent != "unknown"`.

### BLOCKING 2 — `tests/delivery/test_dispatch_project_consent_3030.py` cannot green under any correct fix

Two independent blockers, confirmed by probing the test's own fixture at runtime:

```
  CONSENTING      drain_blocked_reason='saas_disabled'
  NON-CONSENTING  drain_blocked_reason='saas_disabled'
  consenting.repo_slug    = None
  nonconsenting.repo_slug = None
  consent config key written = my-org/engagement-assistant
```

**(a) It contradicts its own sibling file.** `:128` patches `is_saas_sync_enabled → False`, so
*both* rows are stamped `saas_disabled` at capture. `test_dispatch_honours_drain_blocked_3030.py:148`
demands any row with that reason never ship; this file `:177` demands the consenting one ship.
No implementation satisfies both.

**(b) The consent record is inert.** It is written as
`[sync.repo_defaults."my-org/engagement-assistant"]`, and `sync/config.py:190` looks it up via
`repo_defaults.get(repo_slug)` — but `_stub_emitter:109` sets `GitMetadata()` whose `repo_slug`
is `None`, and the only distinguishing field on either event is `project_slug`. **No consent
lookup resolves consent for the event that must ship**, so a correct fail-closed per-project
predicate denies *both* projects and reds `:177`. Greening it would require inventing a fuzzy
`project_slug`→`repo_slug` correspondence — precisely the fragile keying #3031 Defect 2 exists
to eliminate.

**Fix:** give the consenting emitter a real `repo_slug` matching the consent key, and open its
capture gate so its `drain_blocked_reason` is `None`.

### BLOCKING 3 — `tests/sync/test_sync_consent_capture_gap_3031.py` pins a dead API arm

`capture_teamspace_bound` has exactly one production caller — `EventEmitter._capture_to_journal`
(`sync/emitter.py:1961`) — and it passes **neither** `is_teamspace_bound` nor `skip_journal`;
both take defaults. Repo-wide, `grep -rn "is_teamspace_bound\|skip_journal" src/` returns only
the definition site. **No operator reaches `(False, True)`.**

So a two-line `if skip_journal: return event` greens this test while every real capture still
lands in the machine-global journal and #3031 Defect 3 survives intact; and the *real* fix —
gating at the consent layer — leaves it red. This departs from DIRECTIVE_041 (reproduce through
the pre-existing entry point). Compounding: `:98` `assert isinstance(result, Event)` against a
`-> Event` signature forbids the natural `Event | None` fix, so the only greening change is
"build the Event, return it, don't persist it" — a contract nobody would design.

Also, `_NON_CONSENTING_GATE` is decorative: `classify_drain_blocked_reason(gate)` only sets the
`drain_blocked_reason` column, never the write decision, so a fully *consenting* gate produces an
identical result. The one thing in the test expressing non-consent is load-bearing on nothing.

**Fix:** re-anchor on `EventEmitter._emit` / `_capture_to_journal` with a non-consenting routing
state and assert the journal stays empty. Drop the `isinstance` assertion.

### BLOCKING 4 — #3045's own body forbids what this PR did

Issue #3045, verbatim:

> **First implementation step: re-adjudicate `tests/agent/cli/commands/test_charter_cli.py:83-99`.**
> The redding test must *replace* it ... **Writing a new red test alongside it produces a suite
> that asserts both that the command must say "in sync" and that it must not.**

That is exactly what landed. The PR's "coexisting contradiction is the honest state" rationale is
a reasonable general principle, but #3045 already adjudicated this specific case the other way and
the PR neither cites nor rebuts it.

**Fix:** delete or `xfail(reason="#3045")` the `:83-99` assertion here — or state the rebuttal in
the PR body and update the issue. Don't leave the contradiction silent.

**Related, same file:** `:525-526` `assert json_result.exit_code == 0` **and**
`human_result.exit_code == 0` forbid the "honest reporter" remedy #3045 explicitly leaves open
(its own suggested shape: *"must NOT exit 0 with 'already in sync'"*). And `:531` is greenable by
rewording the string in `sync.py:49` without touching any staleness semantics. Assert the
*implication* — `if payload["stale_before"]: assert "already in sync" not in stdout` — and drop the
exit-code pins.

---

### #2993 is the strongest pin in the PR — and the feared failure mode is falsified

The worry that repairing only the idempotency filter (converting Cause B into Cause A) would green
it was tested by simulating both post-fix worlds in pure git:

```
scenario A   (partial fix: byte copy onto coord)  : ancestry rc=1  -> assert A still FAILS
                                                    merge-tree rc=1 -> CONFLICT (add/add)
scenario FIX (ancestry established)               : ancestry rc=0  -> PASSES
                                                    merge-tree rc=0 -> merged tasks.md == amended
```

Assert A is ordered first and correctly resists the shallow fix; both assertions green only under a
genuine ancestry fix. Land this file essentially as-is. Two notes: step 3's helper call is provably
inert (the test fails identically with it deleted — it reproduces **Cause B**), and the A/B
discriminator in the body misclassifies a *correctly fixed* coord as Cause A, so add that caveat
before an operator re-opens #2993 post-fix. Worth adding the **deletion-only** correction case the
ticket calls "the sharp one" — the current Assert B only covers an amendment.

### Non-blocking, worth folding into the body

- **Counts:** body says "nine tests" and "+1217"; actual is **10 test functions**, **+1226**.
- **`mypy` is not clean** on the eight files: `test_review_cycle_rejection_only.py:247`
  `Returning Any from function declared to return "Path"`. It is **pre-existing** (line 247 sits in
  the file's original 345 lines; the PR appends at 343+), so not your defect — but the claim as
  written is overstated.
- **The `test_implement.py:533-545` "inverted pin" bullet is wrong**, and contradicts this PR's own
  new-test docstring (`test_issue_2993_...:44-56`: *"does NOT contradict"*). I read the pin:
  `tasks.md` is written but never `git add`/`commit`ed, so its precondition is the
  *dirty-uncommitted* case #2993 says works correctly. An implementer following the body will delete
  an assertion the code tells them to keep.
- **A third false green, in a file this PR already touches:** `test_review_cycle_rejection_only.py:274`
  and `:307` invoke `agent action implement --feature`, but `workflow.py:1128` declares only
  `--mission`. Every such invoke exits 2 on a usage error, and the test concedes it doesn't check the
  exit code — so it "proves" implement reruns don't inflate the counter *by never running implement*.
  Pre-existing; stronger evidence for your thesis than either example the body names.
- **#3030's ticket names the wrong file.** It pins `sync/batch.py:1064-1080`, the **legacy**
  `OfflineQueue` bridge; the tests correctly target `delivery/dispatcher.dispatch`, which
  `cli/commands/sync.py:296` calls *"the sole event-delivery path"*. The tests are better-aimed than
  the ticket — correct the issue body so an implementer doesn't fix `batch.py` and leave these red.
- **`test_sync_consent_default_deny.py` carries `fast` but not `regression`** — so the blocking job
  shows only a partial view of #3031: the ungated-capture arm reds, the fail-open-default arm (the
  actual 1,322-event incident) never appears. One-token fix, adds no new red.
- **`tests/regression/**` has no entry in the `changes` path-filter block.** The #2993 file runs here
  only incidentally, because this PR also touches `tests/integration/**` and `tests/delivery/**`. A
  future regression-only PR wouldn't trigger it.
- The two **mechanism corrections** (#3033's non-existent diagnostic and non-committing `review`
  command; #2993's two-cause split) are genuinely valuable triage and materially improve both
  tickets. Both independently verified — `MISSION_REVIEW_EVIDENCE_PERSISTENCE_FAILED` has zero
  occurrences repo-wide and was never introduced by any commit.

### Follow-up issue (do not expand this PR)

**Consent ownership: capture-first (FR-017/C-008) vs #3031 Defect 3.** Two accepted contracts assert
opposite things about whether a non-consenting project's event may be written to the journal.
`journal.py:383-402` documents the unconditional write as deliberate; #3031 requires such events
"must never reach the journal". This PR routes around the conflict by pinning the one parameter
combination nobody asserts on. Deliverable: one adjudicated owner, and the losing contract gets
deleted rather than left standing.

### State

Rebased onto `ed470756e` locally and verified there; **not force-pushed** — the branch is still on
the pre-#3007 base, and pushing a rebase would just churn CI ahead of the rework above. Say the word
and I'll push it. The #3007-before-#3050 ordering constraint is now moot (#3007 merged at 17:19Z);
the two PRs have zero file overlap and #3007 touches nothing matching `regression`.

stijn-dejongh:
## Remediation complete — eight folds pushed, all squad blockers closed

Branch rebased onto post-#3007 `main` (`ed470756e`) and force-pushed with a lease.
Tip: `726a8593c`. All four blocking findings from the review above are folded in,
plus the non-blocking items.

### Blocking findings — closed

| # | Finding | Fold |
|---|---|---|
| 1 | #2996(b)'s two new tests were mutually unsatisfiable | `5232f58fb` |
| 2 | #3030's two `tests/delivery/` files pinned contradictory contracts; the consent fixture could not go green under any correct fix | `d869587cf` |
| 3 | #3031's pin drove an API arm no production caller reaches | `125e4b770` |
| 4 | #3045's stale pin was left standing, against that issue's explicit first implementation step | `aa00560c7` |

**On (1)** — resolved toward the loud guard, per the operator's standing preference
for loud failures with clear instructions. The guard now asserts the message is
*actionable*, and `match=` was tightened (the pre-existing "not found"/"is empty"
messages both contain `feedback` and satisfied the old regex). The duplicate-body
test now drives an ordinary feedback file whose *content* duplicates a prior cycle
and compares body-to-body — closing the cheap-green where stripping frontmatter
would have greened it with the fabricated cycle still written.

**On (2)** — `drain_blocked_reason` is a capture-time snapshot of machine-global
gates, so a predicate over it implements no per-project consent. That file is
rescoped to the **#3031 Defect 5** pin its own docstring already claimed and
renamed `..._3031.py`. Fixture proof after the fold:

```
  CONSENTING      drain_blocked_reason=None    repo_slug='my-org/engagement-assistant'
  NON-CONSENTING  drain_blocked_reason='saas_disabled'    repo_slug=None
  consent config key written = 'my-org/engagement-assistant'
```

The two events are now distinguishable on a field a consent predicate can key on.

**On (3)** — re-anchored on `EventEmitter._emit`. The test no longer passes
`is_teamspace_bound`/`skip_journal` anywhere, so `if skip_journal: return event`
cannot reach it; non-consent (`checkout_enabled=False`) is the only variable
between the two emitters, and a consenting sibling is asserted to still capture —
so "disable capture wholesale" cannot look correct either.

**On (4)** — stale pin removed, not annotated. Exit-code preconditions dropped so
*honest reporter* is no longer forbidden, and the test now renders both surfaces
from a single invocation — killing the reword cheap-green and an ordering bug that
would have redded a correct repair-mode fix.

### Also folded

- **`8011129f8`** — #2993's **deletion-only** leg, the case the ticket calls "the
  sharp one" (the existing leg only covered an amendment). Validated by pure-git
  simulation: under a genuine ancestry fix the deletion survives the merge and the
  assertion passes; under a shallow byte-copy, Assert A still fires first. Also
  corrected two docstring overclaims — the test reproduces **Cause B**, and step 3's
  helper call is provably inert (verified by removing it: identical failure).
- **`d18f581a9`** — `--feature` → `--mission` in two pre-existing vacuous tests.
  They exited 2 on a usage error, so they "proved" implement reruns don't inflate
  the counter *by never running implement*. Now real, with the exit code asserted
  so 2 can never masquerade as a no-op again. **They pass** — the contract holds;
  no hidden red was behind it.
- **`726a8593c`** — `regression` marker on `test_sync_consent_default_deny.py` (its
  fail-open arm, the one behind the live 1,322-event leak, was absent from the P0
  index), and `tests/regression/**` + `tests/regressions/**` claimed in the
  `core_misc` changes filter — the `misc` shard owned them but the filter never
  activated on them, so an issue-pinned reproduction was structurally uncollectable
  on a PR that touched only that directory.
- **`a6f6c1029`** — the last `mypy` error. Root cause was config, not code:
  `follow_imports = "skip"` on `specify_cli.*` makes the properly-typed
  `_persist_review_feedback` arrive as `Any`. Narrowed at the boundary, no
  `# type: ignore`.

### Verification on the pushed tip

- **15 failed, 30 passed** across the nine changed test files — 10 intended pins
  plus the 5 pre-existing `test_sync_consent_default_deny.py` reds. Every failure
  is on its substantive assertion; no collateral.
- `ruff`: all checks passed. `mypy`: **no issues found in 9 source files**.
- `test_ci_collection_completeness.py` + `test_gate_coverage.py` +
  `test_no_legacy_terminology.py`: **81 passed** — the rename didn't orphan the
  file, and the changes-filter edit didn't shift the orphan surface.

### Filed, not folded

- **#3053** — *Consent ownership: capture-first durability (FR-017/C-008)
  contradicts #3031 Defect 3.* Two accepted contracts assert opposite things about
  whether a non-consenting project's event may be written. The #3031 pin asserts
  the issue's stated outcome and its docstring names the tension; resolving it is
  an architecture decision, not a test fold.
- **#3030 mechanism correction** posted as a comment (that issue is authored by
  @MOES-Media, so its body is left intact): the ticket pins `sync/batch.py`, the
  legacy `OfflineQueue` transport, while the live path is
  `delivery/dispatcher.dispatch`. A fix following the ticket literally would leave
  the #3050 pin red.

### Still open for the author

The `test_implement.py:533-545` inverted pin is deliberately untouched — but note
the caveat now in the PR body: its precondition is the *dirty-uncommitted* case,
which #2993 says works correctly, so it may survive the fix. Re-adjudicate rather
than deleting it reflexively.

Watching CI. The reds below are the deliverable; the PR is merge-ready once the
expected set is confirmed.

stijn-dejongh:
## CI classification on `726a8593c` — every test red is an intended pin

Ten jobs report failure. **Nine are the deliverable**, and I verified each by
pulling its `FAILED` node IDs rather than inferring from the job name:

| Job | Failures | Verdict |
|---|---|---|
| `regression tests (blocking)` | all 15 pins | deliberate |
| `fast-tests-core-misc (core-misc)` | 3× `tests/delivery/` | deliberate |
| `fast-tests-sync` | #3031 capture gap + 5× `default_deny` | deliberate |
| `integration-tests-agent` | #3045 | deliberate |
| `integration-tests-review` | 2× `test_cycle.py` | deliberate |
| `integration-tests-core-misc (integration)` | #2996(a) | deliberate |
| `integration-tests-core-misc (misc)` | #2993 | deliberate |
| `integration-tests-cli` | #3033 | deliberate |
| `quality-gate` | rollup of the above | consequential |

**Zero collateral failures.** No test outside the pinned set fails anywhere.

Two visibility improvements are now confirmed live rather than merely intended:

- `integration-tests-cli` now carries the **#3033** red. Previously that pin
  surfaced *only* in the dedicated `regression tests` job — its natural shard
  never showed it.
- The blocking job now reports **15** pins rather than 10: the five
  `test_sync_consent_default_deny.py` reds joined the P0 index via the
  `regression` marker fold, so #3031 now shows both arms.

### The one failure that is NOT a pin

`Clean install verification (FR-010)` fails on its **NFR-003 latency regression
gate** — a perf budget, not a correctness signal:

```
pre_cutover_median:  0.745s (745ms)      <- baseline
post_cutover_median: 1.264s (1264ms)
runs: ['1.129', '1.127', '1.264', '1.568', '1.622']
verdict: FAIL
```

Classifying it as a **perf-budget flake on a shared runner**, on four pieces of
evidence:

1. This PR's diff is **test files plus one CI path-filter block**. Nothing in it
   is importable by the CLI, so it cannot move CLI startup latency.
2. The sample spans 1.129s–1.622s — a **44% spread across five runs** of the
   same binary. That is runner noise, not a step change.
3. The job **passes on `main` at the identical base commit** `ed470756e`, and on
   the two `main` runs before it.
4. This same PR **passed this job** on its previous (pre-#3007) base, with no
   change to any file the gate measures in between.

Per the flakiness policy this is *"note it, watch for recurrence, tune the budget
at the root if it repeats — never retry-to-green."* So I have **not** re-run it
and have not touched the baseline. If it reproduces on the next push, the right
response is to re-pin `nfr-003-baseline.json` with a dated rationale — not to
retry until it passes.

Flagging rather than silently absorbing it: if anyone believes the CLI import
path genuinely regressed, that is a `main` question, not a #3050 question.

stijn-dejongh:
## Second remediation pass — the seven residual squad findings are now folded

Tip `4923d6fcc`. Five more folds on top of the first pass; the review residue that
was recorded but never actioned is closed.

### The two MAJORs

**#2996(a) licensed the fabrication this same PR pins against** — `bd762719e`.
The test invoked `move-task --to approved --note …` with **no `--agent`**, then
asserted `reviewer_agent != "unknown"`. Satisfying that from an invocation which
declares no reviewer requires *inferring* provenance from ambient state — defect
(b)'s pathology with the verdict flipped, two files away from where this PR pins
against it. Nothing was asserted about `body` / `affected_files` /
`reproduction_command`, so a synthesized empty artifact passed.

Now: the invocation **declares** `--agent reviewer-renata`
(`agent/tasks.py:610`), and the assertion is an equality echo of the declared
identity plus a non-empty `body`. An inferred reviewer no longer matches the
literal declared value, and a fabricated-empty artifact is caught separately.
`cycle_number == 2` relaxed to `> 1` — the ticket's contract is number-agnostic
and `next_cycle_number` is `len(glob) + 1`, a known double-increment source that
would have redded a correct fix landing at cycle 3.

**#3033 could not distinguish the seam fix from the call-site fix** — `666c4ec19`
+ `b3904f7a8`. Two commits, because the first was honestly insufficient.

`666c4ec19` added a read-back (`git show <ref>:<path>`), a ref-resolvability
check, and a classification precondition guard — the last one closing a real
vacuity risk: if `retrospective.yaml` is ever reclassified out of PRIMARY, the
old test would have gone **silently, accidentally green** with the hole open.

But the implementer then reported that a `_resolve_commit_target` HEAD-fallback
would *still* pass, because `HEAD` is by construction resolvable and readable.
That is correct, and it is a structural limit of a black-box CLI test — not
something more assertions at that entry point can close.

So `b3904f7a8` adds a **seam-level pin**,
`test_resolve_placement_only_rejects_pruned_target_branch_3033`, which calls
`mission_runtime.resolve_placement_only` directly and asserts the ref it returns
actually exists in the repo. It never enters `safe_commit_cmd.py`, so a patch
confined to `_resolve_commit_target` cannot execute on its path. Only an
existence-checked / lifecycle-aware resolution in `get_feature_target_branch` or
`resolve_placement_only` greens it — which is what #3033 §7's *"fixing it at one
call site is whack-a-field"* requires. The two #3033 pins must be discharged
together.

### The five MINORs

- **`58acfe177`** — #2993 parametrized over `placement_ref ∈ {None, CommitTarget}`.
  Production takes the `CommitTarget` path (`implement.py:1706`); only the legacy
  `None` fallback was pinned. The arms provably exercise different code:

  ```
  legacy-fallback : coord tip unmoved,   git log <coord> -- kitty-specs/ empty   -> Cause B (no-op)
  placement-ref   : coord tip 4f16bfbe -> 2cb774dc, log non-empty                -> Cause A (copy without ancestry)
  ```

  Both still fail Assert A first, for independently-fixable reasons — matching
  #2993's two-cause triage exactly.
- **`58acfe177`** — added a coord-descent guard. Assert A alone would be satisfied
  by branching lanes off `main`, which contains the planning artifacts but severs
  the lane from the coordination branch and discards coord lifecycle state.
- **`4923d6fcc`** — the starvation pin re-expressed through `_run_dispatch_batches`
  (`sync.py:807-828`), the loop `sync now` actually traverses, rather than a single
  `dispatch()` call whose shape a legitimate restructure (pushing the predicate into
  `SELECT_ALL_SQL`) would break for non-behavioural reasons. This turned out to be a
  *stronger* pin: tracing the loop's stop condition (`selected == 0 or not advanced`)
  shows a filter-after-`LIMIT` fix yields `delivered=0` **and** no
  `retryable_event_ids`, so the loop breaks on the first batch and strands the
  consenting event permanently — invisible through one `dispatch()` call.

### Verification on `4923d6fcc`

- **17 failed, 30 passed** across the nine changed test files. Up from 15: the
  #2993 parametrization contributes a second arm, and the #3033 seam pin is new.
  Every failure is on its substantive assertion; no collateral.
- `ruff` clean; `mypy` **no issues found in 9 source files**.
- `test_ci_collection_completeness` + `test_gate_coverage` +
  `test_no_legacy_terminology`: **81 passed** — the two new test nodes are collected.

### Honest limit

The #3033 CLI-level test still cannot, by itself, force a seam fix — that is why
the seam pin exists alongside it rather than replacing it. If someone discharges
only the CLI test, the seam pin stays red and says so.

PR #3034, opened 2026-07-28
fix(retrospect): normalize wrapped YAML whitespace

samuelgoff:
## Summary

Fixes ruamel.yaml leaving a wrapping separator as trailing whitespace in nested retrospective scalars. The writer now removes trailing horizontal whitespace only when round-tripping the normalized bytes preserves parsed YAML meaning. Literal scalar content ending in a space or tab remains byte-for-byte semantic content.

## Validation

- `pytest tests/retrospective/test_writer.py tests/retrospective/test_writer_atomicity.py -q` — 42 passed
- `ruff check src/specify_cli/retrospective/writer.py tests/retrospective/test_writer_atomicity.py` — passed
- `mypy --strict src/specify_cli/retrospective/writer.py` — passed
- `git diff --check origin/main...HEAD` — clean
- Independent local diff review found and verified the literal-scalar regression fix.

Remote checks and GitHub review are still pending.

PR #3057, opened 2026-07-28
fix(retrospect): normalize wrapped YAML whitespace + extract shared kernel YAML writer (fixes #3058)

### LINKED ISSUE #3058: Review-artifact override writer wraps reviewer prose → trailing-whitespace artifact (sibling of #3057)
### Summary

`src/specify_cli/review/arbiter.py` serializes the `arbiter_override` decision — which carries reviewer prose/rationale — into committed review-artifact frontmatter using `YAML()` **without setting a width** (ruamel default ≈80). It only `.rstrip("\n")` at document level (`:461`, `:470`), not per-line. A long rationale wraps at ~80 columns and leaves the **same wrapped-scalar trailing-whitespace artifact** that #3057 just fixed for retrospectives.

This is the one concrete sibling that shares all three conditions: the ruamel round-trip mechanism, a long-prose payload, and a missing width guard.

### Locations

- `src/specify_cli/review/arbiter.py:450` and `:467` — `YAML()` constructed without `width`.

Lower-risk same-shape sites (no width set, but payloads are short config so unlikely to wrap in practice — glance, not necessarily action):
- `src/specify_cli/review/baseline.py:153`
- `src/specify_cli/charter_runtime/lint/_drg.py:34`
- `src/specify_cli/core/agent_config.py:64,124`

### Fix

Apply the canonical house pattern `yaml.width = 4096` (used by every other prose-bearing writer: `analysis_report.py`, `frontmatter.py`, `review/artifacts.py`, migration writers), or reuse the retrospective normalizer approach. Add a focused regression test that a long override rationale produces no trailing-whitespace line.

### Provenance

Surfaced by an adversarial review (architect lens) during the maintainer landing of #3057 (retrospective whitespace fix, cherry-picked from the now-closed #3034).

stijn-dejongh:
## Summary

Two related pieces:

1. **Retrospective YAML whitespace fix** (cherry-picked verbatim from @samuelgoff's #3034, authorship preserved). `ruamel.yaml` left a wrapping separator as trailing whitespace in nested retrospective scalars; the writer now removes trailing horizontal whitespace **only when** a `typ="safe"` round-trip confirms YAML meaning is unchanged, so literal-scalar content that legitimately ends in a space/tab stays intact.

2. **Extract a shared YAML-writer seam into the `kernel` layer** and adopt it — turning the one-off retrospective fix into an SSOT primitive, and **fixing the sibling bug #3058** (arbiter review-artifact writer wrapped reviewer prose at ruamel's default width → same trailing-whitespace artifact in committed review frontmatter).

## What changed

**New `src/kernel/yaml_io.py`** (dependency-free root layer):
- `CANONICAL_YAML_WIDTH = 4096`
- `serialize_mapping(data, *, width=4096) -> bytes` — canonical rt config (`default_flow_style=False`, `preserve_quotes=True`, width) + the trailing-whitespace normalizer. Pinned to `typ="rt"` (a safe dumper can't represent the `CommentedMap` the arbiter mutates — enforced by test).
- `write_mapping_atomic(data, path, *, width=4096, mkdir=False)` — durable atomic write (O_EXCL tmp + `fsync` + `os.replace` + best-effort dir `fsync`); a durability-enhanced sibling of `kernel.atomic.atomic_write`.
- The trailing-whitespace normalizer moved here verbatim from `retrospective/writer.py`.

**Adopters:**
- `retrospective/writer.py` → delegates to `write_mapping_atomic(width=120)`, keeping `WriterError` translation. **Width stays 120** — issue #3059 owns that decision; the normalizer (not the width) is what cures the bug.
- `review/arbiter.py` → both frontmatter dump sites use `serialize_mapping` (80→4096 + normalizer). **Fixes #3058.**
- `review/artifacts.py` → `write()` migrated to `serialize_mapping` (exact-match config; byte-stable within the 4096 wrap width, pinned by test) — closes the #3058 class for the whole review-cycle artifact family.

Deliberately **not** touched: `frontmatter.py` (its `preserve_quotes=False` + custom indent would churn every WP task file), plus charter/migration writers — left as opportunistic follow-ups.

## Validation

- `ruff` ✅ · `mypy --strict` ✅ (all 4 touched src files, zero suppressions)
- New tests: `tests/kernel/test_yaml_io.py` (13), arbiter regression (+1, red-first confirmed), `tests/review/test_artifacts_yaml_seam.py` byte-stability pin (2). Touched-area suites: **106 passed**.
- Layer rules: `kernel/yaml_io.py` imports only stdlib + ruamel — `test_layer_rules.py` kernel-is-root invariant passes.
- Two `tests/review/test_cycle.py` failures are **pre-existing on upstream/main** (confirmed on the true base), unrelated to this change.
- **Two adversarial-review passes** (correctness + SSOT/boundary on the retro fix; behavior-preservation on the consolidation): clean, no blockers. One wording overstatement folded.

## Provenance

Maintainer landing of #3034's retrospective fix (that PR was a stale 13-commit stack bundling an unrelated charter change; closed with a note). This PR carries only the genuine retrospective fix plus the kernel consolidation it motivated.

### DIFF
--- a/src/kernel/yaml_io.py
+++ b/src/kernel/yaml_io.py
@@ -1,0 +1,140 @@
+"""Shared YAML serialization and durable-write primitives for the kernel layer.
+
+Canonical home for the house YAML-dump conventions used by every writer that
+persists a mapping (dict / ``CommentedMap``) as YAML: round-trip dumper (so
+``ruamel`` comment/quote/anchor fidelity and ``CommentedMap`` inputs survive),
+a wide fixed line width (so wrapped prose never fragments a scalar across
+lines), and a post-dump normalizer that strips the non-semantic trailing
+whitespace ``ruamel`` leaves on wrapped-scalar continuation lines.
+
+Adopted by ``specify_cli.retrospective.writer`` (moved here verbatim) and
+``specify_cli.review.arbiter`` (#3058) so both surfaces share one seam instead
+of hand-rolling the same dump-then-normalize dance with drifting width
+settings.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import os
+from collections.abc import Mapping
+from io import BytesIO
+from pathlib import Path
+from typing import Any
+
+from ruamel.yaml import YAML
+from ruamel.yaml.error import YAMLError
+
+# Matches the house "prevent line wrapping" width used across writers
+# (e.g. ``specify_cli.review.artifacts._make_yaml``): wide enough that no
+# reviewer-prose or retrospective-detail scalar wraps under normal use.
+CANONICAL_YAML_WIDTH = 4096
+
+
+def serialize_mapping(data: Mapping[str, Any], *, width: int = CANONICAL_YAML_WIDTH) -> bytes:
+    """Serialize a mapping to canonical YAML bytes.
+
+    Uses a round-trip (``typ="rt"``) dumper, not a safe dumper: a safe dumper
+    raises ``RepresenterError`` on a ``ruamel.yaml.comments.CommentedMap``
+    (arbiter and other frontmatter-merge callers pass one in, since they load
+    existing frontmatter with ``preserve_quotes=True`` before mutating it),
+    and it would also strip the comment/quote fidelity round-trip is for.
+
+    The dump is post-processed by :func:`_normalize_nonsemantic_trailing_whitespace`
+    to remove the trailing-space/tab artifacts ``ruamel`` leaves on wrapped
+    scalar continuation lines, without touching semantically-significant
+    trailing whitespace inside literal block scalars.
+    """
+    yaml = YAML(typ="rt")
+    yaml.default_flow_style = False
+    yaml.preserve_quotes = True
+    yaml.width = width
+
+    buf = BytesIO()
+    yaml.dump(data, buf)
+    return _normalize_nonsemantic_trailing_whitespace(buf.getvalue())
+
+
+def _normalize_nonsemantic_trailing_whitespace(serialized: bytes) -> bytes:
+    """Remove writer wrapping artifacts only when YAML meaning remains unchanged.
+
+    Falls back to the original bytes if the safety re-parse cannot confirm the
+    normalized form is semantically identical, so normalization is never worse
+    than emitting the un-normalized dump.
+    """
+    normalized = b"\n".join(line.rstrip(b" \t\r") for line in serialized.split(b"\n"))
+    if normalized == serialized:
+        return serialized
+
+    yaml = YAML(typ="safe")
+    try:
+        original_data = yaml.load(serialized)
+        normalized_data = yaml.load(normalized)
+    except YAMLError:
+        return serialized
+    return normalized if normalized_data == original_data else serialized
+
+
+def write_mapping_atomic(
+    data: Mapping[str, Any],
+    path: Path,
+    *,
+    width: int = CANONICAL_YAML_WIDTH,
+    mkdir: bool = False,
+) -> None:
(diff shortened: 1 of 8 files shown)


stijn-dejongh:
## Adversarial review — summary (2 lenses)

Ran a 2-lens squad over the focused diff. **Verdict: clean to land. No blockers.** Two review minors folded into `1482cd2f3`; two out-of-scope findings filed as follow-ups.

### Correctness lens (round-trip guard, exception paths, test integrity)
- The safe-load round-trip equality guard neutralizes every corruption vector probed: root is always a mapping (single-doc), no custom tags/anchors, CRLF-safe (`yaml.dump` emits LF; semantic `\r` is emitted as an escape, never a raw trailing byte).
- Literal-scalar trailing whitespace is correctly preserved (parametrized test proves it).
- Red-first confirmed sound.

**Folded:**
1. `_normalize_nonsemantic_trailing_whitespace` now falls back to the original bytes on `YAMLError` from the safety re-parse, so normalization is never worse than emitting the un-normalized dump (was: abort the write via `WriterError`).
2. The strip test now asserts the un-normalized dump genuinely wraps and leaves a trailing-whitespace line, so it can't silently go vacuous if a future ruamel stops emitting the artifact. (Re-verified: still reds correctly when the fix is neutralized.)

### SSOT / pattern-completeness lens
- The helper is genuinely new — no pre-existing per-line YAML whitespace seam to reuse. Placement in `writer.py` is fine.
- **Root-cause observation (follow-up #3059):** `retrospective/writer.py` uses `width=120`, the lone outlier vs the canonical `width=4096` every other prose writer uses — that setting is what *causes* the wrapping. The normalizer is a correct fix regardless (and a no-op if width is later aligned), so this is a recorded-decision follow-up, not a blocker.
- **Sibling bug (follow-up #3058):** `review/arbiter.py` serializes reviewer prose with no width guard → same trailing-whitespace exposure on committed review artifacts.

### Gates (local, on the folded tip)
- `ruff` ✅ · `mypy --strict` ✅ · `pytest tests/retrospective/ -q` → 42 passed ✅ · red-first re-verified ✅

Ready for the operator to merge once remote CI is green. I won't merge it.

stijn-dejongh:
## Consolidation folded in + reviewed

Grew this PR from the minimal retrospective fix into the kernel-writer consolidation (per maintainer direction), folding in **#3058**.

**Adversarial review (reviewer lens, behavior-preservation focus): CLEAN, no blockers.** Independently verified:
- `retrospective/writer.py` output is **byte-identical** (width still 120, normalizer moved verbatim, full durability sequence intact, `WriterError` translation preserved with no error path escaping raw).
- `arbiter.py` quote/anchor/comment fidelity survives the dump-instance swap (rt round-trip); `.decode().rstrip("\n")` matches the old contract; 80→4096 is the intended #3058 fix.
- `write_mapping_atomic` is genuinely atomic + durable (same-dir tmp, fd fsync before replace, cleanup-on-failure, no fd leaks).
- rt-pinning is **enforced by test** (safe dumper → RepresenterError on `CommentedMap`).
- Arbiter regression **red-first empirically reproduced** (old path wraps ~370-char prose across 8 lines, 4 with trailing whitespace).

One wording overstatement in the artifacts.py byte-stability claim was folded (`d76945f9b`): it's byte-identical for scalars within the 4096 wrap width; beyond that `serialize_mapping` *improves* on the old path by stripping the same latent whitespace.

**Gates:** ruff ✅ · mypy --strict ✅ · touched-area suites 106 passed · kernel layer invariant holds. The 2 `test_cycle.py` reds are pre-existing on upstream/main.

Ready for the operator to merge once remote CI is green. I won't merge it.

Issue #3078, opened 2026-07-30
mypy --strict: no-any-return in charter.context._render_profile_tactics


stijn-dejongh:
## What

`uv run mypy --strict src/charter/context.py` reports one pre-existing error, unrelated to any in-flight WP:

```
src/charter/context.py:2061: error: Returning Any from function declared to return "list[str]"  [no-any-return]
```

(Line number is post-WP04-extraction; on `kitty/mission-doctrine-delivery-activation-01KYQVQK` pre-WP04 it is line 2238.)

## Where

`_render_profile_tactics` (`src/charter/context.py`), on its `return _render_profile_selector_refs(...)` statement. `_render_profile_selector_refs` is imported (aliased) from `charter.context_renderers.profile_sections` (WP12/T067 extraction) — something in that call chain resolves to `Any`, which mypy strict flags when the enclosing function is declared `-> list[str]`.

## Why this is pre-existing, not caused by the current change

Confirmed via `git stash` on `kitty/mission-doctrine-delivery-activation-01KYQVQK-lane-d` (WP04, mission `doctrine-delivery-activation-01KYQVQK`): the same single error is present with the unmodified file (`context.py:2238`), before any WP04 edits. WP04's scope is the 12 `# type: ignore[attr-defined]` sites (`ArtifactRepository` Protocol typing) plus two unrelated extractions (reference-pointer + delivery-table helpers to `context_renderers/`); this `no-any-return` finding is neither an attr-defined ignore site nor inside code WP04 touches, so fixing it is out of scope for that WP per its own anti-scope-creep guidance.

## Command to reproduce

```bash
uv run mypy --strict src/charter/context.py
```

## Suggested next step

Trace the `Any` leak through `charter.context_renderers.profile_sections.render_profile_selector_refs` (or its return-type annotation) and tighten it, or confirm the loose typing is intentional and add a narrowly-scoped, justified `# type: ignore[no-any-return]` per the repo's typing standards.

robertDouglass:
Triage validation (2026-08-02): KEEP OPEN. Current source/planning records still show this as an actionable recent slice, with no completion or supersession evidence. Retain it in the active queue and add a focused repro/acceptance fixture plus owner/dependency links before implementation; do not close on age alone.

uuzzrm:
I would like to take this issue.

[...]

PR #3070, opened 2026-07-29
Doctrine delivery reachability (mission #187) + #3063 DRG topology authoring

stijn-dejongh:
## What this delivers

Two tightly-coupled pieces, shipped as one slice:

Closes the "declared-but-never-reaches-its-consumer" defect class at the **delivery layer**. Predecessor #3007 wired 4 edges to sources that were themselves unreachable, so the wired artefacts still reached nobody; this mission makes reachability a **measured** property and fixes the rails that carry doctrine to an agent's context:
- asset resolution + the first shipped doctrine **asset** (`common-docs-structural-lint`);
- charter activation reaching the **action boundary**;
- the DRG **writer registry** (per-field writers) so a new top-level graph field can't drop silently;
- **per-channel reachability** measurement (`action_channel_reachable` / `resolve_context`), pinned as named sets;
- **progressive disclosure** (links-not-bodies default cadence) and **every-load delivery**;
- the **profile channel**.

All 15 WPs were independently mutation-reviewed. Issue-matrix #3038 settled to `deferred-with-followup` (the `_KIND_TO_NODE_KIND` project-tier remainder stays on #3038).

### 2. #3063 — DRG topology authoring (operator interview, families A–F)
The C-007 deferred-reachability decision surface, ruled family-by-family by the operator and authored into `HAND_AUTHORED_EDGES`. The authoritative record is [`docs/plans/doctrine/delivery-reachability-wiring-table.md`](docs/plans/doctrine/delivery-reachability-wiring-table.md).

| family | disposition |
|---|---|
| **A** DDD | **delivered at specify** (`software-dev/specify --scope--> domain-driven-design`; +12 action-reachable) |
| **B** refactoring | new `DISCIPLINED_REFACTORING` directive + 14 edges, **inert topology** (delivery via fast-follow profile-walk) |
| **C** C4 / arch-docs | new `USE_C4_MODEL_TECHNIQUES` directive + 9 edges, **inert topology** |
| **D** testing / BDD / mutation | **delivered at implement/review** via existing action-scoped 034/030; 5 new artefacts + 54 edges (−10 from deferred set) |
| **E** analysis / terminology / reasons-canvas | 9 edges, **inert topology** |
| **F** orphan directives | **ACCEPT** 038/039/035 as intentionally charter/mission-gated; 044 → #3064 (default-charter always-on) |
| depth-gated (4) | **ACCEPT** (reachable-in-principle, beyond eager horizon by design; depth-tuning → fast-follow) |

[...]

Issue #3064, opened 2026-07-29
Follow-up WP: empty-charter dispatch behaviour — generic agent + warning + default charter asset


stijn-dejongh:
Follow-up work package capturing the operator's **2026-07-29** decision on empty-charter dispatch behaviour. Supersedes the earlier "retire the `resolver.py` routing fail-open" framing (see Note below).

## Context

Mission `doctrine-delivery-reachability-01KYMXD6` fixed **FR-018** at the **delivery** boundary: absence of `activated_<kind>` keys now resolves to **EMPTY**, not "all built-ins", plus a migration that writes explicit `[]`.

The invocation-**routing/dispatch** layer, however, still falls back to **"all built-ins"** when a project is unconfigured. That fallback is a **deliberate, tested contract** — `tests/specify_cli/invocation/test_registry_builtin_activation_parity.py::test_no_activation_key_admits_all_builtins_in_routing`. Retiring it in-mission was correctly **deferred**: it is a cross-subsystem behaviour change needing product sign-off, and ~**17 routing fixtures** depend on the current behaviour.

## Operator decision (2026-07-29) — the design for this follow-up WP

1. When the charter is empty/unconfigured, the tooling dispatches to a **generic agent** and does **NOT** use charter/doctrine at all. This replaces **both** the current "all built-ins" fallback **and** the naive "hard-fail" alternative.
2. The user MUST be **warned** of this behaviour and advised to activate their repository's charter before running.
3. The **only** always-on doctrine artefact is the **`software-dev` mission type**.
4. Ship a **default `charter.yml`** in the built-in doctrine as an **asset** that a user can activate/copy if they want to forego authoring their own charter. This leverages the asset-resolution + `doctrine asset` surface that mission `doctrine-delivery-reachability-01KYMXD6` just delivered (WP03–05).

## Note

This **supersedes** the earlier "retire the `resolver.py` routing fail-open" framing. The routing fallback changes from "all built-ins" to **"generic agent, no doctrine, with a warning"** — **not** to a hard failure.

## Scope

- Product sign-off on the generic-agent / warning UX.
- The routing-layer change (retire the "all built-ins" fallback; ~**17 fixtures** in `test_registry_builtin_activation_parity.py` and siblings depend on the current contract and must be re-baselined to the new behaviour).
- The built-in `charter.yml` **asset** (default charter shipped via the `doctrine asset` surface).
- Keep the `software-dev` mission type always-on.

## Cross-references

[...]

Issue #3063, opened 2026-07-29
Operator interview: 72 activated-but-unreachable artefacts deferred under C-007


stijn-dejongh:
Filed from mission `doctrine-delivery-reachability-01KYMXD6` (WP09, FR-015). Tracked so the **C-007** decision surface is not lost when the mission closes.

## The decision surface

WP09 produced the FR-015 wiring table. **72 activated artefacts remain action-unreachable at d=2 with no C-007-passing source** — i.e. no attested relationship **and** action-reachable source. They are enumerated in `docs/plans/doctrine/delivery-reachability-wiring-table.md` (lands with the mission).

C-007 defers these to an **after-mission operator interview**: each of the 72 needs an operator decision on whether it should be wired to an action-reachable source (and how), deactivated, or accepted as intentionally unreachable.

## Why tracked

Without a home this decision evaporates when the mission closes, leaving 72 activated artefacts permanently reachable-to-nobody — the exact defect class the programme exists to close, sitting just below the line the mission could fix in scope.

## Related

- #3056 — the *other* deferred half of this programme (progressive-disclosure fetch instruction + 118 `when`-edge backfill). **Distinct**: that concerns utilization / guidance quality of an *existing* link set; this is the operator decision on 72 artefacts that have **no** passing source at all.
- Wiring table: `docs/plans/doctrine/delivery-reachability-wiring-table.md`.

PR #3081, opened 2026-07-30
feat(doctrine): activate profile-channel suggests-edge delivery (fast-follow to #3070)

### LINKED ISSUE #2977: A second field-by-field DRG writer in rewrite_opposed_by silently drops new edge fields from org packs
## The gap

`src/specify_cli/migration/rewrite_opposed_by.py:338-357` declares its **own** `_node_to_dict` / `_edge_to_dict`, hand-restating the field list:

```python
def _edge_to_dict(edge: DRGEdge) -> dict[str, Any]:
    out = {"source": edge.source, "target": edge.target, "relation": edge.relation.value}
    if edge.reason: out["reason"] = edge.reason
    if edge.when:   out["when"] = edge.when
    return out
```

It reads an org pack's `*.graph.yaml` through `DRGGraph.model_validate` and writes it back through this. **So a new `DRGEdge` field authored in an org pack is silently deleted by `spec-kitty migrate rewrite-opposed-by`.**

Same defect class as the extractor writers that mission `doctrine-silence-guards-01KYFV7Q` WP04 just closed — different surface. WP04 derived the extractor's writers from `model_fields` so the class closes rather than the instance; this copy was not in its ownership map and was reported rather than swept.

## Why it matters now

Mission **B1** adds `Relation.IMPACTS` and `is_symmetric` to `DRGEdge`; **B2** adds `aliases`. WP04's fix means those survive the built-in graph's write path. They will **not** survive this one.

The built-in tree is safe because it does not run through this migration. Org and project-tier packs do, so the exposure is real but scoped to external packs.

## Related, found in the same pass

`DRGGraph` itself declares no `model_config`, so an **extra top-level key** in a fragment is still silently ignored. WP04 measured zero occurrences today and deliberately left it: FR-004 named three models, and tightening `DRGGraph` would break external packs that carry extra top-level keys. Worth deciding rather than inheriting.

## Suggested direction

Promote one canonical serializer instead of a third hand-restatement. WP04's note is the honest constraint: doing so means exposing a public serializer across the `doctrine` / `specify_cli` package boundary, which is an architectural call rather than a refactor — hence an issue rather than a fold.

Whatever lands should be covered the way WP04 covered the extractor: assert emitted keys equal `set(DRGEdge.model_fields) - <explicitly withheld>`, so a field added tomorrow is either emitted or deliberately withheld, with no third outcome. WP04's `test_the_hand_written_writer_shape_would_drop_the_new_field` is the pattern — it runs the *replaced* writer shape against the same edge and asserts it drops the field, pinning the defect rather than only the fix.

### LINKED ISSUE #3062: `DRGGraphSchemaError` surfaces as an uncaught traceback at pack-load sites instead of a structured ValidationIssue
Filed from mission `doctrine-delivery-reachability-01KYMXD6` (WP02 follow-up; the fix below is out of WP02's owned surface).

## Context — the intended behavior is correct

WP02 added `extra="forbid"` to `DRGGraph` plus a typed `DRGGraphSchemaError`, **deliberately not** a `DRGLoadError` subclass, so a stray top-level key fails closed *past* the silent-degrade handlers (intended NFR-006 fail-closed behavior). That part is working as designed and should stay.

## The gap — it fails closed, but not legibly

At the consumer pack-load sites — `src/specify_cli/doctrine/pack_validator.py` and `src/specify_cli/doctrine/pack_assembler.py` — `DRGGraphSchemaError` surfaces as an **uncaught traceback** rather than a structured `ValidationIssue` during `doctrine validate`. Failing closed is right; presenting it to an operator as a raw Python traceback is not.

## Follow-up

Catch `DRGGraphSchemaError` at those two pack-load sites and structure it as a `ValidationIssue`, preserving the fail-closed semantics — validation still fails, it just fails *legibly*.

## Minor related note (fold in if convenient)

WP04's `AssetRepository._pre_validate` records `_source_paths[id]` **before** model validation, so `source_path(id)` can return a path for an id that `get()` treats as absent. Cosmetic fail-closed-hygiene, not a correctness bug — recorded so it is not rediscovered cold.

## Related

Adjacent to #2977 (the DRGGraph strictness / `model_config` discussion) but **distinct**: #2977 debates *whether* to tighten `DRGGraph`; WP02 already tightened it, and this issue is purely the **consumer-UX** of the resulting error at the load sites.

### LINKED ISSUE #3075: Follow-up (PR #3070 squad): DRG writer-discovery gate + 2 unregistered DRGGraph emit sites + repo Protocol typing
Follow-up from the adversarial review squad run during the PR #3070 landing pass (doctrine delivery-reachability mission). These are MINOR/structural findings deliberately **not** folded into the landing pass to keep it reviewable; capturing them here so they aren't lost. None is a correctness blocker.

### 1. DRG writer registry has no *discovery* gate (completeness asserted by fiat)
`src/specify_cli/drg_writers/registry.py` + `tests/specify_cli/drg_writers/test_registry_completeness.py`.
The mutation gate is genuinely non-vacuous for **registered** writers, but there is no gate that *discovers* writer sites (no AST/`rglob` scan for `_node_to_dict`/`_edge_to_dict`/graph-dict-shaped functions). A future hand-restating writer added in a new module and simply not appended to the `Final` tuple is invisible — the "a writer nobody registered" blind spot, relocated from *field* to *writer site*. The module docstring frames explicit wiring as *avoiding* that blind spot when it only *relocates* it.
**Fix:** add a discovery test that scans `src/` for functions returning node/edge/graph dicts and asserts each is a registry member; or soften the docstring to state the gate covers registered writers only (with strict-model derivation as the new-writer mitigation).

### 2. Two `DRGGraph` document-emit sites bypass the derived writer (concrete instance of #1)
`src/charter/synthesizer/project_drg.py` (`_serialize_graph`) and `src/specify_cli/migration/rewrite_opposed_by.py` (`_write_graph`) both hand-restate the five `DRGGraph` top-level keys instead of routing through `graph_document_to_dict`, and neither is a `DOCUMENT_WRITERS` member. A new `DRGGraph` top-level field would be emitted by the extractor but silently dropped by these two. (Node/edge fields within the payloads are safe — they route through the derived `model_to_graph_dict`; only document-level keys are at risk.)
**Fix:** replace both literal payload dicts with `graph_document_to_dict(graph)` and register both as `DocumentWriter` members.
**Note:** the `rewrite_opposed_by._write_graph` half is already tracked by #2977 — dedupe against it; this issue adds the `project_drg._serialize_graph` half + the discovery-gate (#1).

### 3. New `# type: ignore[attr-defined]` from `object`-typed repositories
`src/charter/context.py` (~3297, 3436–3439) and `src/charter/progressive_disclosure.py` (~258–259) carry `attr-defined` ignores only because the repository/service surfaces are typed as bare `object` and then duck-typed (`repository.get` / `get_provenance`; `service.directives/tactics/styleguides/toolguides`).
**Fix:** introduce a `Protocol` (e.g. `SupportsArtifactLookup` with `get`/`get_provenance`) threaded through `collect_typed_artifacts`/`build_disclosure_payload`, deleting the ignores.

---
Context: PR #3070 · mission doctrine-delivery-reachability (#3063 lineage) · related #2977. Line numbers approximate to the PR #3070 tip; re-anchor when picked up.

### LINKED ISSUE #3079: Pre-existing red: test_context_org_governance.py::test_builtin_profile_path_unchanged_and_deterministic
## What

```bash
uv run pytest tests/charter/test_context_org_governance.py::TestNoOrgPacksGovernanceRegression::test_builtin_profile_path_unchanged_and_deterministic -q
```

fails deterministically (not flaky — reproduced twice) with:

```
AssertionError: assert 'python-pedro' in 'Governance:\n  - Template set: software-dev-default\n  - Paradigms: (none)\n  - Tools: git, spec-kitty\nDirective IDs...'
```

Note ``Tools: git, spec-kitty`` in the rendered output — missing ``mypy``/``pytest``/``ruff``, which suggests the profile ("python-pedro") is not resolving against the built-in doctrine catalog for the test's fresh ``tmp_path`` repo (no ``.kittify/config.yaml``, no org pack), so the profile id string never appears in the rendered governance text.

## Where

Branch: `kitty/mission-doctrine-delivery-activation-01KYQVQK-lane-d` (mission `doctrine-delivery-activation-01KYQVQK`, WP04), but confirmed pre-existing — reproduced identically on the tip of `kitty/mission-doctrine-delivery-activation-01KYQVQK-lane-d` *before* WP04's own edits (commit `d0838843d`, via `git stash`).

## Why this is not WP04's regression

WP04's scope is `src/charter/context.py` (reference-pointer + delivery-table extraction to `context_renderers/`, `ArtifactRepository` Protocol typing) and `src/charter/progressive_disclosure.py` (2 documented type-only sites). This test exercises the dispatch governance-context / profile-resolution path (`_governance_text` → `build_charter_context(..., profile=...)` → `_load_agent_profile` → activation-aware profile map) and is unrelated to either extraction or to any of the 12 retyped call sites. The failure reproduces identically with WP04's changes fully reverted (`git stash`).

## Command to reproduce

```bash
uv run pytest tests/charter/test_context_org_governance.py::TestNoOrgPacksGovernanceRegression::test_builtin_profile_path_unchanged_and_deterministic -q
```

## Suggested next step

Trace why the built-in `python-pedro` profile doesn't resolve (or its directive/tactic references don't render) against a repo with no `.kittify/config.yaml` and no org pack — likely a doctrine-root / built-in-catalog resolution regression from a recent doctrine-tree move, independent of this mission.

stijn-dejongh:
## Summary

Fast-follow to the doctrine-delivery-reachability slice (#3070, merged). PR #3070 authored the A–E `suggests` edge families and the delivery-rail forward API, but the **profile-channel reachability walk still traversed only `{requires, specializes_from}`** — so every authored edge was **inert**. This mission animates the topology: the profile channel now follows `suggests` and surfaces each edge's `when` clause as the applicability condition, delivered as links-not-bodies. It then reconciles the reachability pins + wiring table + forward-API allowlist to the new reality, delivers the Family C templates and Family B anti-patterns by edge, and closes the trailing hygiene tickets.

Live end-to-end proof (aggregate squad): resolving `architect-alphonso`'s profile channel now reaches 129 artefacts including `domain-driven-design` and the C4 paradigms, each rendered with its `when` label — the delivery the #3063 topology needed.

## Work packages (7, all independently reviewed)

- **WP01 — profile-channel `suggests` walk + `when` delivery** (core). `PROFILE_CHANNEL_RELATIONS += SUGGESTS`; a charter-layer projection reuses `link_references(roots=seeds, …)` (no re-walk) so seed-sourced Family A/B edges surface their `when`; a new `render_profile_suggested_doctrine` renderer delivers non-procedure kinds as `when`-labelled links (NFR-003). Wires `agent_profile_seed_urns`.
- **WP02 — DRG topology authoring.** Family-A `when` backfill; a C4 `template:instantiates` edge from `action:documentation/design`; **7 refactoring anti-patterns** via `tactic --REJECTS--> anti_pattern`, each grounded in the tactic's attested `problem`/`when` (**11 deferred** — un-attested, no fabrication, C-004); cardinality/histogram goldens + composition ledger.
- **WP03 — reachability reconciliation (terminal).** `_PROFILE_UNREACHABLE` 153→60, `_PROFILE_RESCUES` 2→30, measured only via the canonical helper; wiring table (neither-channel deferred **48→20**); allowlist sweep. Per-member NFR-002 ledger + a non-vacuous cross-check test. (The reviewer independently re-measured every pin member-by-member.)
- **WP04 — `context.py` extraction + Protocol typing.** Reference-pointer + delivery-table helpers extracted to `context_renderers/` (byte-identical, 153 consumer tests green); `ArtifactRepository` Protocol removes all **12 `# type: ignore[attr-defined]`** (net-removal, `mypy --strict`).
- **WP05 — DRG writer registry unification + discovery gate.** All **three** document-emit sites routed through the canonical `graph_document_to_dict` (incl. `pack_assembler`, which was silently **crashing `yaml.safe_dump` on the `Relation` enum**); registered; a non-vacuous AST discovery gate proven RED independently for **both** the dict-literal and `.model_dump()` shapes.
- **WP06 — `DRGGraphSchemaError` UX + asset source-path.** Structured `ValidationIssue` at both `pack_validator` sites (fail-closed preserved); a base `_post_validate` success-path hook fixes the asset `_source_paths` split-brain (the AgentProfileRepository "twin" was verified a no-op — regression test added, no manufactured diff).
- **WP07 — hermetic delivery-test fixture.** `test_every_load_delivery` fixture excludes the ambient `context-state.json` (was a local-only false-red).

## Issue closure

Closes #3075
Closes #2977
Closes #3062
Closes #3079

- **#3075 / #2977** — all three writer sites routed + registered + the discovery gate + the Protocol typing (C-007 whole-fix).
- **#3062** — schema-error UX + asset `_source_paths` ordering.
- **#3079** — surfaced by the pre-merge aggregate squad as **mission-introduced** (the suggests-walk enriches `python-pedro`'s profile-citations past the per-section budget → delivered as a fetch pointer, dropping the inline profile-id). This is the intended NFR-003 context-bloat behavior, so the governance guard was re-pinned on its real invariants (determinism, org-isolation, profile resolves via a stable directive anchor). Originally mis-filed as pre-existing by a WP-internal-base check; reclassified + resolved here.

## Refs (not closed)

- **Refs #3063** — the A–E `suggests` families are now delivered (its core ask); ~20 broader activated-but-unreachable artefacts remain a documented follow-on. Marked resolved in the mission issue-matrix; please confirm full closure.
- **Refs #2532** — WP04 extracts a slice; full de-god of the 3528-line `context.py` remains.
- **Refs #3070 / #3077** — parent slice (already merged); this builds on it.
- **Refs #3064** — out of scope per spec C-006 (separate mission).
- **Refs #3078** — genuinely pre-existing `mypy no-any-return` in `_render_profile_tactics` (red on the mission base `10e970ed2`); filed, not fixed.

## Follow-ups

- Polish the generated `when` prose (`profile_sections.py`): it prepends "When you " to clauses that are already full sentences (`"When you designing or reviewing…"`). Functional but low prose quality on the primary delivered surface — worth a `when_clause` normalization ticket.
- FR-006 retired **1** forward-API symbol (`agent_profile_seed_urns`), not the spec's ≈2–3 estimate — the other symbols legitimately gained no `src` consumer here (no manufactured importers).

## Test evidence

- Rebased onto current upstream/main (`b04da00e1`, +3k lines); **all mission gates green on the new base: 107 passed** (reachability, profile-suggests-delivery, no-dead-symbols, drg-writer-discovery, no-authored-applies-edge, every-load-delivery, context-org-governance, terminology guard).
- Every WP passed independent opus review; a 2-lens pre-merge aggregate squad verified cross-WP composition (delivery chain live end-to-end; seams clean; goldens honest — the D19 divergence was measured, not green-washed).

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3081"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787992029&installation_model_id=427282&pr_number=3081&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3081&signature=6241bc3998cd6143ef1017fe4cc7f74776af8378d6a15de5da149b9b38006dd7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


stijn-dejongh:
### CI classification (after the fix push)

**Mission-caused reds — fixed in this PR:**
- `arch-adversarial (shard 3)` golden-count 271>270 — WP05's `pack_assembler` regression added 2 cardinality asserts; annotated `# golden-count: cardinality-is-contract`.
- `fast-tests-charter` — WP04 narrowed `_render_directive_include`/`_render_tactic_include` to take the repo, moving `service.directives` to the call site; `test_kind_cascade_exhaustive`'s bare-`object()` stub was coupled to the old internal structure (production is unaffected — real services always have `.directives`). Updated the stub.

**Pre-existing honest-P0 reds — NOT mission-caused (rigorously verified against base `b04da00e1`):** the mission touches **none** of `charter/sync.py`, `merge/executor.py`, `review/cycle.py`, or the dispatcher — all UNCHANGED. These are the known red-first repros that gate releases while their P0s stay open:
- `regression tests (blocking)` / `fast-tests-core-misc`: #3030 / #3031 (drain/consent), #3045 (charter-sync surfaces), #2804 (merge clobbers gate artifacts), review-cycle rejection. Red on main by design.

**Advisory:** `diff-coverage` / `quality-gate` are dominated by this PR's own new code — not padded.

Per the red-main release policy, the remaining reds are the honest baseline, not a regression introduced here.

stijn-dejongh:
**CI settled** — the 2 mission-caused reds are now green: `arch-adversarial (shard 3)` ✓ and `fast-tests-charter` ✓ (plus docs-freshness ✓). The remaining `CI Quality` reds (regression-blocking, fast-tests-core-misc, diff-coverage, quality-gate) are exclusively the pre-existing honest-P0 repros in mission-untouched code (#3030/#3031/#3045/#2804/#3033/review-cycle) + advisory coverage, per the classification above. This PR introduces no regressions; ready for review.

stijn-dejongh:
**Ready for review.** The one mission-caused CI red (`diff-coverage`, the PR's own new code at 89%) is now green — added real focused tests for the new `context_renderers/` edge/error branches + a `pragma` on the uncoverable `ArtifactRepository` Protocol stubs (commit `f6c29cc1b`), and a CHANGELOG entry (`24a275b06`).

The remaining reds are exclusively the pre-existing honest-P0 red-first repros in code this PR never touches — #3030/#3031 (delivery/sync consent), #3045 (charter-sync), #2804 (merge), #3033 (safe-commit), review-cycle — plus the `quality-gate` aggregate over them. Verified against base `b04da00e1`: `charter/sync.py`, `merge/executor.py`, `review/cycle.py` and the dispatcher are all UNCHANGED. This PR introduces zero regressions.

Issue #3086, opened 2026-07-30
merge never clears coordination_branch from meta.json after deleting it — every merged mission later crashes retrospect create with CoordinationBranchDeleted


robertDouglass:
## Summary

`spec-kitty merge --mission <slug>` (with the default `--delete-branch`) deletes the mission's coordination branch (`kitty/mission-<slug>-<mid8>`) from git on every successful run, but **never clears the corresponding `coordination_branch` key from that mission's `kitty-specs/<slug>/meta.json`**. This leaves every successfully-merged mission permanently inconsistent: `meta.json` declares a `coordination_branch` that git no longer has.

Any later command that routes through the canonical status-surface resolver (`resolve_status_surface_with_anchor`, `src/specify_cli/coordination/surface_resolver.py`) for that mission then hits `CoordState.DELETED` and raises `CoordinationBranchDeleted` — a class designed to signal *accidental data loss* from a deleted branch, but here it's actually the guaranteed, correct end-state of a successful merge, misclassified as an error.

Confirmed real-world impact: `spec-kitty retrospect create --mission <slug> --update` crashes with an unhandled traceback for any such mission.

## Evidence this is not a synthetic edge case

In a real downstream project (`Priivacy-ai/spec-kitty-saas`), **every** historical mission whose `meta.json` declares a non-null `coordination_branch` (~10 missions, spanning months, unrelated to any single session) already has that branch deleted from git — **100% hit rate**. One was verified in detail (`team-model-private-and-collaborative-01KW1VSE`): merged via a real GitHub PR (#353) months ago, `target_branch` also long gone (normal post-PR cleanup), yet `meta.json` still carries a `coordination_branch` pointing at a branch git has never had reachable since. This is the guaranteed steady state of every mission merged with `--delete-branch` (the default) — not a rare edge case.

## Root cause (v3.2.6, commit `b04da00e1`)

- `src/specify_cli/merge/executor.py:1291-1325` — the `if run.delete_branch:` block in `_phase_cleanup_worktrees_and_branches` deletes the lane branches AND the mission/coordination branch (`lanes_manifest.mission_branch` — the same value stored as `coordination_branch` in `meta.json`, per `src/specify_cli/core/mission_creation.py:61-66`). No write-back to `meta.json` happens anywhere in this function, the rest of `merge/executor.py`, or `src/specify_cli/cli/commands/merge.py` (grepped both files for `coordination_branch`/`flatten`/`flattened`: zero hits).
[...]

Issue #2953, opened 2026-07-26
import-history: pending-exit gate is one-shot — dedup launders unmaterialized events on re-run


stijn-dejongh:
## Context

PR #2884 review (correctness lens) — `sync import-history --apply` now fails closed on pending events (exit 1), but the failure is structurally one-shot:

- Import event ids are deterministic and frozen (`synthesize.py` `_IMPORT_ID_NAMESPACE`), and the server dedups on `event_id`.
- On re-run, previously-pending events map to `duplicate` (`delivery/receivers.py` batch outcome mapping), `UploadReport.ok` flips True, and the command exits 0 — **whether or not the projection ever materialized them**.
- `pending` can also arise from a 200 batch response that simply omits an entry for an event ("absent entry → pending"), which no operator action can resolve.

The landing pass fixed the *message* (it now says exactly this and points at the dashboard/projection as the authoritative check — commit `342f7905b`), but the structural gap remains: nothing on the client can distinguish "duplicate because previously materialized" from "duplicate because previously pending".

## Proposed direction

Record pending event ids in a receipt (provenance manifest sidecar or a small ledger) at the moment of the pending exit, so a re-run can:
1. re-check exactly those ids against the projection (or a status endpoint), and
2. fail closed again if they are still unmaterialized, instead of laundering them through dedup.

Direct import delivery deliberately does not journal (that is its design premise), so this should be a minimal receipt scoped to the import command, not a re-introduction of the sync journal.

## Acceptance sketch

- `--apply` with N pending → exit 1 and a receipt containing those N event ids.
- Re-run while server still has not materialized them → exit 1 again (not 0-via-duplicate).
- Re-run after genuine materialization → exit 0 and the receipt is cleared.

stijn-dejongh:
## Triage (planner pass, 2026-07-26)

**Type** Bug · **Labels** sync, saas, reliability, priority:P1 · **Milestone** 3.2.x · **Parent** #1800 (SaaS sync & event-envelope hardening)

**Bug**, not Task: the operator-visible outcome is a false green — exit 0 on a re-run whose events were never materialized. That is the same defect class as #2264 (sync status reporting success blind to remote/import state), which is why P1 rather than P2.

[...]

Issue #1800, opened 2026-06-09
Epic: SaaS sync & event-envelope hardening


stijn-dejongh:
## Purpose

Harden the SaaS synchronization path and the event envelope so local and remote mission state converge correctly, safely, and without data loss. Groups the protocol-correctness, envelope-completeness, and concurrency-safety work on the sync boundary that currently lives as scattered `feat(sync)`/`feat(events)`/`fix(sync)` tickets with no owning epic.

## Scope

- **Event envelope** — required identity fields (project / authenticated user / session), PII hygiene before logging or queueing, and ordering/watermark metadata.
- **Sync protocol** — batch-endpoint error handling, idempotency, and commit/state notifications back to clients.
- **Concurrency & safety** — vector-clock watermarks, pre-claim advisory checks against the server, and correct drain-block behaviour for team vs no-team contexts.
- **Local durability** — committing local event/decision logs into the mission record so sync never loses authored state.

## Out of scope

- Team Kitty launch gating (#1091) — this epic is the sync *plumbing* beneath it.
- The CLI-facing evidence surface (see the CLI user experience epic).

stijn-dejongh:
## First hosted-sync drain of an established 324-mission corpus (2026-07-27) — measured rejection census

Recording the numbers once here so the child issues can reference rather than each restate them.

**Setup.** 13,391-event local backlog, oldest event 42 days old, 324-mission dogfood corpus, TEAMSPACE mode. The backlog had never drained: `Delivered (cur/prev) 0/0`.

**Three gates blocked `sync now` in series**, each only visible after clearing the one before it:

1. Rollout flag `SPEC_KITTY_ENABLE_SAAS_SYNC` unset — checked first in `saas/readiness.py`, short-circuiting *before* the server URL is read, so a correctly-configured target still reports `Server: Disabled`.
2. Legacy queue rows — cleared by `sync migrate`, and re-accumulates within minutes because every command emits events.
3. TeamSpace mission-state blockers — 305 missions. Root cause was #2995 (repair hard-errored on `kind:annotation` rows); fixed in #2998, after which `doctor mission-state --fix` returns `errors=0` and the dry-run synthesises 14,589 envelopes instead of 0.

**Drain result.** 9,250 delivered, 0 transient, 0 terminal failures — and **4,141 permanently rejected**:

| Count | Reason | Class | Issue |
|------:|--------|-------|-------|
[...]

PR #3098, opened 2026-07-31
fix(sync): gate every consent-bypassing egress path and make the boundary structural (#3030)

### LINKED ISSUE #3004: sync doctor and sync status derive queue truth independently of target_authority — divergent scope DB paths and false-green Queue size 0 after sync migrate
## Summary

`spec-kitty sync doctor` and `spec-kitty sync status` each derive queue truth independently, bypassing `target_authority.resolve_sync_target` — the resolver that exists specifically to make divergent scope resolution impossible. This produces two operator-visible symptoms from one root cause:

1. **Divergent scope DB paths.** In a single environment and a single session, `doctor` reported `/home/stijn/.spec-kitty/queues/queue-1b1f4056beca5d51.db` while `status` reported `/home/stijn/.spec-kitty/queues/queue-62c2403b647fc027.db`.
2. **A false-green `Queue size 0` after `sync migrate`.** `migrate` relocates rows out of the source queues into the event journal. `doctor` then reports `Queue size 0 / 100,000 (0%)` — which reads as "everything delivered" and actually means "nothing in *this* sink". The honest number is `sync status`'s `Delivered (cur/prev)`, which read `0/0` at the same moment, with 13,297 events retained and undelivered.

## Mechanism

In `src/specify_cli/cli/commands/sync.py`:

- `status()` (:2981) constructs a bare `OfflineQueue()` at **:3043** and reads `queue.size()` at **:3053**
- `doctor()` (:3595) constructs its own bare `OfflineQueue()` at **:3620** and reads `stats.total_queued` at **:3624**

Neither routes through `src/specify_cli/sync/target_authority.py::resolve_sync_target`, which is consumed correctly elsewhere (`sync.py:528`, `preflight.py:424`, `owner.py:380`, `saas/readiness.py:188`). That module's docstring is explicit that `derived_queue_scope` / `queue_db_path` are always derived from the resolved target and that a cached scope "is never used as authority" — two independent bare constructions plus two different depth derivations defeat both guarantees.

## Why it matters

These are the two commands an operator reaches for to answer "did my events sync?". Getting a reassuring answer from the wrong sink is worse than getting no answer. In the incident that surfaced this, `doctor` reported a healthy empty queue while 13,297 events sat undelivered — the discrepancy was only caught by cross-checking `sync status`.

## Related, but distinct

- **#2750** (retire the legacy `queue.db` write path) fixes the *write* side; the reporting divergence survives it for as long as each diagnostic picks its own sink independently.
- Both functions live in `cli/commands/sync.py`, the 3,913-LOC god-module tracked as **#2971** — so this is a campsite opportunity as well as a fix.
- A related diagnostic inconsistency worth folding in here rather than filing separately: `doctor mission-state --fix` reports `errors=0` while `--teamspace-dry-run` on the same corpus reports `valid: false` with 93 errors. Two modes of one command disagreeing about health, currently owned by no issue.

## Suggested fix

Route both `status()` and `doctor()` through `resolve_sync_target`, derive the queue from `queue_db_path`, and make the depth figure name its sink (or report retained-vs-delivered rather than a bare queue size, so a post-`migrate` zero cannot read as success).

## Context

Surfaced during the first hosted-sync drain of an established 324-mission corpus (2026-07-27), alongside #2999, #3000, #3001.

### LINKED ISSUE #3005: Permanently-rejected events are not counted as terminal failures — sync reports Terminal failures 0 while 4,141 events are undeliverable
## Summary

Permanently-unretryable event rejections are counted as `rejected`, never as `terminal_failed`. Operator-facing diagnostics report `Terminal failures 0`, which is true by classification and false in fact — so a drain that permanently lost 4,141 of 13,391 events presented as healthy.

## What happened

During the first hosted-sync drain of an established 324-mission corpus:

```
Event sync (TEAMSPACE): delivered 6750  duplicate 0  pending 0
                        rejected 4141  transient 0  terminal-failed 0  (selected 10891)
```

and, from `spec-kitty sync status`:

```
Event Sync
  Mode                      TEAMSPACE
  Retained events           13391
  Delivered (cur/prev)      9250/0
  Terminal failures         0
```

4,141 events will never be delivered without an upstream product change (#3001, #2999, #3000). No surface says so. `Terminal failures 0` says the opposite, and the only way to discover the loss is to pass `--report` to `sync now` and read the JSON.

## The distinction that is being lost

`transient` and `rejected` are not the same kind of failure, but the summary treats only `terminal_failed` as permanent:

- **transient** — retryable, and retrying works. In this drain, 10,634 transient failures (`access_token_expired`) all succeeded on the next run.
- **rejected** — the server evaluated the event and refused it on a rule. Re-sending the identical payload produces the identical refusal, forever. That is terminal by any operational definition.

The fields already exist and are already distinct (`summary.rejected` and `summary.terminal_failed`, `src/specify_cli/cli/commands/sync.py:355`, `:389-400`, `:930`). The defect is the bucketing, not missing infrastructure.

## Why this should be fixed before #3001, not as part of it

#3001 (the 3,391 historical transition-rule rejections) needs a product decision and will therefore sit for a while. This does not: it is small, independently fixable, and it makes #3001's impact *visible to every operator immediately*, regardless of which option #3001 eventually picks.

It is also the reason nobody noticed the loss in the first place. Any project onboarding an established corpus hits this silently.

## Suggested fix

Either count rule-based rejections toward `terminal_failed`, or surface `rejected` with equal prominence in `sync status` and `sync doctor` and have a non-zero count register as a health finding rather than passing silently. Whichever way, `Terminal failures 0` must stop being reachable while thousands of events are permanently undeliverable.

## Context

Split out of #3001, where this was a two-sentence parenthetical attached to an issue that needs a product decision. Sibling rejection classes: #2999 (687), #3000 (63). Same drain.

### LINKED ISSUE #3030: P0: sync drain authorizes per-checkout but delivers per-journal — one opted-in checkout ships every project on the machine

### LINKED ISSUE #3031: P0: hosted-sync consent is opt-out and fail-open, recorded outside the project, and capture is ungated

### LINKED ISSUE #3053: Consent ownership: capture-first durability (FR-017/C-008) contradicts #3031 Defect 3 — adjudicate which owns the journal write
## Summary

Two accepted contracts in this repo assert **opposite things** about whether a
non-consenting project's event may be written to the machine-global event
journal. Neither is wrong on its own terms; they have simply never been
adjudicated against each other. Until one owns the decision, any fix for #3031
Defect 3 is either a contract violation or a no-op, depending on which document
the implementer happens to read first.

Surfaced while landing #3050 (the red-first P0 evidence PR) — see
https://github.com/Priivacy-ai/spec-kitty/pull/3050 and its review comment.

## Contract A — capture-first: the write is unconditional (FR-017 / C-008)

`src/specify_cli/event_journal/journal.py`, `capture_teamspace_bound`:

> Durably capture a Teamspace-bound fact *before* the delivery gates.
> The journal write is unconditional for Teamspace-bound families; `gate`
> only decides the recorded `drain_blocked_reason` (delivery eligibility),
> **never whether the write happens** (FR-017, contract §2). A request to skip
> the write for a Teamspace-bound family fails loudly (C-008, T018).

`TeamspaceBoundDropError` exists specifically so that a silent drop is
impossible. This is a deliberate durability guarantee, not an oversight.

## Contract B — #3031 Defect 3: capture must be gated

Issue #3031, Defect 3 ("capture is ungated"):

> The consent gate is consulted at emit and at upload, but the journal write is
> unconditional ("capture-first"). `~/.spec-kitty/event_journal/` is one
> machine-global store that every checkout writes to.

Remediation, as stated on that issue:

> A non-consenting project's events **must never reach the journal.**

## Why this is load-bearing rather than academic

The sole production caller, `SyncEmitter._capture_to_journal`
(`src/specify_cli/sync/emitter.py:1961`), passes neither `is_teamspace_bound`
nor `skip_journal` — both take defaults, so **every** captured event travels
Contract A's unconditional path. The capture gate's only observable effect today
is the `drain_blocked_reason` column.

That means:

- Implement Contract B literally and you break C-008's durability guarantee for
  Teamspace-bound facts, and the `TeamspaceBoundDropError` machinery becomes a
  guard against a thing the product now does on purpose.
- Honour Contract A and #3031 Defect 3 cannot be closed at capture at all — the
  confidentiality control has to live entirely at drain, which is where #3030
  already shows it is absent.

There is also a live incident behind Contract B: 1,322 events belonging to 5
projects that never opted in were delivered to a hosted instance
(`Priivacy-ai/spec-kitty-saas#585`). Durability and confidentiality are both
real requirements here; the question is which one owns the *write* decision.

## What a resolution looks like

Pick **one** owner of the consent decision and make the other contract defer to
it explicitly. Plausible shapes, not a ruling:

1. **Capture owns it** — gate the write on project consent, and narrow C-008's
   "never silently dropped" guarantee to *consenting* Teamspace-bound families,
   with the refusal still loud for the rest.
2. **Drain owns it** — keep capture unconditional, and make the journal
   per-project (or add a project predicate to the drain read) so a
   non-consenting project's rows are structurally undeliverable. Durability is
   preserved; confidentiality moves entirely to selection.
3. **Split by family** — Teamspace-bound facts stay unconditional per C-008;
   everything else is consent-gated at capture.

Whichever wins, **the losing contract's text gets deleted, not left standing.**
Two documents asserting opposite invariants is the actual defect being filed
here.

## Non-goals

- Redesigning the journal schema.
- Changing the drain predicate (that is #3030).
- Re-litigating the consent *keying* problem — `repo_slug` vs `project_slug` is
  #3031 Defect 2 and is tracked there.

## Evidence / references

- `src/specify_cli/event_journal/journal.py` — `capture_teamspace_bound`, `TeamspaceBoundDropError`, `classify_drain_blocked_reason`
- `src/specify_cli/sync/emitter.py:1935-1968` — `_capture_to_journal`, the sole caller
- #3031 (Defect 3), #3030, `Priivacy-ai/spec-kitty-saas#585`
- #3050 — `tests/sync/test_sync_consent_capture_gap_3031.py` pins Contract B's
  stated outcome through the real emit path, and its docstring names this
  tension explicitly so the implementer meets it before writing code.

---

# Decision (operator, 2026-07-28): **do both**

Adjudicated. This supersedes the "pick one owner" framing above — the three
shapes were a false choice. Scope the journal **per project** *and* gate the
**write** on consent. Defence in depth, not either/or.

## Rationale

- **The gate is the control; the scoping is the blast radius.** Gating stops a
  non-consenting project capturing at all. Scoping bounds the damage for the
  case that actually bit us — where the gate is right but the *consent record*
  is wrong (an accidental opt-in, a stale `repo_slug` key, a fresh clone
  inheriting a default).
- **Per-project scoping makes the #3030 leak structurally impossible.** A drain
  handles exactly one project's capture at a time, so cross-project delivery
  stops being a predicate you can forget and starts being a shape the store
  cannot express. The drain predicate becomes defence in depth rather than the
  sole control.
- **It buys a manual backstop.** With a visibly per-project journal path, an
  accidental opt-in is recoverable by inspection: the ledger is written, spotted
  on disk, and deleted — one project's file, without touching anyone else's.
  That is only possible if the artifact is (a) written somewhere and (b) not
  commingled. A write-gate alone gives you nothing to find; a machine-global
  store gives you nothing you can safely delete.

## Consequences to implement

1. **Journal path gains project scope.** Today `get_journal(team_slug=...)`
   (`sync/emitter.py:1961-1962`) keys by team; `test_journal_path_is_producer_scoped_not_server_scoped`
   (`tests/event_journal/test_capture_first.py:151`) already pins producer-scoping
   as an invariant. This extends that axis rather than inventing one.
2. **Narrow FR-017 / C-008 rather than delete it.** The unconditional-write
   guarantee applies to **consenting** Teamspace-bound families.
   `TeamspaceBoundDropError` stays exactly as-is for that case — a skip request
   for a *consenting* Teamspace-bound fact must still fail loudly.
3. **A non-consenting no-write is a decision, not a silent drop.** It must be
   observable — a diagnostic naming the project and the consent state that
   produced it, per the standing preference for loud failures with clear
   instructions. C-008 exists because silent drops are unacceptable; that
   principle survives this change, it just stops being expressed as "always
   write".
4. **Migration** for existing machine-global / team-scoped journals.
5. **#3030's drain predicate is still required** — demoted from sole control to
   second layer, not dropped.

## Effect on the existing pins

- `tests/sync/test_sync_consent_capture_gap_3031.py` (#3050) asserts the
  **gate** half through the real `EventEmitter._emit` path and remains valid
  and correctly aimed.
- `tests/delivery/test_dispatch_project_consent_3030.py` (#3050) asserts the
  **drain** half and remains valid — under per-project scoping its premise
  (two projects sharing one journal file) becomes the legacy shape the
  migration retires, so it should be re-adjudicated at that point rather than
  deleted.
- No pin in #3050 needs changing to accommodate this decision.

Non-goals above are unchanged, except that "redesigning the journal schema" no
longer excludes **path/scoping** changes — those are now in scope; the row
schema still is not.

### LINKED ISSUE #3112: sync consent: make the unreadable-project-local fail-closed invariant local instead of call-ordering (defence-in-depth)
## Summary

`read_project_local_consent` (`sync/consent.py:367`) deliberately collapses an unreadable/unusable project consent file to `None`, and `_build_checkout_sync_routing` (`sync/routing.py:190-218`) then falls through `local_sync_enabled → repo_default → default-deny` — a checkout-level grant could stand in for a committed refusal. What actually closes the leak is that **both** `resolve_checkout_sync_routing` and `..._readonly` call `_routing_for_unreadable_project_config` (`routing.py:70,84`) **before** `_build_checkout_sync_routing`, denying on any `project_local_consent_fault`.

So the invariant "an unreadable project-local consent file must fail closed" currently lives in **call-ordering**, not in the function's own return contract. Every production entry is fenced today (verified: `read_project_local_consent` has no unfenced caller), but a future caller that invokes `_build_checkout_sync_routing` / `read_project_local_consent` directly — skipping the fence — reopens the exact stale-grant leak.

## Done when (defence-in-depth, pick one)

- An architectural test asserts the fault fence precedes `_build_checkout_sync_routing` at **every** entry point; **or**
- The fault check is folded into `_build_checkout_sync_routing` itself, so the fail-closed invariant is local to the function rather than an ordering convention.

## Where

- `src/specify_cli/sync/consent.py:367`, `src/specify_cli/sync/routing.py:70,84,190-218`
- Raised as a review NOTE (reviewer-renata, NOTE 1) during the #3098 landing pass.

---

MOES-Media:
## What this fixes for you

**Before this change, Spec Kitty could send one project's data to a different project's hosted workspace — including projects you never connected to sync.**

A machine keeps **one shared local event journal** across every Spec Kitty checkout on it. When you ran `sync`, the consent check only asked *"is sync enabled for this checkout?"* — and once the answer was yes, it delivered the **entire** journal. So working on two projects on one laptop was enough to ship an unrelated project's events, mission names, and verbatim agent prompts to the server. In this product `mission_slug` values are client-engagement names and envelopes can carry prompt text, so the *metadata is the confidential content*.

On 2026-07-27 a single `sync now` delivered **1,322 events from five projects the operator never opted into**, alongside the one intended project.

**After this change:** every path that can send data off your machine is gated **per project, by the data's own identity**, and it **fails closed** — if consent can't be determined (missing, unreadable, or malformed config), nothing is sent. A project that never opted in has its events held locally and never delivered. Two new operator tools help you see and clean up what was collected:

- **`sync doctor`** now tells you the *real* state of each project's consent — and distinguishes "no record" from "your config is unreadable/malformed" (previously it told you "no consent record" when the truth was "your file won't parse", sending you to re-record consent you'd already given).
- **`sync purge`** removes locally-pooled data (journal, delivery ledger, body uploads, local-commit frames), dry-run by default, and is honest about what it can and cannot reach rather than implying total erasure.

This closes the cross-project egress leak and the related defects logged against the same mission — consent must gate *capture* not only delivery (#3031, absorbed here), terminal delivery failures mis-reported as zero (#3005), `sync doctor`/`status` reading queue truth from a retired store (#3004), and the journal-ownership adjudication for capture-first durability (#3053). See the closing block at the end for the full list.

## Known limits (stated, not implied away)

- **Direct tracker connectors (Jira/Linear) are a separate, still-open surface.** They push issue titles under a machine-scoped credential and currently have **no per-project refusal** — a project with `sync.enabled: false` can still push issue titles. This is off the automatic `sync now` path and is tracked for a follow-up (C-006 / E20).
- **At-rest pooling is unchanged.** A consenting project's payloads still share one machine-global store; this change secures *egress* and *future collection*. Historical local collection is an operator `sync purge` action.
- **The live two-project canary (SC-008) needs an operator to run it.**

---

### The bug class, stated once
*A path reaches a consent answer through something other than the data's own `project_uuid`.* Substitutes found and closed: the current working directory, `repo_root`, machine-global arming, daemon scope, and a checkout-level grant. The original diagnosis (both #3030 and `saas#585`) pinned the defect at `sync/batch.py`; the real drain was `delivery/dispatcher.py`'s `_select_undelivered`, whose universe was every row of every project with no project predicate — a fix where the issue pointed would have passed its tests and leaked in production.

### Egress paths closed (8; five were discovered during the work)
Journal dispatcher (consent pushed into the SQL read), body uploads, local-commit frames, WebSocket publish, daemon publish endpoint, `import-history`, tracker SaaS client (gated at `_request`, the chokepoint all 12 endpoints share), and the widen SaaS client (gated before the URL is built).

### The structural change — stop enumerating, make it unforgeable
- **`ConsentedBatch`** — `DeliveryReceiver.deliver()` accepts only a frozen type whose sole constructor requires a resolved consent answer per `project_uuid`. A new sender **cannot compile** without checking.
- **`tests/architectural/test_egress_consent_boundary.py`** — scans for **sinks, not names** across 72 sites in 28 files, each allowlisted against the seam that covers it; the allowlist is ratcheted, and blinding it reds all 72 sites (proven non-vacuous).
- **`tests/architectural/test_unfiltered_journal_read_boundary.py`** — a per-symbol guard on the unfiltered journal read.

### Fail-open defects fixed
Ten occurrences of one rule — *inability to determine X is never permission for X*: unreadable machine index, unreadable project-local file deferring to a stale grant, the same shape in `routing.py`, invalid config crashing `load_identity`, unusable `uuid` crashing routing, mis-spelled `sync.enabled` silently voiding a committed refusal (`no` is the string `"no"` under YAML 1.2), the invocation propagator reading undetermined consent as consent, a corrupt index write destroying other projects' records, a corrupt `owner.json` passing preflight, and the evidence-mode gate promoting on malformed `mode_of_work`.

### Suggested reading order
1. `kitty-specs/journal-project-consent-3030-01KYKWQS/egress-inventory.md` — every network sink in `src/`, traced caller chains, ranked ungated-first.
2. The boundary guard + `ConsentedBatch` (the structural claim), then `sync/consent.py`, then the individual gates.
3. `.../tracer-tooling-friction.md` — how measurement went wrong and the rules adopted in response.

### Test evidence
Full sweep across `tests/sync`, `tests/delivery`, `tests/event_journal`, `tests/specify_cli/sync`, `tests/specify_cli/identity`, `tests/invocation`, `tests/architectural` on a quiet tree: `4438 passed, 6 failed, 22 skipped`. Every acceptance pin was proven load-bearing by mutation testing. The sharpest single result: a mutant stamping **project B's uuid** into a request body left the count, URL, body bytes and queue depth all unchanged — the pre-conversion `assert len(sent) == 1` *passed*. That is the incident in one line: shipping one event belonging to the wrong project satisfies a count assertion perfectly. The converted content assertions red on it.

---

## Maintainer landing pass

Rebased tip is current `upstream/main` (no rebase needed). A profile-loaded adversarial squad reviewed the security boundary across three lenses (structural/contract, SSOT/duplication, fail-closed correctness):

- **Verdict: no MAJOR structural or security defect.** The boundary is **one seam, not eight** — all egress sinks funnel through a single default-DENY resolver; `ConsentedBatch` is a genuine compile-time-ish gate (mint-witness + `@final` + `isinstance`); **no consent-bypass survives on the `sync now` hot path.** Squad NOTES (a duplicated `project_egress_refusal` wrapper across two `egress_consent.py` files; the `decision widen` attribution site; a positional-args transport-call blind spot in the guard) are filed as one follow-up, not merge blockers.

**Folds on this branch (one commit each):**
1. `regen cli-commands.md for sync purge + migrate/doctor help` — the machine contract carried the new commands; the human reference was stale (docs-freshness gate).
2. `re-pin CLI visible-count census tolerance for sync purge` — dated rationale for the one new command.
3. `mypy dict type-args in sync/config.py` — one missing generic parameter.
4. `resolve canonical-producer lint in #3030 test fixtures` — 9 hand-rolled event dicts in new tests, annotated with the sanctioned exemption (no baseline padding).
5. `graduate resolved #3030/#3031 egress regression tests to module suites` — 7 red-first pins are green now that the P0 is fixed; the `regression` marker (a deliberate open-P0 red signal) is removed and the red-first framing updated to past tense. They stay as normal module tests.
6. `harden #3030 sync CLI tests against cross-worker scope-cache pollution` — the new CLI tests passed in isolation but read an empty journal under the full parallel shard; the fixtures now reset the process-global `TokenManager` (mirroring the existing journal-cache reset) in setup and teardown.
7. `add bandit nosec B608 to static-identifier SQL in sync/retention` — the mission made the bandit gate enforcing; the flagged queries interpolate only static module-constant table/column names (values always bound via `?`), so they carry a per-line `# nosec B608` with rationale alongside the existing ruff `# noqa: S608`.
8. `fold the unreadable-project fault check into _build_checkout_sync_routing` (#3112) — the "unreadable project-local consent file fails closed" invariant now lives inside the routing builder rather than depending on callers running the fence first; behaviour-preserving, fences retained, with a direct-call regression test.

Plus documentation updates codifying landing-pass practice (consumer-focused BLUF PR bodies, changelog style, one-commit-per-fold, delegate-remediation-to-subagents, clean-history) and this PR's consumer-focused changelog entry.

**Pre-existing reds on `main` this broad PR merely un-skips (confirmed red on the merge-base, NOT introduced here):** the `accept` command-file snapshot drift (twelve-agent-parity + command-renderer), `test_charter_widen_integration` malformed-JSON, the `#3045` charter-sync surface, three review-cycle tests, and the `#2804`/`#3086` merge regression red-firsts (deliberate open-P0 signals — left red per ADR 2026-07-17-1). These are not this PR's to green.

---

Closes #3030
Closes #3031
Closes #3004
Closes #3005
Closes #3053
Closes #3112

Addresses (partially, via FR-014) `Priivacy-ai/spec-kitty-saas#585` — server-side rejection is out of scope here and cannot be closed from this repo.

Related but **not** closed here (tracked separately): #3045 (charter-sync silent-success — a pre-existing red of a different defect class), #3086 (merge `coordination_branch` strand), #2953 (import-history pending-exit dedup laundering — distinct from the consent gate), plus two disclosed open surfaces with no issue yet — E20/C-006 (tracker connectors, no per-project refusal) and FR-032 (`_ws_client` phantom).


MOES-Media:
## Remediation summary

**Verdict: DRAFT for operator review.** Not un-drafted, not merged.

### Base and history
Rebased onto current `upstream/main` (0 behind at slice time). History compressed per the
charter's `clean-linear-commit-history` tactic — **admin bunched, code by slice**, not one squash:
`492bf3ba9e` governance dossier (31 files) + `da3d176c33` code slice (149 files). The sliced tree
is **content-identical** to the 244-commit working branch apart from one ruff line-wrap (verified
by empty `git diff`).

| Gate | Result |
|---|---|
| `commitlint --from upstream/main --to HEAD` | clean |
| `ruff check --no-cache` (144 changed `.py`) | **All checks passed** |
| `pytest tests/architectural/` | **1465 passed, 4 failed, 4 skipped** |
| Full sweep, 7 roots (pre-slice) | **4438 passed, 6 failed, 22 skipped** |

### Pre-existing failures — reproduced, not fixed in-PR, not retried to green
The 4 architectural failures are all `tests/architectural/test_tid251_enforcement.py`. Reproduced
on a pristine `upstream/main` worktree with `PYTHONPATH=$WT/src` and import resolution **verified
inside the worktree** — this repo's editable install otherwise imports the main checkout, which
would have made the comparison a tautology. Identical 4 failures there.

Also pre-existing, confirmed the same way: `test_charter_io::test_get_mission_id_returns_none_when_meta_json_malformed`,
`test_charter_package_exports::…`, two `test_safe_commit_cmd::…_3033`, and
`test_doctor_ops::test_sweep_nfr_002_10k_files_under_5s` (wall-clock, fails under load).

The other 2 sweep failures **were ours** and are fixed in-PR: a `/tmp` literal in a new test, and
golden-count ratchet regrowth — resolved by judgement, **seven sites converted, two annotated,
none re-frozen**.

### Red-first proof
Every gate in this PR was built red-first and proven load-bearing by mutation, with plugins loaded
via `PYTHONPATH` and never source edits. The sharpest single result: a mutant stamping **project
B's uuid** into a request body left the count, the URL, the body bytes and the queue depth *all
unchanged* — the pre-conversion test **passed**. Only the converted content assertion reds. That is
this incident in one line: `assert len(sent) == 1` is satisfied by shipping one event belonging to
the wrong project.

### Review cycles
10 of 11 WPs approved. WP07 took **five cycles**; its cycle-4 rejection found that `sync doctor`
printed a project label its own recommended remedy refused (`--project acme/app` → *"No project
matches"*), with a docstring asserting the false justification **and a green test defending it**.
Cycle 5 verified the fix across 7 collision shapes — no shape resolves to a wrong project, and the
only reachable change is *resolve → refuse*, the safe direction for a destructive command.

### Known limits — in the PR body, not implied away
Operator-configured tracker connectors remain an **open collection surface** (machine-scoped
credential, project-scoped binding, **no expressible refusal** — a project with a committed
`sync.enabled: false` can still push issue titles to Jira). Both guards state what they cannot see.
At-rest pooling is unchanged. **The live two-project canary (SC-008) has not been run** — it needs
the operator, and it is the one remaining work package.

### One hazard worth fixing before CI trips on it
Adapter registration pollutes across test files: a fixture calling `reset_adapters()` in teardown
left the process with no consent resolver, failing later suites' **positive controls** with the
production text *"no hosted-sync consent resolver is registered"*. Fixed here. Flagging the shape
because it is deterministic in alphabetical order, and on a consent mission that text reads as a
gate defect rather than fixture teardown ordering.

stijn-dejongh:
## Maintainer takeover — landing pass claim

Picking this PR up for a maintainer landing pass (P0 sync/egress consent hardening, closes #3030). Plan, per the [maintainer PR-landing runbook](https://github.com/Priivacy-ai/spec-kitty/blob/main/docs/development/pr-landing.md):

1. Isolated worktree off the PR head, rebase onto current `upstream/main`.
2. **Classify every red check** (CI is broadly red right now — AST lint, docs-freshness, lint, quality-gate, regression, and several fast-test shards) into: PR defect / contract legitimately crossed / pre-existing main breakage / perf flake.
3. Launch a **profile-loaded adversarial review squad** over the structural claim (the `ConsentedBatch` seam + the two architectural egress guards) and the individual gates, since this is an architectural / security-boundary PR.
4. **Fold** remediations as separate, single-purpose `landing fold: …` commits on this branch (`maintainerCanModify` is on).
5. Red-first verification on the consent gates where applicable, full local evidence, then hand off green — **the operator merges; I never run `gh pr merge`.**

Evidence trail to follow in this thread. Thanks @MOES-Media — the write-up and the mutation-tested acceptance discipline are exemplary; this claim is about getting it green and independently reviewed, not re-litigating the design.

stijn-dejongh:

Rebased tip was already current `upstream/main` (2 commits ahead, 0 behind — no rebase needed). Nine commits folded on top; pushed as a fast-forward. **The PR body has been rewritten consumer-focused (impact first).** Handing off for the operator to merge on green — I do not merge.

### Adversarial review verdict (profile-loaded squad, 3 lenses)
- **architecture / structural claim** (architect-alphonso): the claim **holds within its stated bounds**. `ConsentedBatch` is a genuine unforgeable-by-accident gate (mint-witness + `@final` + `isinstance`); the egress guard keys on the **sink call**, not names, with non-vacuity proven by negative controls. Not enumeration-in-disguise.
- **SSOT / duplication** (paula-patterns): **one seam, not eight** — every egress sink funnels through a single default-DENY resolver; the fail-closed rule is centralized, not copy-pasted.
- **fail-closed correctness / security** (reviewer-renata): **no consent-bypass survives on the `sync now` hot path**; all ten fail-open fixes verified to deny on the indeterminate path; the admitted "known limits" are honestly out-of-scope of the incident trigger.

**No MAJOR structural or security defect.** Squad NOTES → one follow-up issue (below): duplicated `project_egress_refusal` wrapper across the two `egress_consent.py` files; the `decision widen` attribution site; a positional-args transport-call blind spot in the boundary guard; the `read_project_local_consent` fence-ordering latent footgun (NOTE 1). None block merge.

### Folds (one commit each, `landing fold:` prefixed)
| # | Commit | Why |
|---|---|---|
| 1 | `baba73729` regen `cli-commands.md` for `sync purge` + migrate/doctor help | machine contract carried the new commands; human reference was stale → **docs-freshness** red |
| 2 | `9742f4acf` re-pin CLI visible-count census tolerance (259→260) | `sync purge` saturated the band; dated 2026-07-31 rationale |
| 3 | `6ee48c32b` mypy dict type-args in `sync/config.py:632` | one missing generic parameter → **lint/quality-gate mypy** red |
| 4 | `9cd7260ad` resolve canonical-producer lint in `#3030` test fixtures | 9 hand-rolled event dicts in new tests → sanctioned exemption, **no baseline padding** → **AST-lint** red |
| 5 | `c63ab82b4` graduate resolved `#3030`/`#3031` regression tests | 7 red-first pins are green now the P0 is fixed; `regression` marker removed, red-first framing → past tense; they remain as normal module tests (`-m regression` no longer collects them) |
| 6 | `4f8e4ca78` harden `#3030` sync CLI tests vs cross-worker scope-cache pollution | new CLI tests passed in isolation but read an empty journal under the full parallel shard; fixtures now reset the process-global `TokenManager` in setup+teardown (mirrors the existing journal-cache reset) |

Plus three `docs:` commits: landing-pass guideline updates (consumer-focused BLUF PR bodies, changelog style, one-commit-per-fold, delegate-remediation-to-subagents, clean-history), a retrieval-index regen, and this PR's consumer-focused **changelog** entry (`#3030`).

Remediations were implemented by profile-loaded subagents (charter-loaded profiles, `implement`→sonnet), maintainer-adjudicated.

- **docs-freshness** `--ci`: exit 0, 0 errors (remaining warnings are localhost link probes).
- **census + terminology**: `tests/docs/test_check_cli_reference_freshness.py::...within_tolerance` + `tests/architectural/test_no_legacy_terminology.py` → **11 passed**.
- **mypy** `--strict`: the `sync/config.py` type-arg error is gone; no new errors introduced. (The 28 errors in `migration/backfill_runtime_state.py` are **pre-existing** — byte-identical to the merge-base — and are left untouched.)
- **canonical-producer lint**: 0 new violations; baseline untouched.
- Graduated `#3030`/`#3031` tests: **52 passed**; ruff clean.
- `#3030` CLI suites: 65 passed in isolation and inside the broad shard-matching slice.

### Pre-existing reds this broad PR merely UN-SKIPS (confirmed red on the merge-base `21aec3fe6` — NOT introduced here; not this PR's to green)
- `accept` command-file snapshot drift — `tests/specify_cli/regression/test_twelve_agent_parity.py` (accept-*) + `tests/specify_cli/skills/test_command_renderer.py` (\*-accept). Out of mission scope (touches 12 generated agent copies).
- `test_charter_widen_integration.py::TestGetMissionId::test_returns_none_if_json_malformed` — raises via untouched `mission_runtime`.
- `#3045` `test_sync_command_..._do_not_contradict_3045` — open charter-sync red-first.
- three review-cycle tests (`test_review_cycle_rejection_only`, `test_cycle` ×2).
- `#2804` and `#3086` merge regression red-firsts — deliberate open-P0 signals, left red per ADR 2026-07-17-1.

These fail identically on a fresh `upstream/main` worktree; they surface here only because this PR touches broad paths that un-skip the affected CI shards (the shard path-filter effect).

### State
Folds pushed; **watching CI — merge-ready on green** (green defined as: the CI Quality checks this PR is responsible for; the pre-existing reds above are out of scope and tracked separately). Follow-up issue for the squad NOTES to be filed under the egress/sync epic.

Thanks again @MOES-Media — the mutation-tested acceptance discipline and the egress-inventory made this an unusually reviewable P0.

stijn-dejongh:
## Test-isolation / potential-flake findings (declared for transparency)

After the folds above, I re-classified the remaining red CI against a pristine `upstream/main` worktree. Two classes of test failure are **not** product defects in this PR and are called out here so nobody re-chases them:

### 1. Shard-parallel isolation flakiness — this mission's own new tests (tracked: #3115)

A set of this mission's new tests **pass in isolation** (and in small `xdist` groups) but **fail under CI's full parallel shard** (`fast-tests-cli`, `fast-tests-sync`):

- the `sync status/doctor/purge/migrate` per-project CLI tests read an **empty journal** under the shard (`<uuid> is in the journal but 'doctor' did not name it`, panel shows `Queue 0 event(s)`) — the command resolves a different producer-scoped journal path than the seed, i.e. a process-global left dirty by a sibling test on the same worker;
- `tests/sync/tracker/test_saas_client.py::…test_429_respects_retry_after` shows `Expected 'sleep' called once. Called 173 times.` — a leaked module-global retry/backoff config.

It was **not reproducible locally** even with the exact CI selection (`tests/cli/ tests/specify_cli/cli/ -m "fast and not windows_ci"`, various `-n`), which is distribution-dependent poisoning. The landing-pass attempt (`4f8e4ca78`, resetting `TokenManager` in the fixtures alongside the existing `reset_journal_cache()`) did **not** clear it — the culprit is a third global still to be pinned. Rather than push unverifiable guesses, this is tracked for a focused investigate-and-remediate pass in **#3115** (parented under #1800, milestone 3.2.x). The tests are correct in what they assert; the gap is fixture isolation under parallelism.

### 2. Local-machine-only artifact — NOT a CI failure

`tests/sync/test_sync_consent_default_deny.py::test_unresolvable_routing_does_not_consent_to_sync` fails on **some developer machines** because the checkout resolver walks up and treats **`/tmp` itself as a repo root** (`resolve_checkout_sync_routing_readonly()` returns a `/tmp`-rooted object instead of `None`). It **passes on CI** and reproduces on **pristine `upstream/main`** — a machine-specific filesystem-root-walk quirk, not a code defect. Folded into #3115's scope as a hardening note.

### Everything else red is pre-existing

The remaining red checks (the `accept` command-file snapshot drift ×14, `test_charter_widen_integration` malformed-JSON, `#3045`, three review-cycle tests, `#2804`/`#3086` regression red-firsts, `test_mission_status_aggregate` ×2, `test_registry_builtin_activation_parity`) all reproduce **identically on a fresh `upstream/main` worktree** — this PR only un-skips the shards that run them (the shard path-filter effect). They are honest main reds per ADR 2026-07-17-1, not this PR's to green.

**Net:** the only genuinely-this-PR failures are the shard-isolation ones in class 1, now tracked in #3115. The consent/egress product change and its gates are green in isolation and mutation-proven.

Issue #3115, opened 2026-07-31
test(#3030): investigate & remediate shard-parallel test-isolation flakiness in sync CLI/consent tests


stijn-dejongh:
## Summary

Several of the `#3030` mission's **new** tests **pass in isolation** (and in small `xdist` groups) but **fail under CI's full parallel shard** (`fast-tests-cli`, `fast-tests-sync`, run with `-m "fast and not windows_ci"` + `-n auto --dist loadfile`). This is cross-test global-state pollution, distribution-dependent, and was **not reproducible locally** even with the exact CI file selection — which is why it survived the mission's own sweep (that sweep ran `-p no:randomly` and did not hit the poisoning worker assignment).

This ticket is to **investigate and remediate if needed** — i.e. confirm the leaked global, reset it at the right layer, and prove the fix under a shard that matches CI's worker count.

## Affected tests (all this mission's own, all green in isolation)

- `tests/cli/commands/test_sync_status_per_project_3030.py::test_status_names_every_project_with_count_age_and_consent`
- `tests/cli/commands/test_sync_doctor_per_project_3030.py::test_doctor_names_every_project_with_count_age_and_consent`
- `tests/cli/commands/test_sync_doctor_consent_health_3030.py` (4 param cases)
- `tests/cli/commands/test_sync_migrate_backfills_h4.py::test_the_consent_backfill_reports_records_it_could_not_resolve`
- `tests/cli/commands/test_sync_purge_3030.py::TestPurgeAll` (2)
- `tests/sync/test_consent_write_refusal_3030.py` (3 param cases)
- `tests/sync/tracker/test_saas_client.py::TestRetryBehaviors::test_429_respects_retry_after`

## Symptoms → likely cause

- **CLI/journal tests read an EMPTY journal** under the shard: `AssertionError: <uuid> is in the journal but 'doctor'/'status' did not name it`, with the rendered panel showing `Queue 0 event(s)`. The command resolves a **different journal path** than the seed wrote — i.e. a process-global that determines the producer scope / journal path (`resolve_journal_path` reads `SPEC_KITTY_HOME` live and keys the DB on `_producer_token(user_id, team_slug)`) is left dirty by a sibling test on the same worker.
- **The retry test** shows `AssertionError: Expected 'sleep' to be called once. Called 173 times.` with an exponential-backoff sequence — pointing at a **leaked module-global retry/backoff config** a sibling test mutated and did not restore.

[...]

Issue #2782, opened 2026-07-18
test_mission_create_json_strict_when_sync_skips_ingress consistently red — auth-failure/connection-refused instead of the private-team skip path


stijn-dejongh:
**Consistent CI + local failure** (red on `main` @ 7bc635aa7, not introduced by any recent PR). Surfaced while auditing green-vs-red `@regression` tests during the doctrine-activation-freshness landing (PR #2781).

`tests/sync/test_strict_json_stdout.py::test_mission_create_json_strict_when_sync_skips_ingress` expects stderr to carry `direct ingress skipped` / `direct_ingress_missing_private_team` (the private-team-missing skip path), but instead gets:

```
sync_diagnostic severity=warning diagnostic_code=sync.server_auth_failure fatal=false sync_phase=final_sync message=Final sync failed after local command success. ... Detail: HTTP transport error: [Errno 111] Connection refused
```

i.e. the sync layer **attempts a connection** (→ refused) rather than short-circuiting on the missing private team. Either the product doesn't skip ingress in this state (product P0) or the test's env/auth setup doesn't force the intended skip path (test-isolation bug). Needs a sync-owner triage.

Interim: marked `@pytest.mark.regression` and routed off the blocking `integration-tests-sync` job (that job's selector now excludes `regression`, consistent with the fast-tests jobs and ADR 2026-07-17-1); it remains visible in the non-blocking `regression visibility` gate.

stijn-dejongh:
Correction to the interim handling: the test is **not** deselected from `integration-tests-sync` — that would be fake-green. It stays `@regression` as a *signal* (known-tracked P0, don't fix unless in scope) and **fails honestly** in its blocking job; the non-blocking `regression visibility` gate catalogs it. Red CI here is the intended honest signal until this is triaged.

robertDouglass:
Spec Kitty mission assertive-test-suite-sanitation-01KZME3P is now the active sanitation workflow for this issue. Its evidence ledger will record the terminal disposition before PR merge.

robertDouglass:
Pre-existing failure reconfirmed while validating PR #3297 against `main` at `5d49d31ed`.

Both the latest `main` CI run and PR #3297's blocking regression job fail the same test:

```text
tests/regression/test_issue_2782_sync_strict_json_ingress_skip.py::test_mission_create_json_strict_when_sync_skips_ingress
[...]

Issue #2749, opened 2026-07-17
NFR-003 clean-install latency gate regressed by step-authority (#2723): spec-kitty next median 1.25s > 1.05s ceiling


stijn-dejongh:
## Summary

The `Clean install verification (FR-010)` job's **NFR-003 latency regression gate** fails: `spec-kitty next` cold-start median has regressed to **1.247s**, above the **1.05s** CI ceiling. Every one of the 5 measured runs exceeds the ceiling, so this is a consistent, reproducible slowdown — not runner noise.

## Command run (CI)

```
python scripts/check_nfr_003_latency.py \
  --python /tmp/clean-venv/bin/python \
  --fixture /tmp/fixture \
  --baseline kitty-specs/shared-package-boundary-cutover-01KQ22DS/nfr-003-baseline.json \
  --runs 5
```

Measured command: `spec-kitty next --agent test --mission clean-install-fixture-01KQ22XX --json`

## Failure report

```
NFR-003 latency gate report
  pre_cutover_median: 0.745s (745ms)
  post_cutover_median: 1.247s (1247ms)  (n=5, runs=['1.326','1.369','1.131','1.161','1.247'])
  delta: +67.4% (ceiling: <=40.9%)
  ceiling_seconds: 1.050s
  verdict: FAIL
```

Locally reproduced (uv/editable): median **1.463s** (runs `1.487, 1.504, 1.462, 1.456, 1.463`).

## Why this is pre-existing / upstream, not introduced by the observing PR

- The gate **passed** on the pre-rebase commit `164b4a663` (CI run 29556399670, "Clean install verification (FR-010) pass") and **fails** after rebasing onto `upstream/main` at base `3a9048c18`.
- The only CLI-code delta across that rebase is the three step-authority commits: `e7c196782` (S-A / #2722), `15fb436a2` (**S-B / #2723 — "step authority — step.yaml single source; action_sequence/template_set projected"**), `3a9048c18`.
- The observing PR (#2733, primary/merge terminology) is import/runtime-neutral: it only touches docstrings, doctrine prose/data files, tests, and internal renames — nothing on the `spec-kitty next` execution path.

## Where the time goes

- Pure `import specify_cli` is ~50ms (not the regression).
[...]

PR #3126, opened 2026-07-31
Bundle B — the egress consent boundary (#3110, #3111, #3109, #3177)

### LINKED ISSUE #3109: sync egress: resolve the phantom token_manager._ws_client — wire it or delete the three dead senders (FR-032)
## Summary

`token_manager._ws_client` is a **phantom attribute** that silently disables three WebSocket senders. It is read via `getattr(token_manager, "_ws_client", None)` at `sync/__init__.py:373`, `sync/local_commit.py:683`, and transitively by `invocation/propagator.py` (via `adapters.get_saas_client` → the factory at `sync/__init__.py:411`) — but **nothing in `src/` ever assigns it** (no `=`, no `setattr`; only tests inject it). The genuinely live client is a *different* attribute on a *different* owner: `SyncRuntime.ws_client` (`sync/runtime.py:242`, handed to the emitter at `:252`/`:379`).

Consequences today: the propagator send (FR-025) and `emit_local_commit`'s immediate send (E16) are **dead in production**, and `_saas_client_factory` always returns `None`.

**Why this is not merely dead code:** the moment anyone assigns `_ws_client` — a plausible one-line "fix" for "the propagator never sends" — all three paths go live simultaneously, and the FR-025 / WP12 consent gates become the only thing between them and the #3030 incident. It is also invisible to `tests/architectural/test_no_dead_symbols.py` (a string-literal `getattr` is a string, not an AST reference).

## Resolve deliberately (pick one)

- **Wire it** to the real client and accept that the FR-025/WP12 gates are load-bearing (add a test asserting the gate fires on this path); **or**
- **Delete** the three phantom readers and the `_saas_client_factory`, removing the latent multiplier.

Either way, add a guard so a never-assigned `getattr`-by-string attribute read cannot re-appear unseen.

## Where

- `sync/__init__.py:373,411`, `sync/local_commit.py:683`, `invocation/propagator.py`, `auth/manager.py`
- Origin: FR-032 in `kitty-specs/journal-project-consent-3030-01KYKWQS/spec.md` (Security / Medium / **Open**).

---

### LINKED ISSUE #3110: sync egress: consolidate duplicated project_egress_refusal wrapper + UNDETERMINED_PROJECT_REFUSAL across saas_client/tracker
## Summary

The two `project_egress_refusal` helpers in `saas_client/egress_consent.py:92-145` and `tracker/egress_consent.py:147-211` are near-identical copies: same `None`→refusal guard, same `import specify_cli.sync` registration dance, same four-branch verdict→operator-message mapping (`DENIED`/`NO_RESOLVER`/`UNANSWERABLE`/fallback). The `UNDETERMINED_PROJECT_REFUSAL` constant is duplicated verbatim (`saas_client/egress_consent.py:85-89`, `tracker/egress_consent.py:140-144`).

The consent **chain** is single (both delegate to `invocation.adapters.resolve_egress_consent`) — what is copy-pasted is the **wrapper**. This fails **closed** (the fallback refuses), so the risk is operator-message drift, not a safety hole. Drift has already started: the tracker copy carries a "future member" comment the saas copy lacks.

## Suggested consolidation

Hoist a single `egress_refusal_for(root, *, subject_noun) -> str | None` next to `resolve_egress_consent` in `invocation/adapters.py` (both packages already import it and are on the INTEGRATION side, free to depend on it). The only real divergence — the DENIED noun ("decision identifiers" vs "engagement identifiers") — becomes one parameter.

## Where

- `src/specify_cli/saas_client/egress_consent.py`, `src/specify_cli/tracker/egress_consent.py`, `src/specify_cli/invocation/adapters.py`
- Raised as a review NOTE (paula-patterns) during the #3098 landing pass.

---

### LINKED ISSUE #3111: sync egress: harden 'decision widen' consent to key on the decision's owning project, not the current checkout
## Summary

The `spec-kitty decision widen` egress path is the **weakest of the four** `EgressConsent` construction sites. It resolves the project root as `locate_project_root() or Path.cwd()` while the `decision_id` is an operator-supplied argument — so consent is answered for **the checkout the operator is standing in**, not provably the project that **owns the decision** being widened. That is exactly the bug class the #3030 mission chases: consent answered by *where the code stands* rather than *whose data moves*.

It fails **safe today** only because that endpoint's body carries only `invited_user_ids` and the id is a ULID, not a slug. The exposure reopens the moment a slug-accepting form is added to this path.

## Done when

- The `decision widen` consent decision is keyed to the project that **owns** the decision (resolve-to-owner), not the current checkout; **or**
- A guard makes it impossible to add a slug-carrying field to this path without a per-owner consent check.

## Where

- `src/specify_cli/cli/commands/decision.py`, consent construction documented at `saas_client/egress_consent.py:59-69`.
- Raised as a review NOTE (paula-patterns) during the #3098 landing pass.

---

### LINKED ISSUE #3177: decisions: unstattable mission directory loses its diagnosis on Python 3.14+ (is_dir() EACCES divergence)
## What is wrong

`resolve_decision_ownership` silently loses the diagnosis for an unstattable mission
directory on Python 3.13+, and reports the *wrong operator action* instead.

`src/specify_cli/decisions/ownership.py:387-388`:

```python
if not resolved.is_dir() or not resolved.is_relative_to(specs_root):
    continue
```

The recording of a dropped candidate — `unreadable.append(candidate.name)` — hangs off
the `except OSError` on the following lines. That handler only runs if `is_dir()`
**raises**.

## Measured, with a control first

`euid=1000`, mode bits enforced, control asserted before anything was attributed:

```
             CONTROL readable.is_dir()   symlink -> unsearchable dir
3.11.15      True                        RAISES PermissionError(errno=13)
3.12.13      True                        RAISES PermissionError(errno=13)
3.14.4       True                        returns False
```

So on 3.13+ control takes the bare `continue`, the drop is never recorded, and
`unreadable_ledgers` stays empty. The operator-facing consequence, measured directly:

```
3.11 / 3.12   "... could not answer: m-link. To fix: this is not a missing checkout
               — `git pull` will not fix it ..."
3.14          "(no missions were found under .../kitty-specs). To fix: run `git pull` ..."
```

The 3.14 message is the exact misdiagnosis that was fixed for 3.11/3.12: the operator is
told to `git pull` for a permission problem.

Test evidence, isolated single-file runs, same `pytest 9.0.3` in venvs on each side so
the HOME/user-site trap is excluded:

| interpreter | `tests/specify_cli/decisions/test_ownership_3111.py` |
|---|---|
| 3.11.15 | `33 passed` |
| 3.12.13 | `33 passed` |
| 3.14.4 | `2 failed, 31 passed` — `assert () == ('m-link',)` |

## Severity: fail-closed, and invisible to CI

`owned=False` and `missions_searched=()` on all three interpreters, so **nothing is
transmitted** — this is a *diagnosis* defect, not a consent or egress defect.

But `requires-python = ">=3.11"` makes 3.13 and 3.14 supported, and **no CI job runs the
test suite on 3.13+**. CI does provision 3.13 in `build-wheel`
(`.github/workflows/ci-quality.yml:3848`) and `clean-install-verification` (`:3898`), and
neither runs pytest — while `clean-install-verification` *does* import `specify_cli` and
run `spec-kitty next` on 3.13. So 3.13 is a supported surface that no suite covers.

## Why it is filed rather than fixed

Found by the mission's own verification WP after the owning WP was already approved and
merged. Fixing it means changing the contents of `unreadable_ledgers` — a published field
in the `--dry-run` payload that already has a documented mixed-kind problem — which needs
its own red-first tests on an interpreter CI does not run.

## Suggested fix

Record the drop on **both** exits rather than only in the handler, or route the candidate
through the same explicit-probe idiom `_read_ledger` already uses — which is this module's
own documented answer to this exact `EACCES` divergence. The module removes
`Path.exists()` three times *because* it is divergent, states the `is_dir()` divergence
explicitly, and then reads 3.14's non-raise as "a clean refusal on 3.14". It is not a
clean refusal.

Add a 3.13-or-3.14 row to the test so the regression cannot return unseen.

## Falsifier

If `is_dir()` raised on 3.13+. It does not — probed with a readable control in the same
run, on all three interpreters.

MOES-Media:
One consent boundary, three sharp fixes, plus a self-filed bug reversed and closed for good.

Closes #3110
Closes #3111
Closes #3109
Closes #3177

## What changes for you

- **`decision widen` no longer trusts a decision it wasn't asked about.** Previously, widening a decision could resolve ownership from whatever the ledger happened to hand it rather than the mission you actually named — so a decision owned by another mission could be widened and have the egress gate treat it as consented, leaking that project's identifying text. It now refuses whenever the acting mission and the decision's owner disagree.
- **The consent-refusal gate that decides whether project-identifying text may leave your machine has one home now**, not two near-identical copies that could silently drift apart (fixing one and missing the other). `saas_client/egress_consent.py` and `tracker/egress_consent.py` are gone; `src/specify_cli/egress.py` is the single definition site, and it doesn't pull in the transport or sync subsystems just by being imported.
- **Listing missions on Python 3.14 no longer lies about permission errors.** An unreadable mission directory used to be silently dropped, and the CLI told you "no missions found — try `git pull`" for what was actually a permission problem. It now reports the directory as unreadable. (This only affected 3.14; the CLI's pinned/shipped interpreters are 3.11–3.13, so exposure was limited — see the floor-correction note below.)
- **A stale docstring is now honest.** The saas-client factory registrar's docstring claimed re-registration only replaces the factory "when the qualified name matches"; the code actually always replaces it. Behavior is unchanged — only the documentation was wrong.

## The governing rule this PR establishes

**Identity is a detector, never an actuator.** A component may use identity to report who owns a decision; nothing may use identity to grant the consent that lets data leave. #3110 and #3111 are two faces of the same boundary — recorded in `docs/adr/3.x/2026-08-04-1-egress-consent-boundary.md`.

## The self-filed bug, reversed

#3177 was filed by this mission against its own #3111 fix, then deferred as fail-closed (MEDIUM, since 3.14 wasn't shipped/pinned anywhere in CI). The deferral is reversed here rather than carried forward: no CI job runs pytest above 3.12, so a deferred fix would have sat somewhere no gate could ever rediscover it. Two corrections to the original report, both measured on 3.11.15 / 3.12.13 / 3.13.12 / 3.14.4: the divergence floor is **3.14, not 3.13** (`Path.is_dir()` raises through 3.13 and only swallows `EACCES` starting 3.14), and the mechanism is 3.14's rewrite of the predicate to `os.path.isdir()`, not the removal of a private `_ignore_error` helper (that hypothesis conflated a 3.13 packaging change with the actual 3.14 behavior change). The standing AST guard that let this ship despite review is also tightened, with a paired known-bad control so it's tested against the shape it exists to catch, not only against source it already passes.

- Rebased onto current `upstream/main`; one landing fold added (`docs(landing): rewrap two lines so MD018 stops misreading them as headings` — a markdownlint false-positive on two hard-wrapped lines that happened to start with `#NNNN`; this is a soft-fail lint step, not a blocking regression).
- Red-first verification re-run independently for #3177: reverting the fix's predicate (`S_ISDIR(resolved.stat().st_mode)` → `resolved.is_dir()`) on a real Python 3.14.6 venv reproduces exactly the claimed 4 test failures; restoring the fix returns to 35/35 passing.
- Remaining CI reds are pre-existing and unrelated to this PR (matched by test id + assertion text against the current `upstream/main` tip, not the stale merge-base): `regression tests (blocking)` (6 honest-red open P0s — #2782, #2804, #2996 ×2, #3086 — none touched here), `fast-tests-sync` (already-filed flake #3187), `fast-tests-cli` (pre-existing render-width brittleness tied to the open #3115 test-isolation investigation), `Clean install verification (FR-010)` (documented perf-budget flake #2749, NFR-003 latency gate noise). None of these are folded here; see the landing-summary PR comment for the full evidence trail.
- No regression-marked test in `tests/regression/` is touched, closed, or greened by this PR's product changes — checked directly against the fresh CI run, not assumed.


MOES-Media:
## Remediation summary

**Verdict: draft, for operator review.** All three artifacts accepted; no escalation pending.
Base: `bb2020fea924d6e5b157974f27a7cab1a77ad259` (`ahead-of-base: 0` — branch is on current
`upstream/main`; no rebase was required). History compressed 15 → 1 commit: admin bunched, and
there is no code slice because there is no code.

### Squad verdicts, by round

| Point-cut | Lenses | Verdict | Above MEDIUM |
|---|---|---|---|
| post-specify R1 | paula-patterns, reviewer-renata, debugger-debbie, architect-alphonso | 4 × ACCEPT-WITH-CHANGES | 2 CRITICAL, 3 BLOCKER, 17 HIGH |
| post-specify R2 | renata, debbie | ACCEPT-WITH-CHANGES ×2 | 0 BLOCKER; several HIGH |
| post-specify R3 | renata, debbie | renata **0**; debbie **2 HIGH** | **gate tripped → escalated** |
| post-plan R1 | alphonso, debbie | CHANGES REQUESTED ×2 | 4 HIGH (2 each, all distinct) |
| post-plan R2 | alphonso, debbie | APPROVE / ACCEPT-WITH-CHANGES | **0 / 0** |
| post-tasks R1 | renata, planner-priti | CHANGES REQUESTED ×2 | 1 BLOCKER, 6 HIGH |
| post-tasks R2 | renata, priti | APPROVE-WITH-MEDIUMS / PASS | **0 / 0** |

The spec's 3-round gate tripped with 2 HIGH standing. Per the standing rule I halted rather than
attempting a fourth round or lowering a severity; the operator authorised the closing fixes without
re-review. **That risk is carried explicitly** — those edits are the least-scrutinised text in the
document, and the PR body says so.

### Folds that changed the design

- **The primary acceptance criterion was vacuous-passable.** It said "from a checkout that does not
  own the named decision produces zero outbound requests" and never said *consenting* — a
  non-consenting checkout already sends zero. An implementer could have written that test, watched
  it pass against unfixed code, and shipped nothing.
- **A criterion instructed the defect it was added to prevent** — requiring the guard to *match* a
  callee shape its predicate excludes by design, so the only way to green it was the silent-widening
  direction two other requirements forbid.
- **The compatibility clause had a repair that reinstates the leak.** The design substitutes "this
  checkout's ledger lists the decision" for "the server knows this decision" — not the same set. The
  obvious fix for the resulting red is *fall through to the acting root*. The spec now forbids that
  repair by name.
- **A CI observation was structurally unobtainable**: any PR carrying the routing fix edits the
  workflow file, which is itself in the filter group under test. The proposed workaround produced no
  CI run at all (trigger filters on the PR's *base* ref). That obligation is now necessarily
  post-merge.
- **The plan's own fix would have authored instance seven of the bug class** — the ownership
  resolver takes slug, `cwd` **and** `env`, and asserts nothing about containment.
- **Two work packages could not land green at all** — both add `docs/` pages against a 1:1 lockfile
  (685/685 clean, every drift row blocking) that no lane owned.
- **219 lines of load-bearing rationale were being deleted with no destination**, including the only
  written record of a high-impact requirement that deliberately has no criterion.

### Findings raised and judged wrong — recorded so they are not re-raised

- *"The CI glob edit avoids running the whole suite forever."* Measured backwards: adding the module
  to one group **stops** the unmatched-fallthrough and strands the tracker's tests, which run nowhere
  else. The single-group form is a silent coverage loss — worse than the default.
- *"Assert the SaaS guard does not match `mod.SaasClient(...)`."* That predicate is already the
  stricter of the two; widening it can only make the guard see **more** sites. The pin would red on
  a coverage improvement.
- *"`#3113` bounds this mission's guard-coverage claims."* It does not — its blind spot is in the
  egress-boundary guard, not the attribution guards.

Five orchestrator rulings were overturned by measurement during the loop; each is recorded with the
measurement that overturned it, so a successor does not restore them.

### Local gates

`commitlint` **EXIT=0** (1 commit). `ruff`/`mypy`: **0 Python files changed** — no in-scope input,
reported as an input count rather than a pass. Docs shards: **0 `docs/` files changed** — not
triggered. `pytest tests/architectural/ -q`: `19 failed, 1438 passed, 4 skipped, 26 errors`,
**identical failing node sets on branch and pristine base** (45 = 45, symmetric difference empty,
control detected a planted difference). Dominant cause `ModuleNotFoundError: No module named 'typer'`
— the known environmental class. **Zero PR defects. Nothing fixed in-PR, nothing retried to green.**

### Red-first proof

**N/A — this PR contains no product change.** The `#3111` leak has never been reproduced end-to-end;
the implementer's red-first proof will be the first observation of it. Stated as an open item.

### Pre-existing failures

All 45 architectural reds reproduce on pristine `upstream/main`; none is filed as new because the
dominant cause is the environmental class already documented for this repo. Four are on the
operator's known pre-existing list. **No upstream issue filed** — filing one for a known
environmental interaction would be noise; if a maintainer disagrees, say so and I will file it.

### State

Draft, for operator review. Implementation gated on Bundle A (`#3115`, `#3113`), both measured OPEN.
Follow-up issues from §"Deliberately not fixed" in the body are not yet filed — awaiting the
operator's preference on whether they are parented under a Bundle B epic or filed standalone.

MOES-Media:
## Landing summary — verdict: **ready, with every red accounted for**

Rebased onto `upstream/main` **abca7ec96** (current as of this comment). History compressed **66 → 12** commits: governance dossier bunched into one, code sliced one commit per work package, folds last. All 47 owned paths verified byte-identical to the approved lane content; the three generated docs indexes were **regenerated** against the new base rather than carried over, because both 1:1 lockfiles run at `severity=error`.

### Red-check classification — 8 jobs failed, 7 pre-existing, 1 flake

`upstream/main` at **abca7ec96** — this PR's exact base — fails **12** jobs in the same workflow. This PR fails **8**. Set difference:

| Job | Verdict | Evidence |
|---|---|---|
| `regression tests (blocking)` | pre-existing | main's 6 failing test ids are **exactly** these 6 |
| `arch-adversarial (arch_shard_2)` | pre-existing | 1 failing id, identical to main |
| `arch-adversarial (arch_shard_3)` | pre-existing | 2 failing ids, identical to main |
| `fast-tests-cli` | pre-existing | 1 failing id, identical to main |
| `lint` | pre-existing | `pip-audit`: `cryptography 49.0.0` / **CVE-2026-69247** (fix 50.0.0). **Reproduced on a pristine `upstream/main` worktree.** This PR touches neither `pyproject.toml` nor `uv.lock`. `bandit` passed. |
| `fast-tests-sync` | pre-existing (#3115 pollution) | see below |
| `quality-gate` | aggregator | fails because the above did |
| `Clean install verification (FR-010)` | **flake, re-measured green** | see below |

**`fast-tests-sync` is non-deterministically red on `main`:** six consecutive main runs produced **six different failing sets** — union of 11 distinct `tests/sync/` tests, and one run fully green. `test_429_respects_retry_after` (one of this PR's three) is red on main run `30736019209`. The remaining delta is an `FR-007` leak-guard ERROR, the same guard hitting a different victim each run. This is #3115's documented shard-parallel pollution, whose victims the mission dossier already names as this mission's own test surface. **Confirmed by isolated runs** — the trustworthy measurement here: `test_saas_client.py` **53 passed**, `test_saas_client_origin.py` **20 passed**, `test_background.py` **19 passed**.

**`Clean install verification` was the only job red here and not on main — and it is a perf flake, not a regression.** The `NFR-003` latency gate reported median **1.075s** against a **1.050s** ceiling (+44.3% vs 40.9%) — 2.4% over, with a **1.737s cold-start outlier** in a 5-run median. Re-run on the identical commit: median **0.781s, +4.9%, PASS** (runs 0.738–0.820, tight). **The budget was not touched.** Independently: `specify_cli.egress` is **not** imported by `import specify_cli` nor by the CLI entrypoint, so the new module adds zero startup cost, and this PR deletes two modules.

### Folds applied during landing, each with its reason

1. **`fix(decisions)` — reversed #3177's deferral.** `_mission_dirs` classified candidates with `resolved.is_dir()`, which returns `False` on EACCES from 3.14, so the recording branch under it was **present and inert** — LOW-8's original defect reintroduced by LOW-8's own fix. Two corrections to the original report, both measured on four interpreters: the floor is **3.14, not 3.13** (3.13.12 still raises, so the filed falsifier *fires*), and the mechanism is 3.14 delegating to `os.path.isdir`, **not** `_ignore_error` being removed. The "3.13+" came from probing `hasattr(pathlib, "_ignore_error")`, which flips at 3.13 only because 3.13 makes `pathlib` a *package*. Fixed rather than carried because no CI job runs pytest above 3.12 — deferring would park a known regression where no gate could rediscover it, with two permanently red tests as its only marker.
2. **`test(decisions)` — the guard that missed it.** The standing AST guard permitted the whole stat family inside a `try`, and a `try` around a non-raising call handles nothing, so it was **GREEN with the defect live**. Split now on measured raise behaviour: all four predicates banned outright, `stat`/`lstat` still permitted in a `try`. Plus the **paired known-bad control** the finding asked for, known-bad cases first.
3. **`fix(gates)` — the four architectural ratchets this branch tripped**, identified by measuring the base first: `JOB_GROUPS` drift from my own routing change; six `/tmp` sentinels → `/nonexistent`; two raw `kitty-specs` literals → `_SPECS_DIRNAME`; six golden-count sites **annotated `cardinality-is-contract` rather than re-frozen**, each judged in context (five are exactly-once egress assertions already paired with content assertions; the sixth is an anti-vacuity input floor). Re-freezing would have raised the ceiling and hidden the next regrowth.
4. **`chore(mission)` — `status_phase`**, from `cutover-guard`'s own prescribed remedy. Value checked against precedent rather than taken on faith: 335 of 337 missions on `upstream/main` carry exactly `"1"`. The migration reported `Flipped: 0` while writing the field, so the diff was read instead of the summary.

### Local verification

| Check | Result |
|---|---|
| Mission test set, **3.11.15** (a CI interpreter) | **152 passed** |
| `test_ownership_3111.py` on 3.11.15 / 3.12.13 / 3.14.4 | **33 passed** each — 3.14 was `2 failed, 31 passed` before the fold; **34** with the new control |
| `tests/architectural/` (full, proper env) | 7 failed / 1582 passed / 0 errors — **none mine**; both marker failures name `tests/runtime/test_runtime_bridge_identity.py`, which upstream changed in `943058143` |
| `commitlint --from upstream/main --to HEAD` | **RC=0** over all 12 |
| `ruff check` (changed files) | clean |
| `mypy --strict` (changed src) | 3 errors, **all pre-existing** — reproduced at 150/151/351 on a pristine base, appearing here at 162/163/365 shifted by exactly my insertions. Count unchanged 3→3. |
| Docs gates (inventory lockfile, retrieval index, description length, freshness) | clean — 3 warnings, all external link probes |

**Red-first / falsifiability.** Restoring `is_dir()` returns both EACCES tests to red on 3.14 and green on 3.11 — the divergence itself under control. Planting `resolved.is_dir()` back inside the `try` reds the tightened guard, naming the line. The #3111 differential was proven falsifiable under a plugin that drops the slug: `1 failed, 11 passed`, the failure quoting the captured request line. FR-013 verified empirically, not by reading: `specify_cli.egress` imports with **zero** leaked module-scope imports of `sync` or either transport.

### Filed rather than fixed here

- **#3178** — the pre-merge gate's remaining MEDIUM (security): `decision widen` resolves the destination team from ambient env vars *before* the per-project auth file, so an operator holding `SPEC_KITTY_TEAM_SLUG` for a team they do not own sends this checkout's decision identifier to that team. The **reverse direction** from #3111. Also records that `spec.md` folds FR-007 into FR-002 and that fold reads as discharged when it is not.
- **#3177** retitled to the measured floor (3.14+) with the mechanism corrected on the issue.
- `docs/adr/3.x/README.md` is missing the row for `2026-07-29-1-lane-base-recorded-planning-commit.md` **on `upstream/main`**. Untouched here, fails no check (`freshen_adr_inventory --all --check` exits 0), one command fixes it — noted rather than filed.

### Two deliberate deviations from our landing process, both measured

- **Did not run `git push origin upstream/main:main`.** The fork's `main` is **263 ahead / 380 behind** `upstream/main`; that push would discard 263 commits on a shared fork. It is also unnecessary — this PR's base is `Priivacy-ai:main`, so the fork's main never enters the commit range. (The sibling Bundle A landing reached the same conclusion independently.)
- **`landing fold:` appears in commit bodies, not subjects.** `landing` is not a valid conventional-commit type and the base's own commitlint rejects it.

### State

Rebased on current `upstream/main`, adversarial self-review folded, every red attributed. Flipping from draft to **ready-for-review** per the operator's standing go — which is also what re-runs the draft-skipped suites, so expect further pre-existing red from the five jobs that fail on `main` but had not been selected on the draft (`integration-tests-core-misc` ×3, `integration-tests-review`, `sonarcloud`). I will classify those the same way.

**The operator merges — never us.**

MOES-Media:
## Follow-up: the last unattributed red resolved, and one self-review fold

### `Clean install verification (FR-010)` is not this PR — filed as #3186

It recurred (2 of 3 attempts), so I stopped treating it as a flake and measured it instead of
re-running it. Three attempts on the **same commit** `de744d29a`:

| attempt | median | verdict |
|---|---|---|
| 1 | 1.075s (+44.3%) | FAIL |
| re-run | 0.781s (+4.9%) | PASS |
| 3 | 1.268s (+70.2%) | FAIL |

Then run locally with the **same interpreter (3.11.15)**, a **byte-identical fixture**, `--runs 9`,
each tree importing its own source:

```
pristine upstream/main (abca7ec96)   median 1.110s  (+48.9%)  FAIL
this branch                          median 1.124s  (+50.9%)  FAIL
```

**The base fails the same gate**, 1.3% apart, against a within-tree spread of ~80% (main alone:
0.834–1.502s). And `-X importtime` on the measured command gives **1039 modules on both trees** —
identical — with neither `specify_cli.egress` nor `specify_cli.decisions.ownership` imported by it
at all. This PR cannot have changed that command's startup cost.

The gate's ceiling sits below the runner's noise floor. It is 8/8 green on `main`'s CI runs because
that runner is faster than the ceiling, not because the ceiling has margin. **I did not touch the
budget** — raising a shared tolerance from inside an unrelated PR is bumping-to-green, and the
measurement shows the base is affected equally. Details and suggested directions: **#3186**.

### Final red classification: 11 here, 12 on `main`

The only job failing here and not on `main` is the latency gate above. `integration-tests-cli`,
`mutation-testing` and `slow-tests` are **skipped** on this run — an earlier report of them as red
was a cancellation caused by my own push, not a failure.

### Self-review fold — a defect I introduced, caught by re-reading the hunk

`S_ISDIR(resolved.stat().st_mode)` is what makes EACCES observable, and that part stands. But
`stat()` also **raises** for the absence-like failures where `is_dir()` merely returned `False`, and
I routed both to the same handler. So a **dangling** mission symlink raised `ENOENT`, landed in
`unreadable_ledgers`, and produced a *"could not read it"* refusal naming something that is simply
not there — **LOW-8's missing-vs-unreadable conflation running backwards**, reintroduced by the fix
for its fourth instance.

Measured before the fix: a lone dangling symlink yielded `unreadable_ledgers=('m-dangling',)` and a
refusal naming it.

The discrimination is now on errno — `_ABSENT_ERRNOS = {ENOENT, ENOTDIR, EBADF, ELOOP}`, deliberately
the exact set `pathlib._ignore_error` used through 3.13, reproduced rather than imported because that
helper is private and 3.13 moved it out of the top-level namespace. The intent is narrow: make EACCES
observable **without** changing how absence-like failures are classified.

The test asserts both halves in one place, because half a discrimination proves nothing. Half 1: the
dangling candidate is dropped and **not** recorded. Half 2 is the control that stops the test being
satisfiable by code that records nothing at all — the same shape with EACCES must still be recorded,
and it skips honestly rather than passing if the process can stat through a `0o000` directory.

**35 passed on 3.11.15**; ruff and `mypy --strict` clean.

I am re-running CI on this head and will re-classify; the change is confined to
`decisions/ownership.py` and its test module.

MOES-Media:
### Re-classification after the errno fold — unchanged

CI re-ran on `00807ece7`. The failing job set is **identical** to the previous run (excluding the
`quality-gate` aggregator), so the `_ABSENT_ERRNOS` fold changed nothing in CI:

- **This branch: 11 failing jobs. `upstream/main` at the same base (`abca7ec96`): 12.**
- The **only** job failing here and not on `main` remains `Clean install verification (FR-010)` — the
  NFR-003 latency gate, filed as **#3186** with the measurement showing pristine `upstream/main`
  fails it identically (1.110s vs 1.124s, 1.3% apart, 1039 modules imported on both trees).
- Every other red was verified same-reason against `main` at test-id level: `regression tests
  (blocking)` is an exact 6-for-6 match; `arch-adversarial` shard_2/shard_3 and `fast-tests-cli`
  have identical failing ids; `lint` is the `cryptography` CVE reproduced on a pristine worktree;
  `fast-tests-sync` is #3115's non-deterministic pollution (six main runs, six different sets), with
  isolated single-file runs green at 53 / 20 / 19.

Local verification on this head: **154 passed** on 3.11.15 across the mission's test set (up from 152
— the two tests this fold added), **35 passed** on 3.14.4 in the ownership module, ruff and
`mypy --strict` clean, commitlint RC=0 over all 15 commits.

Ready for review. **The operator merges — never us.**

LynnColeArt:
Splitting this PR's red CI into inherited vs. owned, because most of it is not yours and chasing it would cost you a session.

**Inherited from `main`** (failing identically on `main`'s own CI Quality run [30912775008](https://github.com/Priivacy-ai/spec-kitty/actions/runs/30912775008), so unaffected by anything on this branch):

- `fast-tests-sync` — `test_saas_client_origin.py::TestSearchIssues::test_429_retries_then_raises`, `test_saas_client.py::TestPolling::test_exponential_backoff_intervals`, `test_saas_client.py::TestRetryBehaviors::test_429_defaults_to_5s_when_missing`
- `regression tests (blocking)` — six, including `test_issue_2782_sync_strict_json_ingress_skip`, `test_issue_2804_merge_resets_gate_artifacts`, `test_issue_2996_approval_after_rejection_writes_no_verdict`, `test_issue_3086_merge_delete_branch_flattens_coordination_metadata`
- `fast-tests-cli` — `test_sync_purge_3030::TestPurgeAll::test_help_does_not_promise_machine_wide_erasure`
- `quality-gate`, `sonarcloud`

**Owned by this branch** (green on `main`):

- `lint` is mypy, not ruff. `_event_payload_without_at`, `_matches_legacy_contract`, `_seed_field_label` and `_seed_row_mismatch` are each receiving `object` where a `StatusEvent` is expected, plus a `StatusEvent | InnerStateChanged | None` union being treated as having `.get`.
- `arch-adversarial (arch_shard_2)` — `test_inline_meta_read_gate.py::test_routed_load_meta_floor`
- `arch-adversarial (arch_shard_3)` — `test_pytest_marker_correctness.py::test_fast_marker_must_not_apply_to_subprocess_users` and `::test_subprocess_git_users_must_carry_git_repo_marker`
- `integration-tests-core-misc` (integration / misc / specify-cli-heavy) — names not pulled yet
- `Clean install verification (FR-010)`

**Probably one fix, not two.** #3135 shows a byte-identical mypy error set and the same single `test_routed_load_meta_floor` failure. The two branches are siblings off the egress cluster, so whatever is handing `object` to those four helpers is very likely the same seam in both, and one correction should clear the `lint` and `arch_shard_2` columns on each.

One correction I owe on my own reading: I initially had `fast-tests-sync` down as caused by these branches. It is not — I had read `main`'s failure set off the commit `check-runs` summary, which omits the sharded CI Quality jobs, instead of opening `main`'s workflow run. The three tracker retry tests were already red.

Separately and not yours to fix: those six regression tests are each named after the issue they were written to pin, and they are red on `main`. Worth surfacing on its own thread rather than inside this PR.

stijn-dejongh:
Claiming this PR for landing: rebase onto upstream/main, classify red checks, fold remediations (boy-scout — one commit per fold, not deferred), run the review squad for the consent/ownership boundary this PR establishes, and verify the regression-marked tests this PR is expected to green (relocating/remarking them per tests/regression/README.md exit rules rather than leaving them red-first for a bug this PR fixes). Evidence to follow.

stijn-dejongh:
## Maintainer landing summary

**Verdict: merge-ready.** Rebased onto current `upstream/main`, classified every red check, folded five remediations, ran an independent red-first verification of the #3177 fix, dispatched a profile-loaded adversarial squad (verdict: safe to merge), and confirmed the regression-marked test set is unchanged throughout. CI is green except four checks that are pre-existing, tracked elsewhere, and reproduce identically on `main`'s own current tip.

### Rebase

Rebased 15 commits from the original merge-base (`abca7ec96`) onto current `upstream/main` (`1ca57887`, 73 commits later) — clean, no conflicts.

### Original 4 issues (author's own work, verified independently)

- **#3110** — duplicated consent-refusal gate consolidated into `src/specify_cli/egress.py`. Confirmed by `git grep`: no remaining importer of the two deleted modules anywhere in the tree.
- **#3111** — `decision widen` no longer trusts a decision it wasn't asked about; refuses when the acting mission and the decision's recorded owner disagree.
- **#3109** — docstring-only fix, behavior unchanged.
- **#3177** — **red-first verified independently**, not just taken on the author's word: built a real Python 3.14.6 venv, ran the ownership test suite green (35/35), then reverted the fix's predicate (`S_ISDIR(resolved.stat().st_mode)` → `resolved.is_dir()`) and confirmed it goes red (4 failed, matching the PR's own claimed count), then restored the fix and confirmed clean green again with zero diff.

### Folds (5, one commit each, `<type>(landing): ...`)

1. **`docs(landing)` — MD018 markdownlint fix.** Two hard-wrapped lines happened to start with `#3111`/`#3098` at column 1, read as ATX headings. Rewrapped, no wording changed. (Non-blocking soft-fail step, but a real finding — fixed anyway.)
2. **`docs(landing)` — twin docstring left uncorrected by #3109.** `register_egress_consent_resolver`'s docstring carried the identical false "replaces only when qualname matches" claim that #3109 fixed on its sibling function three lines below — the fixed function's own docstring even says "identical in the sibling registrar above," but the sibling itself was never updated. Found by the architect-alphonso squad lens.
3. **`docs(landing)` — CHANGELOG entries.** None of the four closed issues had one, despite every other recently-merged mission in the Unreleased section carrying one; #3111 in particular is security-relevant (cross-mission consent laundering). Found by the architect-alphonso squad lens.
4. **`refactor(landing)` + `test(landing)` — generalized the #3111/#3177 EACCES-safe stat guard (closes #3194).** `safe_is_dir()`/`safe_is_file()` extracted to `src/specify_cli/core/utils.py`, applied at all 8 originally-flagged call sites (`retrospect.py`, `_coordination_doctor.py`, `mission_type.py`, `migrate/charter_encoding.py`, `migrate/backfill_provenance.py`, `audit/engine.py`, `init.py`) plus 2 more found while writing regression tests (`retrospective/summary.py`, `status/identity_audit.py`). Each site gets its own dedicated EACCES regression test (errno-mocked, interpreter-independent, plus a real permission-bits test with an honest skip — mirroring `test_ownership_3111.py`'s technique). `decisions/ownership.py`'s own call site deliberately left untouched to avoid destabilizing its AST-guard-verified shape. Found by the paula-patterns squad lens; **issue #3194 closed** (folded here rather than left as a follow-up).

**Evidence for fold 4:** `ruff check` + `mypy` clean on every touched file; ~800 tests passing across the touched suites; full `tests/architectural/` run by the implementing agent: 1678 passed, 2 skipped, 2 xfailed, 0 failed; re-verified independently on the landing branch itself (314 targeted tests, ruff clean); `-m regression` run before and after — same 6 pre-existing opens, identical test IDs and assertion text, nothing flipped.

### Adversarial squad (pre-merge, architectural/API-surface change)

Two profile-loaded lenses dispatched read-only against the landing worktree:

- **architect-alphonso** (design/contract adherence): **safe to merge.** Verified the single-definition claim, the import-graph claim (transport/sync genuinely lazy-imported, not module-scope), and that `resolve_decision_ownership` closes the #3111 hole via actual code + test inspection, not just narrative. Findings: the twin docstring (folded, #2 above), the missing CHANGELOG (folded, #3 above), and that "identity is a detector, never an actuator" is enforced by tests/docs only, not a mechanical guard (filed as **#3195**, not folded — this mission already converted an analogous invariant into an AST guard for #3177, so the precedent for eventually doing the same here exists, but designing that guard is its own scoped task).
- **paula-patterns** (SSOT/duplication/boundaries): **safe to merge.** Confirmed no live duplicate of the consolidated consent logic anywhere in the tree (including docs/tests/agent-copy surfaces), confirmed the detector/actuator split is structurally clean, and found the EACCES-divergent pattern recurring at ~8 other call sites (filed as #3194, then folded — see above — rather than left open).

**Remaining open follow-up: #3195** (identity-detector-never-actuator needs a mechanical guard, not just docs/tests) — architecture-debt, correctly out of scope for this landing pass, tracked for a future pick-up.

### CI state — 4 known, pre-existing, unrelated reds

Matched by **test id + assertion text** against `upstream/main`'s own current-tip CI run (not the stale original merge-base), so these are provably not this PR's fault:

| Check | Cause | Why not folded |
|---|---|---|
| `regression tests (blocking)` | 6 honest-red open P0s: #2782, #2804, #2996 (×2 tests), #3086 | Deliberate red-first signals per ADR 2026-07-17-1; none touch anything this PR changes. Confirmed identical test IDs/assertions on `main` tip and unchanged after every fold in this pass. |
| `fast-tests-sync` | `test_saas_client_origin.py::test_429_retries_then_raises` and siblings — already-filed flake **#3187** ("SaaS tracker retry tests assert on sleep call counts and flake non-deterministically on CI") | Mission-scoped, already tracked; the real fix needs rethinking the retry-count assertion strategy, not a one-line fold. |
| `fast-tests-cli` | `test_sync_purge_3030.py::test_help_does_not_promise_machine_wide_erasure` — Typer `--help` rendering wraps `--all` across lines on CI's terminal width, unrelated to the app's own pinned-width console seam | Pre-existing (reproduces identically on `main` tip), tied to the broader `#3115` render-surface-width investigation already underway in this codebase (see `tests/conftest.py`'s WP02 seam). |
| `Clean install verification (FR-010)` | NFR-003 latency gate: `post_cutover_median 1.317s` vs `1.050s` ceiling | Documented, tracked perf-budget flake **#2749** — the gate's own baseline file explicitly says "NOT caused by unrelated PRs... do NOT recalibrate to mask it." One arch shard also showed a one-off `missions-root hardcode scan took 33.97s (>30s budget)` timing flake on an earlier push of this pass; did not reproduce on the final run. |

`quality-gate` fails only as an aggregate reflection of the above (it's in `main`'s own required-jobs list unconditionally) — this exactly matches `main`'s current state, not a regression this PR introduces.

### Regression-marked test set — unaffected

Checked directly, not assumed: `-m regression` across the whole tree returns the identical 6 failing test IDs with identical assertion text on `main` tip, on this PR before any folds, and on this PR after all 5 folds (including the EACCES-helper generalization touching 10 source files). None of this PR's product changes (#3110/#3111/#3109/#3177's fixes, or the #3194 generalization) intersect with any currently-open regression-marked test.

### State: watching for merge

Green except the 4 documented, cross-referenced, pre-existing exceptions above. PR is not a draft. Evidence trail is this comment plus inline commit messages. **The operator merges — I have not and will not run `gh pr merge`.**

Issue #3130, opened 2026-08-01
test(sync): 11 process-global and live-thread leaks in tests/sync, surfaced by the FR-007 leak guard


MOES-Media:
## Summary

The `#3115` bundle mission added an autouse leak guard to `tests/sync/conftest.py` (FR-007) that snapshots inventoried process-global state and the live-thread set around each test and **fails the test that leaves them dirty**. On its first full run it surfaced **11 stable, previously invisible defects** in `tests/sync/` — plus 1–2 more that appear only under `--dist loadfile`.

These are **pre-existing**. They are not caused by the guard, and they are not flaky.

## Evidence they are real

Verified twice, independently:

- **Serially reproducible**: `12 passed, 11 errors in 57.07s` on the flagged node-ids alone.
- **Confirmed by an instrument that is not the guard**: the mission's reviewer neutralised the guard at hook level (asserting the mutation bound) and ran its own census, written independently. It flagged **the same 11 nodes, the same symbols, the same threads** — and re-checked each leaked thread 1.0s later, finding every one `STILL-ALIVE-after-1.0s`. Not teardown-timing races.
- **With the guard inert, all 12 tests pass** (`12 passed in 56.86s`), so these are green today and the guard is what makes them visible.

## The 11

| # | Node-id | Leaks |
|---|---|---|
| 1 | `test_issue_598_hang_fixes.py::TestBackgroundStopBounded::test_stop_does_not_hang_when_sync_is_slow` | live `Thread-1`, target `_guarded_final_sync` |
| 2 | `…::test_stop_emits_structured_warning_when_sync_times_out` | live `Thread-2`, same target |
| 3 | `test_runtime.py::TestSyncRuntime::test_starts_background_service` | live `spec-kitty-sync-async-loop` |
| 4 | `test_runtime.py::TestUnauthenticatedBehavior::test_no_websocket_when_unauthenticated` | same |
| 5 | `test_target_authority.py::test_all_fields_populated_under_env_equals_config` | `os.environ['SPEC_KITTY_SAAS_URL']`: `None → 'https://config.example.com'` |
| 6 | `test_background_body.py::TestTimerBodyQueue::test_timer_triggers_when_only_body_queue_has_tasks` | live `Thread-3` |
| 7 | `…::test_timer_skips_when_both_queues_empty` | live `Thread-4` |
[...]

Issue #3113, opened 2026-07-31
test(arch): egress-boundary guard misses all-positional transport calls — name the limit / tighten the AST match


stijn-dejongh:
## Summary

The egress-boundary guard's injected-transport rule `TRANSPORT_CALL` (`tests/architectural/test_egress_consent_boundary.py:295-306`, `_transmits_a_body`) requires a `headers` argument **and** a body **as keywords**. A future sink written with all-positional arguments — `poster(url, body, hdrs)` — or without an explicit `headers=` keyword evades both the method-name rule and the transport-call rule. The guard's own bite-test only exercises the kwargs form (`poster(url, data=body, headers=hdrs, ...)`, ~line 922).

This is inherent to AST matching and low-likelihood (idiomatic HTTP calls use kwargs), so it is a completeness gap, not a live leak. But it should be a **named** claim, not a silent one: the guard's stated limit list (getattr-by-string, empty registries, dynamic import, subprocess, at-rest, bare `.put`, multi-sink-per-file) does not currently mention the positional-args case.

## Done when

- The positional-argument transport call is added to the guard's stated completeness limits (docstring), and — if cheap — the AST matcher is tightened to also flag a body passed positionally; the bite-test grows a positional-form case.

## Where

- `tests/architectural/test_egress_consent_boundary.py:295-306` (and the bite-test ~line 922)
- Raised as a review NOTE (architect-alphonso, NOTE 5) during the #3098 landing pass.

---
Follows up from #3098 (P0 sync/egress consent hardening).
Part of #1800 — Epic: SaaS sync & event-envelope hardening.

PR #3135, opened 2026-08-01
fix(#3108): refuse tracker egress unless both consent channels permit

### LINKED ISSUE #3108: sync egress: tracker connectors (Jira/Linear) have no per-project consent refusal (E20/C-006)
## Summary

Operator-configured tracker connectors (Jira/Linear) send issue **titles** to a third-party tracker under a **machine-scoped** credential, with a **project-scoped, git-tracked** binding — but nothing keys the egress on the data's own project, and **no per-project refusal is expressible**. A project with a committed `sync.enabled: false` can still push its issue titles. The machine-global credential and the project-scoped intent are reconciled nowhere.

This is a real open collection surface, but it is **off the automatic `sync now` / daemon-drain hot path** — it is operator-invoked and targets a third party, not spec-kitty SaaS — which is why the #3030 mission scoped it out and disclosed it rather than fixing it.

## Where

- `src/specify_cli/tracker/local_service.py` via `src/specify_cli/cli/commands/tracker.py`
- Disclosed as **E20 / C-006** in `kitty-specs/journal-project-consent-3030-01KYKWQS/egress-inventory.md` and in PR #3098's "Known limits" section.

## Done when

- A project's committed consent refusal (`sync.enabled: false`) is honoured on the tracker-connector path, keyed by the project the issue belongs to.
- A regression test proves a refusing project's issue titles are not pushed.

---

### LINKED ISSUE #3174: No executed test that a refusing project keeps its ungated tracker commands working (#3108 follow-up)
Found during `#3108` (Bundle C) review. Scoped as a follow-up by the reviewer rather than absorbed, because the claim is measured-true today and the harm from it being wrong is low.

## The gap

`#3108` gates three tracker commands on a two-channel egress verdict: `tracker sync pull`, `sync push`, `sync run`. It deliberately leaves `status`, `bind`, `unbind`, `map add` and `map list` (without `--provider`) **ungated** — a project that refuses egress must keep its local-only commands fully working.

The upgrade note `docs/migrations/tracker-egress-refusal.md` states this to operators. **Nothing executes it.** There is no test that runs the ungated commands against a refusing fixture and asserts they succeed.

## Why it matters more than a coverage tick

"No over-gating" is the property that distinguishes a correct gate from one that simply refuses everything. A gate that refused unconditionally would satisfy *every* refusal assertion in the mission's acceptance suite — the suite pins that refused things are refused, not that permitted things are still permitted, for the ungated set.

The mission's positive controls cover the consenting-push path. They do not cover "a refusing project can still run `status`".

## Evidence it is true today

Measured during WP04's review, as a probe rather than a committed test:

```
project with tracker.egress: refused
  status      -> OK  {'configured': True, 'provider': 'beads', ...}
  sync_pull   -> REFUSED by the egress gate
  sync_push   -> REFUSED by the egress gate
  sync_run    -> REFUSED by the egress gate
```

Structurally corroborated by the call-site census: exactly 6 `tracker_egress_verdict` call expressions across 5 enclosing functions, none of them in the ungated commands.

## What would close it

One cell in `tests/sync/tracker/test_tracker_egress_refusal_3108.py` (or `test_local_service.py`) driving `status` and `map list` against a `tracker.egress: refused` fixture and asserting exit 0 — the converse of the totality assertions already there.

---

Observed at `abca7ec9` + the `#3108` branch. Not fixed in that mission to avoid further amendment of already-approved files.

MOES-Media:
Closes #3108
Closes #3174

## What this fixes, for an operator

If you use a **local `beads`/`fp` tracker binding**, `spec-kitty tracker sync pull`/`push`/`run` used to ship your issue titles, bodies, labels, and assignees — which in this product *are* client engagement names — to an operator-named subprocess **with no consent check at all**. A project that had committed `sync: {enabled: false}` still shipped them. That gap is now closed.

**After this change:** a local tracker binding needs a recorded consent decision, and **absence denies**. Nothing about your binding changes — only a consent decision is now required. Two ways to keep syncing:

- record `tracker.egress: permitted` in your project's own `.kittify/config.yaml` — grants the **local** subprocess path only, and consents to hosted SaaS sync **not at all**; or
- opt into hosted sync (`spec-kitty sync opt-in` / `sync.enabled: true`).

`spec-kitty sync doctor` tells you exactly which channel is refusing and how to fix it. Full remediation paths: [`docs/migrations/tracker-egress-refusal.md`](docs/migrations/tracker-egress-refusal.md).

> ⚠️ **Breaking change.** Every existing `beads`/`fp` binding stops syncing on upgrade until its project records one of the two decisions above. This is deliberate: absence-permits would bet confidentiality on the unverified premise that `bd`/`fp` are local-only, when the executable name is operator-overridable from a machine-global file.

## The consent model

Two channels are joined at one call site, `tracker_egress_verdict(root, *, destination, identifiers)`:

- **Channel 1** — the existing hosted-sync consent chain, keyed on `project_uuid`. **Absence denies.**
- **Channel 2** — `tracker.egress` (`refused` | `permitted`) in the project's own committed config. Absence defers to Channel 1; any other value is a fault that refuses (no case-folding, no `yes`/`on`/`1`).

The asymmetry is the design: at the **local subprocess** destination a `permitted` grants independently of Channel 1 (the subprocess is the operator's own machine); at the **hosted service** destination `permitted` is a no-op and never a grant, because that transport reaches spec-kitty's own hosted service — the exact destination the original incident concerned.

Note on scope: the issue's original premise (that the hosted Jira/Linear path leaked) does **not** hold — that path was already gated by #3030. This PR closes two *different* defects: the ungated local subprocess path (Gap B), and the lack of a separable per-destination consent key (Gap A).

## Landing-pass remediation (maintainer)

An adversarial squad (@robertDouglass, three lenses) returned `BLOCK_MATERIAL` with three HIGH findings; all three were adjudicated and folded before hand-off:

- **HIGH-1 — HTTP before refusal (fixed).** A refused *hosted* sync still emitted a `urllib` HEAD probe to the tracker host, because `_check_sync_readiness` ran the networked readiness probe before the consent gate. The hosted verdict is now consulted at the CLI pre-flight, ahead of any network probe — so refusal precedes any HTTP attempt. Red-first proven; both call-site censuses (G4 + `TestZeroBlastRadius`) re-pinned for the new audited chokepoint.
- **HIGH-3 — local `bind` over-gated (fixed).** The migration note promised a local bind keeps working, but `bind_command` ran hosted readiness first, so an unauthenticated project's local `bind` exited at `missing_auth`. A local `--provider beads`/`fp` bind now skips hosted readiness; SaaS-provider binds still require it. Red-first proven; migration note made explicit.
- **HIGH-2 — two-authority verdict (adjudicated → tracked as #3287).** Enforcement is single-authority (`project_egress_refusal`); only the *diagnostic* label re-resolves through sync internals, and that split can never flip the enforced decision (traced at every write site, pinned by a new defense-in-depth regression). The honest fix needs the Bundle-B resolver-contract change, out of scope here — filed as **#3287**.

A second, **independent** squad then re-reviewed the folded tree blind to this work (three fresh lenses, code + spec only) and returned **3/3 ACCEPT — zero BLOCK_MATERIAL, zero MAJOR**: no leak reproduced, every malformed/unreadable input fails closed, enforcement confirmed single-authority. Two further hardening folds came out of it:

- **Root consistency (fixed).** The CLI pre-flight and the transport each resolved the repo root independently; a caller whose cwd ≠ the data-owning root could have the two hosted gates judge different projects. The four sync commands now resolve the root once and thread it into both. Red-first proven.
- **Residual nits (#3291).** The import-isolation probe now skips a package-less interpreter instead of false-failing; an unreadable `.kittify` *directory* is carried as a fault instead of raising a stray traceback (a real "never raises" contract fix). One efficiency residual (redundant per-verdict git-identity resolution) is left to **#3291**, tied to #3287.

Nine landing folds total, each a single-purpose `<type>(landing):` commit; contributor history is code-by-slice with the governance dossier bunched.

## Tests

- Egress suites green (`tests/sync/tracker/test_tracker_egress_*_3108.py`), guards green (`tests/architectural/test_tracker_egress_guards_3108.py`), mock-boundary CLI tests green. `ruff`, `mypy --strict`, `commitlint`, terminology, docs-freshness, markdownlint all clean on the folds.
- #3174's "no over-gating" promise is executed: `TestUS7NoOverGatingUngatedCommandsSurviveRefusal` proves `status`/`map add`/`map list`/`unbind` keep working on a refusing project.

## Deferred, with issues filed

- **#3287** — Bundle-B resolver contract / two-authority diagnostics.
- **#3290** — `runtime_bridge → sync` hard-import coupling (surfaced by an architecture-boundary review; out of scope here).
- **#3291** — residual efficiency nit (redundant per-verdict git-identity resolution).
- #3167 · #3168 · #3169 · #3170 · #3172 — the contributor's original deferrals.

Largest residual: nobody has established what a real `bd`/`fp` binary does with the argv, so Gap B's *severity* is unestablished — the gate is defensible either way.


MOES-Media:
## Remediation summary

**Verdict: draft for operator.** Planning-only; no product change; board halted at 8 × `planned`.

### Folds that changed the design, and why

| # | Finding (measured) | Fold |
|---|---|---|
| 1 | The issue's premise is false — Jira/Linear already gated, absence already denies | Mission re-scoped to the two gaps that actually survive |
| 2 | `_resolve_saas_backend_for_provider` overrides the on-disk provider **in memory**; three commands drive the hosted transport from a `beads` repo | `destination` became a **required keyword-only parameter**, never derived; new guard pins that no call site computes it from a config read |
| 3 | Guard count declared "exactly three" against a design with five call sites | Corrected to 5 enclosing functions / 6 call expressions, counted separately |
| 4 | Positive control captured **1 argv, no sentinel** on an empty store | Store seeding became a pinned requirement; argv count corrected 2 → 3 |
| 5 | Default ownership mode skips `create_issue` entirely | Every fixture pins `spec_kitty_authoritative`, with the reason |
| 6 | Un-armed CLI aborts: exit 1, 0 subprocess, 0 HTTP — satisfying every refusing assertion with nothing built | Fixtures set the arming var explicitly and assert refusal **text** |
| 7 | The house pattern patches out the gate's own method — plugin-injected gate measured **bind count 0, 519 tests green** | Recorder injection point became a requirement; gate carries a bind counter asserted non-zero |
| 8 | A module-qualified call site passed **both** guards with the input count *rising*, while killing every specified mutant | Matchers resolve `ast.Name` **and** `ast.Attribute`; third mutant added |
| 9 | Erasure inventory was 3 sites; the real count is 9, of which 2 are broken *by this Mission's own field-shape change* | Inventory re-derived and split by class, with a landing rule |
| 10 | Three "reds" do not exist on the base | Each re-stated honestly; two reclassified as defence-in-depth with **no** red-first pin |

### Squad verdicts

- **post-specify** — 4 lenses, **4 × REJECT**. Remediated.
- **post-plan** — 3 lenses, **3 × REJECT**. Remediated, then **re-reviewed**: 3 × ACCEPT-WITH-FINDINGS, all blockers closed.
- **post-tasks** — 3 lenses, 3 blockers + 2 highs. Remediated.

Two operator escalations were raised mid-flight where the decision was genuinely the operator's, and
one orchestrator refinement was flagged rather than applied silently: taken literally, two operator
answers combined would have allowed a grant to spec-kitty's hosted service with hosted-sync consent
absent. Polarity was therefore bound to the destination, and the operator was told.

### Gates

- `commitlint --from upstream/main --to HEAD` → **exit 0**
- `pytest tests/architectural/ -q` → `19 failed, 1438 passed, 4 skipped, 26 errors`
- Pristine `upstream/main` worktree, same suite → `19 failed, 1438 passed, 4 skipped, 26 errors`
- **Failing test-name sets diffed: 45 vs 45, zero difference in either direction.** Sets, not tallies.
- `ruff`/`mypy`: no changed Python files. The prescribed field shape and verdict signature were
  separately prototyped under `mypy --strict` during review — `Success: no issues found in 1 source
  file`, zero suppressions.

**No upstream issue filed for the 45.** They are unchanged by this PR, but I have not established
whether they are red on CI — the dominant local cause is `/usr/bin/python: No module named pytest`
in subprocess probes, which is environment-shaped. Claiming main is broken without that evidence
would be a false report.

### Red-first

N/A — no product change. The committed evidence is the measurement record in
`tracer-evidence-base.md`, which a second lens independently re-ran and reproduced.

### Base

`upstream/main` @ `bb2020fea`. Zero upstream commits since the merge-base; upstream never touched the
dossier path. History compressed to one governance commit (all-admin); tree verified byte-identical
to the pre-compression working history.

### State

Fork CI validating. **Draft for operator. Not un-drafted, not merged.**

MOES-Media:
## Correction: 41 of the 45 local reds were my environment, not the repo

My earlier comment reported `19 failed / 26 errors` and classified all 45 as pre-existing on the
strength of an identical set-diff against `upstream/main`. The set-diff was sound, but the *cause* I
offered was wrong, and the conclusion was weaker than it needed to be. Root-caused and fixed.

### Root cause

`pytest`, `typer` and `specify_cli` were resolving only from the **user site** (`~/.local/...`).
This clone's `.venv` had been created without the test extra. The architectural suite deliberately
**isolates `HOME`**, so every subprocess those tests spawn lost its imports.

Controlled probe:

```
CONTROL: real HOME      →  pytest OK from /home/jeroennouws/.local/lib/python3.14/.../pytest
SUBJECT: isolated HOME  →  ModuleNotFoundError: No module named 'pytest'
```

That single mechanism explains the whole distribution: 27 × `gate-coverage collection did not
complete cleanly` (spawned `--collect-only` sweeps dying), 8 × `No module named 'typer'`,
4 × `move-task … failed` (spawned CLI), 3 × `assert 'TID251' in ''` — that last one asserting on
**empty subprocess output**, which is this repo's own "an empty output file is no measurement" rule
biting its own suite.

### Fix and verification

`uv sync --extra test`. Control confirms the fix addresses the actual mechanism: all three modules
now import **under isolated `HOME`**, the exact condition that broke them.

Re-ran the 15 previously-failing files:

```
before:  45 failed/errored
after:    4 failed, 176 passed in 1372.03s (0:22:52)
cleared: 41
new:      0   (set-diff, not count-diff)
```

The **4 that remain are all `test_tid251_enforcement.py`** — precisely the documented
known-pre-existing entry ("4 tests, proven pre-existing on `origin/main` in a pinned worktree").
Nothing is unexplained, and nothing new was introduced.

### What this changes

- The local architectural gate is **green modulo the documented known-red roster**. That is a
  materially stronger statement than my earlier "45 pre-existing, cause unknown".
- It also **withdraws the concern** about whether these are red on CI. They are an artifact of a
  user-site install meeting `HOME` isolation; CI installs into a venv, so the mechanism cannot fire
  there. No upstream issue is warranted, and none was filed.

### Method note

The re-run initially reported `exit 0` with a 0-byte output file. Both were meaningless — the exit
status belonged to the wrapper shell, and the empty file was pytest buffering, while the real
process was still alive. Caught by checking the PID rather than the exit code. The earlier full-suite
attempt was killed at a 10-minute timeout and was **not** interpreted; it was re-run narrowed
instead. Runtime rose from ~9:45 to 22:52 precisely *because* the subprocess tests now execute
rather than failing fast.

Still a draft. Nothing merged, nothing un-drafted.

MOES-Media:
## Remediation summary — implementation landed on the planning dossier

**Verdict: all 8 work packages approved. Draft, for operator review.** Rebased onto `upstream/main` `abca7ec96`; history recompressed to **9 commits — dossier bunched, code by slice**.

### Every review round found something real. The folds that changed the code:

| WP | Round 1 | What the fold was |
|---|---|---|
| WP01 | **REJECT** ×2 HIGH | A cell named "hosted drain unaffected by tracker key" gave **neither arm a project** — it compared a fixture with a clone of itself, green and asserting nothing. A bind-counter cell was guaranteed to false-red once WP04 landed, because its two arms embedded different fixture names into the recorder path that becomes `argv[0]`. |
| WP02 | ACCEPT-WITH-FINDINGS | The FR-010 null residual's recorded reason was **factually wrong** — the reviewer built the constructor-level fix and got byte-identity, so the deviation stands on a different ground (it would widen the unaudited `_extra` surface). Closed-set vocabulary exported so WP03 could not re-spell it. |
| WP03 | **REJECT** ×2 HIGH | `except TrackerConfigError` was too narrow: an unreadable `.kittify` **directory** raised `PermissionError` out of a function whose contract is "never raises". And `channel1_state` reported `recorded_refusal` for a project whose Channel 1 **permits** — invisible behind 100% line coverage, because the line was executed everywhere and asserted nowhere on the permitting path. |
| WP04 | **REJECT** ×1 HIGH | A forward CHANGELOG link turned two green docs gates red, unreported. Kept as a named both-ends window rather than allowlisted — that allowlist is deliberately empty and forbids entries for links that will resolve. Closed by WP08: **2 failed → 44 passed**. |
| WP05 | **REJECT** ×1 HIGH | Removing the old seam symbol left the E2 egress allowlist **certifying a symbol that no longer existed** — the file was still permitted to transmit while the thing that made that safe had gone. The guard fired exactly as designed. |
| WP06 | **REJECT** ×1 HIGH | `verdict.message` quotes the operator's own `tracker.egress` value verbatim (C-020 demands it) and was printed through rich markup: a value of `[refused]` rendered as `''` — **a false statement about the operator's own file, on the surface built to stop exactly that** — and `[/bold]` exited 1 with `MarkupError`. |
| WP07 | ACCEPT-WITH-FINDINGS | Peak RSS **1,205,972 → 96,056 kB**. G3 accepted a gate nested as a call argument, which executes a credential read *before* the gate — the very consequence G3 exists to prevent. |
| WP08 | ACCEPT-WITH-FINDINGS | Refused to document a remedy with no executed test behind it: SC-007 names two remedies for the identity-less state and only one was executed. Proved the other, reported the gap; the missing test is now committed. |

### Three CI gaps the guards' sweep exposed — all the mission's own false-green shape, aimed at its own evidence

- **The acceptance harness was orphaned.** Its `pytestmark` was a bare `skipif` with no selection marker, so CI would **never have run the suite that proves the leak is closed.** Now `integration` (not `unit` — those cells spawn a real subprocess and open sqlite).
- **The verdict suite claimed `fast` while shelling out**, violating a marker-correctness rule on an always-on pole. It was adding a violator; now `integration`. The one remaining violator is pre-existing on `upstream/main`.
- **Golden-count regrowth** in three trees. 40 sites annotated where the count *is* the assertion; **3 were misclassified and converted** — `len(PROBED_VALUES) == 15` pinned no labels, so deleting the `list` case and duplicating `true` kept the count and red nothing, silently dropping coverage of an unhashable value.

### One operator decision

`decisions/DM-FR016-hosted-byte-identity.md`. FR-016 (byte-identical hosted refusal) and a WP03 review finding (recompose the message with a no-op note) demanded opposite things. **FR-016 wins** — a mission closing the local gap must not perturb the already-shipped hosted gate. Accepted cost: the hosted raise site renders `message` alone, so the verdict's remedies do not appear there; they resurface in `sync doctor`, which reads the same verdict and is not bound by byte-identity.

### Gates on the rebased tip

```
commitlint  upstream/main..HEAD      clean
ruff check  16 changed py files      All checks passed!
mypy --strict 8 changed src files    5 errors == 5 on base, zero new (diffed vs base worktree)
mission surface                      780 passed, 0 failed
tests/architectural/                 7 failed, 1588 passed — all 7 pre-existing, each reproduced
                                     red on a pristine upstream/main worktree
docs link gate                       44 passed (was 2 failed, 42 passed)
```

### Honest limits

- `#3115` remains open; its shard-pollution risk to `tests/sync/tracker/` is mitigated but not zero and is not locally reproducible. Any CI red there should be attributed by name before being treated as ours.
- `#3113` was closed by **documenting** its gap, not tightening the matcher, so the egress-consent scanner does not catch all-positional transport calls. G1–G8 do not lean on it.
- Preconditions with **no executable guard**, recorded as prose: the tracker path staying operator-invoked, and a third transport reusing `HOSTED_SERVICE`. Neither G4 nor G5 decides polarity.
- **Largest residual:** nobody has established what a real `bd`/`fp` binary does with the argv. The gate is defensible either way; Gap B's *severity* is unestablished.

**Staying a draft. Not un-drafting without your explicit go, and not merging.**

MOES-Media:
## CI triage — every red attributed; two were mine and are fixed

Three pushes. **`cutover-guard` and `docs-freshness` are now green** — both were genuinely ours. CI also caught **two real regressions eight rounds of local review missed**, both now fixed and pinned where they originate.

### The two that were ours

**1. An unbound checkout was told something was wrong with it.** The `sync doctor` renderer appended to doctor's `issues` list whenever the verdict refused — including for a checkout with **no tracker bound at all**. Absence of both channels refuses a transmission nothing is attempting. It also made the renderer's contribution depend on ambient checkout state, which is how it broke `tests/sync/test_sync_doctor.py::test_doctor_healthy`: a heavily-mocked unit test that nonetheless resolves the real project root. Gated the **append only** — both rows still render their REFUSED verb, because suppressing them would rebuild the false green where "tracker egress is fine" and "I never looked" read identically.

**2. The fix for (1) reintroduced WP03's HIGH-1, one level up.** The binding probe called `load_tracker_config` unguarded, and that raises on an unparseable `.kittify/config.yaml` — so `sync doctor` aborted mid-render on exactly the checkout it exists to diagnose. The symptom read like a wording change and was a **truncated command**: `REPAIR THE FILE'S SYNTAX` dropped 4 → 2 because every line after the block stopped printing. `tracker_egress_verdict` is defended against this internally (NFR-003); a second, direct config read was not.

### The remaining reds, each attributed by measurement

**Upstream `main`'s CI Quality is failing at `abca7ec96` — the exact commit this branch is based on** (run `30874653016`), with the same job set. Job-by-job:

| Check | Attribution |
|---|---|
| `lint` | `pip_audit` dependency finding. Red on main. This PR changes no dependency (`uv.lock`/`pyproject.toml` untouched). |
| `arch-adversarial (arch_shard_2)` | `test_inline_meta_read_gate::test_routed_load_meta_floor` (census 110 vs floor 117). Red on main, and reproduced red in a pristine `upstream/main` worktree locally. |
| `arch-adversarial (arch_shard_3)` | `test_pytest_marker_correctness` — names only `tests/runtime/test_runtime_bridge_identity.py`, which is pre-existing. **This mission adds no violator**; a marker violation it *did* introduce (`[unit, fast]` on a subprocess-invoking file) was found and fixed before this PR. |
| `fast-tests-cli` | `test_sync_purge_3030::test_help_does_not_promise_machine_wide_erasure`. Red on main at the same commit. |
| `fast-tests-sync` | `test_429_retries_then_raises` + `test_429_defaults_to_5s_when_missing` — the `#3115` retry class; main is red on the same file/class. Plus 2–4 FR-007 leak-guard **errors** in `test_background.py` / `test_legacy_queue_guard_3030.py`, pinned by the guard itself to filed issue `#3130`. Those vary run-to-run (4 then 2, different node ids), and running this mission's three new test files *before* `test_background.py` in one process gives **239 passed** — so these tests are not being polluted by ours. |
| `regression tests (blocking)` | Red on main. Its assertions name `#2804` and `#3086` themselves (merge / acceptance-matrix reproductions). |
| `quality-gate` | Cascade — it aggregates the above and reports `FAIL — job failure` for each. |
| `Clean install verification (FR-010)` | **Perf-budget flake, filed as `#3180`.** Measured median **1.058s** against a **1.050s** absolute ceiling — an **8ms (0.76%)** overshoot, with the five samples spanning 0.968–1.103s and **two of five below the ceiling**. The baseline's own note records a prior 19ms overshoot that required recalibrating 1.000 → 1.050. The gate measures `spec-kitty next`, and `runtime.next` does not import `specify_cli.cli.commands.sync` — verified — so this PR's diff is outside its reach. It also **passed at 27s on the immediately preceding push**, whose only delta was a function-local `try`/`except` in that renderer. Not bumped to green. |

### Landing-guide note

The guide prescribes `landing fold: …` commit subjects. This repo's `commitlint.config.cjs` requires a conventional type, so those commits were rejected outright — rewritten as `chore/docs/fix(tracker-egress)` with "Landing fold" in the body. Flagging so the guide can be corrected.

**Local state after the folds:** mission surface **1415 passed**; `commitlint` clean; `ruff` clean; `mypy --strict` zero new errors against a base worktree; all three docs gates green (`errors=0`); `cutover-guard` green.

MOES-Media:
## Rebased onto current `upstream/main`, Bundle B port completed, all 19 red checks re-classified

**Head `39339ac82` — 15 ahead / 0 behind.** Safety tag `prerebase/3135-1786066851` → `0a772e345` retained.

### Why the rebase needed work beyond replaying commits

Bundle B (already on `main`) moved `project_egress_refusal` and made `identifiers` **required**. Five call sites were still on the pre-Bundle-B arity. Resolving them surfaced three things not visible from the diff:

1. `tests/sync/tracker/test_tracker_egress_verdict_3108.py:38` **imported the deleted module** — a collection-time `ImportError`, not merely a TypeError risk.
2. **48 further test call sites** omitted `identifiers`, plus `_resolve_channel1(root)` and `project_egress_refusal(root)`. Ported by AST transform, each passing its destination's owning fragment via a `_IDENTIFIERS_FOR` map imported from the transports rather than restated. Verified by AST probe: **54 calls carry `identifiers`, 0 missing**, with a control proving the probe can detect a miss.
3. `test_no_saas_client_import` asserted `"saas_client" not in source` — banning the token in **prose**, so it false-red'd on an explanatory docstring. Re-pinned over *parsed import statements*: strictly stronger (it now catches `from specify_cli.tracker import saas_client` and `import … as _c`, which the literal check missed) and indifferent to comments. Controlled against 5 known-answer cases before being trusted.

Upstream's rule is *"each transport passes its own identifier-set fragment as an argument"*, so `local_service.py` got its own `LOCAL_SUBPROCESS_EGRESS_IDENTIFIER_KINDS` rather than importing the hosted constant — same value today, but a module-level import would drag the hosted HTTP stack into local-connector import time, and either fragment may change independently. `sync doctor` reports rather than transmits, so each row uses **the owning transport's** fragment; a doctor-local set would print a refusal differing from the one the gate raises, which is the exact failure `egress_verdict.py` exists to prevent.

### The 19 checks

| Check | Verdict |
|---|---|
| `lint` ×2 | **Pre-existing main breakage** — not ruff. The failing step is pip-audit: `cryptography 49.0.0`, CVE-2026-69247, fix 50.0.0. `ruff check src tests` prints `All checks passed!` in the same job. **A rebase cannot clear this; a dependency bump can.** |
| `arch_shard_2` ×2 | **Pre-existing main breakage** — `test_routed_load_meta_floor`, reproduced on a pristine `upstream/main` worktree with a byte-identical message (floor 128, margin 4, live 133). Not fixed in-PR. |
| `arch_shard_3` ×2 | **Staleness — evaporated.** Passes on the rebased branch *and* on pristine main. |
| `fast-tests-sync` ×2 | **Measurement artifact of #3136** — mechanism proven from source, not assumed. `@patch("specify_cli.tracker.saas_client.time.sleep")` resolves to the **stdlib** `time` module attribute, so the recorder is process-global. The CI call list `[0.001, 0.002, 0.004, 0.008, 0.016, 2.0]` contains exactly one call the test made (`2.0`); the doubling series is a concurrent backoff loop in another thread. **Two** tests affected, not one. This PR's `src` diff contains zero `time.sleep`/retry lines. Both pass serially here. |
| `integration-tests-core-misc` ×3 | **Pre-existing main breakage** — review-cycle, #2804 merge-gate reset, #3086 coordination metadata, registry activation parity. None in this diff's scope. *Classified from CI evidence + diff scope; not reproduced locally.* |
| `regression tests` ×2 | **Pre-existing main breakage** — 6 failures. The one sync-adjacent failure was verified rather than assumed: `tests/sync/test_strict_json_stdout.py` → **4 passed**. |
| `Clean install (FR-010)` | **Measurement artifact** — NFR-003 latency gate, median 1.058 s vs a 1.050 s ceiling. Missed by 8 ms, and it **passed on a newer run of the same commit**. |
| `fast-tests-core-misc (rest-2)` | **Measurement artifact** — `test_sweep_nfr_002_10k_files_under_5s`, 5.105 s against a 5 s wall-clock budget. |
| `fast-tests-cli` ×2 | **Not a PR defect** — `--all` absent from `purge --help` (CI rendered a degenerate usage line). Zero `purge`/`--all` lines in this diff; passes locally on the branch. |
| `quality-gate` ×2 | **Derived** — pure aggregator; 5 of its 12 blocking verdicts are "improperly skipped" suites whose dependency failed. |

**Zero PR defects. Zero contract crossings** — nothing needed re-pinning, because the port composes upstream's mechanism rather than moving a seam.

- `ruff check` on changed files → `All checks passed!`; `--select TID251` → `All checks passed!`. (`ruff format` never run.)
- `mypy --strict src/specify_cli src/charter src/doctrine` → **0 errors / 1133 files**, against pristine main's 0 / 1132 → **zero new errors**.
- `tests/sync/tracker/` → **704 passed**, 0 failed, `^ERROR tests/` = 0.
- `tests/architectural` → 1757 passed / 5 failed, corrected to **1 real failure** (the `load_meta` floor, identical on main). The other 4 shell out to `sys.executable -m ruff`, and this venv has no ruff — control: all **9 pass** under a ruff-provisioned interpreter against the same worktree.

### One thing deliberately not done

`fast-tests-core-misc-nodeids.txt:8336` is a frozen selection census that has **already drifted on pristine main** — the nodeid is generated from the filesystem and cannot be produced there either (line 8169 names a second deleted file). It was **not regenerated.** Refreezing a selection baseline is precisely how a real selection regression gets laundered into green, and this PR does not move that census.

Residuals ledgered in `kitty-specs/tracker-egress-refusal-3108-01KYWF1R/bundle-b-port-residuals.md` (R-1…R-5). No issues filed, per operator direction — the ledger is reviewed for folding once the mission reaches PR.

**State:** rebased, pushed, CI re-validating on the corrected base. Draft, for the operator.

MOES-Media:
## Two folds: `#3174` closed, and the `#3110` contract crossing re-pinned

Pushed `e873f20ce`. Zero `src/` changes in either commit — `git diff --stat HEAD~2..HEAD -- src/` is empty.

### 1. `#3174` — the no-over-gating promise is now executed

`#3108` gates `sync pull`/`push`/`run` and deliberately leaves `status`, `bind`, `unbind`, `map add`, `map list` ungated; `docs/migrations/tracker-egress-refusal.md` states that to operators. Nothing executed it.

**Why that mattered, as a number rather than an argument.** Mutating the gate to *also* refuse the four ungated commands, the pre-existing suite ran:

```
41 passed, 5 deselected
```

**Every existing refusal assertion is satisfied by a gate that breaks the documented promise.** The suite pinned that refused things are refused; it never pinned that permitted things are still permitted. That is exactly the failure "no over-gating" distinguishes, and it is why this belonged in the mission rather than in a follow-up.

New class `TestUS7NoOverGatingUngatedCommandsSurviveRefusal` — 5 cells in `tests/sync/tracker/test_tracker_egress_refusal_3108.py`, the literal converse of `TestUS4TotalityAcrossThreeEntryPoints` which lives in the same file. Fixture permits Channel 1 (`project.uuid` + `sync.enabled`) and refuses only `tracker.egress`, so the tracker key is the sole refuser. Every cell first runs `_assert_gate_is_live`: the *same* root must refuse `sync pull` with the verbatim Channel-2 string, so no cell can pass because the gate was inert.

- **status** — exit 0, and `issue_count == 1` reads the seeded sentinel back *through* status.
- **map list** — exit 0, payload `== {"mappings": [], "pending_binding_upgrade": None}`.
- **map add → map list roundtrip** — the write is read back through a *separate* command, not trusted from `map add`'s echo.
- **unbind** — exit 0, then asserted from `load_tracker_config(root).provider is None` rather than from `status`, because post-unbind `status` exits 1 at `readiness=missing_auth` (no local binding → no `_is_local_binding()` short-circuit). That is `#3172`'s pre-flight, not this gate.
- **counter cell** — delegating (never stubbed) counters on all three module bindings of `tracker_egress_verdict`: all four ungated commands leave every counter at **0**; `sync pull` on a fresh root then moves `local_service`'s to exactly **1**. The census, executed.

Red-first: all five fail under the mutation with `assert 1 == 0` and the verbatim Channel-2 refusal. Reverted byte-identically (`cmp` IDENTICAL, `sha256 583a0c84…` unchanged, `git diff` empty).

**`bind` is excluded, and measured rather than assumed:** `bind_command` calls `_check_readiness` directly with no `_is_local_binding()` short-circuit, so it exits 1 at `readiness=missing_auth` before mission code — `#3172`'s class, out of scope here. Ledgered **R-6**.

**Census re-derived after the Bundle B port** (predicate: `ast.Call` whose `func` resolves to `tracker_egress_verdict`, excluding the `def`, the imports, and ~14 docstring mentions a grep would have counted): still **6 call expressions across 5 enclosing functions** — `sync.py:1968`/`:1973`, `local_service.py:212`/`:235`/`:253`, `saas_client.py:341`. Only the line numbers moved. None in the ungated commands.

### 2. `#3110` SC-004 clause 3 — a contract crossing, re-pinned and strengthened

`test_sc004_clause3_both_deciding_modules_bind_the_one_definition_site` failed with `AttributeError: module 'specify_cli.tracker.saas_client' has no attribute 'project_egress_refusal'`. That is **caused by this PR** and it is a genuine contract crossing, not a stale test.

The delegation chain is now `tracker/saas_client.py::_request` → `tracker_egress_verdict` → `_resolve_channel1:383` → `specify_cli.egress.project_egress_refusal`, and `egress_verdict.py:143` states it in the product: *"This module holds the Mission's only module-level import of `project_egress_refusal`."* So `tracker.saas_client` is a **caller of the one decider**, not a decider. Restoring a direct binding there would be the wrong fix — it would give the hosted path two Channel-1 decisions and a second composer in front of FR-016's byte-identical refusal string.

Clause 3 hardcoded its module set as two import aliases at the file top; it derived nothing. It now **derives** the binder set by AST over `src/` (module-level `ImportFrom` only — the sole form that can produce a stale attribute), checks it against a pinned expectation, then applies the unchanged `is` comparison to every derived binder. Dated rationale recorded in the house convention.

Two mutants, both reverted byte-identically:
- **MUT-A**, the partial-consolidation shape → `assert <function project_egress_refusal at 0x…6156c0> is <function project_egress_refusal at 0x…b21580>` — only the `is` comparison sees it.
- **MUT-B**, a third binder appears → `Extra items in the left set: 'specify_cli.tracker.local_service'` — **what the hardcoded version could not catch at all.**

### `integration-tests-sync` — pre-existing, verified at the merge base

4 errors, 0 failures, all FR-007 leak-guard *teardown* errors leaking `[E26] sync.runtime._runtime` / `[E27] sync.background._service`. The merge base `709a59534` (run `31128128294`, job `92707752017`) fails with the **same four node-ids**: `209 passed, 4 errors` there vs `366 passed, 4 errors` here — the delta is this mission's four new tracker files being selected, not new breakage. This PR touches zero `src/specify_cli/sync/` files. Ledgered **R-7**.

### Gates

| Check | Result |
|---|---|
| `test_tracker_egress_refusal_3108.py` | **46 passed** (41 + 5 new) |
| `test_egress_consolidation_3110.py` | **13 passed** (was 1 failed / 12 passed) |
| `tests/sync/tracker/` | **709 passed** |
| `ruff check` (both files) | All checks passed |
| `mypy --strict` | 4 errors, **identical to the HEAD baseline** measured on a `git show HEAD:` copy — **zero new** |

`tests/architectural`: 7 failed / 1755 passed, all classified and none from this diff — 1× `test_routed_load_meta_floor` (pre-existing on main, cleared by #3247), 4× `test_tid251_enforcement` (`No module named ruff` — a local venv artifact, not CI), 2× `test_wp_prompt_build_latency` (**pass 2/2 in isolation at `-n0`**; a contention artifact of the local `-n auto` run).

One housekeeping note: residual entries **R-6** and **R-7** both landed in the second commit; R-6 logically belongs with the first. Not worth rewriting history for.

robertDouglass:

Three independently profile-loaded lenses reviewed the aggregate: `reviewer-renata`, `debugger-debbie`, and `architect-alphonso`.

**Verdict: `BLOCK_MATERIAL`.** The examined payload/subprocess chokepoints fail closed and focused controls are strong, but the broader zero-egress/one-authority/no-overgating contracts are not satisfied. Material issues were not fixed.

- **High — `src/specify_cli/cli/commands/tracker.py:1053` (also push/run/publish):** authenticated SaaS sync runs networked readiness before service-level consent. Reproduced a `urllib` reachability `HEAD /` before `TrackerEgressRefusedError`. This violates the mission NFR that refusal precedes any HTTP attempt.
- **High — `src/specify_cli/tracker/egress_verdict.py:377-433`:** Channel 1 is enforced through canonical `project_egress_refusal`, then independently re-resolved through sync internals for state/remedies. One verdict can combine enforcement from one authority evaluation with diagnostics from another.
- **High — `docs/migrations/tracker-egress-refusal.md:199`:** docs/#3174 closure promise local `tracker bind` remains working, but `bind_command` performs hosted readiness first and the no-overgating suite explicitly excludes bind after observing `readiness=missing_auth`.
- **Medium — `tests/sync/tracker/test_tracker_egress_refusal_3108.py:41`:** hosted tests bypass CLI and patch only `httpx`, so they cannot observe the pre-gate `urllib` request.
- **Medium — architecture:** local enforcement is repeated at three service methods and guarded by exact syntax/call census, rather than owned at one local transport boundary; identifier disclosure remains duplicated caller-supplied prose.

**Evidence:** focused verdict/config/architectural suites pass on the rebased tree (`191 passed`); additional focused acceptance controls passed. No confidential payload was observed in the pre-refusal HEAD, and later tracker HTTP/subprocess transmissions were blocked. CI reds inspected on the old head were unrelated/pre-existing; they were not used as correctness evidence.

**Maintenance only:** rebased conflict-free onto current `main`, fixed five pre-existing touched-file strict-mypy `Any` returns with boundary casts, and rerolled `e873f20ce → 23dec9963`. Verification: ruff clean, strict mypy clean across eight source files, `191 passed`, diff check clean. No material finding was modified.

stijn-dejongh:
## Landing triage (maintainer): **HOLD — not landable as-is; remediate the 3 HIGHs or rescope first**

Picking this up for the current landing pass. Before investing in remediation I ran viability triage, because the PR is a week old and carries a fresh squad block.

**Verdict: not clearly landable.** @robertDouglass's 3-lens adversarial squad returned `BLOCK_MATERIAL` yesterday (2026-08-09) and then did *maintenance only* — rebase + strict-mypy boundary casts — with the explicit note *"No material finding was modified."* The three HIGH findings are violations of this mission's own NFRs and still stand on the current head (`23dec9963`).

### Triage axes

- **Superseded? No.** `tracker_egress_verdict` / `egress_verdict.py` do not exist on `upstream/main`; #3108 (P1) and #3174 are both still open. The two-channel consent model and the local `beads`/`fp` subprocess gate are this PR's unique contribution — #3030 gated only the hosted path.
- **Compatible? Structurally, yes.** The branch rebased conflict-free onto current `main` and the port composes Bundle B's `project_egress_refusal` mechanism rather than moving a seam.
- **Still needed? Yes.** The local subprocess egress path is ungated on `main` today; the collection surface #3108 names is real, and the focused subprocess/payload controls in this PR do fail closed.

### Why it's a HOLD, not a LAND

The value is real, but the squad's three HIGHs are core to the PR's promise, not cosmetic:

1. **HTTP before refusal** (`tracker.py:1053`, also push/run/publish): networked readiness (`urllib HEAD /`) runs *before* the consent gate, so a refused hosted sync still touches the tracker host. That's a leak on the exact surface this PR claims to seal — its stated NFR is "refusal precedes any HTTP attempt."
2. **Two-authority verdict** (`egress_verdict.py:377-433`): enforcement comes from canonical `project_egress_refusal`, but state/remedies are re-resolved through sync internals — one verdict combines enforcement from one authority with diagnostics from another, against the one-authority contract.
3. **`bind` over-gating vs. docs** (`docs/migrations/tracker-egress-refusal.md:199`): the #3174 closure promises local `tracker bind` keeps working, but `bind_command` runs hosted readiness first and the no-over-gating suite explicitly excludes `bind` after observing `readiness=missing_auth`.

Fixing these is heavy, security-sensitive work on the contributor's design — reordering the gate ahead of network readiness across 3–4 commands, unifying the verdict authority, and reconciling `bind`/docs. These are not one-line folds.

### Recommendation

**RESCOPE or remediate-then-land — operator's call.** Holding for direction before any heavy remediation; I have not touched the branch. The current CI reds (`integration-tests-cli`, `regression tests (blocking)`, `quality-gate` aggregator) are consistent with the contributor's prior pre-existing-main classification and are secondary to the squad block.

stijn-dejongh:
**Maintainer landing notification.** Per operator direction, I'm taking #3135 into the landing pass on the *remediate-then-land* track: I'll fold fixes for @robertDouglass's three HIGH squad findings (HTTP-before-refusal, two-authority verdict, `bind` over-gating vs. docs), re-verify, refresh the PR body + changelog, and hand off green for the operator to merge. Working in an isolated worktree off the current head (`23dec9963`); maintainer folds will appear on the fork branch `bundle-c-tracker-refusal-3108`. Evidence to follow.

stijn-dejongh:
## Landing remediation summary — 6 folds pushed, squad HIGHs resolved

Head moved `23dec9963` → `dcb5d41e6`. Worked in an isolated worktree off the PR head; no rebase needed (already current on `upstream/main`).

**Verdict: merge-ready on green.** All three of @robertDouglass's HIGH squad findings are resolved — two folded, one adjudicated to a filed follow-up with proof it is not a security defect. PR body rewritten BLUF; changelog + migration note updated.

### Folds (each its own single-purpose commit, commitlint-clean)

| Commit | Fold | Why |
|---|---|---|
| `bef00995c` | `fix(landing)`: refuse hosted egress before the readiness network probe | **HIGH-1.** `_check_sync_readiness` ran `evaluate_readiness(probe_reachability=True)` — a real `urllib` HEAD to the tracker host — before the consent gate. Now consults `tracker_egress_verdict(HOSTED_SERVICE)` first, so refusal precedes any HTTP. Re-pins G4's call-site census (5→6 enclosing / 6→7 expressions) with dated rationale for the new audited chokepoint. |
| `9cf58c102` | `fix(landing)`: let local `bind` proceed without hosted readiness | **HIGH-3.** `bind_command` ran hosted readiness unconditionally, so an unauthenticated project's local `beads`/`fp` bind exited `missing_auth` — contradicting the migration note. Now skips hosted readiness for local `--provider`; SaaS binds unchanged. Migration note made explicit. |
| `df30af21c` | `test(landing)`: pin that the reporting split never flips enforcement | **HIGH-2 defense-in-depth.** Pins that `_classify_channel1`'s diagnostic re-resolution can never change `refused`/`refusing_channels`. |
| `e0ff3e0f1` | `test(landing)`: stub `tracker_egress_verdict` in mock-boundary CLI tests | HIGH-1 fallout: the new CLI gate correctly refused 10 pre-existing mock-boundary tests that mock `_service()` and carry no consent. Extended their autouse fixture (egress is out of scope there — it has its own suite). |
| `809940947` | `test(landing)`: join the `TestZeroBlastRadius` census to the HIGH-1 gate | Second call-site census (separate file) also pinned `tracker.py: 0` — re-pinned to `1`, kept in step with G4. |
| `dcb5d41e6` | `docs(landing)`: changelog local-command survival + pre-probe refusal; fix MD049 | User-facing notes for HIGH-1/HIGH-3; fixes two pre-existing MD049 emphasis violations in the PR's own changelog entry that the markdownlint gate flags. |

### Squad adjudication

- **HIGH-1 / HIGH-3** — folded, red-first proven (probe-fires-then-refuses → no-probe; `missing_auth` exit → success). 
- **HIGH-2** — **not a security defect.** Enforcement (`refused`, `refusing_channels`) is computed solely from the single authority `project_egress_refusal`; only the diagnostic *label* re-resolves through sync internals, and I verified at every verdict write site (`egress_verdict.py:692,709,728,752`) that the split can never flip the enforced decision. The honest structural fix needs the Bundle-B resolver contract to carry a decision value (the module's own documented retirement condition), which is out of scope for a local-gap-closure PR. Filed as **#3287**, parented under #1800.
- MEDIUM findings (test patches httpx not urllib; local enforcement repeated at three methods) — subsumed: HIGH-1's fix moves the gate ahead of the network at the CLI boundary; the three-method placement is guarded by the G-series census this PR keeps honest.

### Local gates on the folded tip

```
ruff (changed files)            All checks passed
mypy src/.../tracker.py         Success: no issues found
commitlint upstream/main..HEAD  clean (all 6 folds + contributor commits)
egress suites + guards          green modulo 1 pre-existing baseline red (below)
mock-boundary CLI tests         64 passed
terminology guard               10 passed
docs-freshness --ci             errors=0 (external-URL link WARNINGS only)
markdownlint (changed docs)     0 errors
```

### Pre-existing reds, classified (not this PR)

- The three CI reds carried over from before the folds — `integration-tests-cli`, `regression tests (blocking)`, `quality-gate` (aggregator) — all reproduce on **main's own CI Quality at the exact merge-base `28ae75ea9`** (run `31329585486`, which is redder still: also `fast-tests-doctrine`/`fast-tests-charter`, path-filtered off this PR). Zero PR defects; matches the earlier classification and the squad's note that the old-head reds were unrelated/pre-existing.
- `test_tracker_egress_verdict_3108.py::TestModuleShape::test_module_imports_cleanly_without_specify_cli_sync` — pre-existing baseline red (bare-subprocess `ModuleNotFoundError`, env artifact), reproduced identically on the base `23dec9963`.
- I will re-classify after CI reruns on `dcb5d41e6`, since the CLI-touching folds may un-skip a shard.

### State

Folds pushed to `bundle-c-tracker-refusal-3108` via `--force-with-lease`. **Watching CI; merge-ready on green. Draft/hand-off for the operator — I do not merge.**

Contributor note (@MOES-Media): the canonical changelog is `docs/changelog/CHANGELOG.md` (root `CHANGELOG.md` is a symlink); your egress entry had two MD049 emphasis violations that the markdownlint gate flags on the changed file — fixed in `dcb5d41e6`.

stijn-dejongh:
## Post-fold CI re-classification (head `dcb5d41e6`, CI Quality run `31352719410`)

Re-ran the classification now that CI has completed on the folded tip. **The folds added zero new red** — the failed-*job* set is byte-identical to the pre-fold base run, and every shard the folds touch (`fast-tests-cli`, `fast-tests-sync`, `fast-tests-status`, `integration-tests-core-misc`, `arch-adversarial`) is **green**.

| Failing check | Attribution |
|---|---|
| `integration-tests-cli` | **Pre-existing, not tracker-related.** Same test, assertion, and counts on base `23dec9963` and folded `dcb5d41e6`: `test_agent_mission_commit_to_branch.py::test_commit_to_branch_treats_empty_safe_commit_shape_as_unchanged_when_dirty` → `RuntimeError: safe_commit: git commit failed` (`1 failed, 1057 passed` both sides). This is the `commit_for_mission`/`safe_commit` subsystem — the #3269 area in main's own tip — a CI-sandbox git-commit failure. Zero `src/specify_cli/tracker/` or `sync/` lines in this diff reach it. |
| `regression tests (blocking)` | **Pre-existing.** Red on main's CI Quality at the merge-base `28ae75ea9` (run `31329585486`); the deliberate red-first P0 reproductions (ADR 2026-07-17-1). |
| `quality-gate` | **Derived** — pure aggregator over the two above. |

All three reproduce on **main's own CI Quality at the exact merge-base** `28ae75ea9`, which is in fact redder (also `fast-tests-doctrine`/`fast-tests-charter`, path-filtered off this PR). **Zero PR defects; zero contract crossings beyond the two censuses re-pinned in-PR.**

**State: green-modulo-honest-reds. Merge-ready. Hand-off for the operator — I do not merge.**

stijn-dejongh:
## Independent adversarial re-review — 3 lenses, blind to the landing work → 3/3 ACCEPT

A fresh squad reviewed the head **cold**: three profile-loaded lenses (`reviewer-renata`, `debugger-debbie`, `architect-alphonso`), each briefed on the mission's own claims and the code only — not on the folds, the prior review, or any of this thread (they were told not to read it, to avoid anchoring). Each reached its own verdict.

**Result: zero BLOCK_MATERIAL, zero MAJOR. All three ACCEPT.**

- **reviewer-renata (contract↔code + test quality):** took all six mission claims as hypotheses to refute and refuted none; judged the tests non-vacuous — "I did not find a test that would pass against a broken implementation." Independently verified the real subprocess-recorder + `httpx` trip-wire with positive controls.
- **debugger-debbie (hands-on leak-hunt):** *no leak reproduced.* With live trip-wires + positive controls: a refusing local project spawns **zero connector/network**; a refusing hosted project makes **zero httpx/socket before refusal**; the CLI pre-flight refuses with **zero network** ahead of the reachability probe; every malformed/unreadable/`None` input **fails closed** (refuse, never permit, never raise); ungated commands keep working on a refusing project.
- **architect-alphonso (single-authority / seam):** independently traced every `refused=`/`refusing_channels=` write site and confirmed enforcement is single-authority; destination polarity is enforced structurally by the `_JOIN` table + G5/G6; the re-pinned censuses are non-vacuous and not weakened.

### One more fold from the squad (pushed: `0b6594b84`)

`fix(landing)`: **thread one resolved repo root through the sync egress pre-flight and transport.** alphonso flagged that `_check_sync_readiness` (the HIGH-1 pre-flight) and `SaaSTrackerClient._request` (the transport) each resolved `require_repo_root()` independently — additive-safe today, but a caller whose cwd ≠ the data-owning root could have the two hosted gates answer for **different projects**. The four sync commands now resolve the root once and thread it into both `_check_sync_readiness(root=...)` and `_service(root=...)`. Red-first: `test_sync_pull_pre_flight_gate_and_transport_share_one_resolved_root` — a stateful `require_repo_root()` makes the pre-fix code judge project `p0` at the gate and `p1` at the transport (assertion fails); the fix makes both `p0`. Guards/censuses unchanged (no new call site); ruff/mypy/commitlint clean.

### Residuals filed (none blocking)

- **#3287** — the two-authority *diagnostics* contract (enforcement is already single-authority; needs the Bundle-B resolver-contract change to retire the classifier's second read).
- **#3291** — three squad NOTEs: harden the environment-fragile import-isolation test (a category-3 local false-red, green on CI), trim redundant per-call git-identity resolution, and silence a caught `PermissionError` traceback on unreadable-config paths.

**State: 7 folds total, independent squad clean, CI re-validating on `0b6594b84`. Merge-ready on green. Hand-off for the operator — I do not merge.**

stijn-dejongh:
## Two more folds — #3291 residuals 1 & 3 (head `c31c0d85d`)

- `d8ed2398a` `fix(landing)`: **carry an unreadable `.kittify` directory as a fault instead of raising.** `_read_project_local` promises "never raises," but its `is_file()` existence probe sat outside the try/except, so an unreadable enclosing directory raised `PermissionError` straight out — the verdict still refused correctly, but a full traceback printed to stderr. Now carried as the same `unreadable` fault as an unopenable file. Red-first (`test_a_project_config_in_an_unreadable_directory_is_unreadable`). Also a boundary cast on a pre-existing strict-mypy `Any`-return (present verbatim on `upstream/main`) that this file-touch pulls into the changed-file gate — same treatment as the other touched files.
- `c31c0d85d` `test(landing)`: **make the sync import-isolation probe skip a package-less interpreter.** The `subprocess.run([sys.executable, …])` probe false-red under `uv run` (bare pyenv interpreter, no editable install). It now runs a positive-control preflight: skip when the interpreter has no install, fail only on a genuine `specify_cli.sync` import edge. Green on CI's venv either way.

**Item 2 (redundant per-verdict git-identity resolution) deliberately not folded** — its clean fix is the two-authority consolidation in **#3287** (enforcement + reporting each resolve checkout routing → ~6 git subprocesses/verdict); a standalone process-wide cache would risk daemon staleness. Kept in **#3291**.

9 folds total. Local gates clean (ruff/mypy/commitlint/targeted suites). CI re-validating on `c31c0d85d`. **Merge-ready on green — I don't merge.**

stijn-dejongh:
## Final landing verdict — merge-ready, green-modulo-honest-reds

**Head `c31c0d85d`.** Handing off for the operator to merge; I do not merge.

### Verdict

Landable. The mission closes a real, previously-ungated egress surface; all three of @robertDouglass's HIGH squad findings are resolved, an independent second squad confirmed the result blind, and CI on the final head carries only pre-existing reds.

### Folds — 9, each single-purpose `<type>(landing):`

| Commit | Fold |
|---|---|
| `bef00995c` | HIGH-1: refuse hosted egress before the readiness network probe (+ G4 census re-pin) |
| `9cf58c102` | HIGH-3: local `bind` proceeds without hosted readiness |
| `df30af21c` | HIGH-2: pin that the reporting split never flips enforcement |
| `e0ff3e0f1` | stub the gate in 10 mock-boundary tests it correctly refused |
| `809940947` | join the `TestZeroBlastRadius` census to the HIGH-1 call site |
| `dcb5d41e6` | changelog: local-command survival + pre-probe refusal; fix 2 pre-existing MD049 |
| `0b6594b84` | thread one resolved repo root through the sync pre-flight + transport |
| `d8ed2398a` | carry an unreadable `.kittify` dir as a fault instead of raising (+ boundary cast) |
| `c31c0d85d` | make the import-isolation probe skip a package-less interpreter |

### Two independent squads

- **Original (@robertDouglass, 3 lenses):** `BLOCK_MATERIAL`, 3 HIGH — all resolved (2 folded, 1 adjudicated single-authority + tracked as #3287).
- **Independent re-review (3 fresh lenses, blind to the folds, code + spec only):** **3/3 ACCEPT — zero BLOCK_MATERIAL, zero MAJOR.** No leak reproduced; every malformed/unreadable/`None` input fails closed; enforcement confirmed single-authority by trace.

### CI on the final head (`c31c0d85d`, run `31355897212`)

Failed jobs: `integration-tests-cli`, `regression tests (blocking)`, `quality-gate` (aggregator) — **byte-identical to the merge-base set**, all reproduce on main's CI-Quality at `28ae75ea9`. `integration-tests-cli` is a pre-existing `safe_commit` git-sandbox failure in the commit subsystem, not this diff. **Every shard touching the 9 folds is green** (`fast-tests-cli`, `fast-tests-sync`, `integration-tests-sync`, `integration-tests-agent`, all `arch-adversarial`). **Zero PR defects, zero new red.**

### Local gates

`ruff` / `mypy --strict` / `commitlint` clean across the fold range; targeted egress + guard + consent suites green; terminology, docs-freshness (`errors=0`), markdownlint clean.

### History

27 commits: 18 contributor commits (code-by-slice, governance dossier bunched — no wip/fixup/revert noise) + 9 single-purpose landing folds. Linear, bisectable, each commit revertible on its own.

### Filed follow-ups

**#3287** (two-authority diagnostics contract) · **#3290** (`runtime_bridge → sync` coupling, assigned) · **#3291** (residual nits) — plus the contributor's #3167/#3168/#3169/#3170/#3172.

### Landing-order constraints

None. This PR stands alone; no structural cutover rides it.

**Draft/hand-off for the operator. Not merging.**

robertDouglass:
Post-merge reconciliation from the #3262 consent mission:

- The two substantive red checks visible on this PR were not tracker-consent regressions. `test_mission_create_json_strict_when_sync_skips_ingress` was retired/redesigned by `eeda3df5f`; `test_commit_to_branch_treats_empty_safe_commit_shape_as_unchanged_when_dirty` was corrected by `47cc8f3b2` because hook rejection is an error, not an empty changeset. Both commits are ancestors of current `main`.
- Because #3135 was merged externally at `5d49d31ed`, it cannot now be "refreshed before merge" as originally planned. The remaining semantic reconciliation is owned by #3262 / draft PR #3293 WP07: tracker-hosted calls must consume the final per-project context, stable native identity, durable WP06 attempt/lease/final gate, and correlated terminal refusal. Tracker Channel 2 remains narrowing-only and may never grant hosted sync.
- This merged PR is therefore historical context, not the active repair branch. No attempt is being made to reopen or amend its merged head.

Issue #47, opened 2025-12-07
Could not start dashboard (vscode, windows)


lein1013:
after initialising the dashboard can not be startd

windows, vscode

```
spec-kitty init ccam --ai copilot 
...
spec-kitty dashboard 
...
Warning: Could not start dashboard: module 'signal' has no attribute
'SIGKILL'
```

PR #44, opened 2025-11-26
in acceptance_support.py, manage "Unexpected UTF-8 BOM"

sdementen:

Issue #26, opened 2025-11-13
"SyntaxError: '{' was never closed" when running `.kittify\scripts\powershell\Set-TaskStatus.ps1`


sdementen:
When running the script `.\.kittify\scripts\powershell\Set-TaskStatus.ps1 -TaskId WP01 -Status done` I get the error message:
```
WARNING: [spec-kitty] python.exe :   File "<string>", line 8
At ...\.worktrees\003-...\.kittify\scripts\powershell\common.ps1:140 char:15       
+     $output = & $python.Path -c $script $RepoRoot 2>&1
+               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (  File "<string>", line 8:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError

    print(json.dumps({error: fUnable
                     ^
SyntaxError: '{' was never closed
```
It seems it is due to using " instead of ' in the python code https://github.com/Priivacy-ai/spec-kitty/blob/main/scripts/powershell/common.ps1#L107C1-L137C27. The " in the python code may conflict with the " from powershell. Replacing " by ' in the python code solves the issue

Issue #22, opened 2025-11-10
Community Question/Discussion: Should .kittify be added to .gitignore by default?


robertDouglass:
One thing that I noticed when running `spec-kitty init` on an existing project is that it creates a lot of files which fall under version control by default. Furthermore, upgrades to spec-kitty cause further changes, particularly in .kittify. I'm rolling a patch right now that will prevent any of the agentic folders (eg .codex, .claude, .opencode) from being tracked by version control by default. My question is whether people would rather have .kittify tracked, or not? 

PROS:
.kittify is your project history... all of the research, specs, planning, tasks, and the history of the kanban workflow, as well as the quickstart, contracts etc. This is VALUABLE project information. Keeping it would make a lot of sense.

CONS:
You'll need to commit changes to git and that creates a blurry boundary between the software you're writing that the tools you're writing it with. Plus, you'll have to manually be responsible for whether or not the LLMs have written any sensitive information into .kittify. They're not meant to, per se, but it is the internal dialog between you, your LLMs, and their own musings, which can be quite verbose, and manually auditing it all is cumbersome and not always otherwise necessary. 

I'm leaning towards including .kittify in .gitignore by default. 

What do you think?

sdementen:
kitty-specs are indeed key to keep in the repo (very nice history of the project development)

for .kittify, it would nice to pin the version of spec-kitty-cli (and maybe name the python package just spec-kitty under PyPI). Updates would be done via the python package manager.
Before any new AI session, ensure that the version of spec-kitty used during "spec-kitty init" hasn't changed and if so update the .kittify folder.

sdementen:
from my learning curve feedback, it would have been clearer to me if the spec-kitty specific folders would be mainly stored under a /kitty-specs/config folder alongside the features folders) without the need of the folder /.kittfy.

For instance (assuming python project with github copilot):
- /.github/prompts ==> no changes
- /.github/copilot-instructions.md ==> make it an agent prompt referring to a /kitty-specs/config/agent-instructions.md
- /.kittify/memory/*, /.kittify/active-mission, /.kittify/.dashboard and similar with project specific info ==> move into /kitty-specs/config/
- specifying which python virtual env to use for running kitty-spec command and/or running tests/etc ==> create into /kitty-specs/config/tools (or something a bit cross language)
[...]

PR #3152, opened 2026-08-02
feat(conformance): executable conformance suite for doctrine, profiles, skills and cross-layer composition

MOES-Media:
Adds an executable conformance suite that verifies Spec Kitty's own doctrine, agent profiles, skills and cross-layer composition against a real model, using [muster](https://github.com/garrison-hq/muster) as the grader.

Five missions, developed and reviewed on `MOES-Media/spec-kitty`, replayed here onto current `main` as one reviewable branch. Purely additive: `conformance/` does not exist upstream, and exactly one existing file is modified.

## What this is for

Spec Kitty ships doctrine, agent profiles and skills as artifacts. Nothing currently checks that a real model, handed a projected agent body, actually behaves the way the profile says it will. This suite does that: it runs manifests through muster, grades replies against rubrics quoted verbatim from muster's rubric doc, and ships discrimination controls that must fail so a green run means something.

| Suite | What it checks |
|---|---|
| `conformance/skills/` | Skill manifest completeness; trigger-routing behaviour across query families |
| `conformance/doctrine/` | 13 directive manifests, lint-graded, with a drift gate against the source directives |
| `conformance/crosslayer/` | Persona/SOP/skill composition; RFC-1 persona conformance |
| `conformance/behavioral/` | Five agent profiles x four doctrine axes, judge-graded against a live model |

## Commits

Sliced so each mission is reviewable on its own. Governance dossiers are bunched first, then one commit per mission, then CI wiring, then two commits adapting to a `main` change that landed while this was in flight.

```
e34a40c7e chore(conformance): spec-kitty mission governance dossiers
7e5bf1dee feat(conformance): static skills-manifest completeness gate
43a0447bf feat(conformance): doctrine rule manifests and drift gate
cc9681726 feat(conformance): cross-layer persona, SOP and skill composition suite
6c505c96b feat(conformance): skill trigger-routing behavioral suite
278b338d2 feat(conformance): judge-graded doctrine behavioral suite
412c025c9 ci(conformance): route conformance suites into ci-quality path filters
ba1ad5c1b fix(conformance): resolve built-in doctrine at the packs/built-in seam
553c78579 docs(conformance): repoint built-in profile path labels at packs/built-in
```

The single modification to an existing file is `.github/workflows/ci-quality.yml`: three insertions routing `crosslayer.yml` into the `pull_request`/`push` path filters and the `changes` job's `e2e` group. Nothing else in that file is touched.

## What the suite found in this repository

This is the part worth reviewing first, because it is the point of the exercise.

### Projected agent bodies silently drop three declared fields

`_render_helpers.py::_body_lines()` emits name, purpose, and `## Specialization` with `primary_focus` and `avoidance_boundary`, then stops. `collaboration.handoff-to`, `collaboration.canonical-verbs` and `capabilities` are parsed from the source `.agent.yaml` and never reach the deployed agent. `specialization.secondary-awareness` and `success-definition` are also unrendered, and no test pins the body's contents.

Measured, five profiles, `gpt-4o-mini`, k=5 runs per case, zero errored runs:

| Axis | Graded field | Reaches the agent? | Result |
|---|---|---|---|
| avoidance-boundary | `avoidance-boundary` | yes, verbatim | 24/25 |
| capabilities | `capabilities` | no (rubric permissive by design) | 24/25 |
| handoff-to | `handoff-to` | **no** | **0/25** |
| canonical-verbs | `canonical-verbs` | **no** | 16/25 |

The pass rate tracks whether the graded field reached the agent. `canonical-verbs` makes it sharp: `implementer-ivan`'s verbs are `implement, fix, refactor, test, debug`, words a model would use unprompted, and it scores 5/5. `debugger-debbie`'s are `investigate, falsify, bisect, diff, trace, converge`, and it scores 0/5.

The rubric is not at fault. Replaying muster's judge call byte-for-byte against hand-written replies, at both order-swap positions: an ideal reply passes 10/10, a bad reply fails 10/10, and a reply that names the correct role but also narrates the downstream lifecycle fails 10/10. A causal delta test appending the roster to the body moved 1/15 to 5/15.

An agent running inside Spec Kitty can confirm this in its own system prompt, which is the projected body and contains no handoff roster.

Filed as MOES-Media/spec-kitty#47. Not fixed here: this PR adds the measurement, it does not change the renderer.

### Two further defects, filed not fixed

- **MOES-Media/spec-kitty#46** — `accept`'s strict metadata gate requires an `agent` value that only the interactive `planned -> claimed` hop can write, so no orchestrator-driven mission can pass it. `--policy` reaches `policy_metadata`, but it is coerced into a fixed 8-field dataclass with no `agent` slot, so a supplied value is silently dropped. `--lenient` also disables path-convention validation, so the escape hatch relaxes two unrelated gates at once.
- **MOES-Media/spec-kitty#49** — mission lifecycle commands emit commit messages that fail this repository's own commitlint gate (`Accept <slug>`, `Record acceptance commit for <slug>`, `Finalize acceptance artifacts for <slug>`, `Add spec for <slug>`). `commitlint.config.cjs` already exempts some tool bookkeeping; these templates fall outside the pattern. Rebuilding history for this PR is what makes it commitlint-clean.

## Adapting to `packs/built-in`

`873832aa1` relocated built-in doctrine while this branch was in flight. The suite hardcoded the old paths, so three gates failed on the new base. Fixed in the last two commits.

Worth reviewing carefully, because "fix a path so a gate stops throwing ENOENT" can make a gate green by making its input set empty. Every gate reports its own input count, and each was re-proved against inputs it must still reject:

- **completeness guard**, 5 mutations, all exit 1: a dropped directive rule, a deleted manifest, a renamed `integrity_rules:` key, a probe attached to a directive rule (the documented interlock), and a wrong directive dir which throws rather than silently passing
- **drift gate**, 4 mutations, all exit 1: injected drift, a `STRUCTURAL_ABSENCE`, a de-drifted control (`discrimination control is dead`), and a glob pointed at a nonexistent directory
- **persona gate**, 4 mutations, all non-zero, including tampering the second loop iteration to prove it does not short-circuit, and tampering the source at the new location to prove it reads `packs/built-in/` rather than a stale copy

Three things were deliberately left on the old paths:

- `source.supporting:` permalinks are `blob/<sha>/` URLs pinned to commits predating the relocation, so the old path is correct at those SHAs. Rewriting them would 404 every citation.
- Recorded `muster sop run --json` evidence under `conformance/behavioral/evidence/raw-*/` is verbatim output from dated real runs. Rewriting it to look like it came from the new layout would be falsifying evidence.
- `kitty-specs/**` dossiers are historical records of what was authored at the time.

`028-search-tool-discipline.directive.yaml` did change content upstream, but the rewritten line is `procedures[4]` and the manifest quotes `integrity_rules[0..2]`, which is byte-identical. Checked across all 14 manifests: 49 ruleTexts, none quote a `procedures` bullet, so the rewrite is invisible to the suite. No requote was needed.

## Verification standard

Every check here has a recorded **rejection** run, not only a success run. A check with no rejection run was treated as unverified.

That standard exists because this work found **twenty verification commands that reported green while verifying nothing**, and every one was caught by running it against input it should reject rather than by reading it. Representative causes: `grep -qx '[PASS]'` matching a one-character line; a bare `|` in a BRE; `(.passThreshold // .k) == .k` reading true exactly when the field is missing; an empty glob leaving `ok=true`; and a script committed `100644` while invoked by path, which is `exit 126` in CI and which `bash script.sh` never reproduces.

Two were found only by executing the workflows under `act`; static analysis passed both files clean:

- The behavioral suite's `CONTROL-JUDGE-IMPOSSIBLE` was an **unfalsifiable conjunct** — an impossible rubric fails under a healthy judge, a degenerate model, and a judge stuck at FAIL alike. Since every rule in every profile manifest is judge-graded, a total judge outage would have been certified as genuine discrimination. Demonstrated by running the control manifest under a muster build whose known defect is a stuck-FAIL judge: the report was byte-identical on every field the guard reads, and the guard printed "genuine discrimination confirmed". Closed with a positive judge control that passes 3/3 on a healthy judge and 0/3 on a stuck one.
- The trigger-routing suite's control returned `passed: false, runsErrored: 0` for a permanently-triggering model, because muster computes `passed = shouldTriggerAxis.passed && nearMissAxis.passed` and the two axes have opposite predicates on the same rate. Byte-shape-identical to a healthy run. Closed by additionally asserting the near-miss axis passed.

## Known gaps, stated rather than buried

- **The cadence workflows have never run on GitHub Actions.** Both are `workflow_dispatch`-only, and dispatch only registers from the default branch, so they cannot run until this merges. Both pass `actionlint` and `shellcheck` and have been executed locally under `act` across live-credential, no-credential and dead-endpoint arms. One manual dispatch each after merge closes this.
- **`handoffDiscipline` will read 0/25 until #47 is fixed.** The axis is not unpassable; the measured per-run rate is roughly 5 to 10 percent, so 0/5 at threshold 3-of-5 is the likely draw rather than an impossibility.
- **`avoidanceBoundary` and `capabilityContainment` returned 5/5 for every profile** with no negative control for those two axes, so the run contains no internal evidence they can fail.
- **An inconclusive run can overwrite conclusive committed evidence.** The merge job is `if: always()`, so an endpoint-dead cadence run commits its artifact over the previous one. `runsErrored` is recorded per axis, but nothing prevents the overwrite.
- **The suite hardcodes this repository's internal layout**, which is why `873832aa1` broke it within a day. `src/doctrine/pack_paths.py::built_in_dir()` is the right seam, but it is Python-only while two of the three gates are Node and Bash, and the 14 `sopFile:` keys are resolved by muster relative to the manifest and cannot use a resolver at all. Adopting it needs either a printable CLI (`doctrine pack-root --kind directives`) or a generated-paths module with its own drift gate. Deliberately not attempted here.
- **`conformance/**` is not in ci-quality's trigger paths**, so a PR editing only a conformance script will not re-run the tests guarding it. Inherited from an existing sibling test rather than introduced here.

## Provenance

Full per-mission history, including every review cycle and the commit SHAs cited in the acceptance matrices, is preserved on `MOES-Media/spec-kitty` in PRs #22, #23, #26, #44 and #48. This branch is a clean replay of that work onto current `main`; the fork remains the audit trail.

Mission dossiers travel with the code under `kitty-specs/`, including specs, plans, tasks, work-package prompts, acceptance matrices and review cycles.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3152"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788263053&installation_model_id=427282&pr_number=3152&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3152&signature=39cbaa53b21b3f2483a5bd1e15328d0ab76f50fff03e744d41b1648049905037"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


robertDouglass:
## Adversarial PR audit — [#3152](https://github.com/Priivacy-ai/spec-kitty/pull/3152)

Method: three independent profile-loaded reviews in isolated worktrees, followed by arbiter inspection of the source, tests, branch topology, and current CI.

**Arbiter verdict: `DRAFT_NOT_REVIEWABLE`.** This large conformance programme is not yet admissible evidence: cadence workflows have not run in GitHub Actions, several controls lack negative controls, inconclusive results can overwrite conclusive evidence, `conformance/**` is absent from relevant CI triggers, and handoff discipline is recorded as 0/25.

**Next evidence:** add discriminating mutants/negative controls, monotonic evidence handling, owned CI triggers, and a real hosted run before architecture/security rereview.

robertDouglass:
Maintenance reroll only: the existing profile-loaded squad verdict remains `DRAFT_NOT_REVIEWABLE`; no material issue was addressed.

Rebased conflict-free onto current `main` and rerolled `553c78579 → e1304c444`. Ruff is clean across all changed Python/conformance test files. `git diff --check` remains non-clean because this draft intentionally includes a whitespace-only JSON negative-control fixture plus pre-existing trailing whitespace in mission/evidence artifacts; I did not rewrite test evidence or mission records as “maintenance.” Current CI will be the next mechanical signal.

robertDouglass:
Automated review — condensed after validation. No code was modified.

**Verdict: approve as an additive draft; close two evidence-integrity gaps before relying on the cadence as authoritative.**

- **Projected-artifact drift:** `conformance/behavioral/projected/*.md` is graded as generated evidence but has no source-to-projection drift check. Add a gate analogous to `check-persona-drift.sh`.
- **Evidence overwrite:** the commit job runs under `if: always()`; an errored/dead-endpoint run can replace previously conclusive evidence. Commit only when the run is conclusive and the discrimination guard passes.

The suite is valuable and already exposed a real renderer defect. These are durability controls, not objections to the added test surface.

MOES-Media:
## Rebased onto current `main` — `MERGEABLE`, zero attributable failures

Was `CONFLICTING`/`DIRTY`, **281 commits** behind. Now rebased, 12 commits, 279 files.

### The rebase itself was nearly clean

Despite the size, only **one** file overlapped upstream's changes: `.github/workflows/ci-quality.yml`. The conflict was purely additive — upstream's `corpus` path globs vs this branch's `crosslayer.yml` entry — so both were kept.

### Six gates repaired, each at the source it named

None was fixed by relaxing a gate.

1. **`cutover-guard`** — the two missions this PR *adds* carried event-log runtime evidence (68 and 56 events) with **no `status_phase` key at all**. `cutover_eligibility.py:19` is explicit: evidence without `status_phase == "1"` is un-cut-over, so the guard was right to refuse. Set to `"1"`, matching the three sibling missions this same PR adds. `is_cut_over()` now returns `True` with no reasons for all five.

2. **Contract round-trip** — five unmarked ` ```yaml ` blocks, each now carrying an explicit `# round-trip: skip:` naming what it is (four GitHub Actions workflow-step fragments, one manifest shape sketch with `<placeholder>` tokens). *Note for future readers: the gate collects **only** ` ```yaml ` fences — the ` ```sh ` invocation snippets earlier in the same files are not the blocks it reports.*

3. **`unit-contract-residual`** — `test_profile2soul.py` read the built-in profile from `src/doctrine/agent_profiles/built-in/`, a path deleted by the doctrine-pack layout move. The test died on `FileNotFoundError` rather than on anything about the projection it exists to check. Repointed at `packs/built-in/` — the same move this mission's own final commit made for the `PROJECTION.md` labels, but missed here.

4. **SOP doctrine conformance** — directive `042-common-docs` gained a fifth `integrity_rule` upstream (the `audience:` resolution rule); the manifest still had four. Added `042-r5` with `ruleText` verbatim. Checker now reports **13 manifests / 46 directive-derived rules OK**.

5. **Persona drift** — both committed `Soul.md` personas regenerated. Only the `source-hash` line moved; the projected bodies are byte-identical, so upstream's profile edits touched fields the projection does not carry. The gate was reporting genuine staleness, not a content change.

6. **Ratchet baseline** — self-inflicted and caught correctly: the five new skip markers grew `_SKIP_MARKED_BLOCKS` 13 → 18. Bumped in the same PR with a per-block rationale, per that gate's own stated policy, rather than a silent re-freeze.

Also converted one new `len(controls) == 1` to an identity assertion (`test_check_control_discrimination.py`) rather than annotate or re-freeze the golden-count ratchet: `len == 1` is satisfied by *any* single control, so a swapped fixture still read as well-formed — the exact failure that module exists to detect. Verified by construction: a wrong id reds 6 of its 8 tests.

### Verification

`tests/architectural`: **1401 passed**. Conformance gates — cutover-guard, contract round-trip, doctrine drift gate, doctrine manifest completeness, persona drift — all pass locally.

CI: 28 pass, 2 fail. Both — `quality-gate` and `regression tests (blocking)` — are **red on `main` itself** (since 2026-08-12); `quality-gate` cites only `regression-tests`. Neither is attributable.

LynnColeArt:
Fresh maintainer review at head `5a2af5b24` while sweeping PRs today. The 08-13 rebase resolved devin-ai's findings 1, 2 and 5, but five are verifiably still present, and two of them were Robert's explicit 08-11 merge conditions — so holding this one until a fixup lands. The concrete list:

**Must fix before merge:**
1. `conformance/behavioral/evidence/2026-08-02-01KYW5XK.json` — the aggregate declares `gpt-4o-mini`/`api.openai.com`/muster 1.2.2, but the raw reports it aggregates record `baseUrl: mock://test`, `model: mock`, and leak a `/home/jeroennouws/...` absolute path. Regenerate from a live run or remove the artifact — for a suite whose thesis is evidence integrity this one undercuts the PR itself.
2. `.github/workflows/crosslayer.yml:65,139` — pins muster `1.2.1`, the exact version `behavioral.yml:11-15` in this same PR declares has the stuck-at-FAIL judge defect. Consolidate to 1.2.2.
3. `behavioral.yml:635` — the evidence-merge job is `if: always()` with `contents: write`, so an inconclusive run overwrites conclusive committed evidence. This was Robert's 08-11 condition #2 (commit only when conclusive); his condition #1 (a drift gate verifying the `.sha256` sidecars for `conformance/behavioral/projected/*.md`) is also still absent.

**Fast-follow is fine:**
4. `behavioral.yml` — mutable `@v4` action tags on the one workflow holding `MUSTER_API_KEY` (siblings pin SHAs), and its credentialed jobs lack a `permissions:` block.
5. `crosslayer.yml:32-33` — the weekly cron burns live credentials against a manifest whose own banner says zero real cases exist; note the cron is inert until merge, so merging is what arms it.

Everything else looked genuinely strong — the discrimination-guard tests are real and mutation-verified, the workflow that runs on `pull_request` is secret-free and SHA-pinned, and nothing here conflicts with main's current direction. Fix 1-3 and I'll land it same-day.

robertDouglass:
## Adversarial squad review — 2026-09-12

Reviewed `5a2af5b2418766ad7e86fb131a860dad76e62dc9` with three independent, CLI-loaded profiles: Architect Alphonso (boundaries), Debugger Debbie (regression evidence), and Reviewer Renata (contracts/tests).

**Outcome: Substantive fixes needed; trivial portability/formatting fixed separately.**

- **[P1] Mandatory conformance gate references the retired skill root** — [conformance/scripts/check-manifest-completeness.mjs:133](https://github.com/spec-kitty/spec-kitty/blob/5a2af5b2418766ad7e86fb131a860dad76e62dc9/conformance/scripts/check-manifest-completeness.mjs#L133). The checker and conformance/skills/manifest.yaml still reference src/doctrine/skills. That tree exists at this old PR head but is removed from current main, whose canonical source is src/charter/offering/skills. The mandatory PR workflow will break on integration unless the suite is migrated. **Next step:** Rehome all manifest, projection, and script references together, regenerate the current inventory, and rerun the mandatory gates against the integrated tree.
- **[P2] Evidence validator accepts values that violate its schema** — [conformance/scripts/check-evidence-artifact-shape.mjs:72](https://github.com/spec-kitty/spec-kitty/blob/5a2af5b2418766ad7e86fb131a860dad76e62dc9/conformance/scripts/check-evidence-artifact-shape.mjs#L72). A live negative probe with timestamp=12, model=false, endpointHost={}, numeric case ID, string booleans, nonnumeric triggerRate and null threshold printed OK and exited 0. The handwritten check does not enforce the committed JSON schema types. **Next step:** Validate against the schema or enforce every field type and range; add negative mutation cases that must fail.

These require substantive implementation work and remain open; this review does not merge or approve the PR.

**Trivial fix pushed:** [94cb43ba9b51](https://github.com/spec-kitty/spec-kitty/commit/94cb43ba9b513ab2aa56d44403391aad87e3c67d) — Replace GNU-only grep -Pzoq in the control-discrimination guard test with a portable equivalent; format five newly added Python test files with identical ASTs.

**Verification**

- Portable control-discrimination guard: 38 checks passed after the fix; four boundary-discrimination probes passed.
- The 55-case static skills suite passed at the reviewed old head; integration with current main still requires the migration noted above.
- All formatting changes preserve Python ASTs; Ruff formatting and `git diff --check` passed.

**Context / remaining work**

- The 280-file PR received bounded sampling of runners, failure propagation, manifests and integration boundaries; not an exhaustive audit. No paid or credentialed model evaluation ran. Existing maintainer concerns about mock/live evidence provenance and evidence-publishing conditions remain outside these small fixes.

Scope: fixed-head review with targeted checks, not a full compliance/merge gate. External hosted services, native Windows, and browser rendering were not validated unless explicitly stated. Findings are linked to the reviewed commit; small follow-up commits are identified above.

robertDouglass:
Implemented and independently re-reviewed at `cf364b38c427355c0dd6dc77cc326586a64f6805`.

Integrated current main and canonical skill roots; repaired schema type/date/range validation, source/projection drift checks and conclusive-only evidence publication. Removed the aggregate falsely attributed to a live provider. The squad caught a further incomplete-inventory loophole; runner and publication validation now share one canonical suite catalog and require exact profile/axis/doctrine/rule identities.

Validation: Static skills: 56/56; inventory: 55 real skills plus control. Scoped suite: 122 passed, one stale SOP fixture then all 8 SOP tests passed after correction. New publication inventory tests: 18 passed; control guard: 38 passed. Fast baseline: 1787 passed, 5 skipped at the preceding implementation snapshot, with the final inventory delta covered by scoped tests. NOT MERGE-READY: Muster 1.2.2 static crosslayer still reports 0/2 on apparently compatible canonical clauses, and no conclusive paid/live acceptance evidence exists. No gate expectation or source instruction was weakened.

Three independent architecture, contract and regression lenses accepted the remediation; #3152 retains the acceptance blockers stated above. GitHub CI is being checked on the published revision. Optional SonarCloud PR-lookup failures are tracked separately in #4248. No mainline merge or release performed.

CI follow-up: canonically regenerated four stale mission snapshots from their existing event logs. All four projections match the reducer, event logs are unchanged, and the full 436-mission audit reports zero blockers. Three independent reviewers accepted this derived-data correction. The static crosslayer and live acceptance blockers above remain unresolved.

CI corpus follow-up: added one non-normative source-context heading to distinguish foreign NFR citations from this spec's local requirements. Every original word and ID is preserved; the nine declared FR/C IDs are unchanged. Exact failure reproduced red, then 50 owning checks passed. Three independent reviewers accepted final commit `cf364b38c427355c0dd6dc77cc326586a64f6805`; no detector, fixture baseline or tests changed. New-head CI is complete; the terminal result is recorded below. The Muster/static and live acceptance blockers above remain unresolved.

CI completion: All checks are now terminal on `cf364b38c427355c0dd6dc77cc326586a64f6805`: 60 passed, 14 skipped, and the single previously diagnosed Crosslayer static contradiction/drift gate failed (job 103542262624). The repaired corpus and snapshot checks passed. This PR remains NOT MERGE-READY: upstream Muster adjudication and conclusive live evaluation evidence are still required. Neither acceptance expectations nor canonical source instructions were weakened. CI monitoring is paused because no jobs remain; no merge or release performed.

robertDouglass:
Closing under the KISS audit Robert ordered on 2026-09-14. The branch stays; nothing is deleted.

**+51,666 / −1 across 286 files, and zero of them are under `src/`.** 118 files and 23,666 lines are `kitty-specs/` dossiers for six separate missions written to produce this suite. The flagship `behavioral.yml` is 739 lines gated `workflow_dispatch:` only, requiring Node 22, an external `npx @garrison-hq/muster@1.2.2`, three repository secrets and paid LLM calls — so it gates nothing today and costs money every time it fires. It grades our prompts against `gpt-4o-mini`, a model our users do not run. Its committed "live evidence" was found to be `baseURL: mock://test`, carrying a stranger's home-directory path, in a suite whose stated thesis is evidence integrity. It found one real bug in `_body_lines` and explicitly declined to fix it.

It has been in review for six weeks and carries `pr:needs-refresh`.

**Keep exactly one thing:** the roughly ten-line `_body_lines` fix. If someone opens that as its own PR it can land today. The other 51,656 lines should not be carried into a launch.

GOAL.md's binding line is "KISS is binding and we favor progress toward the goal over added ceremony." A conformance suite that cannot run, grading a model nobody uses, is the largest single violation of it in either repository.

PR #3166, opened 2026-08-03
feat(doctrine): ETag skip + Artifactory version for HTTPS fetch

### LINKED ISSUE #3165: [enhancement] doctrine fetch: ETag skip + Artifactory version property for HTTPS packs
## Problem or motivation

Org doctrine packs are often served from Artifactory as `-latest` HTTPS bundles. Today `spec-kitty doctrine fetch` always re-downloads and re-extracts even when the artifact is unchanged, and the only “version” recorded is the HTTP ETag (or nothing useful for operators). Teams need:

1. Fast, safe re-fetches that skip work when Artifactory content has not changed.
2. A human-readable doctrine **version** taken from Artifactory’s standard item `version` property after a successful download, shown in the CLI and stored in `pack-manifest.yaml`.
3. Non-Artifactory HTTPS sources to keep working without requiring or inventing a version.

## Proposed solution

Extend HTTPS doctrine fetch as follows:

- **Conditional fetch:** persist `etag` in `pack-manifest.yaml`; send `If-None-Match` on subsequent fetches. On HTTP `304`, leave the local snapshot **byte-for-byte unchanged** (including the manifest) and report `(unchanged)`.
- **JFrog version (Artifactory URLs only):** after a successful download of a URL under `/artifactory/`, call Storage API `…/api/storage/…?properties=version`, set `pack_version` from that property, print `Version: …`, and write it to the manifest. Fail the fetch if the property is missing/empty so an unversioned JFrog snapshot is never promoted.
- **Non-Artifactory HTTPS:** do not call the properties API; do not set `pack_version` (unless operator `ref` is set); continue using `etag` only for caching.
- **Migration:** if an existing Artifactory snapshot has `pack_version` equal to `etag` (legacy), force one unconditional download to pick up the real JFrog `version`, then resume 304 behaviour.

## Alternatives considered

- **Always re-download** — simplest, but slow and noisy for CI/dev loops against `-latest`.
- **Treat ETag as the doctrine version** — works for caching, but is not the operator-facing version Artifacts carry on JFrog.
- **Require a new `source_type: artifactory` / directory convention only** — useful later for `{name}-latest.{ext}` URL composition; not required to solve ETag skip + version property for existing full HTTPS URLs.
- **Refresh `fetched_at` on 304** — rejected; “unchanged” must mean no on-disk updates.

## Additional context

- Live Artifactory example:
  `GET …/artifactory/api/storage/raf-generic-local/doctrines/doctrine-rnd-latest.tar.gz?properties=version` → `"version": ["0.1.1"]`.
- Operator config today often lives in `~/.kittify/config.yaml` with full HTTPS artifact URLs + `subdir: pack`; project `.kittify/config.yaml` is what `doctrine fetch` reads when run inside a Spec Kitty repo (pack registry is project-scoped).
- Related surfaces: `HttpsBundleSource`, `write_snapshot`, `pack-manifest.yaml` (`etag`, `pack_version`), CLI `Version:` / `(unchanged)` lines.
- Tests: `tests/specify_cli/doctrine/test_sources.py`, `tests/specify_cli/doctrine/test_snapshot.py` (conditional 304, JFrog version property, non-Artifactory silent path).

wonderu:
## Summary
- Skip unchanged HTTPS doctrine pack downloads via `If-None-Match` / HTTP 304 without rewriting the local snapshot or manifest.
- Use explicit `source_type: artifactory` provenance for JFrog Storage API `version` metadata; persist that as `pack_version` while keeping the ETag separate. Generic HTTPS never promotes an ETag to a version.
- Reuse cached validators only when manifest source URL/type match the current configuration; legacy `pack_version` fallback is disabled for pinned refs and Artifactory migration fetches remain unconditional.
- Fail closed for an invalid configured `subdir`, and sanitize credential-bearing URLs in operator-visible metadata errors.

Closes #3165

## Refresh and validation
- Rebased onto current `main` (`8a03733e734f002ab97cc4ee74a006a70f683048`).
- [x] Focused doctrine tests: 66 passed with repository conftest disabled because its global AST gate scans the full tree.
- [x] Ruff checks on changed doctrine source/tests.
- [x] Strict mypy on changed doctrine source.
- [ ] CI green on this PR

The extra `subdir` fail-closed behavior is included as a safety correction from adversarial review; it does not alter the issue linkage above.

Made with [Cursor](https://cursor.com)

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3166"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788352674&installation_model_id=427282&pr_number=3166&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3166&signature=1c8768d1b92210698e53ee573dc0fb95dbe141ac27bca390a1b1a23f7654f3a3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">Autofix with [code]smith</picture></a>


stijn-dejongh:
@robertDouglass : Do we wish to support Artefactory as a first-level way of distributing doctrine / spec-kitty?

robertDouglass:

Three independently profile-loaded lenses reviewed the aggregate diff: `reviewer-renata` (contract/evidence), `debugger-debbie` (fault injection/live paths), and `doctrine-daphne` (canonical authority/provenance).

**Verdict: `BLOCK_MATERIAL`.** Happy-path fetch/304 behavior passes, but provenance and last-good snapshot safety can fail. Per queue policy, material issues were not fixed.

- **High — `src/specify_cli/doctrine/snapshot.py:203`:** cached ETag is reused without binding the existing manifest `source_url`/`source_type` to current config. Reproduced URL A→B with a shared ETag: B returns 304, A doctrine remains and reports unchanged.
- **High — `src/specify_cli/doctrine/snapshot.py:173`:** when promotion and backup restoration both fail, `finally` deletes the only backup. Fault injection left no local snapshot and no recovery copy.
- **High — `src/specify_cli/doctrine/snapshot.py:143`:** root fallback lets a changed archive whose configured `subdir` disappeared replace a valid prior snapshot and report success with zero usable artifacts.
- **High — `src/specify_cli/doctrine/snapshot.py:274`:** ambiguous legacy `pack_version` can be an operator ref yet is sent as `If-None-Match`; reproduced `release-7` being emitted as an ETag.
- **Medium — `src/specify_cli/doctrine/sources/https_source.py:156`:** archive bytes and JFrog version are fetched independently with no checksum/identity binding, so mutable `latest` bytes N can be labeled version N+1.
- **Medium — authority model:** Artifactory is inferred twice from URL substrings while canonical source types remain `git|https|api`; runtime persists `etag` but normative manifest contracts omit it.

**Evidence:** targeted doctrine suites pass (`42 passed`); deterministic probes reproduce the cross-origin 304, subdir replacement, ambiguous-ref, and double-failure rollback defects. Historical CI reds were on unrelated stale surfaces and were not treated as correctness evidence.

**Maintenance only:** rebased conflict-free onto current `main`, fixed two pre-existing strict-mypy `Any` returns in the touched source factory with protocol casts, and rerolled `a178537da → 12c68af20`. Verification: targeted ruff clean, strict mypy clean, `42 passed`, diff check clean. No material finding was modified.

robertDouglass:
Automated review — corrected and condensed after validation. No code was modified.

**Verdict: request changes for one core correctness bug.**

**[MEDIUM] A non-Artifactory HTTPS pack can report its ETag as its version after a `304 Not Modified`.** On the first fetch, `pack_version` is correctly left unset. On the unchanged path, `HttpsBundleSource.fetch` returns `if_none_match`, and snapshot fallback can expose that ETag as `pack_version`, contradicting the PR's “do not invent a version” goal.

Make the unchanged path reuse only a stored real version or `ref`; never fall back from version to ETag. Add a regression test starting from a manifest with `pack_version: null`.

The strict Artifactory-property behavior and wrong-subdirectory behavior are explicit design choices in code/tests. Treat them as follow-up design discussions, not additional blocking defects.

robertDouglass:
🛬 **Maintainer consolidation pass — claimed under Spec Kitty Op `01M0MMZBB7FB197HZCMCZN4TAQ`.**\n\nI’m carrying the valid change onto current `main` in a single reviewed consolidation branch, preserving original authorship via `cherry-pick -x`, folding adversarial findings, proving bugfix tests red-first, and requiring green CI before operator handoff. This PR will be closed as superseded only after the replacement PR is published with evidence.

robertDouglass:
Superseded by #3667. The original community commit is preserved there, with fail-closed snapshot promotion, checksum-bound Artifactory provenance, safe URL identity, ETag scoping, manifest-schema parity, and adversarial regression coverage. Closing this source branch while retaining attribution.

Issue #2468, opened 2026-07-08
Promote mission types + step contracts to full doctrine artifact kinds


stijn-dejongh:
Part of #2466 (Doctrine/Charter extensibility & pack ecosystem).

> **Updated 2026-08-05 after PR #3204 / #3091** (mission `doctrine-consumer-surface-missions-extraction`). That mission relocated the built-in `missions/` **data** to `packs/built-in/missions/` and converged mission-asset resolution onto a doctrine-agnostic kernel primitive, but did **not** touch this issue's promotion contract (verified: `artifact_kinds.py`, `activation_engine.py`, `cascade.py`, `org_pack_loader.py` unchanged; `MissionTypeNotAnArtifactKind` and `_MISSION_TYPE_UNIVERSE_EXTENSION` intact; deferred by that mission's constraint C-002). The relocation did move this issue's source paths and introduce one new sub-decision — see the Design note below.

## Design

`mission_step_contract` is already a first-class `ArtifactKind`; `mission-type` deliberately is NOT — `artifact_kinds.py` raises `MissionTypeNotAnArtifactKind`, and it is only admitted via `_ORG_DRG_KIND_ALIASES["mission_types"]` for augmentation. Promote `mission-type` to a full `ArtifactKind` with its own plural/glob for `packs/built-in/missions/mission_types/` (relocated there by PR #3204 / #3091), retire the `MissionTypeNotAnArtifactKind` side-channel and `_MISSION_TYPE_UNIVERSE_EXTENSION`, and route it through the normal activation/cascade path instead of the `CHARTER_KIND_TOKENS` special case.

[...]

Issue #2467, opened 2026-07-08
Split built-in doctrine into packs + compound packs (KEYSTONE)


stijn-dejongh:
Part of #2466 (Doctrine/Charter extensibility & pack ecosystem).

## Design

`built-in` is today one monolithic bundle and org packs are a flat, un-nested list with no dependency notion between them. This is the keystone: split `built-in` into named packs and give packs (built-in and org) a `depends_on` relationship, so everything downstream in the epic can build against a real DAG instead of a single flat tree.

- Design a pack-manifest schema (`pack_id` / `version` / `depends_on[]`) at the pack-root tier.
- Extend `PackRegistry` / `load_pack_registry` (`org_pack_config.py`) to resolve `depends_on` before per-pack `load_org_pack`.
- Add DAG resolution + cycle detection to `plan_activation` / `commit_activation` (`activation_engine.py`), reusing `cascade.py`'s BFS with a structured error parallel to `UnknownActivationIdError`.
- Split the monolithic `src/doctrine/*/built-in/` tree into named built-in packs so `_default_built_in_dir()` (`agent_profiles/repository.py`) is pack-aware.

## Surfaces

- `drg/org_pack_config.py`
- `drg/org_pack_loader.py`
- `charter/activation_engine.py`
- `charter/cascade.py`
- `agent_profiles/repository.py`
- `doctrine/pack_validator.py`
- `doctrine/pack_assembler.py` / `snapshot.py`

## Depends-on

None — this is the foundation the rest of the epic builds on.

## Sizing

**L**

## Top risk

High blast radius: every surface today hardcodes the one canonical `built-in` tree. Land as a pure refactor + compat shim BEFORE introducing compound-pack semantics, per the shared-package-boundary migration discipline (see `docs/adr/3.x/2026-04-25-1-shared-package-boundary.md` for the precedent pattern).

robertDouglass:
Triage validation (2026-08-02): keep open, but reconcile the delivered slice with the original acceptance contract.

Current planning addenda and commit `873832aa1` establish that the built-in doctrine tree was relocated to the `packs/built-in` seam and that pack-root path handling was hardened. That is evidence for the refactor portion of this keystone.

[...]

Issue #2466, opened 2026-07-08
Epic: Doctrine/Charter extensibility & pack ecosystem


stijn-dejongh:
The doctrine/charter substrate already treats much behaviour as data — the activation plan/commit seam (`charter.activation_engine`), DRG-driven cascade over REQUIRES/SUGGESTS/REFINES edges (`charter.cascade`), lineage-vs-delegation typed relations, and a kind vocabulary that already fences `mission-type` off pending promotion. What's missing is the *packaging* layer: `built-in` is one monolithic bundle, org packs are a flat un-nested list with no dependency notion, and two surfaces customers most want to extend (assets, custom shortcodes) aren't doctrine kinds at all. This epic turns that scattered substrate into a coherent, author-able pack ecosystem so customers can define/refine their own doctrine — gated by one foundational schema change everything else builds on.

**Scope: core pack ecosystem in `3.2.x`; verified-distribution (#2539) is the `3.3.x` extension (operator decision). The dependency order below is sequencing metadata only, not a phasing/scope cut.**

## Dependency spine

```
keystone: pack split
  ├── mission types as doctrine
  │     └── procedure decision (depends on mission-types)
  ├── assets kind
  ├── shortcodes
  └── pack validator

model-discipline pipeline — orthogonal, no dependency
```

The four non-keystone kinds/validators depend only on the pack-split keystone and can proceed in parallel once it lands; the procedure decision additionally waits on mission-types-as-doctrine; the model-discipline pipeline is fully independent and can run any time. See the linked sub-issues for the concrete tickets, sizing, and open decisions.

stijn-dejongh:
Added four sub-issues from the doctrine canonical-structure remediation programme (split from mission `doctrine-canonical-structure-remediation-01KYEYSD`, programme record in `docs/plans/3-2-x-milestone-roadmap.md`):

- #2948 — mission A, silence guards + campsite (first, gates everything else)
- #2949 — mission B1, DRG `impacts` relation vocabulary (blocked by A)
- #2950 — mission B2, DRG edge migration + extractor retirement (blocked by B1)

(Mission C, #2951, is sub-issued under #2935 instead — it delivers that issue directly. Mission D, #2952, has no epic parent; it is blocked by A only.)

Mandatory order: A → B1 → B2 → C, with D depending only on A. See each issue for the load-bearing reason the order can't be relaxed.

PR #3204, opened 2026-08-05
Relocate built-in mission data to packs/built-in/missions + gate preconditions (#3091)

### LINKED ISSUE #3036: Architectural gate requires the repo-coupling the shippable-doctrine rule forbids
## Problem

An architectural gate requires the very repo-coupling that the "Shippable doctrine" rule (`docs/development/review-gates.md`) forbids — so the rule and the gate contradict each other, and the gate currently wins.

## Observed state (verified on PR #3007's branch, `pr-3007-local`)

`tests/architectural/test_no_dead_doctrine_paths.py:456` asserts **exact equality** on a one-element list (line numbers verified against the branch as of this filing):

```python
def test_forbidding_mention_would_false_red_without_its_discriminator() -> None:
    ...
    excluded = sorted((site.path, site.text) for site in scan.forbidding_mentions)
    assert excluded == [
        (
            "src/doctrine/agent_profiles/built-in/doctrine-daphne.agent.yaml",
            "src/doctrine/graph.yaml",
        )
    ], f"A2's effect set moved -- widening it needs a reason: {_render(scan.forbidding_mentions)}"
```

Proven by mutation on the branch: removing that repo-local path (`src/doctrine/graph.yaml`) from `doctrine-daphne.agent.yaml`'s `avoidance-boundary` — i.e. doing exactly what the "Shippable doctrine" rule in `docs/development/review-gates.md` requires (built-in doctrine must be valid and actionable in a consumer repo with no access to the spec-kitty source tree) — turns the test **RED**:

```
FAILED test_forbidding_mention_would_false_red_without_its_discriminator
assert [] == [('src/doctrine/agent_profiles/built-in/doctrine-daphne.agent.yaml',
               'src/doctrine/graph.yaml')]
```

So the gate does not merely *tolerate* the repo-local reference in the daphne profile — it *requires* it, pinned by exact path and matched text (NFR-003's effect-set-pinning discipline, applied to a site the discipline itself should have excluded). Any future cleanup of that profile (stripping the repo-local `src/doctrine/graph.yaml` mention, per the shippable-doctrine rule) cannot land without also updating this gate in the same commit, or CI goes red.

### Second instance of the same shape — Gate C

`test_every_built_in_doctrine_cross_link_resolves` (same file) requires every relative markdown link under `src/doctrine/` to resolve **on our disk**. This cements links that escape the package entirely. Example, verified on the branch:

`src/doctrine/templates/diagrams/README.md` lines 65-71 contain links such as:

```
[`docs/architecture/01_context/README.md`](../../../../docs/architecture/01_context/README.md)
```

`../../../../docs/architecture/...` resolves inside the spec-kitty source tree but is dangling in any consumer repo that installed the built-in template pack without the rest of the spec-kitty repo. The gate's contract, as written, is that such links must never be removed (removing them turns Gate C's "must resolve on disk" check red under the current discriminator set), even though they are exactly the "escapes the package" case the shippable-doctrine rule is meant to catch.

## Why this matters

This is the load-bearing blocker for the follow-on doctrine-decoupling missions the audit page (`docs/plans/doctrine/built-in-doctrine-repo-coupling-audit.md`, §5, §8 item 3) identifies as highest priority: "the gate currently *enforces* the bug." Any daphne-profile cleanup or built-in cross-link cleanup silently breaks CI unless the discriminator itself is redefined first.

## Proposed shape

Reframe the discriminator so it expresses **"this reference is permitted to exist"** rather than **"this reference must exist."** Concretely:

- Drop (or loosen) the exact-list `assert excluded == [...]` pinning in `test_forbidding_mention_would_false_red_without_its_discriminator` so removing the repo-local mention is a *pass* (fewer/no forbidding-mention sites), not a *fail*. The NFR-003 proof only needs to show the discriminator *can* exclude a real site when one exists — it should not also assert that one must currently exist.
- For Gate C, distinguish "this relative link is expected to resolve in every deployment" from "this relative link legitimately points outside the built-in-doctrine package boundary and is therefore exempt from the on-disk-resolution requirement" (rather than silently requiring the escaping link to stay in place forever).

## Known hard parts

- The gate file's own NFR-003 discipline (pinning discriminator effect sets so widening needs a reason, not a regex tweak) is exactly why this contradiction is hard to fix quietly — any relaxation of the pinned assertion needs its own justification recorded in the same commit that changes the underlying doctrine content, per the file's own stated philosophy.
- This should be coordinated with the daphne-profile cleanup and cross-link relocation described in `docs/plans/doctrine/built-in-doctrine-repo-coupling-audit.md` §6 and §8, so the gate change and the content change land together.

## References

- `tests/architectural/test_no_dead_doctrine_paths.py:456` (assertion), same file's `test_every_built_in_doctrine_cross_link_resolves`
- `docs/development/review-gates.md` ("Shippable doctrine" rule)
- `docs/plans/doctrine/built-in-doctrine-repo-coupling-audit.md` (audit page, not yet on `main`; lands with PR #3007)
- PR #3007 (mission A of `doctrine-canonical-structure-remediation-01KYEYSD`)

### LINKED ISSUE #3039: Split test_no_dead_doctrine_paths.py: Gate A is a src/-wide CLI gate wearing a doctrine name
## Problem

`tests/architectural/test_no_dead_doctrine_paths.py` mixes three distinct scopes behind one shared scanner. Gate A scans ALL of `src/` (CLI code included, not just doctrine content); Gates B and C scan only doctrine content under `src/doctrine/`. A future attempt to relocate or extract "the doctrine tests" as a unit would silently drop Gate A's coverage of spec-kitty's own source.

## Observed state (verified on PR #3007's branch)

The module's own module docstring (lines 5-14) names three defect classes ("A", "B", "C") sharing one file:

- **Gate A** — `scan_graph_monolith_paths` is run against `_SRC_ROOT = _REPO_ROOT / "src"` (the whole `src/` tree, not `src/doctrine/`). Its NFR-003 discriminator-proof test, `test_project_tier_graph_path_would_false_red_without_its_discriminator`, pins an exclusion set that names files entirely outside doctrine:

  ```python
  assert excluded == [
      "src/charter/synthesizer/manifest.py",
      "src/charter/synthesizer/project_drg.py",
      "src/doctrine/drg/merge.py",
      "src/glossary/drg_builder.py",
      "src/specify_cli/charter_runtime/freshness/computer.py",
      "src/specify_cli/state/contract.py",
  ]
  ```

  This is a CLI-source-wide gate (`src/charter/...`, `src/glossary/...`, `src/specify_cli/...`) wearing a "doctrine" filename.

- **Gate B** (`scan_shipped_pack_paths`) and **Gate C** (`scan_doctrine_cross_links`) are also invoked against `_SRC_ROOT` / `_DOCTRINE_ROOT` respectively in the assertion tests, but are doctrine-content-specific in nature (the dead `<kind>/shipped/` pack layer, and built-in doctrine markdown cross-links) — a genuinely different scope from Gate A's CLI-wide dead-path scan.

- The audit page `docs/plans/doctrine/built-in-doctrine-repo-coupling-audit.md` (§8, recommendation order item 6, and the closing paragraph) makes this explicit: "split `test_no_dead_doctrine_paths.py` before extracting anything: its Gate A is a `src/`-wide CLI gate wearing a doctrine name, so moving the module wholesale would silently drop real coverage of spec-kitty's own code."

## Why this matters

Several follow-on missions in the `doctrine-canonical-structure-remediation` programme (of which PR #3007 is mission A of 5) are expected to relocate or extract doctrine-specific test coverage as built-in doctrine gets decoupled from the repo (see the other three issues filed alongside this one, from the same PR #3007 landing-pass squad). If "the doctrine dead-path tests" get moved or extracted as one unit without first separating Gate A, spec-kitty's own CLI-source dead-path coverage (the `src/charter`, `src/glossary`, `src/specify_cli` fragment-naming checks) silently disappears with it.

## Proposed shape

Split the module by scope before any extraction work begins:

- Move Gate A (`scan_graph_monolith_paths` + its five assertion/fixture tests: `test_no_source_site_names_the_dead_drg_monolith`, `test_the_migration_hint_names_a_fragment_that_exists`, `test_project_tier_graph_path_would_false_red_without_its_discriminator`, `test_forbidding_mention_would_false_red_without_its_discriminator`, `test_gate_a_rejects_a_planted_violation`, `test_gate_a_discriminators_do_not_swallow_a_planted_violation`) to its own module, scoped explicitly as a `src/`-wide CLI gate (naming reflecting that, not "doctrine paths").
- Keep Gates B and C (the `<kind>/shipped/` pack-layer scan and the built-in-doctrine cross-link scan) together or split further, scoped to doctrine content only.
- Preserve the existing NFR-003 discriminator-proof discipline (the `*_would_false_red_without_*` tests and their pinned effect sets) unchanged across the split — this is a file reorganization, not a discriminator rewrite.

## Known hard parts

- The three gates currently share helper code (`Site` dataclass, `_rel`, `_read_lines`, `_text_files`) — splitting cleanly means either a shared internal helper module or accepting some duplication; either choice should be a deliberate call, not incidental to the split.
- Should be sequenced *before*, not during, the daphne-profile cleanup / cross-link relocation work described in the same audit page, so the split isn't entangled with unrelated content changes in the same PR.

## References

- `tests/architectural/test_no_dead_doctrine_paths.py` (module docstring lines 5-14; Gate A at lines 88-168 and 381-486; Gates B/C follow in the same file)
- `docs/plans/doctrine/built-in-doctrine-repo-coupling-audit.md` §8 item 6 and closing paragraph (audit page, lands with PR #3007)
- PR #3007

### LINKED ISSUE #3091: Phase 1b: relocate the missions/ doctrine tree to packs/built-in
Follow-on to the built-in relocation (`relocate-builtin-doctrine-packs-01KYT87F`). The `missions/` tree was **deferred out of Phase 1** by unanimous post-plan adversarial-squad ruling.

**Why deferred**: `missions/` readers span four layers — including `src/kernel/paths.py` (`files("doctrine")/"missions"`). Routing a *kernel*-layer reader through the doctrine-layer `resolve_pack_root` is a **C-004 upward-import violation**; pushing the resolver into kernel is the C-002 kernel extraction deferred to Phase 2. Several readers are also `Path(__file__)`-relative (invisible to `files()`/string sweeps), and live upgrade migrations read missions content. Graph identity is unaffected (mission-derived nodes are authored in the moving fragments, not scanned from `missions/`).

**First task**: a cross-layer `missions/` reader inventory (doctrine + kernel + charter + specify_cli + upgrade migrations) with a per-reader move/stay + resolution-strategy decision, before any move.

### LINKED ISSUE #3182: Stale TIER-1 command-template overrides teach the banned raw DoctrineService construction and a deleted 'constitution context' command
`.kittify/overrides/missions/software-dev/command-templates/implement.md` and `review.md` are **tracked TIER-1 template-resolver overrides** — they shadow the canonical doctrine templates for this repo's own dogfooding — and both are substantially stale:

1. **They teach the construction pattern mission `charter-sole-door-bypass-closure-01KZ3WAA` just banned**: raw `AgentProfileRepository(project_dir=...)` and `DoctrineService(shipped_root, project_root)`, rather than `charter.doctrine_service_builder.build_activation_aware_doctrine_service(repo_root)`. The five new sole-door architectural gates are AST-based over `src/**/*.py`, so they cannot see Markdown — these overrides are invisible to the enforcement.
2. **They reference `spec-kitty constitution context`, a command that no longer exists** (superseded by `charter context`). Last touched at `fd9965fca`, i.e. pre-charter-rename.

**Why this was not folded during the PR #3175 landing pass:** patching only the two construction lines would leave an otherwise-broken, pre-charter-rename artifact that still instructs agents to run a non-existent command. The file needs a wholesale review against the current canonical templates, which is its own scoped piece of work.

Suggested done-bar: each override either re-based on the current canonical template or deleted if it no longer diverges usefully from it.

### LINKED ISSUE #3183: UnknownMissionTypeError message is self-contradictory: names an id unknown, then lists it as registered
With `activated_mission_types: [my-custom]` in `.kittify/config.yaml` and no resolvable `MissionTypeProfile` on disk for that id, `charter.mission_type_profiles.resolve_mission_type_context()` raises:

```
UnknownMissionTypeError: Unknown mission type 'my-custom'. Registered types: my-custom.
```

The message names the id as unknown and then lists that same id as registered. The cause is a vocabulary collision: "registered" in this error means *activated in `.kittify/config.yaml`*, not *has a loadable profile*, so the two halves of the sentence are measuring different things.

Pre-existing; reproduced during the PR #3175 landing pass while correcting FR-006's framing. Not folded there because the fix is a semantic decision, not a wording tweak: either rename the field/message so the two senses are distinguishable, or make custom mission-type activation validate profile-loadability at activation time and fail there with a useful message.

Note the same landing pass softened a newly-added docstring paragraph in `src/charter/mission_type_profiles.py` that had documented custom-type activation as working end-to-end, since this error contradicts it. That docstring should be revisited when this is fixed.

stijn-dejongh:
## What changes

Built-in mission data (mission-type profiles, step prompts, step contracts, per-type content) now
ships under **`packs/built-in/missions/`**, alongside every other built-in doctrine kind, instead of
being embedded in the `doctrine` Python package. Nothing changes for a user — `spec-kitty init` and
mission resolution behave exactly as before — but the built-in doctrine tree is now consistently
laid out and wheel-shippable, and mission-asset resolution can no longer silently fall back to a
data-less package directory. Ships as one atomically-reviewed relocation plus the gate preconditions
it depended on.

## The mission (7 work packages, each independently reviewed and approved)

- **WP05 — the relocation (`#3091`).** Moves the `missions/` data subdirectories to
  `packs/built-in/missions/`; the 11 `.py` logic modules stay in place. Every reader across kernel,
  doctrine, charter, `specify_cli`, and the upgrade migrations is repointed in the same change, and
  the sibling-path pattern is repointed to the fully-qualified `packs/built-in/missions` so it can no
  longer self-match the now data-less `src/doctrine/missions`. DRG graph fragments regenerated
  byte-identical.
- **WP04 — a doctrine-agnostic kernel resolution primitive.** `pack_paths` and
  `MissionTemplateRepository.default_missions_root()` converge onto one primitive holding no
  doctrine vocabulary, guarded by a kernel-scoped AST gate.
- **WP03 — a committed cross-layer reader inventory** (`docs/plans/doctrine/`), the authority WP05
  consumed.
- **WP01/WP02 — gate preconditions (`#3039`, `#3036`, `#3182`).** The dead-path architectural gates
  are split by actual scope with discriminator proofs on planted fixtures; the shipped
  `doctrine-daphne` profile drops its repo-local `src/doctrine/graph.yaml` reference; Gate C's
  boundary discriminator unions both shipped roots.
- **WP06 — `UnknownMissionTypeError` (`#3183`)** no longer calls an activated mission type both
  "unknown" and "registered".
- **WP07 — the TIER-1 override command templates** refreshed onto the current canonical.

Independent mission-acceptance review: **PASS** across FR-001..FR-007, NFR-001..NFR-004, and
constraints C-001..C-003, with a clean cross-WP interaction sweep. Every pre-existing red was
classified against the pre-mission base (delta zero).

## Post-mission folds (separate commits after the mission)

Small, self-contained, each red-first:
- `fix(resolver)` — resolve **mission-scoped** Tier-1 overrides under
  `.kittify/overrides/missions/<mission>/`, which were previously inert (this activates WP07's
  refreshed dogfood overrides).
- `fix(review)` / `fix(status)` — thread `--reviewer` into the review-cycle artifact frontmatter;
  supersede the stale `review` block on a `--to planned` rollback.
- `fix(kernel)` — bound the `get_package_asset_root` ancestor walk so a broken install fails closed.
- `test(gates)` — make the kernel↔doctrine AST gate segment-aware so it catches the dotted
  `files("doctrine.schemas")` shape; the one pre-existing site is disclosed as an explicit,
  documented exemption.
- `docs(migration)` — correct a stale source-path docstring.

## One deferred item flagged for a design decision

The kernel↔doctrine **schema** coupling (`kernel/schema_utils.py` loads doctrine-owned schema files)
is only *half* addressed here: the AST gate now **catches** the shape and the site is an honest,
documented exemption, but the actual decouple — relocating the schemas or injecting the schema root —
is a design decision (`schema_utils` is a deliberate cross-layer kernel utility) that should not be
rushed into this relocation. It wants its own small follow-up.

## Verification

Rebased onto current `main`; the only integration overlap (`CHANGELOG.md`, `cli/commands/init.py`)
was resolved and re-validated (init integration + template resolution green). Fold and mission
suites green on the integrated tree (regen-roundtrip, the dead-path gates, kernel, resolver, review
provenance, terminology). CI is authoritative for the full sweep.

Closes #3091
Closes #3183
Closes #3036
Closes #3182
Closes #3039

https://claude.ai/code/session_012XRxszZ9PwenoQJxHhSZra

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3204"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788495405&installation_model_id=427282&pr_number=3204&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3204&signature=cab09d09ad7230dc7ec4ce5fbe78c7fe4871d9fcb732b9b359d7e1380907e7a5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


stijn-dejongh:
## Should #2468 remain open after this mission? — Yes: **remain open, update its body**

Two profile-loaded lenses (planner-priti, delivery/scope · architect-alphonso, architecture)
independently reviewed #2468 ("Promote mission types + step contracts to full doctrine artifact
kinds") against this mission's scope and deliverables. They **converge**: the issue stays open, but
its body is now partly stale and gained one genuinely new sub-decision.

### Deliverable overlap: NONE — the promotion is wholly untouched

This mission was a pure **data relocation** (`src/doctrine/missions/{mission_types,mission-steps,…}`
→ `packs/built-in/missions/…`; the 11 `.py` logic modules stayed). It delivered zero of #2468's
ask, and honored spec constraint **C-002** (which explicitly defers #2468/#2652) to the letter.
Verified on the PR head:

- The promotion-contract files have an **empty diff** vs base — `artifact_kinds.py`,
  `charter/activation_engine.py`, `charter/cascade.py`, `drg/org_pack_loader.py` all untouched.
- `ArtifactKind` still has **no** `MISSION_TYPE` member; `MissionTypeNotAnArtifactKind` still raised
  (`artifact_kinds.py:40-48, 218-222`); the side-channels #2468 wants retired are all alive —
  `_ORG_DRG_KIND_ALIASES["mission_types"]` and `_MISSION_TYPE_UNIVERSE_EXTENSION`
  (`org_pack_loader.py:87-129`), still gated by the lockstep drift guard in
  `tests/doctrine/test_org_pack_augmentation.py`.

This is deliberate non-adoption, not an accidental orphan → **#2468 must stay open.**

### What the relocation *changed* for #2468 (why its body needs updating)

1. **New architectural fork (the load-bearing finding).** Every other built-in kind resolves its
   content via the WP04-converged seam `doctrine/pack_paths.py::built_in_dir(kind)` =
   `resolve_pack_root("built-in") / kind.plural` — i.e. a **flat** `packs/built-in/<plural>/`. But
   mission-type data landed **nested** at `packs/built-in/missions/mission_types/` (under a
   `missions/` container, alongside `mission-steps/`, `built_in_step_contracts/`). A naive promotion
   giving `MISSION_TYPE` plural `mission_types` would make `built_in_dir` resolve
   `packs/built-in/mission_types/` — **a path that does not exist.** So #2468 must now decide:
   **flatten** `mission_types/` to a top-level pack dir, **or** teach
   `built_in_dir`/`has_built_in_content_dir` the nested `missions/` container as a mission-tier
   exception. This decision did not exist when #2468 was written.

2. **Stale source paths.** The Design paragraph's "mirroring `src/doctrine/missions/mission_types/`"
   and the Surfaces bullets `missions/mission_types/` + `missions/mission-steps/` now point at moved
   trees.

3. **Partial tailwinds.** The DRG fragments `packs/built-in/mission_type.graph.yaml` +
   `mission_step_contract.graph.yaml` now sit at pack root beside every other kind's
   `<kind>.graph.yaml` (SC-008 pins them byte-identical) — strengthening #2468's "route through the
   normal path" premise. And the **Depends-on #2467** (pack-split keystone) is now effectively
   delivered by the relocation (#3091/#3204) — #2468 reads as unblocked-on-relocation.

### Recommended body edits for #2468 (not applied here — flagged for the issue owner)

- **Surfaces:** repoint `missions/mission_types/` → `packs/built-in/missions/mission_types/` (+ logic
  reader `src/doctrine/missions/mission_type_repository.py`) and `missions/mission-steps/` →
  `packs/built-in/missions/mission-steps/`; **add** `doctrine/pack_paths.py` (`built_in_dir`,
  `has_built_in_content_dir`), `artifact_kinds.py::_HAS_BUILT_IN_CONTENT_DIR` /
  `_NON_AUGMENTATION_ELIGIBLE_KINDS`, and the two `packs/built-in/*.graph.yaml` fragments.
- **Design:** replace the stale source path with `packs/built-in/missions/mission_types/` and record
  the flatten-vs-extend-`built_in_dir` fork above.
- **Depends-on:** note #2467's physical pack positioning is now in place via #3091/#3204; the
  vocabulary/activation keystone work remains the true prerequisite.
- **Top risk** (reversing the tested no-silent-fallback contract) — leave as-is; verified still
  accurate.

_Verdict reached independently by both lenses; deliverable-overlap and contract-intact claims were
each verified against the PR head by `git diff --stat` and direct symbol reads._

stijn-dejongh:
## Landing-pass remediation — CI red adjudication + folds

Reviewed and remediated PR #3204's CI on the rebased tip. **10 fold commits pushed** (`2280d18e5..318d57828`); each is a single-purpose `<type>(landing):` commit (commitlint-clean). Adversarial squad ran (architect-alphonso + paula-patterns) — **no MAJOR findings**.

### Red classification + folds

| CI check (red) | Classification | Fold |
|---|---|---|
| `arch-adversarial (arch_shard_2)` · `test_built_in_location_authority` | PR defect — 10 `packs/built-in` joins added outside the `doctrine.pack_paths` authority | **F1** `b8ef2a31c` — 2 routed through the authority (`pack_manager.py`, `drg/extractor.py`), 8 documented allowlist exemptions (layer-boundary sibling-shapes + caller-supplied-root sites). Gate 17/17. |
| `fast-tests-docs` + `docs-freshness` · `test_docs_structural_lint` | PR defect — reader-inventory doc shipped without `doc_status`/`updated` frontmatter, unregistered in the page inventory | **F2** `e155a72c9` — frontmatter added; page-inventory + retrieval-index regenerated. `check_docs_freshness` exit 0. |
| `integration-tests-core-misc (misc)` | (a) same frontmatter defect (fixed by F2); (b) the slice double-ran `tests/docs` | **F3** `318d57828` — retired the #2359 docs over-coverage (operator decision): `tests/docs` now runs only in `fast-tests-docs`; `tests/regression` left untouched (36 tests still owned). Arch-pin `test_ci_quality_path_filters` updated (26 passed). **Accepted tradeoff:** a src-only change that regresses a docs test is no longer caught pre-merge except on a docs-path/push trigger. |
| `fast-tests-status` · `test_work_package_lifecycle` | **Pre-existing time-bomb** (not this PR) — a test appended an event with a hardcoded `at="2026-08-01"`; once the wall clock passed it, it sorted before the real `now()` events. Fails on main too, today. | **F4** `f1be9e5cf` — anchored the event to real `now(UTC)`. 21 passed. |
| `fast-tests-core-misc (specify-cli-rest)` · `test_nfr004_command_installer_resolves_doctrine_path` | **Stale test** — asserted resolution "through doctrine", but FR-005 deliberately moved the data to `packs/built-in/missions`; resolution is correct, assertion is stale. | **F5** `f6796a93b` — proved the resolved path is the real relocated file, re-pinned + renamed to `..._resolves_relocated_missions_path`. 102 passed. |
| `fast-tests-cli` #1 · `test_tasks_compat_surface` | PR defect (this session's `--reviewer` fold) — `_mt_resolve_reviewer_identity` not registered in the seam compat guard | **F6** `2f8a99a2d` — registered in `SYMBOL_TO_MODULE` + `tasks.py` barrel. 323 passed. |
| `fast-tests-cli` #2 · `test_sync_purge_3030` (`--all` in help) | **Pre-existing, unrelated (#3030)** — Typer forces ANSI color on GH Actions; a raw substring assertion split `--all` across SGR spans. Surfaced only because a fold un-skipped `fast-tests-cli`. | **F7** `95be14ad4` — used the canonical `strip_ansi` helper (per DIRECTIVE_044). 25 passed. |
| `quality-gate` | Aggregator — red because upstream jobs were red | Auto-greens with the above. |
| `regression tests (blocking)` · 6 in `tests/regression/` | **Deliberate open-P0 signal — left red** per ADR 2026-07-17-1 (known coordination issues; a dedicated sibling mission owns them). `test_issue_2996` is the same class as the newly-filed **#3205**. | Not folded, by policy. |

### Campsite folds (squad MINORs)
- **C1** `85621ed4f` — `kernel/paths.py` reuses `_MISSION_ASSETS_SIBLING_PATTERN` instead of re-inlining the shape; drops the now-vacuous gate exemption (one fewer allowlist entry).
- **C2** `0ea77a55d` — corrected the relocated `packs/built-in/missions/README.md` (it wrongly claimed the `.py` logic modules ship in the pack; they stayed at `src/doctrine/missions/`).
- **C3** `c95666737` — refreshed two stale resolution-path comments (`resolver.py` PACKAGE tier, `pyproject.toml` force-include rationale).

### Adversarial squad
No MAJOR / no blocker (both lenses). Cheap MINORs folded (C1–C3). Deep design findings — two-door `get_package_asset_root`, `default_missions_root()` ignoring `SPEC_KITTY_PACKS_ROOT` (env split-brain), the runtime second copy of mission data — consolidated into follow-up **#3210** (parented under #2466), since they need a design decision, not a landing fold.

### Follow-ups filed this pass
- **#3205** (P0) — no approved-verdict `ReviewCycleArtifact` constructor (forces the #1924 override); red repro test still owed.
- **#3206** (P2) — decouple `kernel/schema_utils` from doctrine schemas (retire the AST-gate exemption).
- **#3210** (P2) — unify built-in mission-tree resolution (from the squad).
- **#2468** body updated for the relocation (nested-vs-flat `built_in_dir` fork).

### Evidence
All 10 folds verified green in isolation; the previously-red node-ids confirmed green **composed on the branch tip** (status/nfr004/compat/purge — 4 passed together; arch gate 17/17; docs lint + freshness clean; `test_ci_quality_path_filters` + `test_src_filter_coverage` 26 passed). `ruff`/`mypy` clean on touched sources; commitlint clean on the whole fold range; terminology guard 10 passed.

**State:** pushed; watching CI. Merge-ready on green (the 6 `tests/regression/` reds stay red by policy). The operator merges.

stijn-dejongh:
## Follow-up: `integration-tests-cli` un-masked + remediated

A subtle CI discovery after the first fold batch. `integration-tests-cli` has `needs: fast-tests-cli` + `if: fast-tests-cli.result == 'success'`, so it is **SKIPPED whenever `fast-tests-cli` is red**. On `main` (every recent run incl. tip `96494e5ec`), `fast-tests-cli` is **red** → `integration-tests-cli` has been **permanently skipped** — its failures never surface on main's board.

This PR's folds are the **first commit to make `fast-tests-cli` green**, which un-masks `integration-tests-cli` for the first time and exposes **8 failures that are all pre-existing on `main`** (delta-zero — reproduced identically in throwaway worktrees at `upstream/main` `96494e5ec` and at the pre-fold tip). The PR causes none of them; it reveals debt main was hiding behind its own red gate.

Per operator decision, all 8 are remediated (each a pre-existing bug, folded as campsite):

| Failure | Root (pre-existing) | Fold |
|---|---|---|
| `test_completion_fast_path::test_manifest_matches_live_cli` | stale completion manifest — the `charter apply --compile` help (landed in `pack.py` earlier) was never re-baked | **`195973c76`** regenerated the manifest (1-line drift; PR touches no CLI command) |
| `test_review_git_baseline::test_review_passes_with_notes...` | the review flow now resolves the target branch via real `git symbolic-ref`/`rev-parse`; the test's `_fake_run` mock rejected those | **`3aea58591`** mock now delegates unmatched calls to real `subprocess.run` against the fixture's real repo |
| 6× charter interview/generate (`config.yaml not written`) | `locate_project_root`'s ancestor walk-up finds a stray `.kittify` marker above `tmp_path`, redirecting `repo_root`; the 6 tests invoke the CLI before seeding `.kittify` at `tmp_path` | **`87bf49451`** pinned `SPECIFY_REPO_ROOT=tmp_path` (the established antidote used in ~20 other test files) — hermetic against ambient markers |

**Evidence:** the full `integration-tests-cli` selection (`pytest tests/cli/ tests/specify_cli/cli/ -m 'not windows_ci and (git_repo or integration)'`) is **980 passed, 2 xfailed, 0 failed** locally — and the 6 charter tests pass **even with ambient `/tmp/.kittify` pollution present**, proving the pin is deterministic (CI's clean env will pass too). commitlint clean on all three.

Note for the maintainer log: this is the un-mask class where fixing an upstream gate surfaces a downstream job's long-hidden pre-existing reds. The `fast-tests-cli` reds this PR fixed were themselves a mix of a mission-caused compat gap (`_mt_resolve_reviewer_identity`) and pre-existing (`test_sync_purge_3030` ANSI).

**State:** pushed (`…4512f02e5`). Remaining reds: only the 6 `tests/regression/` (deliberate open-P0, left red by policy). Watching CI; merge-ready on green. The operator merges.

Issue #2404, opened 2026-07-05
spec-kitty accept reads acceptance-matrix.json from stale -coord worktree, not primary surface (coord-topology missions)


stijn-dejongh:
## Context

Discovered while running `spec-kitty accept --mission ci-health-charter-path-and-arch-shard-01KWRTB2` (coord-topology mission) after all 3 WPs were approved.

## The bug

`accept`'s lane-based acceptance-matrix check reports `Acceptance matrix verdict is 'pending'` even after `acceptance-matrix.json` in the primary repo checkout was fully filled in with real evidence (all criteria `pass_fail: pass`, all negative invariants `confirmed_absent`) and committed to the mission's target branch.

Root cause: for a `coord`-topology mission, `spec-kitty` materializes a `.worktrees/<mission>-coord` worktree checked out on the coordination branch (`kitty/mission-<slug>`), separate from the primary repo checkout (on the target branch). Planning-artifact writes made via `spec-commit`/`git commit` directly on an *unprotected* target branch land as a "direct" commit (per `spec-commit`'s own help text: "On an unprotected...primary the commit is direct") — they do NOT propagate into the `-coord` worktree's copy of `kitty-specs/<slug>/`.

`_run_lane_based_checks` in `src/specify_cli/acceptance/__init__.py` then reads `acceptance-matrix.json` via a coord-aware `feature_dir` resolution, which for a coord-topology mission resolves to the **stale `-coord` worktree copy**, not the primary surface with the real edits.

This is the same bug class already found and fixed for the encoding-normalizer (see the comment at `src/specify_cli/acceptance/__init__.py` around line 630-645: "Pre-fix this used `resolve_feature_dir_for_mission` (coord-aware); on a coord-topology mission it scanned the materialized `-coord` worktree, missing the real primary artifacts...`_planning_read_dir` resolves the PRIMARY surface via the same kind-aware seam") — but the acceptance-matrix read path in the same module apparently was never migrated to the same fix.

## Reproduction

1. Create/finalize a coord-topology mission with `acceptance-matrix.json` scaffolded (TODO placeholders).
2. Fill in `acceptance-matrix.json` with real evidence on the primary checkout (target branch), commit via `spec-commit`.
3. Run `spec-kitty accept --mission <slug>`.
[...]

PR #3209, opened 2026-08-05
refactor(sync): retire the dead queue-backed event drain and unblind its cone

### LINKED ISSUE #2804: `spec-kitty merge` re-scaffolds acceptance-matrix.json + issue-matrix.md to placeholders after merge, destroying filled verdicts/evidence

### LINKED ISSUE #3167: Finish #3030's Chain-B migration at the two remaining enforcement sites
Split out of the `#3108` tracker-egress-refusal mission (Bundle C) as an out-of-scope residual.

**Please read the framing note first.** A previous attempt to raise this was framed as *"consolidate the two consent chains"*, and that framing is why it was deferred — it reads as a refactor with no user-visible consequence. It is not. This issue is scoped deliberately as **finishing a migration `#3030` already started**, at the two sites where it was not completed.

## What is unfinished

`#3030` established Chain A (`sync/consent.py::resolve_project_consent` → `project_egress_refusal`) as the authoritative hosted-sync consent chain, keyed on `project_uuid`. Chain B still exists alongside it:

- `src/specify_cli/sync/routing.py:255` — `is_sync_enabled_for_checkout(start: Path | None = None) -> bool`

Chain B answers a version of *"may this data leave"*, but it honours the **repo-slug-keyed** `[sync.repo_defaults]` record that `sync/consent.py:625` explicitly refuses to treat as a consent level:

> `[sync.repo_defaults]` record is not a level, because it is keyed on a mutable [repo slug]

## Where Chain B is still live on a real egress gate

Enforcement (these are the two sites this issue is about):

- `src/specify_cli/sync/batch.py:338` — `return is_sync_enabled_for_checkout()`
- `src/specify_cli/sync/runtime.py:106` — `routing_enabled = is_sync_enabled_for_checkout(project_root)`

Display-only reads (not enforcement; listed so a sweep does not mistake them for the fix):

- `src/specify_cli/cli/commands/sync.py:1964` and `:2081` — both read `routing.effective_sync_enabled` for rendering only.

## Why this is reachable, not theoretical

`_build_checkout_sync_routing` falls through to `get_repository_sync_enabled(repo_slug)` when **both** project-local and checkout-local records are `None`. `enable_checkout_sync` writes that repo-slug record on **every** opt-in.

**Consequence: a fresh clone of an already-opted-in repository drains events that Chain A denies.** The clone has no project-local record — Chain A's absence-denies rule refuses — but Chain B finds the inherited repo-slug default and permits.

## The canonical replacement already exists

`src/specify_cli/sync/body_upload.py:54` — `project_consents_to_hosted_sync(...)`, with the rationale for why it is the correct shape at `:60-88`. The migration is to route the two enforcement sites through it.

## Scope

Two enforcement call sites. The display reads can follow or stay. This is not a request to delete `routing.py` or to unify the modules.

---

Found during `#3108` planning; cited at `bb2020fea` and re-verified at `abca7ec9`. Not fixed there because it is outside that mission's tracker-egress boundary.

### LINKED ISSUE #3187: SaaS tracker retry tests assert on sleep call counts and flake non-deterministically on CI
Three tests in the SaaS tracker client assert on **`sleep` call counts**, which makes them timing-dependent rather than behavioural. They fail non-deterministically on CI with wildly varying numbers:

| Test | main (`abca7ec96`, run `30874653016`) | PR #3175 (`95b2cc316`, run `30895602726`) |
|---|---|---|
| `tests/sync/tracker/test_saas_client.py::TestPolling::test_exponential_backoff_intervals` | `assert 319 == 3` | `assert 54 == 3` |
| `tests/sync/tracker/test_saas_client_origin.py::TestSearchIssues::test_429_retries_then_raises` | `Expected 'sleep' to be called once. Called 202 times.` | `Called 168 times.` |
| `tests/sync/tracker/test_saas_client.py::TestRetryBehaviors::test_429_respects_retry_after` | (not observed) | `Expected 'sleep' to be called once. Called 21 times.` |

The counts differ per run on the same code, which is the signature of a wall-clock-driven retry loop rather than a deterministic backoff assertion. The first two reproduce on `main` independently of any PR.

**Root fix, not a retry-to-green:** the retry path needs an injected clock/sleep seam so the tests can assert the *backoff sequence* (the actual contract — intervals and cap) instead of how many times `sleep` happened to be reached inside a real time budget. Per [testing-flakiness.md](https://github.com/Priivacy-ai/spec-kitty/blob/main/docs/development/testing-flakiness.md), fix correctness flakes at the root; never retry-to-green.

Separately, the same job reports an FR-007 process-global leak-guard error — `tests/sync/test_background.py::TestStartStop::test_stop_idempotent` leaves a live thread (`docs/development/process-global-inventory-3115.md`) — which belongs to the #3115 family rather than here.

Classified during the PR #3175 landing pass as pre-existing sync-family flakiness and deliberately not folded: adding a clock seam to the SaaS client's retry path is sync-surface product work outside that mission's blast radius.

### LINKED ISSUE #3190: scripts/verify_batch_retirement_3167.py: two C901 complexity reds (25 and 21 vs ceiling 15)
Filed as a follow-up rather than absorbed, per the rule that produced the `verification-trust-3115` follow-up programme.

`scripts/verify_batch_retirement_3167.py` landed in mission `chain-b-consent-bypass-3167-01KZ63HK` (WP01) as the committed, re-runnable deletion-manifest closure for retiring `sync/batch.py`'s dead queue-backed senders. It carries two `ruff` C901 findings against the repo's ceiling of 15:

| Function | Complexity | Note |
|---|---|---|
| `_classify_external` | 25 | the load-bearing CODE / STR-TARGET / PROSE classifier |
| `main` | 21 | reduced from 28 → 30 → 21 during WP01 by extracting two helpers, output byte-identical at each step |

**Not CI-blocking today**, and that is the point of filing it: `.github/workflows/ci-quality.yml` runs `ruff check src tests`, so `scripts/` is outside the linted set. `ruff check . --exclude scripts` reports `All checks passed!` — this file is the only `ruff` red in the repo, and nothing reds on it.

**Deliberately not fixed in-mission.** `_classify_external` is the classifier whose correctness the whole deletion rests on; it already caught two resolver bugs that would otherwise have shipped (a seeded-fiat dead set that could not test the mission's own premise, and a vacuous ALIVE tier). Refactoring it for a lint score during the mission it is certifying would trade real risk for a number. No `# noqa` was added — the charter's prohibition on suppression is honoured.

**What a fix must change:** extract the classifier's four resolution modes (direct import / module-alias attribute / string patch target / prose fallback) into named helpers with the same return contract, and confirm `.venv/bin/python scripts/verify_batch_retirement_3167.py` still reports `TOTALS dead=33 (first=24 second=9) alive=21` over 3905 files, with all five enumerations byte-identical.

Reproduce:
```
.venv/bin/python -m ruff check scripts/verify_batch_retirement_3167.py
```

Found by the WP01 reviewer; the mission's own manifest recorded it as a candidate but a candidate is not a follow-up issue.

### LINKED ISSUE #3219: Extract canonical flatten_coordination_metadata primitive (unify executor / doctor / mission-close)
## Problem

The three-mutation "flatten a coordination Mission's `meta.json`" operation — `del coordination_branch` + `pop topology` + `flattened=True` — is **re-inlined at every call site** instead of living behind one canonical primitive:

- `src/specify_cli/merge/executor.py` `_flatten_coordination_metadata_after_branch_delete` (added by #3086 / PR #3218)
- `src/specify_cli/cli/commands/_coordination_doctor.py:816-826` (`doctor coordination --fix`)
- `src/specify_cli/cli/commands/mission_type.py` `_flatten_discarded_mission` (`mission close --discard`) — **partial and divergent**: it calls only `clear_coordination_metadata` (clears `coordination_branch`) and does **not** pop `topology`, so a discarded coord Mission that survives with a stale stored `COORD_*` topology can still route through coordination and hit the same `CoordinationBranchDeleted` crash #3086 fixes.

`clear_coordination_metadata` (`src/specify_cli/mission_metadata.py:721-747`) only performs **1 of the 3** mutations, which is what forces each caller to re-inline the other two. The `topology` / `flattened` keys are semantically owned by `src/specify_cli/migration/backfill_topology.py` (`_TOPOLOGY_KEY`, `_FLATTENED_KEY`), yet no site routes through that owner. This is the 4th touch of the same field-set across #2069 → #2120 (discard) → #2614 (doctor) → #3086 (merge) — a textbook whack-a-field SSOT smell surfaced by the PR #3218 landing-review squad (paula-patterns MAJOR; reviewer-renata / architect-alphonso NOTE).

## Proposed fix

Extract one canonical `flatten_coordination_metadata(feature_dir)` in `mission_metadata.py` that performs **all three** mutations in a **single** load→mutate→`write_meta(validate=False)` (also closing the current double-write / mid-flatten-crash window in the executor path), importing the `topology`/`flattened` key constants from `backfill_topology.py` to preserve ownership. Converge all three call sites onto it, correcting the `mission close --discard` partial-flatten latent bug in the process. Each converged site keeps its own coverage.

**Deliberately deferred from PR #3218** (P0 hotfix): the extraction rewrites three call sites across three modules, well beyond #3086's blast radius. The duplication is real and recurring, so this follow-up is not optional.

## Also captured from the #3218 landing review (smaller, verify during this work)

- **`--push` origin divergence** (architect-alphonso MINOR): `_phase_push` (phase 11) runs before cleanup (phase 12), so the merge-time flatten's bookkeeping commit lands local-only; on a `--push` merge, origin/target keeps the stale `coordination_branch`. #3086's crash is on local resolves so the P0 is fixed, but confirm whether origin is ever a coordination-state resolution source and, if so, close the gap.
- **Name-divergence trust** (reviewer-renata MINOR, #1589 edge): the merge flatten keys its clear on the presence of the `coordination_branch` key, not on confirmation that the specific ref was deleted. It degrades safely (orphaned branch, no crash), but the trust could be documented or tightened.

## References

- Surfaced by: PR #3218 (fix for #3086) landing-review squad.
- Suggested parent epic: #1878 (coordination placement/identity strangler) or #2400 (metadata single-canonical-source) — triage to confirm.

### LINKED ISSUE #3220: Pre-existing failure: test_event_sourced_review_result_this_missions_own_meta_json_fixture asserts on a live, mutable repo fixture
## Command run

```
python -m pytest tests/migration/test_verdict_provenance_backfill.py tests/status/test_reducer.py -q
```//
also reproduces standalone:
```
python -m pytest tests/status/test_reducer.py::TestEventSourcedReviewResultReader::test_event_sourced_review_result_this_missions_own_meta_json_fixture -q
```

## Failure

```
tests/status/test_reducer.py:1036: in test_event_sourced_review_result_this_missions_own_meta_json_fixture
    assert "status_phase" not in meta
E   AssertionError: assert 'status_phase' not in {'baseline_merge_commit': 'd8fefa8c5cbe9c3bf08cfa41492f92858ccca313', 'coordination_branch': 'kitty/mission-review-cycle-verdict-seam-rebuild-01KZ2W7W', 'created_at': '2026-08-03T03:56:20.369377+00:00', 'flattened': False, ...}
```

## Root cause

`tests/status/test_reducer.py::TestEventSourcedReviewResultReader::test_event_sourced_review_result_this_missions_own_meta_json_fixture` reads a LIVE, mutable file from this repository's own tracked corpus as its fixture:

```python
repo_root = Path(__file__).resolve().parents[2]
feature_dir = repo_root / "kitty-specs" / "review-cycle-verdict-seam-rebuild-01KZ2W7W"
meta = json.loads((feature_dir / "meta.json").read_text(encoding="utf-8"))
assert "status_phase" not in meta
```

The test's own docstring frames this as pinning "this mission's OWN, real, un-migrated `meta.json` (no `status_phase` key)" as a literal example of an un-migrated mission. Since that test was written, `kitty-specs/review-cycle-verdict-seam-rebuild-01KZ2W7W/meta.json` has gained a `status_phase` key (visible in the current failure output: `coordination_branch`, `baseline_merge_commit`, `flattened`, etc. are present but the dict was truncated in the assertion diff — `status_phase` is now present too), so the incidental "no status_phase key" precondition the test asserts on no longer holds for that mission.

The test's own comment says the reader "is never gated on `status_phase`" (i.e. `event_sourced_review_result` doesn't care about that key at all) -- so the assertion is testing an accidental property of a live fixture, not the reader's actual behavior. As the repository's own `kitty-specs/` corpus evolves (further migrations, mission lifecycle changes, etc.), any real mission directory used this way is liable to drift out from under the test.

## Why I believe this is pre-existing, not introduced by my change

- Confirmed via `git stash -u` (stashing all changes from WP02 of mission `verdict-seam-write-unification-01KZ9Q35`, which only touches `src/specify_cli/migration/verdict_provenance_backfill.py`, `tests/migration/test_verdict_provenance_backfill.py`, and an out-of-map row in `tests/architectural/verdict_seam_census.yaml`) and re-running the single failing test in isolation -- it fails identically with the stash applied (i.e. on the pre-WP02 tree).
- My change never touches `tests/status/test_reducer.py`, `event_sourced_review_result`, or `kitty-specs/review-cycle-verdict-seam-rebuild-01KZ2W7W/meta.json`.
- `spec-kitty agent action implement WP02 --agent claude --mission verdict-seam-write-unification-01KZ9Q35` itself reported `Baseline: 1 pre-existing test failure(s) captured` at claim time, consistent with this being an already-known-red baseline test, not a regression from this WP.

## Suggested fix

Replace the live-repo-fixture read with a synthetic `tmp_path`-based fixture that constructs a `meta.json` without a `status_phase` key directly (matching the pattern the neighboring `test_event_sourced_review_result_never_reviewed_wp_slot_absent` test already uses), so the test's assertion is no longer coupled to the current, mutable shape of an arbitrary real mission directory elsewhere in the repository.

MOES-Media:
This PR removes dead queue-backed event-drain code from `sync/batch.py`, closing a latent consent-bypass-if-rewired path. The journal dispatcher remains the only live event drain. No behaviour change for existing projects.

Closes #3167, closes #3190, closes #3187, closes #3220, closes #2804, closes #3219.
Advances #3260 (regression and pre-review gate reliability) - this pass makes the regression gate mean exactly one thing, and stops tests from being selected by the wrong CI gate.

## What changed

- Deleted `sync/batch.py`'s queue-backed event senders (`batch_sync`, `sync_all_queued_events`) plus 31 further symbols and 91 test nodes that existed only to support them.
- Removed the now-inert `E15` egress-allowlist entry that had marked this dead code `UNREACHABLE`.
- Fixed the `tests/sync` autouse fixture that had been silently granting "consent" to nearly every test in the folder - the reason a broken gate on this path went uncovered for as long as it did.
- Hardened the standing anti-reintroduction guard (`test_no_queue_drain_constructed_3030.py`) with a positive control, so it can no longer pass vacuously just because the deleted names stopped existing.

## Landing-pass folds

**Remediation:**
- `fix(landing)`: joined the cancelled background timer thread in `BackgroundSyncService.stop()` - closes an FR-007 leak-guard teardown race that was flagging unrelated tests in `fast-tests-sync` / `integration-tests-sync`.
- `test(landing)`: accounted for `verdict_provenance_backfill`'s silent-by-contract `load_meta` call site - a latent pre-existing red on `main` from mission 191, surfaced (not caused) by this branch's rebase.
- `docs(landing)`: added the `CHANGELOG.md` entry this PR was missing.

**Test-suite reliability (advances #3260):**
- Enforced the `tests/regression/` exit rule: the suite now collects exactly one test, the still-open red-first `#2782`.
- Relocated, un-marked in place, or de-duplicated 18 green tests that were still carrying the `regression` marker after their underlying defects were already fixed.
- Relocated 15 further green guards that were squatting in `tests/regression/` with no marker at all, to their functional homes.
- The `misc` CI shard no longer lists `tests/regression/` (singular); the dedicated `regression tests (blocking)` job is now its sole owner.

## Ticket dispositions

The relocated guards cover already-closed issues `#2404`, `#2646`, and `#2996`. `#2804`, `#3219`, and `#2367` remain OPEN, but their reproductions are now green permanent guards living in functional homes - recommend a maintainer verify remaining scope and close. `#2367` may still owe a separate red-first repro for its open remainder (this pass only closed out its Mechanism B strand).

## What stays red (pre-existing on main)

- `regression tests (blocking)` on `#2782` - deliberate red-first P0 per ADR 2026-07-17-1.
- `integration-tests-sync` on `test_dossier_sync_fires_during_mission_creation` - `CharterPackConfigError`, part of the in-flight PACKS_ROOT/charter-provisioning reconciliation on `main`.
- `quality-gate` - aggregates the two jobs above.

Each of these is also red on `main`, independent of this branch.

🤖 Generated with [Claude Code](https://claude.ai/code/session_01Dxg2TtoDLZWs58tQjaLrkS)

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3209"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788503432&installation_model_id=427282&pr_number=3209&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3209&signature=7fb2a8f6fa031b428a6f646ae3996ec20b1927c2902724655d68c87b5eec5dde"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
## Remediation summary

**Verdict: draft, for operator review.** Two commits, sliced per the charter's `clean-linear-commit-history` tactic — governance dossier bunched, code as one reviewable slice. Rebased onto `96494e5ec`, the current `upstream/main` at the time of writing.

### Base hygiene

The branch was 93 commits behind. That mattered more than usual here: `git diff upstream/main HEAD` on the un-rebased branch listed ~180 changed files, because everything upstream had *added* read as a deletion — `decisions/ownership.py`, the charter sole-door tests, the regression-test renames. Applying "the changed code paths" from that list would have reverted 93 commits of other contributors' work. The real change set, taken from the merge base, is **18 code files**. Verified separately that none of those 93 commits touches a file this mission owns.

**The fork's `main` was deliberately not fast-forwarded.** `MOES-Media/main` carries **263 commits `upstream/main` lacks** (and lags by 473). The landing guide's step ⑥ says to fast-forward it; doing so would destroy that work, and an intermediate fork PR would drag hundreds of upstream commits into a range that commitlint scans. Cross-fork draft opened directly instead; its range is measured against `upstream/main` regardless.

| Gate | Result |
|---|---|
| `commitlint --from upstream/main --to HEAD` | **exit 0** (real exit status, not read through a pipe) |
| `ruff check` on all 12 changed source/test files | **All checks passed** |
| `test_batch_drain_retired_3167` + `test_egress_consent_boundary` + `test_batch_split_single_authority` + `test_inline_meta_read_gate` + `test_gate_coverage` + `test_no_legacy_terminology` + `test_no_queue_drain_constructed_3030` | **127 passed, 2 xfailed in 513.95s**, exit 0, `^ERROR tests/` = 0 |

The 2 xfails are pre-existing `#3113` pins carrying their own red-first quotes. `scripts/` sits outside CI's linted set (`ruff check src tests`), which is why the closure script's two `C901`s (#3190) are not CI-blocking — stated here because the dossier did not say it.

### Adversarial passes — four point-cuts, 13 lens dispatches

**Post-specify: the mission was rejected on its own premise and re-specified.** Two lenses independently found that `sync/batch.py`'s senders have no production caller; a third had demonstrated a real POST through the gate, but only by calling `batch_sync` directly — the construction the standing guard exists to prevent. Adjudicated from source rather than averaged. The mission changed from *migrate the gate* to *retire the senders*, and `FR-008` was withdrawn after the same pass showed the docstring it called false is true, because *"the drain"* is a bound term meaning `delivery/selection.py`.

**Post-plan: three lenses converged on two factual errors in the plan's own "measured" table.** A live helper (`_sleep_before_final_sync_retry`) was marked for deletion; two dead ones were marked production-alive on the strength of same-named symbols in other modules. Remedy: `scripts/verify_batch_retirement_3167.py`, a committed re-runnable closure that resolves module-qualified and classifies `CODE` / `STR-TARGET` / `PROSE`. The gate the plan cited as forcing the second tier turned out not to cover the file at all.

**Post-tasks: three deliverables had no file to land in** — the red-first assertion, the `contracts/` artifacts, and the `_leak_guard.py` residual. Ownership fixed before implementation. Grader-free checkboxes ("verified by node id, not by eye") were replaced with commands.

**Pre-merge: two SAFE-TO-LAND, one BLOCK. All three blocking findings remediated in this branch** — C-005's handoff was cited nowhere a successor looks (now a comment on #3130), `SC-003`'s evidence did not exist (measured: `5 passed`, exit 0), and the branch was neither compacted nor rebased. Two dossier numbers that did not reproduce were also corrected: an `851 passed` that reconciles to no nameable selection (real: 29 passed / 2 xfailed, with the command attached), and a post-arm error band of `{6,6,6,6,6}` that a second measurer could not reproduce (real: `{5,6}` at n=7).

### Red-first proof

`tests/architectural/test_batch_drain_retired_3167.py` was replayed against the pre-deletion tree: **`2 failed, 2 passed`**, the red naming all 33 symbols. After: **`4 passed`**. An independent lens re-derived it and separately confirmed the closure's three controls discriminate — an injected caller flips a symbol out of first tier, an injected *seed* caller makes the script **exit 1 rather than certify its own premise**, and dropping `background.py`'s import flips `run_final_sync_with_retries` from alive to dead.

### Pre-existing failures — filed, not fixed, not retried to green

- **#3193** — the `tests/sync` cone's error count is a **distribution** (5–6 at base, `{5,6}` after, n=7 across two measurers), not a scalar. One shape, `live thread … target=None`, produced by `_ChainedTimer`/`threading.Timer` and attributed by the leak guard's `after − before` difference to whichever test spans the thread's lifetime. **The observer moves; the leak does not.** No error in any run implicates a file this PR touches. **Neither measurer asserts this diff had zero influence** — both assert the class pre-exists.
- **#3188** — pre-existing marker reds in `tests/runtime/test_runtime_bridge_identity.py`.
- **#3192** — three live-path branches lost their only pin; their pins tested symbols in the frozen dead set, so keeping them was impossible. Three manifest rows say "partial only" rather than claiming coverage.
- **#3191** — a per-event forbidden-key screen turned out to be **still owed**. Checked, not assumed.

### Residuals filed

#3188, #3190, #3191, #3192, #3193, #3196, #3197, #3198, #3199, #3200, #3201, #3202, #3203 — all open, each with a reproduction and a statement of what a fix must change. Highest-value: **#3196**, `consented_project_uuids` writes machine-global config on read, and the grant survives removal of the project-local grant.

Disclosed rather than papered over: those eight residual citations use the `owner/repo#NNNN` form, which the issue-matrix parser does not match, so the completeness gate **cannot fail** for them. All eight were verified OPEN by hand instead.

### State

Fork CI validating; **PR left as a DRAFT for the operator.** `#3167`'s issue-matrix verdict is `deferred-with-followup`, deliberately not `fixed` — one site was retired rather than migrated, one deliberately unchanged. Not un-drafting without an explicit go, and not merging.

MOES-Media:
## Blocked, and the earlier triage was wrong

Correcting my own classification from the remediation summary above.

I had called the four `fast-tests-sync` reds "pre-existing class, but I cannot rule out our diff." The pristine-worktree reproduction rules *in* our diff:

| Arm | CI's selection `pytest tests/sync/ -m "fast and not windows_ci"` |
|---|---|
| pristine `upstream/main` `96494e5ec` | **`2116 passed, 11 skipped, 271 deselected` — EXIT=0, 0 FAILED, 0 `^ERROR tests/`** |
| this branch (CI) | **`2 failed, 2027 passed, 11 skipped, 2 errors`** |

`2116 − 2027 = 89`, exactly the nodes this PR deletes, so the arms are the same selection. **The shard is green on bare main and red here.** These are not pre-existing; the deletion changed what runs in the process, and the two `time.sleep` call-count assertions in `tests/sync/tracker/` plus two leak-guard teardown errors are composition-dependent. **The observer moved because this PR moved it.** Both nodes pass narrow on both arms, so a narrow run settles nothing.

Operator decision (2026-08-05): **hand `#3136` the evidence and let that mission go first**, rather than have a deletion mission absorb another mission's test-hardening work. The evidence is now on `#3136`, including two corrections to that issue's own record — `test_429_respects_retry_after` **does** now exhibit a failure, which falsifies the inventory column previously marked merely unverified.

**This PR stays a DRAFT until a composition-independent form of those assertions lands.** The previously planned programme order is inverted: the sync-cone mission now precedes this retirement's landing.

### Status of the six reds

| Check | Classification | State |
|---|---|---|
| `lint` | **PR defect — ours** | **Fixed.** The Contextive check found `src/specify_cli/.contextive/orchestration.yml` stale: a *generated* file had been hand-edited. The three drain senses now live in the source, `docs/context/orchestration.md`, and the file is regenerated — `All 13 generated file(s) are up-to-date`, exit 0. Amended into the code slice as `4478deb9b`. |
| `regression tests (blocking)` | **Pre-existing** | The `#2804` marker, `1 failed` on this PR's own base, red lands at `b04da00e13`. Tracked at `#3138`; no duplicate opened. |
| `fast-tests-sync` (2 failed, 2 errors) | **Caused by this diff** | Handed to `#3136`. Blocking. |
| `docs-freshness`, `cutover-guard`, `quality-gate` | **Unclassified** | Not established, and not asserted either way. `quality-gate` is likely a rollup; `docs-freshness` runs unfiltered on all PRs and the `lint` fix touched the same file, so it may clear on the re-run. |

Not un-drafting, and not merging.

MOES-Media:
## Ready for review — rebased onto current `main`, all five CI reds classified

Head `dd0fbab8c`, rebased onto `upstream/main` (0 commits behind at time of writing). Clean rebase, zero conflicts, all four sliced commits replayed.

### Verdict on every red

| Check | Verdict | Evidence |
|---|---|---|
| `docs-freshness` | **PR defect — fixed** | `DOCS-INDEX-DRIFT docs/context/orchestration.md`. Reproduced locally (`exit=1`, 1 ERROR); now `exit=0`, 0 ERRORs. |
| `cutover-guard` | **PR defect — fixed** | `status_phase not flipped despite event-log runtime evidence`. `meta.json` had no `status_phase` key; absent reads as `None`, which the guard correctly fail-closes on. Now `exit=0`. |
| `quality-gate` | **Derivative — not an independent red** | Its 6 blocking verdicts collapse to **2** root causes. `fast-tests-status` carries `needs: [changes, fast-tests-sync]` + `needs.fast-tests-sync.result != 'failure'` (`ci-quality.yml:1303-1309`), and both `integration-tests-{status,sync}` need *it* — so one failure cascades **4** skips, which the gate then reports as `improperly skipped`. |
| `regression tests` | **Pre-existing** | 6 nodes (`#2782`, `#2804`, `#2996` ×3, `#3086`). Proved by a **same-selection** comparison against the **merge ref** — 66 nodes byte-identical to pristine `main`, identical results across five runs (`5 failed, 60 passed, 1 skipped`). |
| `fast-tests-sync` | **Pre-existing** | 3 nodes in `tests/sync/tracker/`. Red on pristine `main` **at this PR's own baseline sha**, on the merge-base, and on **all 25 of the last `main` pushes**. |

### A correction to my own earlier analysis, since it was wrong in this PR's favour

I had classified `fast-tests-sync` as a **PR defect** on the arithmetic `2116 passed − 2027 passed = 89 missing tests`. That subtraction compared a **2127-node** selection against a **2040-node** one. It is not a defect — it is **this PR's declared payload**: 89 `tests/sync` nodes deleted, 2 added, and it reconciles to the digit (`89 − 2 = 87 = 2116 − 2029`, zero residue). Every one of the 91 divergent nodes maps 1:1 onto the diffstat.

Worth recording because the fix does not generalise: switching to the **merge ref** repaired the `tests/regression` comparison (where the divergence was pure staleness) but **cannot** repair this one, because this PR's whole purpose is to change the test selection. **A pass-count subtraction is licensed only after a proven byte-identical node set; for a PR that deletes tests, the only valid attribution path is per-node.**

The three actual failing nodes are `test_429_respects_retry_after`, `test_429_defaults_to_5s_when_missing` and `test_429_retries_then_raises` — all `Expected 'sleep' to be called once. Called 103/160/178 times.` Causal independence is structural: `git diff` over `tests/sync/tracker/` and `src/specify_cli/sync/tracker/` between the trees is **empty**, and the failing code lives in `src/specify_cli/tracker/` — a different package from this PR's `src/specify_cli/sync/{batch,runtime}.py`, importing neither. Root cause is a process-global `@patch("…time.sleep")` recorder counting sleeps from any thread in the xdist worker (filed as `#3115` FR-005); a successor mission for `#3136` is fixing it.

### Folds in this PR, each with its rationale

1. **`docs-freshness`** — regenerated `docs/development/3-2-docs-retrieval-index.yaml`. A `docs/context/*.md` glossary edit derives **two** generated surfaces, not one: the contextive glossary *and* the retrieval index. The earlier `lint` fix regenerated only the first. Regenerated at source, never hand-edited — and regenerated again after this rebase, because upstream registered new pages in it and re-applying a stale copy would have reverted them.
2. **`cutover-guard`** — flipped `status_phase` via the guard's own documented remedy, `spec-kitty migrate backfill-runtime-state`, not by hand-editing `meta.json`.
3. **Complexity** — `scripts/verify_batch_retirement_3167.py` had two functions at complexity **25** and **21** against the repo ceiling of **15**. CI never caught it: `scripts/**` is exempted from `TID251` only, but the lint job runs `ruff check src tests` and never looks at `scripts/`. Refactored by phase to max **12**, behaviour-preserving — stdout byte-identical (42 lines, 2453 bytes), stderr empty, exit 0 both sides, plus a differential control on an injected-caller tree where both versions exit 1 with identical stderr.
4. **The closure script could not refuse** — and this is the one worth a reviewer's attention. `_scan_external` classified only symbols *currently declared* in `batch.py`, but the retirement already deleted `batch_sync` and `sync_all_queued_events`, so those names never reached the classifier and the **exit-1 refusal was unreachable on the tree it ships with**. A re-runnable proof that cannot fail is not a proof. Fixed by classifying `symbols | SEEDS`, and proved as a differential *on the post-retirement tree*: the pre-edit script exits **0** on an injected-caller tree, the post-edit script exits **1** with `*** REFUSED: the mission premise is false for ['batch_sync'] ***`.

- `ruff check src tests` → `All checks passed!`; `ruff check` on all 13 changed Python files → `All checks passed!`
- `commitlint --from upstream/main --to HEAD` → clean
- `spec-kitty cutover-guard --base-ref upstream/main` → `exit=0`, "All diff-touched missions are cut over"
- `check_docs_freshness.py --ci` → `exit=0`, **0** ERROR lines
- `pytest tests/architectural/` on the previous base → `1691 passed, 2 skipped, 2 xfailed`, **0** `^ERROR tests/`
- `tests/sync` was **not** swept locally (a sibling mission holds that window; the two suites must never run concurrently)

### Filed upstream from defects found during this landing

`#3212` — `migrate backfill-runtime-state` reports `Flipped: 0` on a mission it flipped (counters key on `seeded_count`, never on `CutoverResult.flipped`) — on the printed remedy of a fail-closed required check.
`#3213` — a regression gate flips between enforced and skipped depending on which directories are in the selection: `tests/conftest.py:300` sets the flag via an autouse `monkeypatch` (too late for an import-time `skipif`), while six modules set it process-globally at import.

### What is not claimed

The full CI board will not go green on this PR. `regression tests` is designed to be red while open P0s exist (its own workflow comment says so), and `fast-tests-sync`'s three nodes need `#3136`. Neither is caused by this change, and neither is retried-to-green here.

Governance dossier travels with the PR: `kitty-specs/chain-b-consent-bypass-3167-01KZ63HK/**` (spec, research, plan, WP prompts, contracts, matrices, landing handoff).

stijn-dejongh:
Claiming this PR for landing: rebase onto upstream/main, adjudicate red checks, fold fixes as needed. Evidence to follow.

stijn-dejongh:
## Landing pass: remediation summary

**Verdict:** merge-ready pending green CI on `fb29d4cef` (2 maintainer folds pushed). Review below is code + tracker, not a rerun of the adversarial squad — the PR's own four squad passes (post-specify/plan/tasks/pre-merge) are taken as read.

### Red-check classification (merge-base `d0ed802cc`, PR was already rebased onto it)

| Check | Classification | Disposition |
|---|---|---|
| `arch-adversarial (arch_shard_2)` — `test_routed_load_meta_floor` | Pre-existing on main (128 vs live 133, identical text on both) | **Folded** — re-pinned floor 128→130 |
| `integration-tests-core-misc (auth-audit-git)` — `test_sync_batch_uses_factory` | **New on this PR** (green on main at the same commit) | **Folded** — stale assertion, see below |
| `fast-tests-sync` — `test_saas_client.py` sleep-count tests | Pre-existing on main, worse there (4 failed vs 2) | Not folded — tracked flake, #3187 |
| `integration-tests-sync` — dual-write/daemon-retirement leak-guard errors | Identical on main | Not folded — tracked, #3130 |
| `integration-tests-status` — `test_reducer.py` stale fixture | Identical on main | Not folded — already tracked, #3220 |
| `integration-tests-core-misc (integration)` — SC-004 concurrency test | Reproduces on main, **different failure mode each run** (confirmed non-deterministic) | Not folded — **untracked, filed #3235**, parented under #1931 |
| `regression tests (blocking)` — #2782, #2804 | `@pytest.mark.regression`-marked, issue-pinned, identical on main | Left red per ADR 2026-07-17-1 (deliberate P0 signal) |

### Folds (pushed to `pr/batch-drain-retirement-3167`)

1. **`test(landing): re-pin ROUTED_LOAD_META_FLOOR to the current live count`** — pre-existing saturated tolerance band, unrelated to this PR's diff. Re-measured live count directly (133), raised floor 128→130 to restore the established 3-below-live gap, dated entry added to the file's existing convention.
2. **`test(landing): retire the stale get_token_manager check on sync/batch.py`** — the one genuine PR-caused regression. `test_sync_batch_uses_factory` asserted `batch.py` calls `get_token_manager`; this PR deleted the only code in that file that ever called it (the retired transmit primitives), and `batch.py`'s own new module docstring says as much ("holds no transmit surface... performs no I/O of its own"). The assertion was checking a caller that no longer exists — not a reintroduced legacy pattern. Confirmed the module has zero `requests.*` and zero transmit code before folding. FR-017 coverage is unaffected: `test_sync_background_uses_factory` still guards the actual HTTP caller.

Both folds verified green locally (targeted runs only, not a full architectural pass): `test_routed_load_meta_floor` (1 passed), full `test_inline_meta_read_gate.py` (40 passed), full `test_transport_rewired.py` (9 passed), `ruff check` clean on both files, `commitlint` clean across all 6 commits in range, `test_no_legacy_terminology.py` clean.

### PR body / changelog style

- **BLUF:** the opening paragraph leads with internal mechanism (symbol counts, gates, allowlist entries) rather than impact for a user/operator. Worth a rewrite pass per [review-gates.md#pr-body-style](../../docs/development/review-gates.md#pr-body-style-consumer-focused-bluf) — not blocking, since the PR's own transparency about scope is otherwise exemplary.
- **Changelog:** none added, self-disclosed in the body rather than omitted silently. Given the retirement touches a latent (never-exercised) consent-bypass vector, a brief "Security"-category entry would be reasonable for security-conscious operators even though there's no behavior change — operator's call.

### Issue tracker hygiene (this landing pass)

- Filed **#3235** (SC-004 concurrency flake, untracked before this pass), parented under #1931.
- Dispatched `planner-priti` and triaged all **13 residual issues** from the mission (#3188, #3190–3193, #3196–3203) per [manage-issue-tracker.md](../../docs/development/manage-issue-tracker.md): native type, parent epic (#1800 sync epic, #3130 leak-guard umbrella, #1931 test-suite-friction, #1928 lint/Sonar debt, #1797 dead-code/LOC, #1629 glossary), priority + domain labels, milestone `3.2.x` on the P1 sync-hardening items.
- **#3190 (script C901 complexity) is already fixed on this branch** by the two commits that landed after the mission dossier (`_classify_external` 25→12, `main` 21→5, `ruff check scripts/verify_batch_retirement_3167.py` clean) — the PR body's residuals list hadn't caught up. **Updated the PR body**: `closes #3190` added alongside `closes #3167` so it auto-closes on merge. Verified the other 12 stay open (no closing keyword) — correct, since they're genuinely still open follow-ups.

### State

Watching CI on `fb29d4cef`. Merge-ready on green; the only remaining reds are the pre-existing/tracked items above, which are expected to stay red per the never-retry-to-green policy.

stijn-dejongh:
## Landing pass, round 2: stability-motivated fixes beyond this PR's own diff

Per operator direction, went beyond classifying pre-existing reds and fixed the ones directly reachable from this landing pass. Six new commits, `c165a2fee` is the current head.

### #3187 — SaaS retry tests flaking on `sleep` call counts (fixed, closes with this PR)

Root cause (already forensically identified across 3 investigation rounds, preserved in the test's own docstring): `@patch("...saas_client.time.sleep")` patches the single process-wide stdlib `time.sleep` attribute, so the mock recorded calls from *anything* alive in the process during the patch window — `subprocess.Popen._wait`'s internal busy-wait, or a leaked thread from an unrelated test. Fix: `SaaSTrackerClient` now binds `self._sleep = time.sleep` once per instance; both retry/poll call sites use `self._sleep(...)`. Tests patch `client._sleep` directly — immune by construction. Stress-tested under the exact contamination-prone conditions (`tests/sync/` full cone, `-n auto --dist loadfile`): 0 sleep-count failures across 2290 passed.

### #3220 — stale live-repo fixture in `test_reducer.py` (fixed, closes with this PR)

The test read this repo's own `kitty-specs/review-cycle-verdict-seam-rebuild-01KZ2W7W/meta.json` as a fixture, asserting it lacked `status_phase`. That mission was later migrated and gained the key. Replaced with a synthetic `tmp_path` fixture reproducing the same coord-topology-primary-partition shape deterministically.

### #3188 — pre-existing marker reds (closed, no code change)

Already fixed on `main` by an unrelated commit (`40bdfeffc`) before this landing pass started; never closed. Verified from a clean non-dot-prefixed path and closed with evidence.

### #3130 — 11 pinned process-global/thread leaks in `tests/sync` (10 of 11 fixed)

Root-caused and fixed every pinned node except one:
- 4 in `test_background_body.py`, 2 in `test_runtime.py`, 2 in `test_issue_598_hang_fixes.py`, 1 in `tracker/test_saas_client_consent_gate_3030.py` — each missing a `.stop()`/`reset_*()` call after starting real background work (threads, timers, singletons).
- 2 pinned nodes in `test_target_authority.py`/`test_target_authority_wiring.py` — but fixing only those 2 was insufficient: **11 raw `os.environ[...]` writes total** across both files, and the 9 unpinned ones immediately re-dirtied the baseline the pin's `requires_clean_baseline` framing depends on. Converted all 11 to `monkeypatch.setenv`.
- `test_lifecycle_readiness.py`: fixed 2 of its 4 markers outright; narrowed (not removed) the pin for the other 2 — debug instrumentation showed `sync.runtime._runtime` is already back to `None` by the time this test's own cleanup runs, meaning something else (likely a background final-sync retry) resets it mid-test without joining the thread it abandons. Filed as **#3237** (production-code tracing, not test hygiene) rather than chased further.

**Bonus, same class, previously unfiled** (flagged in #3130's own comment thread as "6 more... nobody has said so," and the actual cause of two of this PR's own red CI checks): `test_daemon_self_retirement.py` (2 nodes — a daemon-flagged bootstrap thread with a 1s deferred import that can outlive `monkeypatch`'s fake and bootstrap a real singleton after the test returns) and `test_dual_write_integration.py` (3 nodes — one module-level autouse fixture reset).

Full `tests/sync/` cone under `-n auto --dist loadfile` (the exact conditions #3130 was filed against): **6 errors → 0.** The 3 remaining local failures in that run are an unrelated `/tmp/.kittify` artifact specific to this dev machine, not reproducible on a clean CI runner.

### #3235 — corrected, not "flaky" (test extracted, production bug still open)

Originally filed mid-pass as an unexplained flake. On closer reading, `tests/integration/test_review_durability_matrix.py`'s own module docstring already fully diagnoses this: a real **P0 data-loss race** — `create_rejected_review_cycle`'s commit phase has no protection against two OS processes racing `git add`/`git commit` in the same working tree, so a reviewer's verdict can be silently lost under genuine concurrency (WP15's own "SC-004 — NOT MET" finding). It was never marked `@pytest.mark.regression` or issue-pinned, so it read as ordinary CI flakiness instead of the deliberate signal it is.

Retitled #3235 to reflect the actual finding, reparented under the review-artifact-integrity epic (#3044, not the test-suite-friction one), and extracted the test to `tests/regression/test_issue_3235_sc004_review_cycle_commit_race.py` with `@pytest.mark.regression` added, per `tests/regression/README.md`'s entry rule and the established `#2782`/`#2996` extraction convention. **The production race itself is not fixed** — per the original WP's own test-only mandate ("if a fix seems to need production code, STOP and report"), that's dedicated follow-up work, now correctly tracked and pinned rather than silently red.

### Verification

`ruff check` clean, `mypy` clean (new code), `commitlint` clean across all 9 commits in the PR range, terminology guard clean. Every un-pinned/narrowed leak-guard entry verified via the guard's own self-proving mechanism before the pin was touched.

### State

Pushed to `pr/batch-drain-retirement-3167` (head `c165a2fee`). PR body updated with `closes #3187, closes #3220` alongside the existing `closes #3167, closes #3190`. Watching CI on the new head next.

MOES-Media:
## `arch_shard_1` fold — the renamed reducer node-id, swapped in the NFR-001 committed floor

Pushed `a7a913dd8`. The gate reported:

```
FAILED tests/architectural/test_mission_exit_baseline.py::test_committed_floor_present_in_live_collection
NFR-001 VIOLATION: 1 committed node-id(s) vanished from a live --collect-only of the affected-suites
invocation (… never treat this as an improvement):
['tests/status/test_reducer.py::TestEventSourcedReviewResultReader::
  test_event_sourced_review_result_this_missions_own_meta_json_fixture']
```

**The rename is correct and stands.** This PR renamed that test to `…_coord_primary_partition_slot_absent` because the original read this repository's **own live** `kitty-specs/review-cycle-verdict-seam-rebuild-01KZ2W7W/meta.json` as a fixture. That mission was later migrated and now carries a `status_phase` key, so the live precondition silently stopped holding and the test went red for a reason unrelated to the reader it exists to cover. The replacement builds the same shape synthetically under `tmp_path`. A test whose fixture is a live artifact the repo keeps editing is only true at the instant it was written.

What was missing was the corresponding deliberate update to the committed floor.

### Why a targeted swap rather than a regeneration

- The old id sits at `mission_exit_baseline.txt:2567`, **exactly once**, and the file was byte-identical to `upstream/main` — this PR had not touched it at all.
- It is **not** either hard pin in `test_mission_exit_baseline.py` (`_ISSUE_3157_NODE_ID`, `_ISSUE_3160_NODE_ID`), so `test_known_baseline_pins_are_in_the_floor` is unaffected. Re-confirmed after the edit that both still resolve.
- The new id was proved to collect **by name** from a live `--collect-only` of the real affected-suites invocation (2890 node-ids), not just from its own file.
- **The load-bearing check:** a full set-difference of floor-minus-live returned **exactly one** entry — the old id, and nothing else. That is what proves the swap is sufficient *and* that no second selection regression was hiding behind this red. A wholesale regeneration would have concealed precisely that.

The file's header does document a sanctioned regeneration path — *"Regenerate only when the affected-suites invocation itself legitimately changes scope"* — but that condition is **not met**: the invocation's path list is unchanged. A test rename is not a scope change.

One refinement over the obvious approach: the floor is **globally sorted** (`sort` is a fixpoint), so a literal in-place one-line swap would have broken that invariant. This is a delete-plus-insert at the correct sorted position — one id out, one in, count unchanged at 2886, zero duplicates, sort preserved. Rationale recorded in the header as a `2026-08-06 TARGETED RENAME SWAP` block, matching the file's existing `2026-08-05 REBASE RE-ANCHOR` convention; `_load_baseline` strips `#` lines, and the node-ids inside that comment are deliberately line-broken so a future `grep -F "$OLD_ID"` still returns zero.

**Results:** `test_committed_floor_present_in_live_collection`, `test_baseline_file_is_non_empty`, `test_known_baseline_pins_are_in_the_floor` — 3 passed. Both touched modules in full: **48 passed**. `ruff check` clean across `tests/architectural/`; `mypy --strict` clean on both Python files. No baseline regenerated.

---

## ⚠️ Merge-order hazard with #3247 — please read before merging either

Both this PR and **#3247** re-pin `ROUTED_LOAD_META_FLOOR` in `tests/architectural/test_inline_meta_read_gate.py`, and **both also rewrote the comment block immediately above it**, so the conflict region is the whole block, not one line.

| | value | line | live routed census |
|---|---|---|---|
| `upstream/main` | 128 | 231 | — |
| **this PR** | 130 | 245 | 133 |
| **#3247** | 131 | 326 | 134 |

**#3247 owns this gate** — it is the mission actually adding routed `load_meta*()` call sites, which is why its live count is one higher. So **131 is the correct post-merge value and this PR's 130 is the one that goes stale.** Recommended order: merge **#3247 first**, then re-derive here rather than accepting a textual merge. Whichever lands second must re-measure via `scan_routed_load_meta_calls(SRC_ROOT)` and re-pin; the anti-vacuity margin check will not save a stale value silently.

Also corrected in this PR: the comment block previously read *"not caused by this PR's diff, which does not touch this file"* while commit `3b8f951d8` in this very PR does touch it. Now states what was actually meant — the rest of the diff adds no routed call site — and that claim was verified rather than asserted: `git diff upstream/main...HEAD -- src/` contains **zero** added or removed `load_meta` lines.

Both facts are recorded as **Hazard H5** in `kitty-specs/chain-b-consent-bypass-3167-01KZ63HK/LANDING-HANDOFF.md`. No issues filed, per the standing direction to ledger and assess for folding after PR.

stijn-dejongh:
## 🛬 Maintainer landing notification

Claiming **#3209** for today's landing pass (maintainer runbook: `docs/development/pr-landing.md`).

**Plan:**
- Rebase onto current `upstream/main` (`bccb4b4b5`) — the branch is currently `CONFLICTING`.
- Adjudicate the four red checks (`regression tests (blocking)`, `quality-gate`, `integration-tests-sync`, `fast-tests-sync`), classifying each as PR-defect / contract-crossed / pre-existing-main / perf-flake.
- Fold remediations as separate single-purpose `<type>(landing): ...` commits, delegated to profile-loaded subagents under model discipline (campsite-clean mode).
- Push folds to the PR branch, post the evidence trail, and hand off green + un-drafted for the operator to merge.

I will not merge — the operator merges on green ([step 11](https://github.com/Priivacy-ai/spec-kitty/blob/main/docs/development/pr-landing.md)). Evidence to follow.

stijn-dejongh:
## 🛬 Landing remediation summary

Rebased onto current `upstream/main` (`3ac01d247`) and adjudicated the four red checks. New tip: `f3ea84148`.

### Verdict
Merge-ready on green **modulo the honest P0 regression reds that also red `main`** (see below) — the code changes are sound; the four reds were all pre-existing on `upstream/main`, and this pass either fixes them or confirms them as deliberate red-main signals.

### Red-check classification (all four reproduced on `upstream/main` `bccb4b4b5`/`3ac01d247`)

| Check | Root cause | Disposition |
|---|---|---|
| `fast-tests-sync` | FR-007 leak-guard **teardown errors** — a cancelled `threading.Timer` daemon thread lingered in the live-thread set at teardown (timing-dependent which node is blamed; "fail the polluter"). Pre-existing on main. | **Fixed** (fold below). The 429-retry failure main carries here was already fixed by this PR's `SaaSTrackerClient` sleep-seam commit. |
| `integration-tests-sync` | Same FR-007 leaked-timer class (`test_daemon_self_retirement.py`). Pre-existing on main (main flags *more* leaked-thread nodes than this PR did). | **Fixed** (fold below). |
| `regression tests (blocking)` | `#2782` + `#2804` red-first P0 reproductions. | **Left red — deliberate.** Identical reds on `main`; per ADR 2026-07-17-1 these are honest open-P0 signals, not merge blockers. Not this PR's to green. |
| `quality-gate` | Aggregate of the above (also red on `main`). | Follows once the sync jobs green; the regression-job red is the honest-P0 baseline `main` shares. |

### Fold (1)
- **`fix(landing): join the cancelled timer thread in BackgroundSyncService.stop()`** (`f3ea84148`)
  `stop()` cancelled the timer under `_lock` but never joined its daemon thread, so under full-suite timing the thread could still be live when the FR-007 guard snapshots at teardown. Now: capture → cancel+null under lock → `join(timeout=5s)` **outside** the lock (the timer callback re-acquires `_lock`, so an in-lock join risks deadlock; `_running` is already `False` so no reschedule). Also reaps the raw `Timer` in `test_daemon_self_retirement.py` (`join` after `cancel`) and teaches the two `FakeTimer` test doubles a no-op `join`. Delegated to a profile-loaded `python-pedro` (sonnet) under model discipline; maintainer-adjudicated.
  Local evidence (rebased tip): `tests/sync/test_background.py` + `test_daemon_self_retirement.py` → **45 passed, 0 leak-guard nodes flagged** (4 timing runs); `ruff` + `mypy` clean; `tests/architectural/test_no_legacy_terminology.py` → 10 passed.

### Rebase adjudication — one contributor commit dropped
- **Dropped `test(landing): mark and extract the SC-004 review-cycle commit-race reproduction`** (was `c165a2fee`/`0648b118c`).
  That commit extracted the SC-004 probe into `tests/regression/test_issue_3235_…` as a **red-first P0 reproduction**, on the premise that SC-004 was *NOT MET*. Since then, mission **191 `verdict-seam-write-unification-01KZ9Q35`** (`a5ae131e4`) landed on `main` and **fixed** the race — `main`'s inline `test_sc004_two_concurrent_processes_never_clobber_a_verdict_over_50_iterations` now asserts **SC-004 MET** (event-log durability) and is green. Re-introducing a contradictory red-first repro would reintroduce a red for a criterion `main` already meets. Resolution: **keep `main`'s fixed inline test, drop the stale extraction.** No dangling references to the removed file remain (baselines/shard-maps clean). `#3235` the *ticket* is still open — but its SC-004 test bar is met in code; any residual tracking is mission 191's concern, not this PR's.
- Two `test(landing)` re-pin conflicts (`ROUTED_LOAD_META_FLOOR`, NFR-001 reducer node-id) resolved in favour of `main`'s already-landed #3248 re-pin (both sides converged on floor `130`).

### State
Force-pushed the rebased branch; **watching CI.** Expected end-state: `fast-tests-sync` / `integration-tests-sync` / `quality-gate` green; `regression tests (blocking)` red on `#2782`/`#2804` only — matching `main`'s honest-P0 baseline. Operator merges on that state (I don't merge).

@MOES-Media FYI on the dropped commit above — nothing you authored was lost in substance; `main` simply landed the real fix for the race your commit had pinned as red-first, so the extraction became obsolete.

stijn-dejongh:
## 🛬 Landing update — two more folds + PR body/changelog

Re-based onto latest `upstream/main` (`50685644a`) and pushed. New tip: `b030ea90d`. First CI run (`f3ea84148`) cleared `fast-tests-sync` (the leaked-timer fold worked) and surfaced two more pre-existing reds, both now handled:

### Additional folds
- **`test(landing): account verdict_provenance_backfill's silent-by-contract load_meta site`**
  `integration-tests-core-misc (specify-cli-rest)` → `test_no_unaccounted_load_meta_call_sites` flagged `migration/verdict_provenance_backfill.py::_resolve_mission_id`. This reader arrived with **mission 191 (`verdict-seam-write-unification`)**, which never joined it to the meta-fail-closed census ledger — so the gate was **latently red on `main`** from that merge (the file is absent at `bccb4b4b5`, the shard's last green run; this PR's rebase surfaced it). The reader is `on_malformed=\"none\"` / returns `None` (a backfill must tolerate legacy/bare meta), so the fix is a `silent-by-contract` accounting row with dated rationale — **not** a reroute through `load_meta_fail_closed`. Census gate passes locally with the row.
- **`docs(landing): changelog entry for the dead queue-backed event-drain retirement`**
  Consumer-first `[Unreleased] 3.2.6 / Changed` entry (authored via the `comms-cleo` profile). Normalised to ASCII punctuation to match `main`'s `50685644a` sweep (AGENTS.md UTF-8 discipline — em-dashes render blank on the dashboard). The **PR body** was also rewritten BLUF/consumer-first (impact → what changed → retirement rationale).

### Remaining red after this push — pre-existing, mission-scoped, NOT folded
- **`integration-tests-sync` → `test_dossier_sync_fires_during_mission_creation`: `CharterPackConfigError: CHARTER_PACK_CONFIG_INVALID`.** The FR-007 leaked-thread teardown errors are **gone** (fold worked; now `208 passed, 1 failed`). The sole residual reproduces on `main` (it was in `main`'s own `integration-tests-sync` failing set at `bccb4b4b5`) and locally on the rebased tip. It is the in-flight **PACKS_ROOT / charter-provisioning** reconciliation on `main` (main's own recent commits are actively fixing charter/packs fixture provisioning) — a mission-scope carve-out, not this PR's to fold. Leaving it as honest pre-existing breakage.

### Expected end-state (watching CI on `b030ea90d`)
- Green: `fast-tests-sync`, `integration-tests-core-misc (specify-cli-rest)`, and everything the first run already passed.
- Red (pre-existing, honest): `regression tests (blocking)` (`#2782`/`#2804`); `integration-tests-sync` (`test_dossier_sync` charter-provisioning, main-owned); `quality-gate` (aggregate of those). Each also reds `main`.

Operator merges on that state — the non-green jobs are `main`'s own honest baseline, not defects this PR introduces.

stijn-dejongh:
## 🏁 Landing complete — ready for operator merge

CI on `b030ea90d` settled to the expected **green-modulo-honest-reds** state. Every red is pre-existing on `main` and precisely accounted for; nothing this PR or its folds introduced is red.

### Green
All fast-test shards, all `integration-tests-core-misc` shards, `fast-tests-sync`, `integration-tests-core-misc (specify-cli-rest)` (census fold), `lint`, `arch-adversarial`, `drift-detector`, `docs-freshness`, `clean-install`, wheel/windows/timing — green.

### Residual red — pre-existing, honest, main-owned (NOT merge defects)
| Job | Failing test | Why it stays red |
|---|---|---|
| `regression tests (blocking)` | `test_issue_2782_sync_strict_json_ingress_skip` (1 failed, 18 passed) | Deliberate red-first P0 reproduction per ADR 2026-07-17-1 — leave red until `#2782`'s product fix lands. (`#2804` now passes post-rebase.) |
| `integration-tests-sync` | `test_dossier_sync_fires_during_mission_creation` → `CharterPackConfigError` (1 failed, 208 passed) | Pre-existing on `main` (in its own `bccb4b4b5` failing set) + reproduced locally. In-flight **PACKS_ROOT / charter-provisioning** reconciliation on `main`; mission-scope, not folded. The FR-007 leaked-thread teardown errors that also reded this job are **fixed**. |
| `quality-gate` | — | Aggregate of the two above; red on `main` for the same reasons. |

### Hand-off deliverable
- ✅ Rebased onto current `upstream/main` (`50685644a`); not a draft; full evidence trail above.
- ✅ 3 landing folds (`fix`/`test`/`docs` (landing)), each single-purpose; 1 stale contributor commit (SC-004 `#3235` extraction) dropped because `main` landed the real fix (mission 191).
- ✅ PR body + changelog rewritten consumer-first.
- **Landing-order note:** none — this PR carries no structural cutover that forces an order on other PRs in the pass.

Optional: once `main`'s charter-provisioning reconciliation lands, a final rebase will clear `integration-tests-sync` too. Otherwise this is merge-ready on the honest-reds baseline. **Operator merges — I do not.**

stijn-dejongh:
## 🧹 Campsite fold — regression-suite exit-rule cleanup (7 folds)

Enforced `tests/regression/README.md`'s exit rule: a `@pytest.mark.regression` test that has gone **green** (its P0 defect is fixed) must leave the marker/suite — relocated to its functional home with correct markers, or deleted if a permanent guard already covers it. Delegated to a profile-loaded `python-pedro` (sonnet); maintainer-adjudicated.

**Acceptance (verified locally):** `pytest tests/ -m regression --co` now collects **exactly one** test — `test_issue_2782_...::test_mission_create_json_strict_when_sync_skips_ingress` (`#2782`, still open, still the one true red-first). All 58 relocated tests + baseline gates pass.

| Was in `tests/regression/` | Issue | Disposition |
|---|---|---|
| `test_issue_2782` | #2782 OPEN/red | **Kept** — the only red-first |
| `test_issue_2996_self_referential…` | #2996 CLOSED | **Deleted** — byte-identical dup of `test_cycle.py:170` |
| `test_issue_2996_approval_after_rejection…` | #2996 CLOSED | **Deleted** — the flagship survives verbatim in `tests/integration/test_review_cycle_rejection_only.py`, which `test_verdict_name_truthfulness.py` *requires* by exact file+function (FR-017 AC2). Relocating would have made a second, non-gate copy. |
| `test_cycle.py` (6, in `tests/review/`) | — | **Un-marked in place** (already home); docstrings → permanent-guard |
| `test_issue_2404_acceptance_matrix_write_surface` (4) | #2404 CLOSED | **Relocated** → `tests/integration/` |
| `test_2646_stale_verdict_closes_via_fr001` (2) | #2646 CLOSED | **Relocated** → `tests/integration/` **+ `mission_exit_baseline.txt`/`.py` path-updated** (its only baseline coupling; collected set unchanged) |
| `test_issue_2804_merge_resets_gate_artifacts` (1) | #2804 **OPEN** but green | **Relocated** → `tests/merge/`; docstring flags #2804 still open (broader scope unverified) |
| `test_mission_close_discard_pops_topology` (3) | #3219 **OPEN** but green | **Relocated** → `tests/specify_cli/cli/commands/`; docstring flags #3219 open for the broader primitive-extraction refactor |

Each disposition is its own `test(landing):` fold. Markers swapped per `docs/context/testing-taxonomy.md` (dropped `regression`; kept `git_repo`/`integration`; added type markers). `marker_baseline`, `golden_count`, and `mission_exit_baseline` gates all pass.

**Follow-ups flagged (out of scope here):** `#2804` and `#3219` can likely be closed — their reproductions are green and now permanent guards; recommend a maintainer verifies remaining scope. Also, ~15 files remain under `tests/regression/` that carry **no** `regression` marker (pre-existing) — worth a separate look at whether they belong there.

stijn-dejongh:
## 🏁 Landing pass complete — final state on `a00df9699`

CI settled at the intended **green-modulo-honest-reds** state. The test-suite-reliability folds (regression exit-rule cleanup, 15 green-guard relocations, and the `misc`-shard ownership fix) tripped **no** gate:

- ✅ `arch-adversarial` (all 3 shards), `integration-tests-core-misc` (all shards incl. `misc`), `fast-tests-merge`, `fast-tests-lanes` — green. The relocated tests pass in their functional homes; the repointed cross-file imports (`tests.merge.*`) hold; the `misc` shard no longer owning `tests/regression/` (singular) is correct.
- ✅ `regression tests (blocking)` now collects **exactly one** test (`#2782`) — the exit-rule cleanup is confirmed in CI.

**Residual red — pre-existing on `main`, not this PR's:**
- `regression tests (blocking)` → `#2782` (deliberate red-first P0, ADR 2026-07-17-1).
- `integration-tests-sync` → `test_dossier_sync_fires_during_mission_creation` (`CharterPackConfigError`; in-flight PACKS_ROOT/charter-provisioning reconciliation on `main`).
- `quality-gate` → aggregate of the two.

PR body refreshed with the full fold inventory, `Advances #3260`, and ticket dispositions. **Ready for operator merge** on this honest-reds baseline. Recommend closing `#2804`/`#3219` (and the Mechanism-B strand of `#2367`) after a scope check — their reproductions are now green permanent guards in functional homes.

Issue #3217, opened 2026-08-05
verdict-seam census AST derivation misses helper-constructed records (SC-008 blind-spot)


stijn-dejongh:
Follow-up from the PR #3211 landing pass, surfaced by the architecture review lens.

`tests/architectural/test_verdict_seam_census.py`'s AST/token-scoped `_derive_census` classifies writers/resolvers/readers by matching a predicate on direct call/construction shapes. It **misses record constructions via helpers it does not classify** — the check's own docstring names one live example outside the census: `src/specify_cli/migration/backfill_runtime_state.py::_review_from_frontmatter`.

Consequence: SC-008's guarantee ("the check fails when a new reader is introduced") holds only for members matching the classifier predicate — a helper-wrapped writer/reader can be added without tripping the ratchet. Honest, self-disclosed by the check, but a real coverage gap in the census guarantee.

Part of #3044.

Issue #3216, opened 2026-08-05
Dedup _get_latest_review_cycle_verdict onto the canonical ReviewCycleArtifact.latest reader (FR-012 polarity-preserving)


stijn-dejongh:
Follow-up from the PR #3211 landing pass (review-cycle verdict-seam rebuild), surfaced by the reduction/dedup review lens.

`src/specify_cli/cli/commands/agent/tasks_parsing_validation.py::_get_latest_review_cycle_verdict` hand-rolls the same glob(`review-cycle-*.md`) → sort-by-cycle-number → `split_frontmatter` + `extract_scalar(..., "verdict")` logic as the canonical `src/specify_cli/review/artifacts.py::latest_review_artifact_verdict` / `ReviewCycleArtifact.latest`.

They differ only in **failure polarity**: the status-board reader returns `(None, artifact)` to surface a damaged-vs-absent distinction (FR-012), whereas the canonical path raises `DamagedVerdictRecordError`. A fold onto the canonical reader is possible but **not mechanical** — it needs the polarity translation preserved (the damaged-record distinction is load-bearing for SC-012 fail-closed behavior). Deliberately deferred out of the P0 landing pass rather than risk the polarity distinction.

This reader is declared `active` in the verdict-seam census, so it is disclosed, not hidden.

Part of #3044.

stijn-dejongh:
Closing as **already resolved**.

While scoping the follow-on hardening mission `verdict-seam-boundary-hardening-01KZG179` (which had folded this in), a post-tasks verification against `upstream/main` (`c82f4f1f1`) found that this issue's target — the hand-rolled `_get_latest_review_cycle_verdict` frontmatter reader in `cli/commands/agent/tasks_parsing_validation.py` — was **already retired by the prior mission's WP05** (verdict-seam-write-unification `01KZ9Q35`, FR-003, PR #3245).

Evidence on current `main`:
- `src/specify_cli/cli/commands/agent/tasks_parsing_validation.py:294-299` and `:1029-1030` explicitly document the deletion of `_get_latest_review_cycle_verdict` and its two dead private helpers (`_review_cycle_number`, `_review_artifact_dir_for_wp`); the verdict path there is now event-sourced (`lookup.result.verdict`).
- No live `.glob(`/`read_text`/`ReviewCycleArtifact` frontmatter-reader call remains in that file.
[...]

Issue #1928, opened 2026-06-14
EPIC: Code smells — reduce ruff / mypy --strict / SonarCloud debt (4.x)


stijn-dejongh:
## Intent

Recent PRs and CI runs carry a persistent tail of ruff, mypy (--strict), and SonarCloud findings that pollute CI output, slow down the lint gate, and impose a low-trust maintenance environment. This epic owns the initiative to **measurably reduce that accumulated debt** as part of the 3.2.x release cycle, applying an explicit-baseline-shrink discipline so the debt cannot silently regrow.

## Evidence of the problem

- **`scripts/lint_canonical_producers.py`** (non-CI-gated path) carries pre-existing ruff violations (`UP035` unused-import shim, `SIM` simplification suggestions, `B010` set-attr-with-lambda) plus 2 mypy findings — these are invisible to the normal CI gate.
- **`spec-kitty-events` sibling package** entered 3.2.x with a ~70-error ruff baseline that was accepted as a starting point without a shrink plan.
- **`tests/architectural/test_no_dead_symbols.py::PathValidationError`** is a pre-existing red pin that blocks the suite and normalizes "expected failures" in architectural tests.
- CI fast-test suites (`fast-tests-*`) do not run the architectural gate, meaning forbidden-term and dead-symbol regressions can pass local runs and fail only in `integration-tests-core-misc`. This gap widens the blast radius of each PR.
- SonarCloud hotspot count grows on each PR merge because individual PRs absorb only their own scope; no coordinated debt-reduction pass has been run.

### Phase 0 — Inventory (1–2 sessions)
- Run `ruff check src/ --select ALL` and capture a stable baseline by file.
- Run `mypy --strict src/` and triage findings: genuine type gaps vs. excessive strictness.
- Query SonarCloud for open hotspots/smells by file, grouped by category.
- Document top-10 high-ROI targets (files with the most fixable violations).

### Phase 1 — Critical-path `src/` (main payload)
- Fix all ruff `E`/`F`/`W` (error / pyflakes / warning) violations in `src/specify_cli/`.
- Address mypy `--strict` findings in core modules (`status/`, `merge/`, `cli/commands/`, `lanes/`).
- Fix the `PathValidationError` dead-symbol architectural test pin.
- Update CI baseline files so the new floor is enforced (no silent regrowth).

### Phase 2 — Tests and scripts
- Clean `ruff` in `tests/` (exclude test-only patterns only where justified).
[...]

PR #3218, opened 2026-08-05
fix(#3086): flatten coordination metadata when merge deletes the coord branch

### LINKED ISSUE #3086: merge never clears coordination_branch from meta.json after deleting it — every merged mission later crashes retrospect create with CoordinationBranchDeleted

stijn-dejongh:
## What changes for you

Commands you run against a **merged coordination mission** no longer crash. Previously, once you merged a mission that ran with a coordination branch, the next thing you did with it — create a retrospective, resume implementation, or check its status — failed outright with a data-loss safety error. It hit **every** merged coordination mission (100%: 45/45 in this repo already carry the problem). After this change, merging finalizes the mission cleanly, so it keeps working afterward — no manual recovery step needed.

## What was going wrong

Merging removes the mission's coordination branch, but it used to leave behind a stale pointer to that branch in the mission's own records. Every later command read that pointer, saw a branch that no longer existed, and stopped rather than act on inconsistent state. The only prior workaround was to run `spec-kitty doctor coordination --fix` by hand on each merged mission.

---

## For reviewers

**Root cause.** `_phase_cleanup_worktrees_and_branches` (`src/specify_cli/merge/executor.py`) had two decoupled gates — branch deletion under `delete_branch`, and the only meta-clearing path folded into the `remove_worktree` teardown seam. So a `--delete-branch` merge deleted the coordination branch and stranded `coordination_branch` in `kitty-specs/<slug>/meta.json`, tripping the #1848 `CoordinationBranchDeleted` fail-closed on every later `resolve_status_surface_with_anchor`.

**Fix.** Add the canonical three-mutation flatten (`clear_coordination_metadata` + pop `topology` + set `flattened=True`) — the same one `mission close --discard` and `doctor coordination --fix` use — to the branch-delete gate, decoupled from `remove_worktree`. It runs after lane→target merge-driver reconciliation (`coordination_branch` is a theirs-authoritative planning key), so it is the last writer; it persists through the existing bookkeeping-commit seam, **fail-open** — a commit failure never aborts an otherwise-complete merge, and never restores the pre-flatten snapshot (that would re-strand the marker). Reuses existing primitives; no new mechanism.

**Testing.** Permanent guard `tests/merge/test_coordination_flatten_on_branch_delete.py` (graduated out of `tests/regression/` when this fix greened its red-first pin): red-first verified, asserts the flatten is **committed** (clean tree), plus an idempotent-no-op case for non-coord missions. Targeted merge + coordination suites: **985 passed**. `ruff` clean; `mypy` adds no new errors (2 pre-existing at lines 153/1092, tracked under epic #1928). `commitlint` passes every commit.

**Landing review.** A 4-lens read-only adversarial squad (architect / reviewer / debugger / patterns) ran on the rebased tip — all APPROVE-WITH-FOLDS. Folds landed: fail-open commit handling + docstring correction; commit-persistence + no-op test coverage. The SSOT cleanup (unify the flatten primitive across its three call sites, and fix the `mission close --discard` partial-flatten) is deferred to #3219 as its own scoped mission — out of this P0's blast radius.

**Scope.** Standalone campsite fix — not part of, and not sequenced behind, the review-cycle verdict-seam unification (#2804 / #3216 / #3217 belong to that mission). PR-only to `main`; the operator merges.

Closes #3086
Refs #3219

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3218"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788551368&installation_model_id=427282&pr_number=3218&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3218&signature=dad2d3edcebc9e729d7c4b2c39c305514ec8ff58570a7a705fa1faf43f05ed30"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -312,6 +312,11 @@ _The 3.2.6 development cycle is open. Entries land here as missions merge._
 
 ### 🐛 Fixed
 
+- **Commands run on a merged coordination mission no longer crash (`#3086`).**
+  Once a coordination mission was merged, later commands against it — creating a
+  retrospective, resuming implementation, or checking its status — failed
+  outright, and this hit every merged coordination mission. Merging now
+  finalizes the mission cleanly, so it stays usable afterward.
 - **`decision widen` no longer lets one mission's decision travel under
   another mission's identity (mission `egress-refusal-consolidation-3110`;
   `#3111`).** Previously, ownership was resolved from whatever the decision
--- a/src/specify_cli/merge/executor.py
+++ b/src/specify_cli/merge/executor.py
@@ -1243,6 +1243,70 @@ def _phase_push(run: _MergeRunState) -> None:
     console.print(f"[green]✓[/green] Pushed {lanes_manifest.target_branch} to origin")
 
 
+def _flatten_coordination_metadata_after_branch_delete(run: _MergeRunState) -> None:
+    """issue #3086: clear the coordination marker once the coord branch is gone.
+
+    ``spec-kitty merge --delete-branch`` (the default) deletes a Mission's
+    coordination branch (``kitty/mission-<slug>``) from git but, before this fix,
+    left the paired ``coordination_branch`` key in
+    ``kitty-specs/<slug>/meta.json``. Every later command routing through
+    ``resolve_status_surface_with_anchor`` then hit ``CoordState.DELETED`` and
+    raised ``CoordinationBranchDeleted`` — the deliberate #1848 data-loss
+    hard-fail — a 100% crash on merged coord missions.
+
+    This mirrors the canonical flatten already performed by
+    ``spec-kitty mission close --discard`` and ``doctor coordination --fix``:
+    :func:`~specify_cli.mission_metadata.clear_coordination_metadata` + pop the
+    now-stale ``topology`` + record ``flattened=True``. It reuses those existing
+    primitives rather than inventing a new mechanism.
+
+    Ordering matters. This runs AFTER the lane/mission branches were merged into
+    the target: the spec-kitty-meta merge driver treats ``coordination_branch``
+    as a *theirs-authoritative* planning key (``merge/merge_driver.py``), so any
+    earlier clear would be re-populated by driver reconciliation. Running here
+    makes this clear the last writer, so it wins. The edit is persisted through
+    the same protected-flow bookkeeping-commit seam the merge uses for its other
+    meta.json mutations, so the merged target branch is not left dirty.
+
+    A non-coord Mission (``SINGLE_BRANCH`` / ``LANES``) or an already-flattened
+    one carries no ``coordination_branch`` key, so this is an idempotent no-op
+    that leaves ``topology`` / ``flattened`` untouched.
+    """
+    from specify_cli.mission_metadata import (
+        clear_coordination_metadata,
+        load_meta_or_empty,
+        write_meta,
+    )
+
+    feature_dir = run.target_feature_dir
+    meta = load_meta_or_empty(feature_dir)
+    if "coordination_branch" not in meta:
+        return
+
+    # Canonical three-mutation flatten (parity with ``doctor coordination --fix``
+    # / ``mission close --discard``); ``coordination_branch`` presence above means
+    # meta.json exists, so ``clear_coordination_metadata`` cannot raise here.
+    clear_coordination_metadata(feature_dir)
+    meta = load_meta_or_empty(feature_dir)
+    meta.pop("topology", None)
+    meta["flattened"] = True
+    write_meta(feature_dir, meta, validate=False)
+
+    meta_path = feature_dir / "meta.json"
+    if _paths_have_status_changes(run.main_repo, [meta_path]):
+        commit_merge_bookkeeping(
+            repo_root=run.main_repo,
+            worktree_root=run.main_repo,
+            mission_slug=run.mission_slug,
+            branch=run.lanes_manifest.target_branch,
+            message=(
+                f"chore({run.mission_slug}): flatten coordination metadata "
+                f"after branch deletion (#3086)"
+            ),
+            paths=(meta_path,),
+        )
+
+
 def _phase_cleanup_worktrees_and_branches(run: _MergeRunState) -> None:
     """Worktree removal + lane/mission branch deletion + coordination teardown."""
     from specify_cli.lanes.branch_naming import lane_branch_name, worktree_dir_name, worktree_path
@@ -1323,6 +1387,13 @@ def _phase_cleanup_worktrees_and_branches(run: _MergeRunState) -> None:
         else:
             logger.debug("Mission branch %s does not exist, skipping deletion", lanes_manifest.mission_branch)
         console.print(f"  Cleaned up {len(lanes_manifest.lanes)} lane branch(es) + mission branch")
+
+        # issue #3086: the coordination branch is now gone from git; flatten the
+        # mission's meta.json in the SAME gate so we can never delete the branch
+        # yet strand the paired ``coordination_branch`` marker (the 100%-hit-rate
+        # ``CoordinationBranchDeleted`` crash on every later resolve). Decoupled
+        # from the ``remove_worktree`` gate below on purpose.
+        _flatten_coordination_metadata_after_branch_delete(run)
 
     # -- WP07 / FR-016 / SC-10: Coordination worktree teardown --
     #
--- a/src/specify_cli/merge/executor.py
+++ b/src/specify_cli/merge/executor.py
@@ -1260,13 +1260,26 @@ def _flatten_coordination_metadata_after_branch_delete(run: _MergeRunState) -> N
     now-stale ``topology`` + record ``flattened=True``. It reuses those existing
     primitives rather than inventing a new mechanism.
 
-    Ordering matters. This runs AFTER the lane/mission branches were merged into
-    the target: the spec-kitty-meta merge driver treats ``coordination_branch``
-    as a *theirs-authoritative* planning key (``merge/merge_driver.py``), so any
-    earlier clear would be re-populated by driver reconciliation. Running here
-    makes this clear the last writer, so it wins. The edit is persisted through
-    the same protected-flow bookkeeping-commit seam the merge uses for its other
-    meta.json mutations, so the merged target branch is not left dirty.
+    Placement & ordering. This lives in the ``delete_branch`` gate, co-located
+    with the branch deletion, so the marker is cleared **iff** the branch it
+    names is deleted — the two mutations stay atomic. It is deliberately NOT
+    folded into the earlier, unconditional ``_phase_commit_and_assert``: doing so
+    would (a) require duplicating this gate's ``delete_branch`` guard into a phase
+    that must stay unconditional, and (b) clear the marker *before* the branch is
+    deleted, so a failed deletion would strand the inverse inconsistency (a
+    cleared marker with a still-live branch). It still runs after the lane->target
+    merge-driver reconciliation (which treats ``coordination_branch`` as a
+    *theirs-authoritative* planning key, ``merge/merge_driver.py``), so the clear
+    is the last writer regardless.
+
+    The edit is persisted through the same protected-flow bookkeeping-commit seam
+    the merge uses for its other meta.json mutations. A commit failure here is
+    logged and swallowed (fail-open), never raised: this runs in the post-push
+    cleanup phase, the on-disk flatten has already cleared ``coordination_branch``
+    (so #3086 stays fixed), and aborting an otherwise-complete merge — or
+    restoring the pre-flatten snapshot, which would re-strand the marker — is
+    worse than a locally-dirty meta.json (recoverable via
+    ``spec-kitty doctor coordination --fix``).
 
     A non-coord Mission (``SINGLE_BRANCH`` / ``LANES``) or an already-flattened
     one carries no ``coordination_branch`` key, so this is an idempotent no-op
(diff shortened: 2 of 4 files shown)


stijn-dejongh:
Starting a landing pass on this PR (maintainer runbook, docs/development/pr-landing.md): rebase onto current upstream/main, dispatch a read-only adversarial review squad on the rebased tip, fold any MAJOR findings as `<type>(landing): ...` commits, re-verify red-first, then hand off green + un-drafted for the operator to merge. Evidence to follow.

stijn-dejongh:

**Verdict:** merge-ready pending CI. A read-only, profile-loaded adversarial review squad ran on the rebased tip (4 lenses, all **APPROVE-WITH-FOLDS**, zero REQUEST-CHANGES). Two single-purpose folds pushed; the one genuinely-out-of-scope finding is filed as #3219.

### Rebase
Clean no-op — the branch was already on the current `upstream/main` tip (`1051c430d`), 1 ahead / 0 behind.

### Squad verdicts
| Lens | Verdict | Headline |
|---|---|---|
| `architect-alphonso` (seam/ordering) | APPROVE-WITH-FOLDS | Correct seam + PRIMARY-partition routing verified; the second-commit design is *safer* than folding into `_phase_commit_and_assert` (which would risk a cleared-marker/live-branch inversion). Docstring rationale imprecise. |
| `reviewer-renata` (correctness/anti-laziness) | APPROVE-WITH-FOLDS | Truth-table confirms the fix closes #3086 across all 4 flag combinations; flagged the **unguarded** flatten commit (wrong-abort on recovered-sha; late abort without restore). |
| `debugger-debbie` (live-evidence/coverage) | APPROVE-WITH-FOLDS | **Red-first proven** (revert helper → pin fails on the exact anchor → restore → passes); commit **fires** and leaves the tree clean live, but the pin never asserted it. |
| `paula-patterns` (SSOT/duplication) | APPROVE-WITH-FOLDS | 4th re-inline of the 3-mutation flatten → extract a shared primitive (follow-up, not a P0 fold). |

### Folds pushed (one purpose each, `<type>(landing)`)
1. **`fix(landing)` `a921a04`** — make the flatten's `commit_merge_bookkeeping` **fail-open**: a `SafeCommitRecoveryFailed` carrying a `commit_sha` (commit actually landed via recovery) is treated as success; every other failure is logged, not raised. Deliberately **not** the canonical restore-and-reraise — restoring the pre-flatten snapshot would re-strand `coordination_branch` and re-open #3086, and this runs in the post-push cleanup phase where aborting a completed merge is worse than a locally-dirty meta.json (recoverable via `doctor coordination --fix`). Also corrects the docstring's placement rationale (the real reason for the `delete_branch` gate is **atomicity** — co-locate clear+delete — not the driver-ordering point, which doesn't distinguish this phase from `_phase_commit_and_assert`).
2. **`test(landing)` `28f7e0f`** — pin **commit-persistence** (`git status --porcelain` clean after the flatten, so a regression that drops only the bookkeeping commit is caught — the previous pin asserted on-disk content only, written *before* the commit) and add an **idempotent-no-op** test for the guard's early-return (non-coord mission → `topology`/`flattened` untouched, no commit).

### Rejected finding
- paula-patterns' "no regression test in the PR" — a misread of `git diff --stat`. The red-first pin (`tests/regression/test_issue_3086_...py`) **pre-exists on `main`** and is greened by this fix; the diff shows only `executor.py` because the test isn't part of it. No duplicate test added.

### Filed follow-up
- **#3219** — extract a canonical `flatten_coordination_metadata` primitive and converge the executor / `doctor coordination --fix` / `mission close --discard` call sites onto it (also fixes the `mission close --discard` partial-flatten latent bug and closes the double-write window). Deferred from this P0 by design — rewrites 3 sites across 3 modules. The issue also captures the two MINORs (the `--push` origin-divergence gap; the #1589 name-divergence trust).

### Local evidence (rebased tip, isolated clone)
- **Red-first:** verified live — revert the helper → pin fails on `'coordination_branch' in meta.json` (with `Deleted branch …` proving the delete fired but stranded the marker); restore → passes.
- **Targeted suite:** `tests/merge/` + `tests/specify_cli/coordination/` + `tests/integration/test_merge_cluster_coord_read.py` + the pin → **985 passed** (`-n auto --dist loadfile`).
- **Lint:** `ruff` clean on both files. `commitlint` passes all three PR-range commits.
- **Pre-existing (not this PR):** `mypy src/specify_cli/merge/executor.py` reports 2 `[no-any-return]` at lines 153 / 1092 — identical on `HEAD^`, i.e. pre-existing repo debt (epic #1928). Untouched here.

### State
Un-drafting now; **watching CI, merge-ready on green.** Per the maintainer runbook the operator merges — I do not.

stijn-dejongh:
### Follow-up: pin graduation + changelog (2 commits)

Two commits added since the remediation summary:

- **`test(landing) dae18e1`** — the red-first pin turns green with this fix, so per the `tests/regression/` exit rule it graduates out of that suite (no green tests may live there). Moved `tests/regression/test_issue_3086_…py` → **`tests/merge/test_coordination_flatten_on_branch_delete.py`** (its functional slice — it drives the merge executor cleanup phase), dropped `@pytest.mark.regression` for the canonical `integration` + `git_repo` marks, and reframed the docstring as a permanent guard (keeping the #3086 reference as history). Assertions unchanged. This rides *this* PR by necessity: the fix is what greens the pin, so graduating it anywhere else would land a green test in `tests/regression/` (or a red one in `tests/merge/`).
- **`docs(landing) d72987f`** — consumer-first BLUF changelog entry under `[Unreleased] 3.2.6 / Fixed`: *commands run on a merged coordination mission no longer crash.*

Local verification (targeted; full suite not run in-session): the relocated test + `test_pytest_marker_convention` + `test_marker_job_completeness` → **16 passed**; `ruff` clean; `commitlint` passes both new commits. The architectural shard runs in CI as the release authority. Still un-drafted, merge-ready on green — operator merges.

Issue #3140, opened 2026-08-01
Malformed meta.json raises raw ValueError at 3 call sites instead of failing closed (4 red, plus a 5th routed around for months)


MOES-Media:
Four tests across three independent call sites assert that a corrupt or non-conforming `meta.json` is handled gracefully — returning `None`, or raising a typed domain exception. All four are red on `main`: each call site instead lets a raw `ValueError` propagate.

## The failures

| Node-id | Job | Expected | Actual |
|---|---|---|---|
| `tests/lanes/test_worktree_allocator.py::TestDefensiveHelpers::test_read_coordination_branch_none_on_malformed_json` | `integration-tests-lanes` | `None` | `ValueError: Malformed JSON in …/meta.json: Expecting property name enclosed in double quotes: line 1 column 3 (char 2)` |
| `tests/unit/status/test_mission_status_aggregate.py::TestLoadCoordUnavailableFailsClosed::test_corrupt_meta_fails_closed_instead_of_legacy_fallback` | `integration-tests-core-misc (misc)` | `MissionMetadataUnavailable` | `ValueError: Malformed JSON in …/meta.json: Expecting value: line 1 column 63 (char 62)` |
| `tests/unit/status/test_mission_status_aggregate.py::TestLoadCoordUnavailableFailsClosed::test_non_dict_meta_fails_closed_instead_of_legacy_fallback` | `integration-tests-core-misc (misc)` | `MissionMetadataUnavailable` | `ValueError: Expected JSON object in …/meta.json, got list` |
| `tests/specify_cli/cli/commands/test_charter_widen_integration.py::TestGetMissionId::test_returns_none_if_json_malformed` | `fast-tests-cli` | `None` | `ValueError` |

## Evidence

Stable — all four fail identically in two independent runs on the same SHA:

| Run | Trigger |
|---|---|
| [`30681941495`](https://github.com/Priivacy-ai/spec-kitty/actions/runs/30681941495) | schedule, 2026-08-01T03:24:54Z |
| [`30621215287`](https://github.com/Priivacy-ai/spec-kitty/actions/runs/30621215287) | push, 2026-07-31T09:47:27Z |

Both on `bb2020fea924d6e5b157974f27a7cab1a77ad259`.

## The shape

[...]

Issue #3231, opened 2026-08-06
Acceptance gate: one admitted scaffold row makes `overall_verdict` `pending` and blocks acceptance (pending-poisons-the-aggregate)


MOES-Media:
**Filed by mission `meta-fail-closed-3162-01KZ7FSQ`, WP07 (`SC-009` register row 2; `FR-011`, `C-006`, `C-009`). Filed, NOT fixed — see "Why this is filed and not fixed" below.**

## Symptom

After a mission→target squash merge, the row-aware `acceptance-matrix.json` reconciler admits **both** sides' criterion rows. When the mission branch still carries the `finalize-tasks` scaffold placeholder, the merged document therefore legitimately contains the scaffold row `AC-001` with `pass_fail: "pending"` — and that single admitted row makes the whole aggregate verdict `pending`, which blocks acceptance even though every real criterion is `pass`.

## Mechanism (two lines, both verified by opening them)

1. `src/specify_cli/acceptance/matrix.py:263`, in the computed property `AcceptanceMatrix.overall_verdict` (`:248-271`):

```python
if any(v == "pending" for v in criterion_results + invariant_results):
    return "pending"
```

`pending` **dominates** — one pending row outvotes any number of `pass` rows.

2. `src/specify_cli/acceptance/gates_core.py:525`:

```python
verdict = acc_matrix.overall_verdict
```

feeds that aggregate straight into the acceptance gate, where `:528-529`:

```python
elif verdict == "pending":
    activity_issues.append("Acceptance matrix verdict is 'pending' — criteria or invariants have not been verified")
```

turns it into a **blocking activity issue**.

## Measured

Real fixtures from `tests/regression/test_issue_2804_merge_resets_gate_artifacts.py`, fed to the shipped reconciler `specify_cli.cli.commands.merge_driver.reconcile_acceptance_matrix_documents` (FILLED as *ours*, PLACEHOLDER as *theirs*, empty base — the add/add divergence), control first:

```
CONTROL filled fixture contains marker?  False
merged criterion_ids: ['AC-001', 'FR-001', 'FR-003']
overall_verdict: pending
POST contains SCAFFOLD_TODO_MARKER?      True
```

`FR-001` and `FR-003` are both `pass`; `AC-001` is the scaffold row. The aggregate is `pending`.

[...]

PR #3211, opened 2026-08-05
Review-cycle verdict-seam rebuild: one authority for the current verdict + durable arbiter override

stijn-dejongh:
## What changes for you

A work package's **current verdict now has exactly one source of truth** — the status event log — and an arbiter's override is now a durable, first-class outcome. Before this mission:

- **"Is this WP approved?" had no single answer.** Some consumers (the kanban board, `move-task`'s review-readiness check) read a `review-cycle-N.md` artifact's `verdict:` frontmatter directly, so a reader could disagree with the event log about which verdict was current if the two ever drifted.
- **An arbiter override could silently vanish.** The override writer never git-committed its output under any topology, so an override present in a working tree could disappear from a fresh clone or a branch switch — which is what made `--skip-review-artifact-check` feel unreliable even when it had "worked."
- **An override was indistinguishable from a genuine approval.** Nothing recorded that a standing rejection had been knowingly overridden rather than actually re-approved by a reviewer.

After this mission:

- Every consumer resolves the current verdict through a **reader downstream of the reducer** (the event-sourced snapshot). The review-cycle artifact is authoritative only for *what the reviewer said* (prose, affected files, repro command) — never for *which verdict is current* (FR-001 / SC-011).
- An arbiter override **durably persists through the same commit seam every other verdict-relevant write uses**, so it survives a fresh clone and clears the merge gate without repeating the flag (FR-010).
- An override is **recorded and read back as its own first-class outcome** (a stated actor and reason), never fabricated as an approval nobody performed (FR-011).

Governing ADR: `docs/adr/3.x/2026-08-03-1`. This completes the reframing the predecessor mission (`#3156`/`#2697`/`#990`) left open: the event is the index, the artifact is the payload.

## Closes

Closes #2275
Closes #2996
Closes #990
Closes #2697
Closes #2646

> Epic #3044's children are #2275, #2996, #990 and #3088 (verified, not assumed). This mission does not touch #3088, which stays open; the epic therefore does not close, and this PR does not claim it does.

`#3158` is **not** a child of #3044 (it is rename-debt deferral) and is deliberately not listed.

## How it was verified

- **Full `tests/architectural/` suite: 1570 passed, 0 failed** (the FR-020 census fold and its re-pointed gate live here). The census check now reads the folded `tests/architectural/verdict_seam_census.yaml` as its executable fixture; both its invariants were adversarially mutation-tested (an unregistered writer/resolver/reader still reds it; a `retire` row missing its FR still reds it).
[...]

PR #3155, opened 2026-08-02
Doctrine-Charter Split: single charter read/path authority foundation

stijn-dejongh:
Fixes a class of bugs where the compiled `charter.yaml` was nominally the
governance authority but several surfaces still keyed their "does a charter
exist / what does it say" decision on the display-only `charter.md` — so a
project with `charter.yaml` and no `charter.md` looked broken in the
dashboard, mis-hashed in the analysis report, and could silently resolve
retrospective policy from the wrong file. Also fixes ~two dozen places where
a corrupted `meta.json` could crash a command with a raw, unhandled error
instead of failing predictably.

## What changes for you

- **Dashboard**: the Charter page no longer shows "no charter" for a project
  that has a compiled `charter.yaml` but no `charter.md` (#3150). Presence
  resolution now prefers `charter.yaml` when present and falls back to
  `charter.md` when it's absent — a project with only one of the two files
  still reports a charter, on the dashboard and in the analysis report.
- **`meta.json` corruption**: a broken or malformed `meta.json` now produces
  a clear, typed error — never a raw crash — across every call site this PR
  and its landing pass routed through the canonical `load_meta_fail_closed`
  authority (#3140). A residual, explicitly tracked batch of call sites
  remains genuinely unrouted; see #3162 for the exact enumerated list rather
  than treating this as a blanket "every command" guarantee.
- **Retrospective policy**: if you configure retrospective behaviour in both
  `charter.yaml` and `charter.md` frontmatter and they disagree, `charter.yaml`
  now wins (matching every other charter-governed decision). Configuring it
  only via `charter.md` frontmatter is unaffected.
- **CI**: the `doctrine-charter-tests.yml` workflow now actually covers the
  charter CLI command layer, and the CLI-reference parity check runs instead
  of silently skipping (#3149, #3107, #3102).
- Wheel-packaging groundwork for a future `kernel`/`doctrine`/`charter`
  standalone-wheel split, and an ADR recording that assessment — no wheel
  cutover is performed here (see below).

## Why, and what this deliberately does NOT do

A 3-lens pre-spec squad scoped this as **foundation-first**: the eventual
`kernel -> doctrine -> charter` wheel cutover (#3101) is blocked on a
not-yet-published kernel wheel, and a partial cutover is forbidden by
[ADR 2026-04-25-1](docs/adr/3.x/2026-04-25-1-shared-package-boundary.md)'s
C-007. This PR lands the single read/path authority, the layer-order
[...]

PR #3247, opened 2026-08-06
fix(meta): route the remaining meta.json reads fail-closed, and re-pin #2804's marker (#3162, #3138)

### LINKED ISSUE #3138: Regression: #2804's red marker is failing again on main (merge still resets filled gate artifacts)
`#2804` was closed on 2026-07-25. Its own red marker test is failing again on `main`.

## The failure

```
tests/regression/test_issue_2804_merge_resets_gate_artifacts.py::test_merge_resets_filled_gate_artifacts_to_placeholder
```

```
AssertionError: #2804: spec-kitty merge reset acceptance-matrix.json's overall_verdict
to 'pending' -- the filled, accepted evidence was clobbered by the mission->target
squash merge's '-X theirs' conflict resolution (mission branch's stale placeholder won
over target's already-accepted fill).
```

## Evidence

Stable across two independent runs on the same SHA, and it fails in **two** jobs, so this is not a single-shard artefact:

| Run | Trigger | Jobs |
|---|---|---|
| [`30681941495`](https://github.com/Priivacy-ai/spec-kitty/actions/runs/30681941495) | schedule, 2026-08-01T03:24:54Z | `regression tests (blocking)`, `integration-tests-core-misc (misc)` |
| [`30621215287`](https://github.com/Priivacy-ai/spec-kitty/actions/runs/30621215287) | push, 2026-07-31T09:47:27Z | same two |

Both on `bb2020fea924d6e5b157974f27a7cab1a77ad259`.

## Why this is filed rather than absorbed

Found while building a CI baseline inventory for an unrelated mission, which needed to separate pre-existing reds from PR-caused ones. Under DIR-013 a pre-existing failure gets filed, not treated as accepted baseline.

The distinction that matters: most of `main`'s current reds are legitimate ATDD markers whose issues are **open** (`#2996`, `#3045`, `#3086`, `#2782`, `#3092`). This one's issue is **closed**, which means nothing is tracking it and the assertion text is actively lying about the state of the world — it says "#2804" to a reader who will look up a closed issue and conclude the test is stale.

## What has not been established

Whether the original `#2804` fix regressed, was never complete, or was reverted. The baseline pass classified the failure and checked issue state; it did not bisect. The `-X theirs` conflict resolution named in the assertion is the obvious place to start, but that is the test's own hypothesis, not a fresh diagnosis.

---

*Filed from a CI baseline inventory taken at `bb2020fea9`. Of 41 distinct failing node-ids in that inventory, 29 were tied to an open issue, 1 was this regression, and 18 were unfiled — the other 18 are filed separately.*

### LINKED ISSUE #3162: meta.json fail-closed migration: 13 unrouted call sites left raising raw ValueError (pending-batch-a)
## Context

PR #3155's description claims: "a broken or malformed `meta.json` now always produces a clear, typed error — never a raw crash — across every mission command that reads it (#3140)."

That claim is **false as a universal statement**. Mission `doctrine-charter-split-unification-01KZ0SRB`'s own artifacts — `kitty-specs/doctrine-charter-split-unification-01KZ0SRB/notes/meta-load-census.md` and the paired ledger in `tests/specify_cli/test_meta_fail_closed_full_census_contract.py` — enumerate a bucket of call sites tagged `pending-batch-a` that are still **unrouted**: they call `load_meta(...)` directly (default `on_malformed="raise"`) with no local handling, so a corrupt/non-dict `meta.json` still raises a raw `ValueError`, never the typed `MissionMetaReadError` that `#3140` was supposed to guarantee everywhere.

This issue was discovered during the #3155 landing pass's adversarial review. The review also found that the ledger's own ownership label for this bucket — `"a real routing target owned by WP08's parallel lane"` — is itself inaccurate: `tasks/WP08-meta-fail-closed-route-batch-a.md`'s `owned_files` list does not include any of the files below, and neither does `tasks/WP09-meta-fail-closed-route-batch-b.md`'s. WP08 and WP09 both shipped and closed #3140 without ever touching these files. The mislabel has been corrected in the census/ledger comments (separate landing-fold commit), replaced with a reference to this issue.

## Unrouted call sites (still raising raw `ValueError`, not `MissionMetaReadError`)

- `src/mission_runtime/resolution.py:509` (`_mid8_from_primary_meta`) — divergent-wrapper, `except ValueError`
- `src/mission_runtime/resolution.py:852` (`_resolve_coordination_branch`) — divergent-wrapper, `except ValueError`
- `src/mission_runtime/resolution.py:1106` (`_resolve_mission_id`) — divergent-wrapper, `except ValueError`
- `src/runtime/next/_internal_runtime/planner.py:188` (`_resolve_workflow_for_mission`) — route-unwrapped (the `spec-kitty next` path)
- `src/runtime/next/runtime_bridge_io.py:380` (`_workflow_runtime_template`) — route-unwrapped
- `src/specify_cli/bulk_edit/gate.py:57` (`_is_bulk_edit_mission`) — route-unwrapped
- `src/specify_cli/bulk_edit/gate.py:80` (`ensure_occurrence_classification_ready`) — route-unwrapped
- `src/specify_cli/context/resolver.py:75` (`_read_meta_json`) — route-unwrapped; catches only `FileNotFoundError`, malformed JSON propagates raw
- `src/specify_cli/decisions/service.py:134` (`_resolve_mission_id`) — divergent-wrapper, `except (FileNotFoundError, ValueError)`
- `src/specify_cli/missions/_read_path_resolver.py:846` (`read_primary_meta`) — route-unwrapped
- `src/specify_cli/missions/_read_path_resolver.py:862` (`read_primary_meta`) — route-unwrapped
- `src/specify_cli/missions/_resolve_planning_branch.py:116` (`load_mission_target_branch`) — divergent-wrapper, `except (FileNotFoundError, ValueError)`
- `src/specify_cli/upgrade/feature_meta.py:42` (`load_feature_meta`) — divergent-wrapper, `except ValueError`

13 call sites across 12 functions (one function, `read_primary_meta`, has two call sites).

Every one of these should be routed through `specify_cli.core.paths.load_meta_fail_closed` (FR-007), per the same pattern WP08/WP09 applied to their owned files, honouring the census's DEF A/DEF B disambiguation.

## Raw `meta.json` reads that bypass `load_meta(` entirely (not part of the 13 above)

Found during a later landing-fold pass (also #3163): `notes/meta-load-census.md` is a census of *calls to `load_meta(`*, reconciled against `grep -rn "load_meta("`. It does not — and by construction cannot — enumerate a `meta.json` read that never calls `load_meta` at all. Two such sites exist in `src/specify_cli/git/ref_advance.py`:

- `_committed_meta_object` (~line 192) — reads the committed `meta.json` via `git show HEAD:<path>` and parses the output with the module's own `_parse_meta_object` helper (a private `json.loads` + dict-shape check). Malformed/non-object content silently degrades to `{}`.
- `_meta_change_is_vcs_lock_only` (~line 242) — reads the working-copy `meta.json` via `Path.read_text()` and parses it with the same `_parse_meta_object` helper. A malformed/unreadable working copy is treated as "not lock-only" (blocks the advance) rather than raising.

Neither site calls `load_meta`/`load_meta_fail_closed`, so neither appears in the census's row set, and neither is listed in `tests/architectural/test_inline_meta_read_gate.py`'s allowlist (`inline_meta_read_allowlist.yaml`) — that gate's `test_routed_load_meta_floor`/`test_inline_meta_read_gate_green_against_seeded_allowlist` checks pass today without ever seeing these two sites, because the gate targets a different, narrower inline-read shape than `ref_advance.py`'s `git show` + private-parser pattern.

These two sites are **not** being routed through `load_meta_fail_closed` as part of this issue or the #3155 landing pass — same mission-scope carve-out as the 13 sites above (a full routing pass is its own ATDD cycle, not a fold-in-place fix). They are recorded here so the eventual routing pass for this issue also picks them up, and so `notes/meta-load-census.md`'s scope note has somewhere concrete to point at.

## Relationship to #3140

#3140 is legitimately closed for the sites WP07/WP08/WP09 actually routed (170+ sites verified via the AST-based NFR-003 full-census contract test). This issue tracks the honest remainder that #3140's closure — and PR #3155's "every mission command" description — does not actually cover. Fixing these sites is mission-scoped work (a full routing pass with its own ATDD red/green cycle across several subsystems), not a fold-in-place fix.

## Source

Found via `notes/meta-load-census.md` (WP07, mission `doctrine-charter-split-unification-01KZ0SRB`) cross-referenced against `tests/specify_cli/test_meta_fail_closed_full_census_contract.py`'s `_ACCOUNTED_SITES` ledger (`pending-batch-a` rows) and both `tasks/WP08-*.md` / `tasks/WP09-*.md` `owned_files` lists. The `ref_advance.py` raw-read sites were found during a follow-up landing-fold pass while clarifying `notes/meta-load-census.md`'s scope (also #3163).

MOES-Media:
Closes #3162
Closes #3138

## What this does

Routes the **13 `meta.json` read sites** left unrouted after #3155 routed 170+, so a corrupt or
unreadable `meta.json` **fails closed** instead of degrading silently. Census arms: 29 REFUSE-raw,
16 DEGRADE, 14 REFUSE-typed. Also closes #3138 by re-pinning #2804's returning red marker.

## The behavioural consequence, and why ten handlers moved

`load_meta_fail_closed` raises `MissionMetaReadError`, whose MRO is
`MissionMetaReadError → RuntimeError → Exception → BaseException → object`. **No `ValueError`, no
`OSError`.** So every pre-existing `except (ValueError, OSError)` wrapping a newly routed call
silently stopped absorbing corruption — converting a degrade path into a raising path, the arm change
`C-001` forbids, and invisible to the routed function's own tests.

Ten handlers were widened in three groups:

| group | count | how found |
|---|---|---|
| file-local | 2 | by the WP that routed the sites |
| **stranded on the transitive caller chain** | **4** | only by a *chain-local* sweep; a file-local sweep cannot see them |
| at the degrade sites | 4 | by the WP that owns those sites |

Two of the four stranded arms are **proven** to have dropped `error_code`, `mission_flag`,
`available_missions` and `example_command` from agent-facing JSON payloads while still exiting 1 — a
silent contract change no test caught.

`scripts/sweep_degrade_arms_on_routed_chain_3162.py` is the durable instrument for that class: an
import-resolved fixpoint over `src/`, seeded with the routed functions. It is **self-calibrating** —
`--self-check` replays a recorded 6-arm known answer at the mission's base and returns **2**, refusing
the run, if it does not reproduce. An uncalibrated sweep's silence means nothing.

## Contract crossing: the routed floor, and a red already on `main`

**`upstream/main` d0ed802cc was already RED on this mission's own gate** before the rebase:

```
ROUTED live 133, floor 128, margin 4  ->  band [129, 132]
assert (133 - 128) <= 4  ->  AssertionError: tighten the floor
FAILED tests/architectural/test_inline_meta_read_gate.py::test_routed_load_meta_floor  exit=1
```

Reproduced in a pristine `upstream/main` worktree with the imported `specify_cli.__file__` printed as
the control. Routing landed upstream without the floor being re-derived.

Both prior provenance entries are **retained** — each is true of its own tree, and neither floor is
correct here. PR #3211 measured live 131 and set 128; this mission's WP06 measured live 130 and set
127. The two trees routed **largely disjoint** sites: upstream's `resolution.py` carried **0**
`load_meta_fail_closed(` calls where this mission's carried **3**, yet upstream's total exceeded this
mission's. Measured on the merged tree with the gate's own scanner: **live 134** = upstream's 133 plus
this mission's single allocated net routed call. Same 3-below-live rule as every entry above:
`134 - 3 = 131`, band `[132, 135]`. All three clauses hold, including the **strict** middle one, so a
*fold* reds this gate from below.

**This re-pin clears upstream's red** as a side effect. It is a contract crossing re-pinned with its
measurement, not a ceiling raised to make a test pass.

## Red-first

Every routed site landed **red-then-green**: the routing commit lands *without* the handler change, so
`MissionMetaReadError` escapes `except ValueError` and the fallback test fails **for that reason**; the
handler commit follows. That ordering is the single sanctioned exception to `C-002` and was verified by
**re-checkout at each SHA**, not asserted:

```
commit 0  45b278823  4 passed   exit 0
commit 1  d5f39510f  4 FAILED   exit 1   <- all four are MissionMetaReadError, 0 FileNotFoundError
commit 2  eaa75ecd1  4 passed   exit 0
```

**This mission also turns a pre-existing red green**: #2804's
`test_merge_resets_filled_gate_artifacts_to_placeholder` fails at the merge base and passes here.

## #3138 / #2804

Both assertions were **re-pinned, not deleted**, into one shared predicate called by the marker test and
an SC-010 companion built on the defect's own fixture; assertion 2 is now evidence survival. The product
defect the red pointed at was **filed and deliberately not fixed** (#3231), so the red was not made to
go away without addressing what it pointed at. Mutation probes confirm the companion catches every
weakening — dropping the evidence-survival clause, admitting `"fail"`, rejecting `"pending"`, and letting
the handle leak into the placeholder each turn it red.

## Verification

Local, on this rebased tip:

| gate | result |
|---|---|
| `tests/architectural` | **1777 collected → 1771 passed, 4 skipped, 2 xfailed**, exit 0, `^ERROR tests/` 0 |
| collection equivalence | serial `--collect-only` 1777 == 1771+4+2 — no silent shrink |
| `ruff check` (46 changed `.py`) | All checks passed |
| `mypy --strict` (`src/specify_cli src/charter src/doctrine`) | Success, 1132 files, exit 0 |
| commitlint `upstream/main..HEAD` | exit 0; no body line > 100 chars |
| docs freshness / inventory / description-length | exit 0 / no drift / 0 violations |

**Full-suite measurement (pre-rebase tree, 21,475 nodes across 16 sequential passes)**, every pass
collection-equivalent: 21,422 passed, **13 pre-existing** (red at merge base and head, byte-identical
causes), **6 measurement artifacts** (latency budgets sampled under 6-way contention; two `tests/sync`
workers binding **port 9413** concurrently — all pass serially), and **1 mission-introduced**, fixed.
`mypy --strict` delta was 0 → 1 → **0**: one new error where a widened `except` tuple outgrew its
callee's annotation. Attribution used `96494e5ec` = `git merge-base HEAD main`; `98198e980` is
upstream's *tip* and is **not** an ancestor of HEAD.

## Governance

Full dossier travels with the PR under `kitty-specs/meta-fail-closed-3162-01KZ7FSQ/**` — 8 WPs,
WP01–WP07 approved, WP08 terminal verification, with review-cycle artifacts including the two WPs that
were **rejected and remediated**.

`residual-ledger.md` is the fold-first record: findings are folded into the mission where in scope and
ledgered where not, then assessed after the PR rather than filed as issues during development. It also
records, deliberately, **four errors in the orchestrator's own briefs** that subagents caught by
re-deriving instead of transcribing, and **two attempted fixes that were wrong and were reverted** —
including one that would have dropped 96 missions from a guard meant to cover them, caught by printing
the size of the set its predicate selected.

**State: DRAFT for operator review. Fork CI has not run this branch** (the fork's `main` is 263 commits
ahead of upstream and must not be fast-forwarded, so a same-repo PR's range would include 568 upstream
commits and fail commitlint on upstream's own messages).

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3247"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788633745&installation_model_id=427282&pr_number=3247&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3247&signature=878a2acf1058218d9727afaa3c27d52baf8f524d2a4194594ebb44c3818e725b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
## `diff-coverage` fold — and the reason the red was misleading

Pushed `e46a4cdde`. The gate reported:

```
src/mission_runtime/resolution.py (77.8%): Missing lines 545,1156
Total: 11 lines · Missing: 2 · Coverage: 81%   →  Failure. Coverage is below 90%.
```

**Both lines were already tested.** `tests/mission_runtime/test_wp04_degrade_site_fallbacks.py` (rows 1 and 3) and `test_wp04_sc007_guard_and_handler_contract.py` (SC-007a) hit `:545` and `:1156` today — confirmed by running just those two files under `--cov=mission_runtime`: 12 passed, and neither line appears in the missing-line list.

### Why the gate could not see them

Those tests carry `pytest.mark.integration` + `pytest.mark.git_repo`, which routes them to `integration-tests-core-misc (misc)`. That job's `if:` requires `github.event.pull_request.draft == false` (or a `ci:full` / `ready-for-ci` label). **This PR is a draft, so the job skipped.** `diff-coverage` still `needs:` it — but `needs` on a *skipped* job is satisfied, and the artifact download is `continue-on-error: true`. Run `31137786167` logged `Found 23 coverage report(s)`, none matching `coverage-integration-core-misc-*`.

So an **ENFORCED** gate scored against a silently smaller coverage set, and produced a red that is **indistinguishable from a genuinely untested line**. A contributor who reads that red honestly and adds an integration-tier test gets the identical red back. This is not specific to this PR or this module — it applies to every draft PR and every critical-path file whose coverage lives behind a draft-gated shard. Ledgered as **F22** (CI-topology bucket, alongside F17/C-007); the fix belongs in `ci-quality.yml` gate semantics, not here.

### What was done instead

A new fast-tier file, `tests/mission_runtime/test_wp04_degrade_arms_fast_tier.py`, marked `fast` so it lands in `fast-tests-core-misc (core-misc)` — which runs `--cov=mission_runtime` **and** runs on draft PRs. Probed first that both arms are reachable in-process from a bare `tmp_path` (`get_main_repo_root` falls through when `.git` is absent), so no git repo, no subprocess, no patched reader. Selector check: fast shard collects 7, integration shard collects 0 — no same-tier double-run.

Two things deliberately **not** done:
- **The existing files were not re-marked `fast`.** They genuinely need a git repo; widening a file's markers away from its correct tier is not how this repo buys instrumentation.
- **The `ci:full` label was not applied.** It would have un-skipped the integration shards without un-drafting and turned this red green — while hiding the defect and leaving F22 unrecorded.

### Non-fakeable, with red-first proof

`:545` asserts on the **return value** (`== ""`), never `pytest.raises(ValueError)` around the call — that shape stays green after a narrowing and is precisely the cheat the code comment at `resolution.py:544` names. Two provenance assertions precede it, so the outcome assertion cannot quietly become a no-op if the traversal guard is relaxed and `""` starts arriving down the ordinary `if not meta` path.

| Mutant | Result |
|---|---|
| `:545` narrowed to `except MissionMetaReadError:` | **5 failed, 2 passed** — `ValueError: Not a safe path segment: 'evil/../x'`, failing at the outcome line, not the provenance ones |
| `:1156` `except` arm deleted | **1 failed, 6 passed** — `MissionMetaReadError: Cannot read …/meta.json: Malformed JSON` |

Both mutants reverted byte-identically: `cmp` IDENTICAL, `sha256 96c6bc32…8b61b` unchanged, verified again post-commit. **No production code changed in this fold.**

Green against real code: 7 passed; the missing-line list shifts `438-483 → 547-552` and `1018-1089 → 1159-1161`, i.e. both lines now hit. `tests/mission_runtime` full: **309 passed**. Marker-convention gates: 4 passed. `ruff check` clean, `mypy --strict` zero new errors, no baseline regenerated.

One incidental correction: the comment at `resolution.py:540-542` cites the traversal `ValueError` as arriving from `_compose_primary_feature_dir` via `_read_path_resolver.py:1307`. It actually reaches the `try` through the canonicalizer's **nested** call at `_read_path_resolver.py:469` — same function, same limb, but the line citation is off.

Still a draft, for the operator.

stijn-dejongh:
Closing as **superseded / re-scoped** ✅

Thank you, @MOES-Media — this PR did the load-bearing work of surfacing the meta.json fail-closed routing and the #2804 marker problem, and it directly guided the canonical solution the tracker now reflects. Three things have moved on `main` since it was cut:

1. **Core thesis already landed.** `mission_runtime/resolution.py` now imports and catches `MissionMetaReadError` on `main`, so the central fail-closed behavior this PR introduced is in place.
2. **The gate model changed under it.** The inline-meta-read gate was re-pinned to floor **130** (`06f8c85d9` / `945f882b2`) with batch-B routing (`102acbf47`, #3140) and evasion-vector closure (`640c202ec`) — so this branch's floor-128 / band-[132,135] arithmetic is against a stale gate, and it's now `CONFLICTING`.
3. **The #2804 re-pin is superseded.** `b04da00e1` (row-union write-side placement seam, #3076) is on main — the model #3232 tracks as replacing the deleted #2804 unit gate — so the re-pin here targets a retired model.

The **residual** routing work is now tracked, decomposed, under new epic **#3259** (meta.json fail-closed read routing), whose children reframe it behind a precondition this monolithic branch predates: **#3229** (add the L1 pure-decode seam tier) must land first, then **#3230** routes the four remaining reads onto it (deferred on the missing L1 tier, *not* on structure). #3162/#3138 remain credited here.

Closing so the work continues on the #3259 decomposition rather than a conflicting monolith. Huge thanks — your investigation is what made the canonical path clear.

Issue #3169, opened 2026-08-03
Audit `TrackerProjectConfig._extra` consumers before promoting any further keys


MOES-Media:
Split out of the `#3108` tracker-egress-refusal mission (Bundle C) as a recorded residual. **Mitigated there, not answered.**

## The question

`TrackerProjectConfig._extra` collects unknown keys from a project's `.kittify/config.yaml` `tracker:` block so they survive a round-trip. Nobody has established **who reads it**.

## Why `#3108` cared

That mission promotes a previously-unknown key (`tracker.egress`) to a known field. Promoting a key **removes it from `_extra`**. If any consumer reads `_extra` expecting that key to be present, promotion changes behaviour silently.

`#3108` mitigated its own exposure by making the promotion explicit and pinning the round-trip with tests. It did **not** audit the consumers, so the general question stands for the next key anyone promotes.

## What an answer looks like

An inventory of every read of `_extra` in `src/`, and for each, whether it depends on a key that could later be promoted. If the answer is "nothing reads it", that is a useful and cheap result — say so and close.

---

Raised at `bb2020fea`, still open at `abca7ec9`.

stijn-dejongh:
**Research pass (on `upstream/main` `5d49d31ed`): audit complete → nothing reads it. Closing.**

`_extra` (`src/specify_cli/tracker/config.py:136`) is a **pure round-trip preservation bag**: populated by `from_dict` from unknown keys (`config.py:254`), re-emitted verbatim by `to_dict` (`config.py:169`), and copied forward at 5 reconstruction sites (`service.py:108`, `local_service.py:157`, `saas_service.py:227/283/350`).

**Decisive negative result:** `grep '_extra\['` and `grep '_extra\.get'` across all of `src/` return **nothing** — no code subscripts a specific key out of `_extra` or branches on its contents. So promoting a key to a known field cannot break a `_extra` *reader* (there are none).

The **only** hazard is serialization drop: if a promotion removes the key from `_extra` (via the `_KNOWN_KEYS` filter) without also emitting a known field and carrying it forward, the value is silently lost on the next round-trip — exactly what #3108 had to handle for `tracker.egress`.

**Safe-promotion checklist** (generalized from #3108's FR-011 sites):
1. Add the field + include it in `_KNOWN_KEYS` (`config.py:194`).
[...]

Issue #3136, opened 2026-08-01
test(sync): #3115 sync-half sleep-count producer named — psutil-backed waits excluded, victims corrected


MOES-Media:
## Context

Successor to `#3115`'s sync half, deferred under FR-010's blocking exit clause
(mission `verification-trust-3115-01KYVYWM`, WP14, `deferred-with-followup`).
FR-010 bounds the sync-half hunt to at most 6 agent-hours and 3 candidate
mechanisms, then requires a named-producer, pruned-candidate handoff rather
than a silent close. WP06 spent the budget and produced exactly that; this
issue is the handoff. Cross-references: #3115 (whose victim list this issue
corrects), #3130 (11 confirmed process-global/thread leaks; one — `sync/batch.py:674`
via `background.py`'s final-sync thread — is this investigation's own positive
control, see below).

## Corrected victims

`#3115`'s issue body names
`tests/sync/tracker/test_saas_client.py::TestRetryBehaviors::test_429_respects_retry_after`.
**That test has never exhibited the failure.** The wrong node-id propagated
through the issue, the mission spec, every WP file and every agent brief for
three review rounds before anyone re-opened the CI log. The real failures,
verbatim from CI job `91126025663` (run `30621215287`, `fast-tests-sync`,
Python 3.12.3, `-n auto --dist loadfile`, head `bb2020fea9`):

```
tests/sync/tracker/test_saas_client.py:534: in test_exponential_backoff_intervals
    assert len(sleep_calls) == 3
E   assert 71 == 3                                                            [gw5]

tests/sync/tracker/test_saas_client_origin.py:261: in test_429_retries_then_raises
    mock_sleep.assert_called_once_with(2.0)
E   AssertionError: Expected 'sleep' to be called once. Called 556 times.     [gw2]
```

Any follow-up work must key on these two node-ids, not the one named in #3115.

## Named producer

**CPython's `subprocess.Popen._wait(timeout)` POSIX busy-wait**:

```python
delay = 0.0005                              # 500 us -> initial delay of 1 ms
delay = min(delay * 2, remaining, .05)
time.sleep(delay)
```

CI's `556 = 1 + 6 + 549` decomposes as **one** `Popen.wait(timeout=…)` caught
in flight: the victim's own `call(2.0)`, one complete six-term geometric ramp
(`0.001 → 0.032`), then the loop saturated at its `0.05` cap for the rest of
the patch window. This is **one producer**, not two — a geometric ramp that
[...]

PR #3250, opened 2026-08-06
fix(tracker): refuse `sync publish` on a local binding instead of crashing (#3168)

### LINKED ISSUE #3168: `tracker sync publish` raises an uncaught AttributeError on a beads/fp binding
Split out of the `#3108` tracker-egress-refusal mission (Bundle C) as an incidental live bug found during planning. Not fixed there — it is outside that mission's boundary and absorbing it would have blurred the diff.

## The bug

`spec-kitty tracker sync publish` on a `beads` or `fp` binding raises an uncaught `AttributeError`.

`src/specify_cli/tracker/service.py:202-203` delegates unconditionally:

```python
def sync_publish(self, **kwargs: Any) -> dict[str, Any]:
    return self._resolve_backend().sync_publish(**kwargs)
```

For a local binding `_resolve_backend()` returns `LocalTrackerService`, which **defines no `sync_publish`** (verified: zero definitions in `src/specify_cli/tracker/local_service.py`).

## Why it surfaces as a traceback rather than a clean error

`src/specify_cli/cli/commands/tracker.py:346` — `_run_or_exit` catches only `(RuntimeError, ValueError)` at `:349`. `AttributeError` is neither, so it escapes to the top level.

## Expected

Either a clean "not supported for local providers" message through the existing error path, or an explicit `sync_publish` on `LocalTrackerService`. Deciding which is part of this issue, not presumed by it.

---

Verified at `abca7ec9`.

MOES-Media:
Closes #3168

## The defect

`tracker sync publish` on a local (beads/fp) binding raised an **uncaught `AttributeError`** instead of refusing:

```
tracker.py:1225  payload = _service().sync_publish()
service.py:203   return self._resolve_backend().sync_publish(**kwargs)
AttributeError: 'LocalTrackerService' object has no attribute 'sync_publish'
```

Reproduced from a real CLI invocation (`.venv/bin/spec-kitty`), not an internal call. Recipe is a
`.kittify/config.yaml` with `provider: beads` plus `SPEC_KITTY_ENABLE_SAAS_SYNC=1` — no network, no
auth, no beads binary. `beads` and `fp` behave identically.

**Why local bindings specifically:** `_check_sync_readiness` short-circuits with
`if _is_local_binding(): return`, so local bindings skip the entire readiness chain and land straight
on the raise site. SaaS bindings are stopped earlier by auth/reachability.

**Missing, not `None` or wrong-type**: `dir()` → False, no `__getattr__`, MRO
`['LocalTrackerService', 'object']`, while `sync_pull`/`sync_push`/`sync_run` all exist. An AST sweep
of every facade delegation, `hasattr`-checked against both backends, gives **8 methods × 2 backends =
exactly one missing cell**. 15/16 complete.

## The fix

`LocalTrackerService.sync_publish` refuses with a message naming the provider and pointing at
`tracker sync run`.

**The exception type is the load-bearing detail.** It raises `TrackerServiceError`, **not** the
sibling `LocalTrackerServiceError`. The CLI catches only the former (`tracker.py:461`);
`LocalTrackerServiceError` is a separate `RuntimeError` that appears **nowhere** in the CLI, so
raising it would have reached the operator as a traceback exactly as the `AttributeError` did. The
import is function-local because `service.py` imports this module lazily (`:67`, `:137`) to avoid a
cycle — a module-level import back would invert that direction.

## Why the pin is against the *backend*

The bug shipped green because the covering test at `tests/agent/cli/commands/test_tracker.py:1381`
used a bare `MagicMock()` with no `spec=`, so `sync_publish` auto-existed — **two lines after** the
same test correctly used `MagicMock(spec=SyncConfig)`.

Tightening that mock to `spec=TrackerService` would **still not catch this**, because the facade *does*
define `sync_publish`; only the backend didn't. So the pin targets the backend contract:

- `test_local_backend_implements_every_delegated_facade_method` derives the delegating surface from
  `service.py`'s AST rather than hardcoding it, so a ninth delegation added later is covered without
  editing the test. It reds with the exact defect: *"missing 1 of 8 delegated facade methods:
  `['sync_publish']`"*.
- `test_sync_publish_refuses_cleanly_on_a_local_binding` asserts the caught type.

That mock is deliberately left as-is; the parity test now covers the class regardless of it.

## Verification

| | |
|---|---|
| red → green (2 selected) | **2 failed, exit 1** → **2 passed, exit 0** |
| `tests/sync/tracker` | **474 passed** (472 baseline + these 2), `^ERROR tests/` 0 |
| `tests/agent/cli/commands/test_tracker.py` | **51 passed**, matching baseline |
| `ruff check` | clean |

Run sequentially — `tests/sync` and `tests/cli` never concurrently.

**Pre-existing and NOT fixed**, per the Pre-existing Failure Reporting Rule: two `no-any-return`
`mypy --strict` findings in this file, confirmed identical at the merge base (`:186`, `:209`) and
shifted to `:211`/`:234` only by the 25 lines this change adds.

## Scope, and what is deliberately excluded

This landed as a **governed Op, not a mission** — the pre-spec investigation (included as
`tracker-publish-investigation.md`) concluded the work was ~5 production lines and did not justify a
mission's overhead. The Op record travels as the proof trail.

**Deferred, not forgotten:** whether `tracker sync publish` should exist at all. The investigation
found it refuses on **all six providers**, `SyncEngine` exposes only `checkpoint/pull/push/sync` — so
no publish capability exists to route to — and the CLI success path prints `endpoint`/`status_code`/`ok`,
an HTTP shape no backend returns. That is a user-visible decision (DIR-009) and is out of scope here.

**#3169 is separately answered** by the same investigation and needs no mission: a calibrated AST sweep
over **1202 files** found **7 `_extra` accesses and zero key-level reads**, none on the publish path.
Nothing reads it — it is a write-only round-trip buffer, so it cannot be "read hopefully".

**State: DRAFT for operator review.**

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3250"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788651425&installation_model_id=427282&pr_number=3250&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3250&signature=904c560097e9892e22024aee1a4c5756b5e6004aa3cf6a0fbd8da33129d6960f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
## CI classification — 7 reds, **zero PR defects**

This PR adds one method: `LocalTrackerService.sync_publish`, which refuses by raising `TrackerServiceError` (already a `RuntimeError`, already caught at `tracker.py:461`), plus a parity test deriving the delegating surface from `service.py`'s AST. It starts no runtime, spawns no threads, and touches no gate. Every red below is attributed with evidence rather than assumed.

| Check | Verdict | Evidence |
|---|---|---|
| `arch-adversarial (arch_shard_2)` | **Pre-existing main breakage** | `test_routed_load_meta_floor` — reproduced byte-identically on a pristine `upstream/main` worktree (floor 128, margin 4, live 133). **#3247 clears this** by re-deriving the floor on the merged tree; nothing to do here. |
| `fast-tests-sync` | **Measurement artifact of #3136** | `@patch("specify_cli.tracker.saas_client.time.sleep")` resolves to the **stdlib** `time` module attribute, so the mock's recorder is process-global and counts sleeps from any live thread in the xdist worker. Mission `sync-sleep-count-3136` fixes it with an import-bound module-local alias; this PR's diff contains no sleep or retry lines. |
| `integration-tests-sync` | **Pre-existing main breakage** | FR-007 leak guard: `SyncRuntime` and `BackgroundSyncService` singletons plus live threads (`spec-kitty-sync-async-loop`, `spec-kitty-sync-runtime-start`) left dirty by `test_dual_write_integration.py` and `test_daemon_self_retirement.py`. **The identical four errors appear on #3135**, whose diff is unrelated to this one — which is what establishes it as main's rather than either PR's. |
| `integration-tests-status` | **Pre-existing main breakage** | Fails identically on #3247 and #3135. |
| `regression tests (blocking)` | **Pre-existing main breakage** | Fails identically on #3247 and #3135; traces to review-cycle, #2804 merge-gate reset, and #3086 coordination metadata. |
| `Clean install verification (FR-010)` | **Measurement artifact** | NFR-003 latency gate — median 1.058 s against a 1.050 s ceiling, missed by 8 ms. It passed on a newer run of the same commit on #3135. |
| `quality-gate` | **Derived** | Pure aggregator; its blocking verdicts include "improperly skipped" suites whose dependency failed. Clears when the above clear. |

Nothing was retried-to-green and no baseline was regenerated.

Two of these are cleared by work already in flight rather than by anything here: `arch_shard_2` by **#3247**, `fast-tests-sync` by **#3136**. The remaining five are main's.

Incidentally, the `integration-tests-sync` failures are the same *family* as the `fast-tests-sync` one — process-global state and live threads surviving a test boundary. #3136 addresses one instance of that class (a patched stdlib attribute acting as a shared recorder); the leak-guard errors are a sibling instance (threads and singletons not torn down). Recorded here rather than filed, per the standing direction to ledger rather than open issues.

Still a draft, for the operator.

robertDouglass:

Three independently profile-loaded lenses reviewed the current aggregate diff: `reviewer-renata` (contract/evidence), `debugger-debbie` (live CLI regression), and `architect-alphonso` (tracker seam/capability ownership).

**Verdict: `BLOCK_MATERIAL`.** The immediate `AttributeError` becomes a clean refusal, but exception ownership, operator messaging, and coverage remain materially unsound. Per queue policy, these were not fixed.

- **High — `src/specify_cli/tracker/local_service.py:164-171`:** rationale is false. `_run_or_exit` catches every `RuntimeError`, and `LocalTrackerServiceError` already subclasses `RuntimeError`. Importing facade-owned `TrackerServiceError` from the backend creates an unnecessary reverse dependency/cycle.
- **High — `src/specify_cli/tracker/local_service.py:174-177`:** message says publishing is a hosted-tracker operation, but `SaaSTrackerService.sync_publish` also refuses all hosted providers. It advertises a capability that does not exist.
- **High — `src/specify_cli/tracker/service.py:202-203`:** facade models `sync_publish` as universally supported although both concrete backends reject it. Adding a second refusal duplicates global policy instead of establishing one canonical capability decision.
- **High — `tests/agent/cli/commands/test_tracker.py:1350`:** existing CLI test still asserts a mocked success path. New tests call the backend directly, so they cannot guard the original operator contract: exit code, caught formatting, provider/remediation message, and no traceback.
- **Medium — `tests/sync/tracker/test_local_service.py:438`:** AST/`hasattr` parity checks only local name presence; it cannot detect SaaS omissions, non-callable attributes, signature drift, or incompatible result/error contracts.

**Live evidence:** real beads invocation exits 1 with one refusal line and no traceback; both new tests pass. This confirms symptom relief, not architectural correctness.

**Maintenance only:** rebased conflict-free onto current `main`, fixed the two pre-existing touched-file strict-mypy `Any` returns with boundary casts, and rerolled `e3cf173c5 → f3241bb85`. Verification: targeted ruff clean, strict mypy clean, `2 passed`, `git diff --check` clean. No material finding was modified.

MOES-Media:
## Duplicate: #3294 also fixes #3168

Flagging rather than resolving — this is a reviewer's call, and both PRs are sound.

**#3294** (@stijn-dejongh, opened today, ready for review) fixes the same defect on the same files (`src/specify_cli/tracker/local_service.py`, `tests/sync/tracker/test_local_service.py`) with the same approach: give `LocalTrackerService` a `sync_publish` that refuses cleanly instead of letting an `AttributeError` escape `_run_or_exit`'s `RuntimeError`/`ValueError` catch.

**Where they differ, factually:**

| | #3250 (this PR) | #3294 |
|---|---|---|
| Refusal message | raises `TrackerServiceError` | names the alternative: *"Use `spec-kitty tracker sync push` instead"* |
| Changelog | none | `docs/changelog/CHANGELOG.md` entry |
| Test | **AST-derived parity test** — derives the delegating surface from `service.py`'s AST, so a future backend method added to the facade without a local counterpart fails the test rather than shipping | direct refusal test |
| State | draft | ready for review |

The parity test is the only piece that isn't duplicated. It exists because the original bug shipped green: `test_tracker.py:1380` used a bare `MagicMock()` with no `spec=`, so `sync_publish` auto-existed — and fixing *that* mock with `spec=TrackerService` still wouldn't have caught it, because the **facade** does have `sync_publish`. The pin has to be against the backend, and derived rather than enumerated.

No objection to #3294 landing. If it does, the parity test is worth carrying over separately — happy to open it as a follow-on if that's useful.

stijn-dejongh:
@MOES-Media : please carry over the tests

Issue #3244, opened 2026-08-06
Arbiter override crashes on a conflict-marked review-cycle-N.md (FR-014 downgrade side-effect)


stijn-dejongh:
## Summary

Latent crash surfaced by the verdict-seam-write-unification mission (191) pre-merge adversarial squad (paula-patterns lens). **Not a landing blocker** — narrow reachability — filed as follow-up per the squad's recommendation.

## Defect

FR-014 downgraded the `spec-kitty-review-cycle` merge driver (`src/specify_cli/cli/commands/merge_driver.py`) from fail-closed-abort to **non-aborting (exit 0)**. On a two-verdict collision it now writes git conflict markers (`<<<<<<< / >>>>>>>`) into `review-cycle-N.md` and commits that file. This is acceptable in isolation because the `.md` is non-authoritative post-mission (the verdict lives in the event log).

However, a later **arbiter override** on that WP crashes:

- `src/specify_cli/review/arbiter.py:466` — `ReviewCycleArtifact.latest(wp_subdir)`
- → `src/specify_cli/review/artifacts.py:330` `from_file(candidates[-1])`
- → `artifacts.py:278` raises `ValueError("Review artifact file has no YAML frontmatter: ...")`
- **uncaught** at the arbiter call site → the override path crashes.

## Repro (conditions)

1. A review-cycle `.md` is conflict-marked by the downgraded merge driver and committed (exit 0, non-aborting).
2. An arbiter override is later run on that same WP.
3. `persist_arbiter_decision` → `ReviewCycleArtifact.latest` → `from_file` raises uncaught `ValueError`.

Previously unreachable: the old fail-closed driver aborted the squash, so no malformed file ever landed.

## Suggested fix

Make `ReviewCycleArtifact.latest` / the arbiter call site resilient to an unparseable *latest* file — derive `cycle_number` from the filename (which needs no frontmatter parse) rather than from `from_file()`, or guard the `from_file` call and fall back to the filename-derived cycle number. Add a red-first regression test (conflict-marked latest `.md` + arbiter override → no crash).

## Related

- Mission: `verdict-seam-write-unification-01KZ9Q35` (191), FR-014.
- Sibling gate blind-spot already tracked: #3236 (census wholesale-excludes `verdict_provenance_backfill.py`).

Issue #3193, opened 2026-08-04
FR-007 leak guard: anonymous-thread leaks are attributed non-deterministically, so the tests/sync error count is not reproducible run to run


MOES-Media:
## Summary

The FR-007 process-global leak guard in `tests/sync/conftest.py` intermittently attributes an
**anonymous live thread** (`target=None`) to a different test on each full-cone run. The set
of erroring node ids is not reproducible, which makes the `tests/sync` error count unusable
as a ratchet and makes any per-node cone comparison (e.g. `#3167` SC-006 / NFR-003) show
unattributable differences.

## Observed

Four full serial runs of `tests/sync` (`.venv/bin/python -m pytest tests/sync -ra
-p no:cacheprovider -n0`), each from a tree verified free of leaked daemons via
`pgrep -af 'run_sync[_]daemon'`:

| Run | Tree | Result | Errors |
|---|---|---|---|
| 1 | base `f04ee0a78` | 2376 passed, 19 skipped | **5** |
| 2 | base `f04ee0a78`, the 89 `#3167` nodes `--deselect`ed | 2287 passed, 19 skipped, 89 deselected | **5** |
| 3 | `#3167` WP02 applied | 2287 passed, 19 skipped | **7** |
| 4 | `#3167` WP02 applied, re-run | 2287 passed, 19 skipped | **6** |

The **same 5** errors appear in every run and are the known, filed set:

```
tests/sync/test_daemon_self_retirement.py::TestRunSyncDaemonWiring::test_serve_forever_exits_cleanly_when_server_shutdown
tests/sync/test_daemon_self_retirement.py::TestRunSyncDaemonWiring::test_sigterm_exits_without_deadlocking_server_shutdown
tests/sync/test_dual_write_integration.py::TestDualWriteEventAndFrontmatterConsistent::test_dual_write_event_and_frontmatter_consistent
tests/sync/test_dual_write_integration.py::TestDualWriteMultipleTransitions::test_dual_write_multiple_transitions
tests/sync/test_lifecycle_readiness.py::test_init_emits_project_init_event_offline
```

The **excess** errors differ every time, and every one of them reports only an anonymous
thread:

- Run 3 (+2): `test_background.py::TestSingletonAccessor::test_get_sync_service_returns_same_instance`
  (`live thread name='Thread-7' ident=... target=None`) and
  `test_daemon_self_retirement.py::TestStartSelfCheckTick::test_returned_timer_thread_is_daemon`
[...]

PR #3252, opened 2026-08-07
fix(sync): end the SaaS tracker sleep-count CI flakes + add a repo-wide patch-seam gate (#3136)

### LINKED ISSUE #3136: test(sync): #3115 sync-half sleep-count producer named — psutil-backed waits excluded, victims corrected

MOES-Media:
## What this does for you

Spec Kitty's SaaS tracker retry/poll tests **no longer flake on CI** — and a new architectural gate makes sure the whole class of "a test patches a shared stdlib object and grades the wrong number" bug **can't quietly come back**.

If you run the suite, you'll stop seeing spurious failures like `AssertionError: Expected 'sleep' to be called once. Called 179 times.` landing on test nodes your change never touched.

## What ships

- **The SaaS tracker client is now fully isolated from process-global stdlib state.** Building on [#3187](https://github.com/Priivacy-ai/spec-kitty/issues/3187) — which gave `SaaSTrackerClient` its own `self._sleep` seam — the client now also binds `self._monotonic` and `self._randbelow`. Retry/poll timing and jitter can no longer be perturbed by (or perturb) anything else running in the same test worker. The dead `_poll_jitter_multiplier` helper is removed.
- **A new repo-wide gate** (`tests/architectural/test_shared_module_object_patches.py`, with its census tooling and control fixtures) refuses any `patch()` that mutates a shared module object read by a count/equality assertion. It ships with a frozen, shrink-only baseline of the sites that already exist, so the count can only go down — a new instance of the bug fails review instead of merging.
- A consumer-facing **changelog** entry and a small **venv-safety** doc fix (`.venv/bin/python` for the UI e2e suite, not a venv-destroying bare `uv run`).

## Why this looks different from where it started

This mission's original product fix — a module-local `_sleep` alias — was **superseded by #3187** (an instance `self._sleep` seam) before it could land; the two efforts fixed the same root cause independently and unaware of each other. Per maintainer adjudication the PR was **rescoped**: drop the now-redundant alias (and the ADR/guard that only described it), keep the durable, idiom-independent gate, and extend #3187's seam to finish closing the tracker's remaining stdlib reach-throughs. The full mission dossier records this under a "Landing rescope" banner.

## Known CI reds — neither caused by this PR

- `regression tests (blocking)` — a **by-design** red-first P0 signal ([ADR 2026-07-17-1](https://github.com/Priivacy-ai/spec-kitty/blob/main/docs/adr/3.x/2026-07-17-1-red-main-is-honest-ci-is-release-authority.md)); red is the intended state, cleared only by fixing the pinned product bug.
- `integration-tests-sync` — **pre-existing `main` breakage** in an unrelated subsystem (`test_dossier_sync_fires_during_mission_creation` → `CHARTER_PACK_CONFIG_INVALID`), reproduces on a clean `main` checkout; filed as #3266.
- `quality-gate` is a meta-aggregator that goes green when the two above are addressed.

Everything this PR actually touches — the gate, census, ratchet, tracker seam, docs, terminology, `ruff`, `mypy`, `commitlint` — is green, verified from a non-dot-path checkout and confirmed by the `arch-adversarial` shards on CI.

Closes #3136.
Refs #3187, #3266.


MOES-Media:
## WP07 landing summary — mission `sync-sleep-count-3136-01KZ9B5A`

**Verdict: DRAFT, land-ready for operator review, with one self-inflicted red disclosed below.**
Do not merge or un-draft on my account — both are the operator's actions.

### Local gates on the composed tree

| Gate | Result |
|---|---|
| `ruff check .` | `All checks passed!` · `EXIT=0` |
| `mypy --strict src/specify_cli/tracker/saas_client.py` | 2 errors — **both pre-existing**, identical rule/count on `upstream/main`'s copy (`:162-163` → `:184-185`, a 22-line shift equal to the seam block) |
| `tests/architectural/` | `2 failed, 1778 passed, 5 skipped, 2 xfailed in 769.03s` — both failures classified below |
| `tests/architectural/test_no_legacy_terminology.py` (C-010, CI-only gate) | `10 passed` · `EXIT=0`, run **before** push |
| `tests/sync/tracker/` cone | `466 passed, 11 deselected` · `EXIT=0` · `^ERROR tests/` = 0 |
| `scripts/docs/check_docs_freshness.py --ci` | `exit=0 errors=0` |
| `commitlint --from upstream/main --to HEAD` | `EXIT=0` |

### Red-first proof

```
base saas_client.py  ->  5 failed  EXIT=1   AttributeError: ... does not have the attribute '_sleep'
fixed saas_client.py ->  5 passed  EXIT=0
```

### Every CI failure classified against pristine `main`'s last six `CI Quality` runs

| Job | `main` last 6 | this PR | Class |
|---|---|---|---|
| **`lint`** | **success 6/6** | **fail** | **(d) THIS DIFF** — see below |
| `arch-adversarial (arch_shard_2)` | fail 4/6 | fail | mixed: pre-existing floor + a new integration crossing |
| `fast-tests-sync` | fail 6/6 | fail | (a) pre-existing; the 1 error is the `FR-007` leak class (`#3130`/`#3193`), **no `tests/sync/tracker/` node failed** |
| `integration-tests-sync` | fail 5/5 | fail | (a) pre-existing; all 4 errors are the same leak class |
| `integration-tests-status` | fail 5/5 | fail | (a) pre-existing; the fixture reads another mission's `meta.json` |
| `regression tests (blocking)` | fail 6/6 | fail | (a) pre-existing inverted-red markers on open P0s |
| `quality-gate` | — | fail | roll-up, no independent signal |

**Nothing was retried to green.** `lint` is the only job green on `main` and red here.

### ⚠ The one thing to look at first

`lint` fails with `Broken patch() targets (1 of 5087 checked)` at
`tests/architectural/test_shared_module_object_patches.py:5` — and **the "target" is docstring
prose** explaining the mechanism (``patch("a.b.c.attr")``). `scripts/check_patch_targets.py`'s
`extract_targets` is a regex over raw source with no AST awareness, so it cannot tell a call from a
sentence about a call.

Two defensible fix shapes, two different owners — **deliberately not chosen here**, because picking
between a gate change and a prose change is a design decision and both files are already-approved
deliverables of other work packages:

1. make `extract_targets` AST-aware (the structural fix; removes the whole false-positive class), or
2. break the literal in the docstring (one line; leaves the gate able to redden on future prose).

Recorded as `RL-048`, together with the observation that **this exact class bit four times in this
one work package** — the gate, the `SC-012` suppression grep, `C-002`'s own grep, and finally WP07's
notes file matching the pattern the moment it *quoted* the offending lines. The generalisation, not
the one-line fix, is the useful output.

### Also needing a ruling

`test_no_unregistered_baseline_keys_are_added` is an **integration crossing**, not a defect on either
side: WP05 ships a guard refusing unregistered `_baselines.yaml` keys; `upstream/main` has since
added exactly such a key (`test_verdict_seam_census`). Registering it would make this mission's gate
enforce another mission's ratchet; grandfathering it is structurally blocked by the guard's own
closed frozenset-equality pin. **Needs the gate owner.** → `RL-046`

### Governance

Full `kitty-specs/sync-sleep-count-3136-01KZ9B5A/**` dossier travels with this PR, including
`residual-ledger.md` with ten new entries (`RL-040`…`RL-049`). Per standing operator direction for
this programme, gaps are recorded in the ledger rather than opened as issues, and **no issue number
is invented or reserved**. The `C-001` `tests/sync` window handshake is closed
(`2026-08-06T22:46:46Z` → `2026-08-07T07:02:51Z`).

Lane consolidation (`spec-kitty merge`) has **not** been run — it is the mission's terminus step and
deletes lane branches. This branch is composed from the lanes per the pr-landing recipe, with
`_baselines.yaml` and the ADR index **3-way merged** and both docs lockfiles **regenerated** against
the current base, so upstream's own entries survive rather than being clobbered.

A non-gating CI observation is recorded at `kitty-specs/.../notes/ci-observation-3136.md`. It is
**not** acceptance evidence and is not cited as proof anywhere.

MOES-Media:
### `lint` red fixed — the mission's own phantom-target class, turned on its own prose

**What reddened.** `scripts/check_patch_targets.py` is `[ENFORCED]` at `ci-quality.yml:884` and was the
one CI failure attributable to this diff (`lint` is green on `main`):

```
::error::Broken patch() targets (1 of 5087 checked):
  tests/architectural/test_shared_module_object_patches.py:5: cannot import any prefix of 'a.b.c'
```

**The cause is this mission's own defect class biting the prose that describes it.** The gate extracts
targets with a regex over *raw source*, so a dotted literal written inside a **module docstring** is
mined as if it were a live `patch()` call and then fails to import. WP05's gate docstring spelled one
purely as an illustration of the mechanism. A text-matching probe grading a sentence about a call as
a call — which is the very thing this mission exists to fix.

**The gate's regex was deliberately NOT narrowed.** That was the tempting fix and it is the wrong one.
The `\s*` legitimately bridges newlines for multi-line `patch(\n    "target")` calls. Re-measured on
this PR head rather than transcribed: narrowing it to `[ \t]*` silently drops **1681 of 5086** targets
— roughly a third of an enforced gate's corpus — while looking like hardening. Recorded as `RL-018`.

**Fix applied to the prose**, using the technique WP03 already established in
`test_patch_seam_census_control.py`: a **brace placeholder**, so the regex meets `{` and cannot match.
The docstring now reads `patch("{mod}.attr")`. Explanatory value is preserved in full and extended
with a note saying why the placeholder must stay. It is self-enforcing in the right direction — an
author who later writes a concrete dotted path there gets a loud CI red rather than a silent
mis-patch.

**Corpus count measured on this composed tree: `All 5086 patch() targets valid.` (exit 0).** The total
drops by exactly one from 5087 — the phantom, and nothing else.

**Red-first proof:** restoring the literal reproduces the failure verbatim at 5087; reverting is
byte-identical (`sha256 bc6a2692…97a63a`), and the gate returns to 5086/exit 0.

**Two things worth flagging for review:**

1. **`RL-048`'s suggested fix shape does not work.** It proposed splitting the literal as
   `patch("a.b.c" ".attr")`. Measured: the capture group closes at the first quote, so that still mines
   `a.b.c`, still unresolvable, gate still red. Corrected in the ledger.
2. **Scope is narrower than it looks, and a blanket sweep would have broken CI.** Running the shipped
   `extract_targets` over all of `tests/` finds **12** prose-sited targets but only **1** unresolvable.
   The other 11 are deliberately untouched: `seam_decoy_cases.py` carries a resolvable prose target
   *by design* (SC-015 control fixture), and **Arm F of the census control asserts `regex_only` is
   non-empty** over `tests/sync/` — that set is exactly the five prose sites there, so making them all
   unmineable would fail the arm's own non-vacuity assertion.

This closes one instance, not the class. The class closes only with an is-this-code predicate in the
extractor, which is a separate call and remains unfiled.

**C-002 assessed, not touched:** its grep still returns 3, all three WP prose *asserting the formatter
was never run* — an honest WP fails, a silent one passes. It has **no automated enforcement anywhere**
(zero hits across `.github/workflows/`, `scripts/`, `tests/`), so it reddens nothing here. Recorded in
the ledger rather than worked around.

Other CI reds on this PR were controlled against `main`'s last six runs and are pre-existing or
out-of-scope, not chased to green.

MOES-Media:
## WP07 review cycle 2 — three blockers closed

**1 — the CI note's tripwire had fired and was never discharged.** It is now.
`notes/ci-observation-3136.md` carries a dated addendum re-anchored to `bf68b101b`:
`lint` is **green**, the board is **28 pass / 6 fail / 36 skipping** (was 27/7/36 at
`b0312a438`), and the `lint` row is struck as superseded rather than deleted — the
method that caught it (controlling against `main`) is worth keeping visible.

The invariant itself was the real defect: it pinned *which files moved*, so a
legitimate fix falsified it. Restated in terms of **attribution** —

> Every commit after `b0312a438` is either a `kitty-specs/**` dossier update or a fix
> for a red this note itself classified.

— which a fix satisfies instead of breaking.

**2 — the PR body named a resolved red as the top landing risk.** Item 2 is now struck
and marked RESOLVED, with the measured detail (the string-split shape *doesn't* work;
narrowing the regex drops 1681/5086) and a pointer to the remediation comment. Items 1
and 3 untouched.

**3 — the fix was one branch from the lane that owns it. Ported, not documented.**
`123e2dab3` on `lane-e`. This was worth doing rather than warning about: `feat/` does
not carry `test_shared_module_object_patches.py` at all, so consolidation would have
installed lane-e's mineable copy and reddened `[ENFORCED]` `lint` again *after* this PR
showed it green. Verified the docstring hunk was the **only** delta before porting; the
lane blob and the shipped blob are now the same git object, and no other lane carries
the literal.

**Ledger drift folded.** Both clauses of the ID-allocation header had gone false. `feat/`
stays authoritative for **picking an id**; `pr/sync-sleep-count-3136` is the copy that
**travels**. `bf68b101b`'s addendum is ported back — the two copies are byte-identical
again, with the diff command printed so the next reader can check.

**Five LOWs, each re-measured rather than accepted:**
- the `SC-012` probe must be **path-scoped** — unscoped it reads its own report (49 hits,
  **43 of them dossier prose**, vs 6 scoped). Fifth instance of the class this very file
  generalises; documenting a failure mode does not apply it.
- suppression line numbers are `+10` on HEAD — now located **by command**, not frozen.
  The guessed `:~64` was really `:68`; guesses do not belong in a transcript.
- `test_shared_module_object_patches.py` is **+1020**, not +1010.
- the loud sibling is **2 lines**, not "129 bytes" — `git diff --stat` pads to terminal
  width (123 @72, 131 @80, 139 @100), so the byte figure was never reproducible.
- `RL-040` now **prints the predicate** behind its `48`, per its own rule; `RL-048` names
  the new file, since the gate it chokes on is pre-existing upstream.

Re-verified on the shipped tree: `All 5086 patch() targets valid.` `EXIT=0`;
`ruff check .` → `All checks passed!` `EXIT=0`. The two remaining open reds (`RL-046`
crossing, `RL-047` pre-existing floor) are **unchanged and byte-identical** to the earlier
run — neither retried to green. Still a **DRAFT**.

stijn-dejongh:
🐱 **Maintainer landing — claim**

Claiming PR #3252 for today's landing pass (campsite-cleaning mode). Plan, per the [maintainer landing runbook](https://github.com/Priivacy-ai/spec-kitty/blob/main/docs/development/pr-landing.md):

1. Isolated worktree, rebase onto current `upstream/main` (branch is currently `CONFLICTING`/`DIRTY`).
2. Classify every red against main's own CI at the merge-base first — the three pre-disclosed reds (RL-046 ratchet-baselines integration crossing, RL-047 `test_routed_load_meta_floor` pre-existing, RL-048 resolved) will be re-adjudicated on the rebased tip, not taken on trust.
3. Fold remediations as separate `<type>(landing): …` commits (commitlint-clean), delegating implementation to profile-loaded subagents where sensible.
4. Red-first verification of the `fix(tracker)` seam.
5. Remediation summary + hand-off. **I do not merge** — I hand off green-modulo-honest-reds; the operator merges.

Evidence to follow in-thread.

stijn-dejongh:
## 🐱 Maintainer landing — remediation summary & hand-off

**Verdict: rescoped, not landed-as-written.** This PR's headline product fix was **superseded by [#3187](https://github.com/Priivacy-ai/spec-kitty/issues/3187) before it could land**, so it has been rescoped (per operator adjudication) to salvage the durable, idiom-independent half. Rebased onto current `upstream/main`; the branch was `CONFLICTING`, now `MERGEABLE`.

### Why the rescope

The same root-cause CI flake this mission targets (#3136) was fixed independently on `main` by commit `958baf531` (issue **#3187**, closed 2026-08-08) — but with a **different mechanism**: an **instance-scoped `self._sleep` seam**, not the **module-local `_sleep` alias** this mission's R-1 prescribed. Only one idiom can live in `saas_client.py`; landing the module-local alias would have *reverted* #3187. The two efforts never saw each other (the "neither party saw the other" pattern this PR itself flagged for `RL-046`).

The operator's ruling: **drop everything coupled to the rejected module-local idiom; salvage the durable repo-wide patch-seam gate; and extend #3187 so `saas_client` is fully clean.**

### What ships now (4 commits, all `<type>(landing):`, commitlint-clean `--verbose`)

| Commit | What | Why |
|---|---|---|
| `fix(landing)` | Extend #3187's instance seam: `self._monotonic`/`self._randbelow` + reroute call sites; retarget 10 tracker test patches (9× `time.monotonic`, 1× `secrets.randbelow`) off the process-global stdlib; remove dead `_poll_jitter_multiplier`. | Closes the `saas_client` reach-throughs #3187 left open, so the salvaged gate lands with `saas_client` clean. |
| `test(landing)` | Salvage the repo-wide patch-seam gate + census + control fixtures. **Cut the 8 arms that self-verified the module-local alias** (seam 4a–4d + twins, the seam-split node-id arm, the alias-bypass twin) and their exclusive helpers. Keep the ratchet core (`baseline == gate-flagged-set`) + idiom-agnostic synthetic controls. Regenerate `_baselines.yaml` against `main`. Register both new arch files in `tests/_arch_shard_map.py`. | The gate (refuse a `patch()` that mutates a shared module object read by a count/equality assertion) is idiom-independent and `main` lacks it. The alias-idiom arms are false/meaningless under #3187's seam. |
| `docs(landing)` | `AGENTS.md`/`ui-e2e.md` venv-safety note (`.venv/bin/python`, not venv-destroying bare `uv run`); process-global inventory `#3136` stamp. | Idiom-independent; `main` didn't have the venv note. |
| `chore(landing)` | The mission dossier, **rescoped**: a `LANDING RESCOPE` banner in `spec.md` + landing addendum (`LA-01..03`) in `residual-ledger.md`. | Records honestly that the product half was superseded; the R-1 alias text is retained as historical, marked not-shipped. |

**Dropped** (superseded / coupled to the rejected idiom): the module-local `_sleep` alias product fix, the ATDD guard `test_sleep_attribution_guard_3136.py`, and the ADR `2026-08-06-1-module-local-stdlib-alias-seam.md` (+ its era-index row).

### `RL-046` — gate-owner ruling (resolved)

The salvaged ratchet's reverse-containment arm (`test_no_unregistered_baseline_keys_are_added`) tripped on `main`'s `test_verdict_seam_census` key. It is a **self-contained in-file census** (compared against `census/verdict_seam_IC01.yaml`), read by *no* `test_ratchet_baselines.py` comparison — so it was added to the CLOSED `_GRANDFATHERED_UNREGISTERED_KEYS` set with a dated rationale. Registering it in a comparison list (the arm's default suggestion) would reproduce the "required, never read" defect the arm warns against; grandfathering is the correct bin. This is the visible-diff-in-this-file the self-pin is designed to force.

### The gate's baseline = 11 pre-existing `main` debt sites (recorded, not fixed)

The regenerated `test_shared_module_object_patches` baseline is a frozen, shrink-only record of reach-through patches **already on `main`**, owner `unassigned`:
`git_metadata.subprocess.run` ×6, `body_transport.requests.post` ×2, `batch.time.sleep` ×1, `asyncio.run_coroutine_threadsafe` ×2. **`saas_client` is not among them** — closed by the `fix(landing)` seam extension. These 11 are honest debt this PR records; burning them down is out of scope (see follow-ups).

### Evidence — all from a **non-dot-path** checkout (`/tmp/verify-3252`), not the landing worktree

| Check | Result |
|---|---|
| gate + census-control + ratchet | **25 passed, 1 pre-existing skip** |
| completeness gates (`arch_shard_marker_completeness` + `gate_coverage`) | **44 passed** |
| tracker tests (`test_saas_client{,_origin}.py`) | **73 passed** |
| `check_docs_freshness.py --ci` | **errors=0** (7 warnings, all external-URL link-health) |
| terminology guard | **10 passed** |
| `ruff check` (whole diff) | **All checks passed!** |
| `mypy src/…/saas_client.py` | **0 new** — 2 pre-existing `no-any-return` in `_current_team_slug_sync` (`RL-015`), outside every changed hunk |
| commitlint (4 commits, `--verbose`) | **0 problems, 0 warnings** each |

CI is running now; **merge-ready on green.**

### For the author (@MOES-Media / Jeroen)

Nothing was wrong with your investigation — the forensics in #3136 were excellent and are preserved. The module-local alias simply lost a race to #3187's instance seam, which `main` adopted first. Your durable contribution — the repo-wide patch-seam gate — **ships**, trimmed to be idiom-agnostic, and it even surfaced (and this PR closes) two `saas_client` reach-throughs #3187 left behind.

### Hand-off (@stijn-dejongh — the operator merges; I did not)

- **Left as `draft`.** Given how substantially the scope changed, I did not un-draft — un-draft + merge at your discretion after reviewing the rescope. (Flagging the runbook-vs-PR-convention drift: the runbook lists "un-drafted" as the maintainer deliverable, but this PR's body reserves un-drafting for the operator; I deferred to the latter given the rescope.)
- **No landing-order constraint** — the branch is rebased on current `main` and self-contained.
- **Out-of-scope follow-ups to file:**
  1. Burn down the 11 pre-existing reach-through debt sites the gate now records (`git_metadata`/`body_transport`/`batch`/`asyncio`) — parent under the sync/test-hygiene epic.
  2. The 2 pre-existing `no-any-return` in `saas_client._current_team_slug_sync` (`RL-015`).
- **Worktrees** (`.worktrees/pr-3252-landing`, `.worktrees/pr-3252-pedro`, `/tmp/verify-3252`) + the SSH remote/tracking refs will be cleaned up after you merge.

stijn-dejongh:
## 🐱 CI verdict — green-modulo-honest-reds

CI on `527c0bd1a` finished: **31 pass, 36 skip, 3 fail.** All three reds are classified below; **none are attributable to this PR's diff**, and two of the original PR's reds were *fixed* by the rescope.

### ✅ Reds the rescope FIXED (red on the pre-rescope PR → green now)

- **`arch-adversarial (arch_shard_2)`** — was red because the two new `tests/architectural/` files were never registered in the completeness baselines; the `test(landing)` commit registers them. **All three arch shards now pass** (they run the salvaged gate + census + ratchet).
- **`fast-tests-sync`** — the sleep-count flake itself; now green (the root cause is closed by #3187 + this PR's seam extension).

### 🟥 Remaining reds — all pre-existing / by-design, classified

| Job | Classification | Evidence |
|---|---|---|
| **`regression tests (blocking)`** | **By-design honest red** (ADR 2026-07-17-1). `pytest -m regression` runs `tests/regression/` — intentionally-red, issue-pinned P0 reproductions. Red is the designed state; cleared only by fixing the pinned product bug, never by this PR. | This PR touches nothing under `tests/regression/`; it was red on the pre-rescope PR too. |
| **`integration-tests-sync`** | **Pre-existing `main` breakage** (runbook step-4 bin 3), out of scope. `tests/sync/tracker/test_origin_consumer.py::test_dossier_sync_fires_during_mission_creation` → `CharterPackConfigError: CHARTER_PACK_CONFIG_INVALID` at `create_mission_core` (`mission_creation.py:369`). Charter-pack-config subsystem, unrelated to the sleep-seam/patch-gate. | **Reproduces identically on a clean `upstream/main` (`c82f4f1f1`) checkout** — verified locally. Not foldable here (different subsystem, wide blast radius → carve-out b). |
| **`quality-gate`** | **Meta-aggregator**, not an independent signal — it evaluates the other jobs' outcomes, so it is red *because* the two above are red. It goes green when they do. | — |

### Hand-off (unchanged)

Green-modulo-honest-reds, as promised. The only failing checks are a by-design P0 signal and a pre-existing `main` breakage in an unrelated subsystem — the rescope's own surface (gate, census, ratchet, tracker seam, docs, terminology, ruff, mypy, commitlint) is fully green, verified from a non-dot-path checkout and confirmed by the arch shards on CI.

**@stijn-dejongh** — still `draft`, still yours to un-draft + merge. Added follow-up to file: the pre-existing `integration-tests-sync` breakage on `main` (`test_dossier_sync_fires_during_mission_creation` / `CHARTER_PACK_CONFIG_INVALID`) — filing it lets subsequent landing passes skip re-reproducing it.

stijn-dejongh:
**Un-drafted + follow-up filed** (per operator instruction):

- PR marked **ready for review**.
- The pre-existing `integration-tests-sync` breakage (`test_dossier_sync_fires_during_mission_creation` → `CHARTER_PACK_CONFIG_INVALID`, reproduces on clean `main`) is filed as **#3266** — a stale test fixture vs the WP04 fail-closed mission-create boundary, needs epic-parenting at triage. Unrelated to this PR's diff.

The two remaining CI reds (`regression tests (blocking)` by-design, `quality-gate` meta-aggregator) are classified in the [CI verdict](https://github.com/Priivacy-ai/spec-kitty/pull/3252#issuecomment-5225950436) above. Ready for the operator to merge.

stijn-dejongh:
## 🐱 Maintainer landing — re-pickup

Re-picking up **#3252**: the 2026-08-08 rescope was handed off green-modulo-honest-reds but not yet merged, and `main` has since advanced — the branch is `CONFLICTING`/`DIRTY` again. This pass:

1. Rebase the rescoped branch onto current `upstream/main` (`be2961c76`), resolving conflicts in the canonical locations (changelog, docs retrieval index regen, arch shard-map registrations, gate baselines regen).
2. Re-adjudicate the honest reds on the new tip — confirm #3266 (`integration-tests-sync` / `CHARTER_PACK_CONFIG_INVALID`) still reproduces on clean `main`, and watch for any new red introduced by main's advancement.
3. Re-verify the salvaged surface (patch-seam gate + census + ratchet, tracker seam, docs, terminology, ruff, mypy, commitlint) from a non-dot-path checkout.
4. Re-push (force-with-lease) + refresh the hand-off. **I do not merge** — the operator does.

Evidence to follow.

stijn-dejongh:
## 🐱 Re-pickup — rebased onto current `main`, re-verified, one fewer honest red

The 2026-08-08 rescope had gone `CONFLICTING` again as `main` advanced. Rebased onto current `upstream/main` (`be2961c76`), re-verified the salvaged surface, and re-pushed. **Now `MERGEABLE`** (was `CONFLICTING`); head `269dee001`.

### Rebase

Clean replay of all 5 rescoped commits except **one** conflict — `docs/changelog/CHANGELOG.md`, resolved in the canonical file by keeping **both** entries (main's `#3244` arbiter-override fix and this PR's `#3136` SaaS-tracker fix). No other conflicts; the gate baseline, arch shard-map registrations, and docs index all replayed clean.

### 🟢 Re-adjudication: the pre-existing `integration-tests-sync` red is now GREEN on `main`

Last pass's carve-out **b** was `test_dossier_sync_fires_during_mission_creation` → `CHARTER_PACK_CONFIG_INVALID` (filed as **#3266**). On current `main` that test now **passes** — `main`'s `resolution-activation-foundation` work (single-`PACKS_ROOT` + charter-activation-authority unification) fixed the charter-pack-config path this PR never touched. Verified locally on the rebased tip (`1 passed`); this PR's diff does not touch that subsystem, so the pass reflects `main`, not this PR. **#3266 looks resolvable** — recommend confirming on CI's `integration-tests-sync` and closing it (noted on the issue).

### Re-verification on the rebased tip (evidence)

| Check | Result |
|---|---|
| patch-seam gate (`test_shared_module_object_patches`) | **11 passed, 1 pre-existing skip** |
| ratchet baselines (`test_ratchet_baselines`) | **5 passed** |
| completeness (`arch_shard_marker_completeness` + `gate_coverage`) | **7 + 37 = 44 passed** |
| tracker seam (`test_saas_client{,_origin}`) | **73 passed** |
| `check_docs_freshness.py --ci` | **errors=0** (4 external-URL link-health warnings) |
| terminology guard | **10 passed** |
| `ruff check` (whole diff) | **All checks passed!** |
| `mypy` (`tracker/saas_client.py`) | **0 new** — 2 pre-existing `no-any-return` in `_current_team_slug_sync` (RL-015) |
| commitlint (5 commits, each) | **PASS** |

(The two dot-path-sensitive arch gates root their scan at the worktree via a relative `tests/` / `_REPO_ROOT/tests/sync` and do not filter dot-segments, so these greens are valid; CI's `arch-adversarial` shards are the authoritative non-dot-path confirmation.)

### Remaining honest reds (unchanged classification)

- **`regression tests (blocking)`** — by-design red-first P0 signal (ADR 2026-07-17-1); this PR touches nothing under `tests/regression/`.
- **`quality-gate`** — meta-aggregator; red only because of the regression job. Goes green when it does.

### Hand-off (@stijn-dejongh — the operator merges; I did not)

- Rebased, `MERGEABLE`, un-drafted, self-contained — **no landing-order constraint**.
- Watching CI on `269dee001`; **merge-ready on green-modulo-the-by-design-regression-red**.
- Out-of-scope follow-ups still stand: burn down the 11 pre-existing reach-through debt sites the gate records; the 2 `no-any-return` (RL-015). #3266 now looks fixed on `main` — verify on CI and close.

PR #3294, opened 2026-08-10
fix(#3168): tracker sync publish on a local binding fails cleanly instead of a traceback

### LINKED ISSUE #3168: `tracker sync publish` raises an uncaught AttributeError on a beads/fp binding

stijn-dejongh:
## What this fixes, for an operator

If you have a local `beads`/`fp` tracker binding, `spec-kitty tracker sync publish` crashed with a raw Python `AttributeError` traceback. Local providers have no snapshot-publish transport, but the command delegated to the backend unconditionally, and the CLI's error handler (`_run_or_exit`) catches only `RuntimeError`/`ValueError` — so the `AttributeError` escaped as a traceback.

**After this change** it exits cleanly: *"Snapshot publish is not supported for local providers (beads/fp). Use `spec-kitty tracker sync push` instead."*

## The fix

Add a `sync_publish` to `LocalTrackerService` that raises `LocalTrackerServiceError` (a `RuntimeError` subclass, so `_run_or_exit` renders it as a clean exit 1) — matching the contract the command already documents for local providers.

## Tests

- **Red-first:** `TestLocalSyncPublishUnsupported` (direct call + the real `TrackerService` facade-delegation path) raises `AttributeError` against the pre-fix backend and `LocalTrackerServiceError` after.
- Full `tests/sync/tracker/test_local_service.py` green (31 passed); `ruff` / `mypy --strict` / `commitlint` clean.
- Also carries a boundary cast for a pre-existing touched-file strict-mypy `Any`-return (`_format_local_refusal`, present verbatim on `main`).

Closes #3168.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3294"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788938417&installation_model_id=427282&pr_number=3294&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3294&signature=95c6d8203abe6a097c0834de27ca46f84acfc1709cca255b1fd9c712dd9c1516"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -347,6 +347,13 @@ _The 3.2.6 development cycle is open. Entries land here as missions merge._
 
 ### 🐛 Fixed
 
+- **`spec-kitty tracker sync publish` on a local (`beads`/`fp`) binding now
+  prints a clear error instead of crashing with a Python traceback (`#3168`).**
+  Local providers have no snapshot-publish transport, but the command delegated
+  to the backend unconditionally and hit an uncaught `AttributeError`, which the
+  CLI let escape as a raw traceback. It now exits with a clean
+  "not supported for local providers — use `tracker sync push` instead" message.
+
 - **A spec commit that genuinely FAILS is no longer silently reported as
   "unchanged" (`#3269`).** When a `git commit` failed for a real reason — a
   rejecting pre-commit hook, a lock error — `safe_commit` collapsed every
--- a/src/specify_cli/tracker/local_service.py
+++ b/src/specify_cli/tracker/local_service.py
@@ -86,7 +86,7 @@ def _render_refusal(verdict: TrackerEgressVerdict) -> str:
     them; it is not a path-local message string (FR-012).
     """
     if not verdict.remedies:
-        return verdict.message
+        return cast("str", verdict.message)
     remedy_lines = "\n".join(f"  - {remedy}" for remedy in verdict.remedies)
     return f"{verdict.message}\nRemedies:\n{remedy_lines}"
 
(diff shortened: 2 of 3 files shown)


MOES-Media:
## Note: #3250 is an earlier duplicate of this fix

Not a blocker, and no objection to this landing — flagging so a reviewer isn't surprised to find two PRs for #3168.

**#3250** (draft, opened 2026-08-07) fixes the same defect on the same two files with the same approach. Your operator message is better — it names `sync push` as the alternative, which #3250's doesn't — and yours carries a changelog entry.

The one piece that isn't duplicated is #3250's **AST-derived parity test**: it derives the delegating surface from `service.py`'s AST and asserts every method the facade delegates has a local counterpart, so a future backend method added to the facade without one fails the test rather than shipping.

That matters here specifically because the original bug **shipped green**. `test_tracker.py:1380` used a bare `MagicMock()` with no `spec=`, so `sync_publish` auto-existed — and repairing that mock with `spec=TrackerService` would *still* not have caught it, because the facade genuinely has `sync_publish`. The pin has to be against the backend, and derived rather than enumerated, or the next missing cell ships the same way.

If this PR lands, happy to carry that test over as a follow-on rather than have it die with #3250. Your call.

PR #3346, opened 2026-08-12
fix: isolate explicit owned-checkout mission state

robertDouglass:
## What changes for you

Running `spec-kitty agent mission create` or `spec-kitty next` from a **caller-owned linked git worktree** now works — and each worktree's mission state stays isolated.

Before, invoking either from an explicit linked checkout followed the worktree's `.git` pointer back to the primary checkout and tried to drive the mission there. That failed with `START_BRANCH_FAILED` when the branch was already checked out, and one worktree could read or write another worktree's mission state.

Now the invoking checkout root is validated with **fail-closed git-topology ownership checks** and used as the effective root end-to-end, so two linked worktrees can create and advance distinct missions **concurrently, with zero cross-contamination**, while the primary checkout is left untouched. Managed Spec Kitty coordination and lane worktrees remain refused.

Also fixes the ADR inventory freshener to target the canonical `3.x` era index instead of the redirect README, so newly added ADRs are actually indexed.

Fixes #3328, #3345.

## The fix

- New `src/specify_cli/core/checkout_ownership.py` classifies an explicit checkout claim as `OWNED` / `NESTED` / `FOREIGN` / `BROKEN_POINTER`, reusing the canonical `is_worktree_of` git-topology seam.
- The validated effective root is computed once and threaded through mission creation, resolution, `next` (query, advance, and answer), runtime state, refs, and commits.
- Refusals emit a stable, structured `error_code` envelope (unified across `create` and `next`).

## Tests

- A `checkout_ownership` unit suite, plus a real-worktree **installed-wheel e2e concurrency acceptance** — two linked worktrees, mission-state isolation asserted, 20 iterations.

## Landing notes (maintainer)

- Compacted the 312-commit contributor branch to 3 reviewable commits (byte-identical tree — lossless) and rebased onto current `main`.
- **Adversarial squad** (architect/contract, SSOT/duplication, concurrency/security lenses): unanimous **MINOR-only, no MAJORs** — fail-closed correctness, isolation, path-safety, contract completeness, and ADR alignment all independently verified. Three MINORs folded (see the landing comment); the larger cross-seam SSOT-consolidation findings are tracked in #3373.
- Added a positively-selecting CI job for the distribution+e2e acceptance test (delivers #3343) — see the `ci(landing)` fold.

## Follow-ups

- #3343 is delivered by the CI fold here; #3373 tracks the deferred SSOT/duplication consolidation.
- SaaS #836 depends on this core PR and will provide installed-CLI cross-repo proof after delivery.

[...]

Issue #3307, opened 2026-08-10
CLI emits force=False for review-rejection rollbacks that spec-kitty-events 6.1.0 (vendored by this project) declares contract-invalid


robertDouglass:
## Summary

The CLI's `build_transition_plan` (FR-015 "force provenance") emits `WPStatusChanged` events with `force=false` for evidence-backed backward-edge review-rejection rollbacks (`in_review -> planned`, `in_progress -> planned`, `approved -> planned`, conditionally `in_review -> in_progress` / `approved -> in_progress`). The CLI's own vendored dependency, **`spec-kitty-events==6.1.0`** (pinned in this project), explicitly and unconditionally declares exactly these four-pair "review-rejection family" transitions contract-invalid without `force=True` — with a conformance fixture literally named `wp-status-changed-unforced-in-review-to-planned-invalid`, and a module docstring instructing: *"The CLI emit path in spec-kitty MUST NOT produce unforced backward transitions."*

This surfaced while dogfooding `spec-kitty sync now` on a real project checkout (Priivacy-ai/spec-kitty-saas): 10 real `in_review -> planned` events accumulated locally over the last two weeks, all with `force=false`, and all 10 were permanently rejected by the SaaS ingestion endpoint the moment sync was enabled. (A related, separate defect in how the SaaS batches and reports that rejection is filed as Priivacy-ai/spec-kitty-saas#TBD — linking once filed.)

## Root cause: two independently-authored `validate_transition` functions, same name, opposite answers

- `specify_cli/cli/commands/agent/tasks_transition_core.py::build_transition_plan` imports `validate_transition`/`GuardContext` from **`specify_cli.status`** — a local FSM (`specify_cli/status/wp_state.py`, `specify_cli/status/transitions.py`) that resolves these four pairs **force-free when evidence is present**:
  - `in_progress -> planned`: legal with a non-empty `reason` (`wp_state.py:383-386`)
  - `in_review -> planned`: legal with a `ReviewResult` (`wp_state.py:447-457`)
  - `approved -> planned`: legal with a `review_ref` (`wp_state.py:481-484`)
  - `in_review -> in_progress`: legal with a `ReviewResult` (`wp_state.py:434-435` region)
- The **shared, vendored** `spec_kitty_events` package's own `validate_transition` (`spec_kitty_events/status.py:530-591`) answers the opposite: it fires **"regardless of review_ref / reason"** and appends the literal 400 message the SaaS returns.
[...]

PR #3395, opened 2026-08-13
fix(requirements): scope spec.md requirement extraction to declared ids, not prose citations

### LINKED ISSUE #3394: finalize-tasks treats a cited foreign requirement ID as declared: a spec may never reference another mission's FR-NNN in prose
## Summary

A spec that **cites** another mission's requirement ID in prose is treated as **declaring** that requirement, and `spec-kitty agent mission finalize-tasks` then hard-fails with `unmapped_functional_requirements`. There is no CLI bypass, so the only ways forward are editing an already-reviewed spec to remove the citation, or hand-writing state the CLI owns.

Net effect: **a spec may never reference another mission's FR-NNN in its prose** — including in background/rationale sections explicitly marked as out of scope. Nothing at authoring time warns about this; it surfaces two phases later, at finalize.

## Mechanism (verified first-hand on `main` @ ab0a0b9b5)

`parse_requirement_ids_from_spec_md` (`src/specify_cli/requirement_mapping.py:104-117`) scans the **entire raw spec text**:

```python
all_ids = {req_id.upper() for req_id in _REF_FIND_PATTERN.findall(spec_content)}
functional_ids = {req_id for req_id in all_ids if req_id.startswith("FR-")}
```

There is no scoping to the Requirements table/section and no distinction between a requirement *declaration* and a *citation*. `finalize-tasks` then requires every id the scan produced to be claimed by some work package, and fails when the cited-but-foreign id is (correctly) unclaimed.

## Reproduction

In a mission whose spec declares `FR-001`–`FR-003`, add a background sentence citing another mission's shipped requirement — e.g. referencing `CharterPackManager.activate`'s `FR-021` default-pack materialization as evidence for why an adjacent behavior is *not* in scope. The work packages claim `FR-001`–`FR-003`, which is correct.

```
$ spec-kitty agent mission finalize-tasks --mission <slug> --json
... "unmapped_functional_requirements": ["FR-021"]
```

Checked for an escape hatch: `finalize-tasks --help` and `agent tasks map-requirements --help` offer no scoping, ignore, or override flag.

## Why it matters

- Citing prior art by its requirement ID is normal, useful spec-writing — it is how a spec shows it considered an adjacent mechanism and deliberately excluded it. This makes that a build failure.
- It is discovered at **finalize**, i.e. after spec and plan have both been authored and reviewed. The cheapest remedy at that point is editing a reviewed artifact for a purely mechanical reason.
- Same family as the requirement-ID format constraint: **validate (or scope) where the artifact is authored, not three phases downstream.**

## Suggested fix

Scope the parse to where requirements are *declared* — the Requirements section/table — rather than the whole document; or distinguish declarations from references (e.g. only ids appearing in a `- **FR-NNN**:` declaration position count as declared). Either way, add a regression test with a spec that cites a foreign FR in prose and asserts it is not treated as declared.

A cheaper interim: an explicit, documented escape (e.g. `finalize-tasks --ignore-requirement FR-NNN`) plus a warning naming the file and line, so the failure is at least self-explanatory and does not require editing reviewed content.

## Context

Found while running a real mission on this repo (the mission for #3385). Recorded locally as ledger entry SK-14.

MOES-Media:
A spec.md may now cite another mission's requirement ID in prose without breaking its own
tasks phase.

Before this change, `finalize-tasks` refused to run for any spec whose prose mentioned an
`FR-NNN` it did not itself define — for example, citing an already-shipped requirement as
background evidence for why a bug is easy to miss. The citation was counted as a requirement
the spec *declares*, so it was reported as `unmapped_functional_requirements` and the phase
hard-failed. The only ways out were to map a requirement the work package does not implement,
or to edit an already-reviewed spec to break the match. Citing prior art by `file:line` and
requirement ID is normal practice here, so this could hit any mission.

## What changed

`parse_requirement_ids_from_spec_md` no longer scans the whole document for
`FR-`/`NFR-`/`C-NNN` tokens. It now extracts only IDs written in a position that *declares*
them — one of four shapes found by checking all 366 real `spec.md` files in `kitty-specs/`
rather than assuming the template shape:

- a Requirements table row, with the ID in the first cell (bold `**FR-001**` and
  struck-through `~~FR-006~~` for retired IDs both tolerated)
- a heading naming the ID (`### FR-001`, `### FR-001: Title`)
- a bulleted or numbered list item (`- FR-001: ...`, `- **FR-001**: ...`)
- a bold ID leading a bare paragraph (`**FR-019 — Title.** body`)

A mid-sentence mention matches none of these and is excluded from both the `functional` and
`all` keys. The four shapes were validated twice, independently, against the full corpus: no
real declaration is lost, and every excluded ID is a genuine citation.

Because `all` is what checks a work package's `requirement_refs` entry is *known*, a WP citing
a foreign ID now correctly fails "unknown ref" instead of silently passing. Checked against all
2,329 real `WP*.md` files carrying `requirement_refs`: two reference an ID outside the new
declared set, and both are pre-existing imprecision in those work packages, not breakage.

## Known gap, deliberately left open

Narrowing extraction introduces the opposite risk: a spec whose Requirements section matches
none of the four shapes yields zero declared IDs. To keep that from being silent, this adds a
non-blocking `requirement_extraction_warnings` field (plus a console warning) to
`finalize-tasks` and `map-requirements`, and advisory logging on the `spec-kitty next`
readiness path. Exit codes are unchanged — a warning must never become the hard failure this
PR exists to remove.

A hard-failure detector for that case was prototyped and measured at roughly a 6% false-positive
rate against the corpus, so it is not shipped; the reasoning is recorded in the module comment
rather than left in a session log.

One case remains genuinely unhandled: a spec that declares some requirements properly *and*
writes others as bare prose sentences. Those bare FRs are not counted, and `spec-kitty next`
can advance without them being mapped. Blocking that would mean treating any undeclared token
as unmapped — exactly the behaviour this PR removes — so it needs section-scoped detection and
its own corpus validation. Filed separately rather than half-solved here.

An earlier commit on this branch attempted a narrow block for the zero-declared case; review
established it could not change the advance decision at the CLI level, because the guards that
consume `requirement_mapping_failures` check for work-package files first. It is reverted in
`ae7eba9b2` rather than shipped as a claim the code does not deliver. The advisory logging it
added is kept, and is now isolated so an exception from it cannot reach the readiness gate.

## Tests

Red-first throughout: each behaviour has a test verified to fail with its change reverted.
Coverage includes the original defect, each of the four declaration shapes, the citation forms
excluded, NFR-/C- exclusion, the empty-spec case, and negative-space pins asserting that this
issue's own repro shape does **not** block — the guard against a future change quietly
reintroducing it.

Closes #3394

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3395"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789256068&installation_model_id=427282&pr_number=3395&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3395&signature=eec09f033e84bb28dd2310de6a3743cde87ff9deff74dbc4e32a1252ef1ef9c6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
## Landing pass — evidence trail

Rebased onto `main` @ `7923fda40` (was based on `ab0a0b9b5`). The intervening commit
(#3346, 182 files) adds 467 lines to `tests/next/test_runtime_bridge_unit.py`, which this PR
also edits — the rebase merged cleanly and the affected suites were re-run on the rebased tip
to confirm the merge is semantically clean, not just textually: **164 passed**
(`test_runtime_bridge_unit.py`, `test_bridge_cores.py`, `test_requirement_mapping.py`).

### Red check adjudication

| Check | Bin | Action |
|---|---|---|
| `regression tests (blocking)` — 4 × `tests/regression/test_issue_3307_cli_emits_contract_invalid_unforced_rollback.py` | **Pre-existing main breakage, `regression` carve-out** | Left red, deliberately |

Verified against main's own run rather than assumed: main's `regression tests (blocking)` job
at `7923fda40` fails with the **identical four test ids and identical assertion text**
(`build_transition_plan emitted force=False …`). Deselected counts differ as expected — 37,093
on main vs 37,117 here, i.e. this PR adds 24 tests and zero failures.

The module carries `pytestmark = pytest.mark.regression` and pins itself in its docstring to
open P0 #3307. Per the repo's red-first policy (ADR `2026-07-17-1`) that is a deliberate
red-mainline signal, so it is **not** greened here — doing so would erase an honest release
blocker. This PR touches requirement-ID parsing and nothing in the transition/events path.

**Consequence for merge:** this check cannot go green while #3307 is open — it is red on `main`
too. Every other check is the real signal for this PR.

### Commit hygiene

`spec-kitty profile-invocation complete` auto-generated its bookkeeping commit with the subject
`op(implementer-ivan): generate [...]`. `op` is not in this repo's commitlint `type-enum`, so
that commit fails the repo's own gate — retyped to `chore(ops): …`. Verified the whole range
with CI's own invocation (`@commitlint/cli@19.8.1 --config commitlint.config.cjs`): all 5
commits pass.

Worth flagging to maintainers: CI's `lint` job reported **pass** on the run that contained the
offending commit, so the commitlint gate did not catch a violation that the config plainly
rejects.

### Review trail

Four independent review passes ran before this PR opened, each writing findings to
`kitty-ops/`:

1. **Initial review** — filed 3 findings (1 moderate, 2 nits). All fixed in `62695e578`.
2. **Re-review** — verified those resolved by execution; found **F4 (high)**: the narrowing
   silently defeated `spec-kitty next`'s requirement-mapping readiness gate, where the old
   doc-wide scan had correctly blocked.
3. **Fix round for F4** → `6699c6996`.
4. **Focused review of that fix** — established by execution that the new block **could not
   change the advance decision at CLI level** (all three consuming guards check for
   work-package files first). Rather than ship a commit whose message claimed a block it did
   not deliver, it is reverted in `3823f2b00`; the advisory logging it added is kept and is now
   isolated so an exception from it cannot reach the readiness gate.

The same pass also corrected an overclaim of mine: the remaining mixed case (a spec declaring
some requirements properly and writing others as bare prose) is **not** irreducible.
`find_undeclared_requirement_citations` already does section-scoped detection that could
distinguish it from this issue's repro without doc-wide blocking — it needs corpus validation
before it can gate anything. Filed as #3396 rather than half-solved here.

### Local evidence

- Red-first verified per behaviour, by reverting each source file individually and confirming
  the matching test goes red for the right reason — not a single blanket revert.
- Negative-space pins included deliberately: tests asserting this issue's own repro shape does
  **not** block, so a future change cannot quietly reintroduce #3394.
- `uv run ruff check src/` clean; `mypy` clean on all touched source files.
- Corpus validation run twice, independently, against all 366 `kitty-specs/*/spec.md` files, and
  against all 2,329 `WP*.md` files carrying `requirement_refs`.

**State:** watching CI on the rebased tip. Merge-ready once everything except the #3307
regression pin is green. I am not merging — that is the operator's call.

MOES-Media:
## CI adjudication on the rebased tip — merge-ready

Head `ab15225ea`, rebased onto `main` @ `7923fda40`. **76 checks: 36 pass, 38 skipped, 2 fail.**
Both failures trace to a single root cause that is pre-existing on `main` and must stay red.

### `regression tests (blocking)` — pre-existing, deliberate

4 × `tests/regression/test_issue_3307_cli_emits_contract_invalid_unforced_rollback.py`. Main's
own job at `7923fda40` fails with the **identical four test ids and identical assertion text**.
The module is `pytest.mark.regression` and pins itself to open P0 #3307; per ADR `2026-07-17-1`
that is an intentional red-mainline signal, so it is not greened here.

### `quality-gate` — same root cause, nothing new

`quality-gate` is a roll-up: `##[error] regression-tests: job failure`. Its constituent table on
this PR shows **53 jobs OK and exactly one FAIL** — `regression-tests`.

For comparison, the same gate on `main` @ `7923fda40` fails **four** constituents:

| | This PR | `main` @ `7923fda40` |
|---|---|---|
| `regression-tests` | FAIL (#3307, intentional) | FAIL (#3307, intentional) |
| `clean-install-verification` | **OK** | FAIL |
| `fast-tests-core-misc` | **OK** | FAIL |
| `restart-daemon-nfr-timing` | OK | cancelled |

This branch is greener than the base it targets. Nothing in either red is attributable to this
change, and no red was retried, re-run, or silenced to reach that state.

### Merge-order note

Nothing here forces an ordering against other in-flight PRs — the change is confined to
requirement-ID parsing plus advisory surfaces in `finalize-tasks`, `map-requirements` and the
`next` readiness path. It does add a `requirement_extraction_warnings` key to two commands' JSON
output; additive, following the existing `ownership_warnings` / `concern_coverage_warnings`
pattern, and no `additionalProperties: false` schema or orchestrator-api surface governs those
payloads (checked).

**Merge-ready.** `regression tests (blocking)` and `quality-gate` cannot go green while #3307 is
open — they are red on `main` too, so they are not a signal about this PR. I am not merging;
that is the operator's call.

MOES-Media:
Claiming this PR for today's landing pass: rebase onto current `main`, adjudicate red checks, fold fixes as needed. Evidence to follow.

MOES-Media:

**Verdict: approve the change, land it.** The core diagnosis is correct and its central corpus claim replicates independently. Ten remediation folds were applied; two findings were filed rather than folded, with reasons below.

Rebased onto current `main` (twice — `main` moved mid-pass). All adjudication below is on the rebased tip.

### Red-check classification

| Check | Bin | Action |
|---|---|---|
| `regression tests (blocking)` | pre-existing `main` breakage + `@pytest.mark.regression` carve-out | **left red, deliberately** |
| `quality-gate` | consequence of the above (`regression-tests` is `always-run`, and it is the sole blocking verdict) | none |

Evidence: `main` at `21c2d9966` produces `3 failed, 4 passed`; this branch produces `3 failed, 4 passed` — same three test ids, same assertion text, both locally and on CI. All three are `tests/regression/test_issue_3425_*`, which carries `pytestmark = pytest.mark.regression` and a docstring naming `Open P0: #3425`. Greening it would erase an honest release-blocker signal, so it stays red per ADR `2026-07-17-1`.

Worth noting for anyone tracking the queue: the four `#3307` regression tests **now pass** — `7b4f0a63b` landed that product fix on `main` during this pass. The blocking set is down to `#3425` alone.

### Folds

| Commit | Why |
|---|---|
| `docs(landing)`: changelog entry | Operator-visible behaviour change shipped with no consumer entry in `docs/changelog/CHANGELOG.md` |
| `fix(landing)`: single-source the retrospective FR scan | `retrospective/generator.py:567` still ran the pre-#3394 doc-wide scan, so a prose-only citation was still reported as an unmapped FR — the same bug on a second surface |
| `docs(landing)`: document the field in the tasks-finalize prompt | That prompt is the only surface enumerating `finalize-tasks` JSON fields for the consuming agent; the new field was absent from it |
| `refactor(landing)`: drop the unreachable bold-paragraph pattern | `_BOLD_PARAGRAPH_LEAD_ID_PATTERN` was dead — `[-*]` in the bullet matcher consumes the first `*` of `**`, and the bullet pattern is ordered first |
| `refactor(landing)`: hoist the repeated id alternation | `(?:FR\|NFR\|C)-\d+` appeared 6× in one module, against the repo’s own 3+ constant rule |
| `docs(landing)`: corpus provenance + first-id-per-line rule | A corpus count read as a live invariant; and the `break` in `_declared_ids` is load-bearing, so it needed recording before someone "fixes" it |
| `fix(landing)`: carry the advisory on blocked-operator failure paths | The advisory was computed then discarded by `typer.Exit(1)` — invisible on the exact failure it exists to explain |
| `fix(landing)`: make the advisory shim never-raise by construction | The "can never gate" guarantee held in 1 of 3 callers; the other two turned an advisory crash into exit-1 |
| `refactor(landing)`: home the wrapper next to the function it guards | Follow-on: the wrapper had landed in `runtime_bridge.py`, making two CLI modules import from `src/runtime` for a pure parsing concern |
| `test(landing)`: drop the tasks-finalize baselines #3447 deleted | `b94a005c1` narrowed the 12-agent grid and deleted every `tasks-finalize` baseline; the rebase resurrected them as a delete/modify resolution |

### Squad verdicts

Two profile-loaded lenses reviewed the rebased tip independently. Both cleared the core change; both independently found the retrospective scan and the unreachable advisory, which is why those were folded rather than debated.

The corpus claim was re-measured rather than trusted: 373 specs, 48 ids dropped by the narrowing, **0 real declarations lost**, 1 mission newly hard-failing (`excise-doctrine-curation-…`, ref `NFR-002`) — and that one is a **true positive**: the spec declares `NFR-002a`/`NFR-002b` as table rows and only mentions bare `NFR-002` in prose.

### Local evidence

- Red-first (bugfix proof): reverting the five product files to `main` turns **6 behavioural tests red**, including the `#3394` repro shape; green again on restore.
- Red-first (retrospective fold): reverting it fails with `'FR-999' not in {'FR-001', 'FR-999'}`.
- Behaviour-preservation proof for the pattern deletion: `parse_requirement_ids_from_spec_md` run over all 373 real `kitty-specs/*/spec.md` before and after — **0 differences**.
- `931 passed` across the regression parity grid, the PR's five test files, byte contracts, the retrospective suite and layer rules.
- `tests/architectural/`: `1465 passed, 2 skipped, 2 xfailed`, run from a non-dot path (a run from inside `.worktrees/` can false-green tree-walking gates).
- `ruff check src tests` clean, including the blocking `--select TID251`. `mypy --strict` reports 6 errors in `mission_finalize.py` — **identical 6 on `main`’s version**, lines shifted by one; this PR adds none, and the job is advisory.
- commitlint clean across all 15 commits.

### Filed, not folded

- **#3481** — `missions/_substantive.py:59-127` holds a second definition of "spec.md declares FR-NNN" that now disagrees with this PR on 4 of 5 shapes, so a heading-declared spec passes coverage gating while the auto-commit gate calls it non-substantive. Not folded: that gate governs auto-commit on `mission create` / `setup-plan`, a materially wider blast radius than the bug this PR fixes.
- **#3483** — the advisory is inert in `spec-kitty next --json`, the documented agent flow, because JSON mode raises log handlers to `CRITICAL+1` and `logger.warning` is its only channel. Not folded: surfacing it means adding a field to the pinned `Decision` JSON contract, `DecisionEnvelope`, the `next_cmd` print path, and the parity-oracle STABLE-field ledger.

### State

Watching CI. Merge-ready on green **except** `regression tests (blocking)` / `quality-gate`, which stay red until `#3425` lands its product fix — that is the intended signal, not an outstanding defect on this PR.

No landing-order constraint against the other open PRs. Note for whoever merges: **#3428 contains this PR's 5 commits** (it was built on this branch before being retargeted to `main`), so whichever merges second will need those duplicates reconciled.

I do not merge — over to the operator.

MOES-Media:
### Addendum — an 11th fold, and it was this pass's own defect

The summary above listed ten folds. A ledger sweep afterwards surfaced an eleventh, caused by fold #2 in that list. Recording it plainly rather than quietly amending the count.

**`fix(landing): keep the WP-ref scan digit-compatible with the declared-id parser`**

Moving `_find_unmapped_frs` onto the canonical parser widened the **spec** side to `(?:FR|NFR|C)-\d+`, while `_collect_fr_references` kept `\b(FR-\d{3,})\b` for **WP task files**. Both review lenses correctly said to leave the WP-side scan alone — it is a citation scan, a genuinely different contract — but that reasoning was about *scan style*, and neither side checked the *digit count*. The two then disagreed:

```
spec declares FR-01, FR-001, FR-002; WP01 references "FR-01 and FR-001"

before: unmapped == ['FR-01', 'FR-002']     <- FR-01 covered, reported as a gap
after:  unmapped == ['FR-002']               <- only the genuinely uncovered one
```

So the fold traded #3394's false positive on the retrospective surface for a *new* false positive on two-digit ids. Widening the WP side only ever collects more refs, so it can only reduce false unmapped reports; the citation-scan contract is unchanged. Pinned red-first by `test_unmapped_frs_digit_count_matches_the_wp_ref_scan`, which fails with the narrower regex (`assert 'FR-01' in {'FR-001': ...}`).

`560 passed` across the retrospective suites; `ruff` clean; commitlint clean across all 16 commits.

**What it does not fix:** the local defect ledger records this same regex as a silent fake-green for *lettered* ids (`FR-A`…`FR-D`), which produce zero findings from a populated spec. The canonical parser is digits-only, so that case is untouched and stays open. Flagging it because the two defects live on the same line and a future reader could easily assume one closed the other.

Branch is at 16 commits, `3f176eb2d`. Still not merging — over to the operator.

stijn-dejongh:
Superseded by #3428 (merged 2026-08-18). #3428 landed as a self-contained superset: it carries this PR's declared-id extraction scoping (the #3394 residual — `_declared_ids` + the four `_DECLARED_ID_PATTERNS`) **and** the #3396 bare-prose blocking predicate in one change, on a `main` that had neither.

The pre-merge adversarial squad on #3428 verified the supersession is behaviorally clean: `_declared_ids` is a clean equivalent of this PR's narrowing (same declared set for table-row / heading / bullet / bold cases), not a divergent second implementation. Keeping both open would only conflict on `requirement_mapping.py`.

Closing as superseded. The #3394 foreign-id-citation scoping this PR targeted is now delivered on `main` via #3428, so #3394 can likely be closed too — leaving that call to you. Follow-up #3555 tracks the one architectural de-dup the squad deferred.

PR #3520, opened 2026-08-16
fix(doctrine): thread org-tier resolution into the five doctrine consumers that never had it

### LINKED ISSUE #3516: Org-tier doctrine never reaches its consumers: step contracts, expected-artifacts, and the executor's DRG are all project-scoped or built-in-only
An organisation doctrine pack can be registered, validated and activated, and still deliver none of its step contracts, artifact manifests or graph nodes to the runtime. In most affected repositories the parameter needed to fix this **already exists** — the callers simply do not pass it. **One repository is the exception** (`MissionTemplateRepository`, see defect 3b): it has no tiering mechanism at all, so org-tier support there is new surface, not caller threading.

Verified against `3.2.6rc2`.

> **Reconciled 2026-08-17** to the pinned correction comment below (figures, construction-site count, sizing). The defects are all real and the direction is right; the specific numbers in the original filing were not independently reproducible and have been corrected here.

## 1. `StepContractExecutor` is project-dir-only

`specify_cli/mission_step_contracts/executor.py:160-162`:

```python
self._contracts = contract_repository or MissionStepContractRepository(
    project_dir=repo_root / ".kittify" / "doctrine" / "mission_step_contracts"
)
```

`MissionStepContractRepository.__init__` (`doctrine/missions/step_contracts.py:176-181`) already accepts `org_dirs: list[Path] | None = None` alongside `project_dir`. It is never supplied here.

**Consequence:** an org pack shipping `mission_step_contracts/*.step-contract.yaml` is invisible to composition dispatch. The only way to make those contracts resolve today is to duplicate them into the consuming project's `.kittify/doctrine/mission_step_contracts/`. That duplication is a documented workaround in at least one pack, with this issue named as its removal condition.

**Construction sites (corrected).** There are **six** construction sites, not two. **Five are in scope**; one (`doctrine/missions/step_contracts.py:308`) is **correctly built-in-only** — its docstring documents that as deliberate and the field it populates has no production consumers. Two of the five must move **in lockstep** or the result is worse than not fixing them:

- `review/gate_bindings.py` with the executor — or review gates go silently inert;
- `mission_loader/command.py` with `runtime_bridge_composition.py` — or a valid org-tier `contract_ref` is accepted at runtime and then rejected at load-time validation.

## 2. The executor loads an org-blind DRG

`executor.py:179` calls `load_validated_graph(context.repo_root)` with no `org_root`.

`charter/_drg_helpers.py:39-45` documents `_resolve_org_root` as **intentionally inert** — it always returns `None`, and its own docstring states that callers in `specify_cli` are *"expected to resolve the org root themselves … and supply it explicitly"*. `StepContractExecutor` does not.

**Consequence:** `_resolve_step_delegations` resolves every `delegates_to` candidate against a graph that cannot contain org-tier nodes, so correctly-cited candidates resolve to nothing. A probe measured **347 nodes without the org root and 348 with it** (using the fixture shape the existing suite already uses), the difference including the org pack's own directives. The direction is the point; the specific delta is fixture-dependent and should not be quoted as a target.

## 3a. `expected-artifacts.yaml` and `MissionTypeProfileRepository.for_project`

Both read project/built-in scope only. `MissionTypeProfileRepository` already accepts `org_dirs` (unthreaded but live caller — cheap). `doctrine_service_builder.py:142-197` is an existing helper that self-resolves org roots from config and would serve here.

## 3b. `MissionTemplateRepository` has no tiering (new surface)

Separated from 3a because it is a different class of problem. `MissionTemplateRepository` has **no tiering mechanism at all** — there is no `org_dirs` parameter to thread. Org-tier support here is a **new convention**, not caller threading, and is the single biggest sizing driver in this issue. (This is also the template-tier boundary shared with #3523 — the two issues should cross-reference to avoid double-owning it.)

## 4. Delegation results are computed and discarded

`StepContractExecutionResult.resolved_delegations` and `.unresolved_candidates` are populated by the executor and read by nothing — the sole consumer of that result object reads only `invocation_ids`. The shipped built-in `software-dev/specify` contract declares **7** delegation candidates and leaves several permanently unresolved, and no surface reports it. (The original filing said "5 of 9"; the unresolved-candidate problem stands, but the denominator was wrong.)

Either surface these (a warning, or the `--json` payload) or remove the computation. Silently computing and discarding them means a pack author gets no signal that a citation is dead.

## Live in-repo instance

This repository's own registered org pack under `packs/internal` is currently **invisible to the DRG loader** — a live in-repo instance of defect (2). It separately carries a `fragment.yaml` / `*.graph.yaml` naming mismatch, which is pre-existing and **out of scope** here.

## Why these belong together

All are the same defect in different consumers: doctrine resolution that stops at the project or built-in tier when an org tier exists and is registered. Fixing (1) without (2) still leaves delegation dead; fixing either without (4) leaves the failure silent.

## Scope / sizing (corrected)

**L, not M** — roughly **170–210 production LOC plus 250–300 test LOC**, driven by the five in-scope call sites, the two lockstep pairs, and the new-surface tiering work in (3b). Most of it is caller-side threading of an existing `org_root`/`org_dirs` argument; the `MissionTemplateRepository` tiering is genuinely new.

### LINKED ISSUE #3525: Org-tier doctrine follow-ups (#3520): thread repo_root into the dossier index + hoist one resolve_existing_org_roots primitive
## Context
Deferred findings from the adversarial landing squad on #3520 (org-tier doctrine consumers; mission `up-org-doctrine-consumers-01M05YAB`, closed #3516). #3520 threaded org-tier resolution into the runtime consumers correctly and is landing green; these two items are follow-on architecture work both lenses (architect + architecture-scout) independently surfaced. Neither gated the merge.

## 1. Thread `repo_root` into the dossier completeness-index path so the org `expected-artifacts.yaml` override actually enforces (MAJOR/MINOR)
FR-008 fully builds the org-tier read in `ManifestRegistry.load_manifest(mission_type, repo_root=...)` (org resolution, the order-preserving cache-key fix, tests). But **no production caller passes `repo_root`**: `DossierIndexer.__init__` takes only `manifest_registry` (`src/specify_cli/dossier/indexer.py:89`) and its four `load_manifest(mission_type)` call sites (`:123, :176, :355, :407`) plus `sync/namespace.resolve_manifest_version` (`:98`) are all project-blind.

Net effect: an org `expected-artifacts.yaml` override reaches the **governance/gate** path (`charter.mission_type_profiles._resolve_expected_artifacts_slot`, org-aware) but **not** the **dossier completeness-index** path (built-in only). The two subsystems then disagree about "what artifacts does step X require" for an org-overridden mission type — a single-canonical-authority coherence risk. `DossierIndexer.index_feature` *does* receive `feature_dir` (a `kitty-specs/<mission>` dir), so the repo root **is** derivable — correcting the mission's "sole production caller" premise (`research.md:88-94`, `plan.md:290`, `data-model.md:124`), which is factually incomplete.

**Do:** thread `repo_root` through `DossierIndexer` (and `resolve_manifest_version`) so `load_manifest` is org-aware on the completeness-index path, or scope those callers out with the same rigor `namespace.py` got. Correct the "sole production caller" claim in the mission docs.

## 2. Hoist a single `resolve_existing_org_roots` primitive; collapse the 8 duplicated existence-filter sites (MINOR / SSOT)
The pattern `[root for root in resolve_org_roots(repo_root) if root.exists()]` now appears at **8 sites** — 2 named helpers (`charter/doctrine_service_builder.py:152` `_self_resolve_existing_org_roots`; `specify_cli/dossier/manifest.py:56` `_resolve_existing_org_roots`, **new in #3520**) and 6 inline (`charter/profile_resolution.py:118`, `charter/mission_type_profiles.py:1001` **new in #3520**, `invocation/org_profiles.py:66`, `cli/commands/_doctrine_asset.py:90`, `cli/commands/profiles_cmd.py:108`, `cli/commands/charter/context.py:84`). `resolve_org_dirs` re-implements the same filter loop internally (`org_pack_config.py:491-497`) rather than being built on a shared primitive.

**Do:** hoist one public `resolve_existing_org_roots` in `org_pack_config.py` (beside `resolve_org_roots`/`resolve_org_dirs`), define `resolve_org_dirs = [r/subdir for r in resolve_existing_org_roots(...)]`, and route the 8 sites through it. Documenting the divergence in docstrings (as #3520 does) is not tracking it.

## 3. (already tracked in-code) Multi-pack DRG precedence
`resolve_org_dirs` returns ALL existing packs (later-declared-wins overlay) while the executor's DRG `org_root` takes only the FIRST (first-match-wins). Coherent for the shipping single-pack case; a half-wired hazard for 2+ packs. #3520 added an inline cross-reference at `mission_step_contracts/executor.py` (the construction site) and it is scoped out by C-004. Listed here for completeness — close the multi-pack DRG story as a unit before enabling multi-pack org packs.

---
Surfaced by architect-alphonso + paula-patterns lenses during the #3520 landing pass. All deferred (no release blocker); parent under the org-doctrine/#2467 pack line.

MOES-Media:
## What & why

**An organisation doctrine pack you registered, validated, and activated delivered
none of its step contracts, artifact requirements, or graph nodes to `spec-kitty` at
runtime — it was silently inert past activation.** The runtime consumers resolved
doctrine from the built-in pack plus the project tier only: they *had* the org-tier
parameter, but the callers never passed it. This threads it through, so an activated
org pack actually reaches the runtime.

Closes #3516.
Closes #3525.

## How

A shared `resolve_org_dirs` helper (`doctrine/drg/org_pack_config.py`) threads the org
tier into the consumers that lacked it:

| Consumer | Fix |
|---|---|
| `StepContractExecutor` + its DRG load | `org_dirs` threaded into the contract repository; the DRG load supplies `org_root` |
| `review/gate_bindings` | same, **in lockstep** with the executor (else review gates pass while verifying nothing) |
| `MissionTypeProfileRepository` | `org_dirs` threaded at its live caller (governance profile provenance `builtin` → `org`) |
| runtime dispatch + mission-load validation | same, **in lockstep** with each other (else an org-tier `contract_ref` is accepted at runtime but rejected at load-time) |
| `expected-artifacts.yaml` | new org-tier read for the governance artifact slot |

Each lockstep pair changed in a single commit with a falsifiable regression test.
Two silent-empty-success paths now **warn** (a dropped org root; a malformed org
`expected-artifacts.yaml`) rather than serving the built-in and reporting success.

**Projects with no org pack are unaffected — every touched path is byte-identical
without one** (verified at each site).

## A chain of org packs now works, and the dossier index is org-aware (folds in #3525)

An adversarial review of this PR asked the load-bearing question — *does org-tier
resolution work for a **chain** of org packs?* — and the answer was **no**: every seam
walked the full chain except the DRG graph merge, which took only the **first** pack. So
a second `doctrine.org.packs[]` entry's step contracts loaded into the repository but
their `delegates_to`/graph edges silently vanished (the executor even disagreed with
itself — repo half vs DRG half). This PR now folds in the fix:

- **Multi-pack DRG chain.** `load_validated_graph` merges the **full declared chain** of
  org roots in order (later-declared-wins, matching the doctrine overlay), replacing the
  first-match behaviour. The executor's DRG half, `action_doctrine_bundle`, and the review
  gate binding (which previously carried **no** org roots at all — inert even single-pack)
  all thread the chain now; a malformed pack in the chain is dropped on its own, keeping
  the rest. Charter-*build*-time DRG callers stay intentionally org-inert.
- **Dossier completeness index org-aware.** `repo_root` is threaded through `DossierIndexer`
  (and the `reconcile` / `sync/dossier_pipeline` entry points), so an org
  `expected-artifacts.yaml` override is enforced by the missing-artifact detector, not just
  the governance gate.
- **SSOT.** One `resolve_existing_org_roots` primitive now owns the org-root existence
  filter; the duplicated inline copies route through it.

Pinned by a chain regression suite (N-pack node+edge presence, URN-collision later-wins,
per-root degrade, executor repo/DRG self-consistency). Closes #3525.

*Scope note (honest):* the chain is now coherent end-to-end for the runtime
doctrine-dispatch, dossier-index, and repository-overlay seams. One seam of the same
class is left for a follow-up — `charter activate/deactivate --cascade` loads its DRG
org-inert (pre-existing; not touched here), so an org-pack artifact's `requires`/`suggests`
edges don't cascade-activate. Tracked in **#3527**.

## Tests

Two lockstep guards + a malformed-org-pack degrade test, all falsifiable (mutating one
site turns each red). Delta assertions are relative, not brittle magic numbers (e.g. the
DRG grows by exactly 1 node with an org root, asserted as a delta). No-org-pack
byte-identity is proven at every site. Touched suites green on the rebased tip.

## Landing notes (maintainer)

- Rebased onto current `upstream/main` (clean — none of the intervening commits touched
  the 7 threaded seams). **History cleaned** from 18 mission commits to 2 reviewable
  commits **authored by @MOES-Media** (implementation+tests / mission-artifacts,
  tree-verified identical to the rebased tip), plus maintainer `(landing)` folds.
- An independent maintainer adversarial squad (architect + architecture-scout lenses,
  opus) ran on top of the author's own three-lens review, then a second pass specifically
  on the **chain** question. It proved (live 2-pack fixture) that the multi-pack DRG was
  half-wired, and drove the fold above; a verification re-review confirmed the fix closes
  the incoherence end-to-end (every runtime DRG consumer now walks the chain in the
  overlay's precedence) rather than relocating it. The original two deferred items
  (dossier index, SSOT primitive) plus the multi-pack chain are all **folded into this
  PR** (`#3525`), not deferred.
- Commits: 2 authored by @MOES-Media (implementation+tests / mission-artifacts) + 5
  maintainer `(landing)` folds; `ruff` clean and no new `mypy --strict` findings on the
  touched modules; the chain fold is TDD (red-first regression suite).


MOES-Media:
## Ready for review — CI green

**Head:** `8dbcaa715a8e` · **41 checks passing, 0 failing.**

### One red, and it was ours

The first CI run failed `arch-adversarial (arch_shard_1)` on three architectural tests. Classified as **introduced by this PR**, not staleness, and fixed rather than suppressed:

| test | cause | fix |
|---|---|---|
| `test_runtime_has_no_direct_doctrine_imports` | four `specify_cli` modules imported `doctrine` directly | re-exported `resolve_org_dirs` through the existing `charter.drg` facade; all call sites now import from there |
| `test_runtime_has_no_new_lazy_doctrine_imports` | same, via function-body imports | same |
| `test_convert_sites_do_not_exceed_frozen_baseline` | new tests added golden-count assertions past the frozen baseline | converted to a single content pin; **baseline not regrown** |

No allowlist or suppression entries were added — NFR-005 requires this land with zero, and silencing an architectural rule to pass its own gate is what the suite exists to prevent.

Two things surfaced while fixing that the CI log did not name:

- **`test_facade_all_reexports_are_tabled[charter.drg]`** also failed locally — the facade requires every `__all__` re-export to carry an identity contract in `_FACADE_TABLE`. Added; this is the completion of the facade fix, not a workaround.
- **`src/runtime/next/runtime_bridge_composition.py` had the identical violation and would have passed CI.** Both boundary tests hardcode `_RUNTIME_ROOT = src/specify_cli`, so nothing under `src/runtime/` is scanned for direct `doctrine` imports. Fixed here for consistency with the rule rather than with the test; the gate gap itself is worth a separate issue.

### Verification

3873 tests pass across the touched suites locally; `ruff` and `mypy` clean. One timing flake (`test_spec_kitty_home_pin_budget`) was isolated to parallel-worker contention — passes 4/4 run alone, touches nothing in this diff.

Limitations are stated in the PR description rather than left for a reviewer to find — notably that FR-008 does not yet reach the dossier indexer's completeness gate, and that the `ManifestRegistry` cache-key fix closes a design risk in new code rather than a pre-existing production defect.

Not merged — hand-off only.

MOES-Media:
Claiming this PR for landing: verify rebase currency onto upstream main, classify the red `e2e-cross-cutting` / `quality-gate` checks, fold any landing fixes as needed. Evidence to follow.

MOES-Media:
## Landing status — diagnosis complete, no fold pushed yet

**Head:** `6984044860fb` (`pr/up-org-doctrine-consumers-01M05YAB`)

### Branch currency
Verified against `origin/main` tip `2cffc248f758` — merge-base(HEAD, main) == main tip exactly. No rebase needed; the branch is already fully current.

### What was done
No commits pushed. This pass fetched the actual `e2e-cross-cutting` job log, diagnosed the failure to its exact root cause, and confirmed (via `main`'s own CI Quality run at this PR's merge-base, run `31971089398`) that `e2e-cross-cutting` passes on `main` — so this is not a pre-existing red.

### Classification

| Check | Verdict | Bin | Evidence |
|---|---|---|---|
| `e2e-cross-cutting` | FAIL | **Introduced by this PR** | All 20 parametrized failures in `tests/e2e/test_worktree_owned_root_concurrency.py::test_installed_cli_keeps_two_owned_worktrees_isolated` share one root cause: `DRGLoadError: No DRG graph files found in directory: <repo>/packs/internal`, raised inside `StepContractExecutor.execute()` (`src/specify_cli/mission_step_contracts/executor.py`) via its new `load_validated_graph(context.repo_root, org_root=effective_org_root)` call — uncaught, it propagates up through `runtime_bridge_composition.py`'s catch-all and surfaces as `"kind": "blocked"` for every `next` dispatch. `packs/internal` is this repo's own registered org pack (`.kittify/config.yaml`, pre-existing, unchanged by this PR) whose `drg/fragment.yaml` uses the org-fragment schema and does not match the `*.graph.yaml`/`graph.yaml` layout `load_graph_or_dir` requires at the org root — a pre-existing, previously-inert layout mismatch this PR's own risk register already names (C-005: "not fixed here... does not block verification"). Before this PR, no code path fed `packs/internal` into a strict DRG-graph loader, so the mismatch was never exercised. This PR wires exactly that path in, and the call site does not replicate the graceful-degradation precedent already established a few lines away in `charter/action_doctrine_bundle.py:_resolve_action_bundle` (same `load_validated_graph(..., org_root=...)` call, wrapped in `except DRGLoadError` → WARNING + empty-bundle fallback) — the executor's comment says it "mirrors" that function but only copied the resolution half, not the degradation half. Confirmed not pre-existing: `main`'s own CI Quality run at this PR's exact merge-base (`2cffc248f758`, run [31971089398](https://github.com/Priivacy-ai/spec-kitty/actions/runs/31971089398)) shows `e2e-cross-cutting` = **success**. |
| `quality-gate` | FAIL | Roll-up, downstream of `e2e-cross-cutting` | The job's own log states: `Blocking verdicts (1): e2e-cross-cutting` — every other child check passes. (`main`'s roll-up is separately red at the same merge-base, but for unrelated pre-existing reds — `fast-tests-agent` and `slow-tests` — that this PR's shard path-filters skip entirely, so they don't reach this PR's run.) |

### Why no fold yet
The correct fix touches production exception-handling in the DRG composition path, and there are two defensible shapes with different safety implications: (a) catch `DRGLoadError` around the org-layer contribution and degrade gracefully with a loud WARNING, matching `_resolve_action_bundle`'s precedent, or (b) leave the fail-closed block as intentional (consistent with this codebase's stated NFR-006 fail-closed philosophy elsewhere) and instead fix `packs/internal`'s own registration/layout. That's a product decision, not a mechanical fold, so it's being handed back rather than folded blind.

**State: red, not merge-ready.** No further action taken pending that decision.

MOES-Media:
## Green — `e2e-cross-cutting` resolved

**Head:** `70dd446c873e` · **47 passing, 0 failing.** `e2e-cross-cutting` passed at 18m52s.

### What the red actually was

All 20 failures were one test, parametrized, with one cause:

```
DRGLoadError: No DRG graph files found in directory: <repo>/packs/internal
```

**Introduced by this PR, not staleness.** `StepContractExecutor.execute()` — one of the five consumers wired here — now calls `load_validated_graph(repo_root, org_root=...)`, and that resolves to this repository's own registered org pack. `packs/internal/drg/fragment.yaml` uses the org-fragment schema rather than the `*.graph.yaml` layout the strict loader requires. Confirmed against `main`'s CI at this PR's exact merge-base, where the same job passes.

Worth stating plainly: **this PR's own risk register named that mismatch and misjudged it.** C-005 recorded it as pre-existing and "does not block verification" — true, right up until this change added the first consumer that feeds it into a strict loader. "Pre-existing and inert" describes the current consumer set, not a property of the defect.

### The fix

`StepContractExecutor` now degrades a malformed org-pack DRG to "no org contribution" with a WARNING naming the pack and the error, instead of letting it crash composition. This mirrors `charter.action_doctrine_bundle._resolve_action_bundle`'s established handling of the identical call — which the executor's own comment already claimed to mirror, having copied the resolution half and not the degradation half.

Two things about the shape:

- **Fail-closed is preserved where it matters.** The catch re-raises when no `org_root` was supplied, so a malformed built-in or project DRG still crashes as before. Only the optional org tier degrades.
- **Degrading is not going silent.** The WARNING names the org root and the exception, consistent with how this PR treats every other degradation path — `resolve_org_dirs` warns per dropped root, and a malformed org `expected-artifacts.yaml` warns rather than falling through quietly.

Tests cover both directions: a malformed org pack degrades with exactly one warning, and a well-formed one contributes with none — so an unconditional warn would fail the second half.

### Still true from the description

`packs/internal`'s `fragment.yaml`-vs-`*.graph.yaml` mismatch remains a real defect; it is now non-fatal rather than fixed. It is not touched here — that would widen this PR past its scope.

Not merged — hand-off only.

stijn-dejongh:
🛬 **Maintainer landing pass — claiming.** (@MOES-Media — thorough PR, and thanks for the honest limitations section.)

**Early recommendation: LAND.** Triage:
- **Not superseded** — #3516 ("org-tier doctrine never reaches its consumers") is OPEN; `resolve_org_dirs` is absent on `main`.
- **Compatible** — the branch is only 4 behind, and none of the intervening commits touched any of the 7 seams this PR threads (charter profiles, org_pack_config, runtime bridge, dossier manifest, mission loader, step-contract executor, review gate bindings); clean rebase expected.
- **Still needed** — a registered/activated org pack currently delivers none of its step contracts, artifact manifests, or graph nodes to the runtime.
- CI is green (47 pass / 0 fail); you already ran a three-lens review.

Plan: rebase onto current `upstream/main`, re-verify the touched suites green, run a maintainer adversarial squad on the org-tier resolution contract surface (the two lockstep guards + the flagged `DoctrineService._org_dirs` vs `resolve_org_dirs` existence-semantics divergence), **add the missing `[Unreleased]` changelog entry** (this is user-facing), clean the mission-bookkeeping history while preserving your authorship, then hand off for the operator to merge.

stijn-dejongh:
## 🛬 Maintainer landing — remediation summary

**Verdict: LAND. Operator merges.** Rebased, history cleaned, one squad NOTE folded, changelog added, pushed (`70dd446c8` → `1906c5303`).

### Triage
Not superseded (#3516 OPEN; `resolve_org_dirs` absent on `main`), compatible (4 behind; none of the intervening commits touched the 7 threaded seams — clean rebase), still needed.

### Red classification
CI was green (47 pass / 0 fail). The one local test failure — `test_create_mission_propagates_named_exception_type` — is the **worktree trap** (`create_mission` fails-closed inside a `.worktrees/` checkout); it passes on CI's normal checkout. Not a PR defect.

### History cleaning
18 mission commits → **2 reviewable commits authored by @MOES-Media** (implementation+tests / mission-artifacts, preserved via `--author`, tree-verified identical to the rebased tip) + 2 maintainer `(landing)` folds.

### Adversarial squad (profile-loaded, opus) — both LAND, no release blocker
- **architect-alphonso:** ran the two lockstep guards + degrade tests live (6/6 green); confirmed they are non-vacuous (drop one site → the guard reds), the retain-vs-filter divergence is inert at the wired seams (the base repository skips missing dirs anyway), no-org-pack byte-identity holds everywhere, and the malformed-org degrade is correctly narrow.
- **paula-patterns:** confirmed `resolve_org_dirs` genuinely converges the subdir consumers, `org_expected_artifacts.py` is a real new seam (not duplication), and the delta assertions are falsifiable. Also credited the branch's own `fix(landing)` commit for catching a **real cross-cutting crash** — this repo's `packs/internal` malformed DRG would have crashed *every* composition dispatch once org resolution was wired.
- **Fold:** `docs(landing):` — inline cross-reference in the executor between the DRG (`org_root`, first-match-wins) and the repository (`org_dirs`, later-declared-wins) precedence rules, so a future maintainer doesn't unify one and leave the other (multi-pack hazard, C-004).
- **Filed #3525** (deferred architecture, both lenses): (1) thread `repo_root` through `DossierIndexer`/`resolve_manifest_version` so the org `expected-artifacts.yaml` override reaches the completeness-index path — today it's governance-seam-only, a real gate/dossier divergence — and correct the mission's inaccurate "sole production caller" claim; (2) hoist a single `resolve_existing_org_roots` primitive (the existence-filter is now duplicated across 8 sites, 2 added here).

### Deliverable hygiene
- **Added the missing changelog entry** (`docs(landing):`) — the PR shipped none for a user-facing change; a consumer-focused `[Unreleased]` Fixed entry closing #3516, naming the dossier-index limitation.
- PR body rewritten BLUF-first, stating plainly that `load_manifest`'s org branch is test-only today and the divergence is tracked (#3525).

### Verification
Touched suites green on the rebased tip (~339 tests, minus the 1 worktree-env false-red); `ruff` + `mypy --strict` clean on touched modules (the only `--strict` findings are 3 pre-existing ones in `mission_type_profiles.py`, on `main`; mypy is advisory in CI); terminology + docs-freshness (errors=0) clean; all 4 commits commitlint-clean.

**State:** watching CI; merge-ready on green. I don't merge — handing off for you to merge.

stijn-dejongh:
## 🔁 Folded #3525 in + verified the org-pack chain (per operator request)

The operator asked to **double-check whether org-tier resolution works for a chain of org packs** and to **fold #3525 in**. A profile-loaded squad (architect + architecture-scout, opus) answered the chain question with a **live 2-pack fixture: it did NOT** — every seam walked the full chain except the DRG graph merge, which took only pack #1. So a second `doctrine.org.packs[]` entry's step contracts loaded into the repository but their `delegates_to`/graph edges silently vanished (the executor even disagreed with itself).

**Folded in as 3 TDD commits (red-first), delegated to a sonnet implementer against the squad's exact spec, then opus-reviewed:**
- `refactor(landing):` hoist one `resolve_existing_org_roots` primitive (SSOT).
- `fix(doctrine):` **the chain fix** — `load_validated_graph` folds the full org-root list in declaration order (later-declared-wins, matching the overlay); the executor DRG half, `action_doctrine_bundle`, and `gate_bindings` (previously **zero-org**, inert even single-pack) all thread the chain; the malformed-pack degrade is **per-root**; the C-004 first-match comments are retired.
- `fix(dossier):` thread `repo_root` through `DossierIndexer`/`reconcile`/`dossier_pipeline` so an org `expected-artifacts.yaml` override reaches the completeness index too.

**Verification re-review verdict: the incoherence is closed end-to-end — a real fix, not a relocation** (primitive correct; every runtime DRG consumer walks the chain; repo-half and DRG-half agree; per-root degrade + executor self-consistency pinned by tests). Chain regression suite: N-pack node+edge presence, URN-collision later-wins, per-root degrade, executor self-consistency. All 16 `load_validated_graph` callers compile (back-compat preserved); charter-build callers stay intentionally org-inert.

**One residual, filed honestly as #3527** (not folded — pre-existing, distinct subsystem): `charter activate/deactivate --cascade` loads its DRG org-inert (same class as the `gate_bindings` bug fixed here), so an org-pack artifact's `requires`/`suggests` don't cascade-activate. Beyond #3525's scope.

Changelog + PR body updated (the two former 'known limitations' are now shipped features). Re-watching CI.

stijn-dejongh:
✅ **CI fully green on the folded head** (`5a8fc1b74`, [run 32008575783](https://github.com/Priivacy-ai/spec-kitty/actions/runs/32008575783)): **62 pass, 22 skipped (path-filtered), 0 fail** across every workflow. Both arch shards that the chain fold initially tripped (`arch_shard_1`, `arch_shard_2`) are green after the gate-remediation fold.

**Merge-ready — handing off for you to merge.** Summary of what this PR now is:
- @MOES-Media's org-tier consumer threading (2 commits, authorship preserved) + maintainer landing folds.
- **#3525 folded in** (your call): the org-pack **chain** now resolves in the DRG (was first-only — proven broken by a live 2-pack fixture, fixed with later-declared-wins matching the overlay), the dossier completeness index is org-aware, and one `resolve_existing_org_roots` SSOT primitive. Verified end-to-end by an opus re-review as *closed, not relocated*.
- Arch-gate cascade the fold tripped (facade table, lazy reach-through, dead symbol, golden-count) all remediated; a pre-push governance sweep was clean.
- Auto-closes **#3516** and **#3525**. One pre-existing same-class residual (`charter activate/deactivate --cascade` org-inert DRG) tracked honestly in **#3527**.

I don't merge; once you've merged I'll verify it landed clean and clean up the worktree + branch.

Issue #2853, opened 2026-07-22
test-friction: frozen-baseline ratchet/allowlist gates impose a manual toll on every consequential change


stijn-dejongh:
## Problem

The frozen-baseline **ratchet / allowlist** architectural gates (`tests/architectural/test_ratchet_baselines.py` + `tests/architectural/_baselines.yaml`, and siblings like `test_golden_count_ban.py`, the `test_no_dead_symbols` content-hash allowlist, the `test_example_round_trip` skip-marker baseline, the `test_artifact_kinds` exact-set) are **duct-tape that imposes a manual toll on every consequential change**. Each of them hard-fails CI when a count/set/hash it pins moves, and the only remediation is a hand-authored `_baselines.yaml` diff (or allowlist edit) + a `# justification:` comment **in the same PR**.

They are meant as burn-down ratchets (growth fails, shrinkage warns), but in practice a *legitimate additive change* — a new doctrine kind, a new contract skip-marker, a new golden-count assertion, a new `__all__` symbol — trips several of them at once, none of which the per-WP author can see locally (they only fire in the `integration-tests-core-misc` CI job).

## Concrete cost (observed on PR #2852 / the glossary-pack-doctrine-kind mission, #1418)

A single additive mission (one new `ArtifactKind`) required manual baseline/allowlist edits across **four** of these gates:
- `test_ratchet_baselines::skip_marker_blocks` — one legitimate contract skip-marker → manual bump 6→8 (also had to absorb a pre-existing +1 that landed upstream without a bump).
- `test_golden_count_ban::convert_sites` — 8 legitimate cardinality assertions in new tests → each needed a `# golden-count: cardinality-is-contract` annotation.
- `test_no_dead_symbols` — a symbol whose content changed → stale content-hash → manual hash refresh.
- `test_artifact_kinds` exact-set — the enum grew → manual expected-set edit.

None of these caught a real defect; all were mechanical toll-paying on a legitimate change. They also caused a **base-divergence footgun**: because they pin absolute values, a mid-mission rebase silently reverted upstream additions in the co-touched files, which only surfaced when these gates went red post-merge.

## Ask

Evaluate retiring or restructuring the frozen-absolute-baseline ratchet family:
- Where the burn-down is complete (the thing being ratcheted has reached its target), **delete the gate** — it now only imposes toll.
[...]

PR #3524, opened 2026-08-17
fix(resolver): wire the org doctrine tier into template resolution and FSM discovery

### LINKED ISSUE #3523: Org packs cannot ship templates or mission-FSM content: two forked resolvers, and neither has an org tier
An organisation doctrine pack can ship templates and mission-FSM content, and neither reaches the production resolution path. Two separate defects compound.

## 1. The two template resolvers have drifted, and the production one is missing a tier

`doctrine/resolver.py` and `specify_cli/runtime/resolver.py` are forked implementations of the same 5-tier template chain. `doctrine/resolver.py:177-179` probes a **mission-scoped override** (`.kittify/overrides/missions/<mission>/<subdir>/<name>`); `specify_cli/runtime/resolver.py:281-286` does not.

Production `mission create` and plan setup go through the `specify_cli` one. So templates installed at the mission-scoped path resolve for `charter list` and `show-origin`, and **fail** for the commands that matter.

This is not theoretical. A downstream pack hit it by execution while capturing acceptance evidence: `mission create` failed with `TemplateConfigurationError … maps to unresolved filename 'test-strategy-template.md'` until the templates were copied to the *flat*, non-mission-scoped `.kittify/overrides/templates/` — which works, but is global across mission types, so every shipped template filename must be globally unique.

## 2. Neither chain has an org tier at all

There is no tier between LEGACY and GLOBAL_MISSION for an organisation pack's own templates. The same gap exists in FSM discovery: the core loader is fully root-parameterised (`DiscoveryContext` + `_build_tiers`), so an org tier there is roughly a dozen lines — but there are two forked walks, mirroring the resolver split.

Net effect: a pack can declare `doctrine.org.packs[].local_path`, have every other artifact family resolve from it, and still be unable to ship a template.

## Proposed shape

**Converge first, then add the tier.** Add the missing mission-scoped probe to `specify_cli/runtime/resolver.py` so both chains agree (~6 LOC), then add an org tier to both, sourced from the existing `resolve_org_roots()` rather than a second config key.

Two alternatives were considered and rejected:

- **Sanction `mission_packs:` as the bridge.** It works today for a custom id from a pack-native root, but it is unschema'd and effectively undocumented. Honest sanctioning costs a pydantic model, an `init`/migration writer, a `doctor` check, reader convergence and a doc rewrite — more work than the org tier — and leaves operators declaring the same pack twice with weaker path semantics (no `~`, no `${VAR}`, no `subdir`). Better kept as a documented local escape hatch.
- **Project-tier-only, drop the org tier.** Not the safe status quo it looks like: defect 1 means the project tier's mission-scoped home does not work on the production path today either, so the convergence work is needed regardless. Having paid that, the org tier is a small increment.

Found on `3.2.6rc2`.

MOES-Media:
**An organisation doctrine pack can now ship templates and mission-FSM content, and `spec-kitty` actually resolves them at runtime.** Before this change, an org pack could register, validate, and activate cleanly — and then deliver *nothing*: a template or mission it declared was found by no resolution path, so it was silently inert. Org-pack authors saw activation succeed but their templates/missions never took effect, with no error to explain why.

Closes #3523.

## What was broken

1. **Configured org packs were unreachable from template resolution.** `resolve_org_roots()` existed and worked, but the template chain never called it on the paths that matter — so an org pack's templates were undiscoverable.
2. **FSM mission discovery had the same hole**, at three independent wiring sites (one a separately-maintained duplicate of another), so org-pack missions never showed up either.
3. **The two forked resolvers had drifted.** `doctrine/resolver.py` carried a mission-scoped tier-1 override probe (`.kittify/overrides/missions/<mission>/<subdir>/<name>`) that `specify_cli/runtime/resolver.py` did not — the same nominal lookup behaved differently depending on which resolver ran.
4. **`charter list` mislabelled org entries** — it tagged org-sourced entries `GLOBAL_MISSION` and read them from `org_root/doctrine/missions` while the resolvers read the flat `org_root/missions`. Either bug alone made the listing wrong.

## The fix

An `ORG` tier is inserted between the project-legacy and user-global tiers — in the asset chain (now 6 tiers) and in FSM discovery Walk A (now 8 tiers). The two resolvers' tier-1 mission-scoped override probe is reconverged. `charter list` reports the `ORG` origin honestly and reads the correct directory. Walk B's previously-swallowed FSM template-load failures now surface instead of failing silently.

**The tier is purely additive.** With no org pack configured, `resolve_org_roots()` returns `[]` and every new lookup is a no-op — projects without an org pack are byte-for-byte unaffected, and (measured, see landing notes) pay **zero** added latency.

## Tests

323 focused tests across the resolver, bridge, FSM-discovery, `charter list`, and mission-loader surfaces — including a four-site tier-position parity test that asserts `ORG` sits in the same place in every resolver and both discovery walks, and mutation checks that confirm each new guard fails when broken rather than passing vacuously. All green on the rebased tip.

## Landing notes (maintainer)

- **Rebased** onto current `upstream/main` (was 62 commits behind). This slice is complementary to the recently-landed org-pack runtime work: #3520 threaded the org tier through the *runtime doctrine consumers* and #3525 handled *multi-pack chains*; this PR is the still-unlanded *template-resolution + FSM-discovery* slice. Overlap with #3520 in `mission_loader/command.py` reconciled — `resolve_org_dirs` (step contracts) and `resolve_org_roots` (discovery roots) are complementary, no double-wiring.
- **History compressed** from 68 commits to 18: one mission-planning artifact commit, six reviewable code commits (WP01–WP06), eleven single-purpose `(landing)` folds. Tree-parity verified identical to the pre-cleanup tip.
- **The one red — NFR-003 latency gate — was a pre-existing, known-flaky toll (#2749), not this PR; recalibrated this pass under operator adjudication.** The org-tier wiring adds no latency: measured with the interpreter held constant and only `src` swapped, main median **1.795s** vs this PR **1.783s** (statistically identical). The gate's 1.05s ceiling had been false-redding on CI-runner environment variance (CI medians ~1.25–1.35s with spikes to ~1.78s). Per the operator's call — and the escape clause the baseline's own note reserves ("a deliberate, documented recalibration once #2749 is adjudicated") — the `test(landing)` fold raises `ci_target_median_seconds` 1.05s → **1.60s** (~20% headroom over the observed CI median, still catching a genuine step-change regression), with the measured evidence recorded in the baseline note. The deeper step-projection optimization remains #2853's territory. `quality-gate` was red only because it aggregates the clean-install job.


stijn-dejongh:
**Maintainer landing pass — claiming this PR.**

Picking this up for today's landing pass: rebase onto current `upstream/main`, adjudicate the two red checks (`Clean install verification (FR-010)`, `quality-gate`), fold fixes as needed, and hand off green-modulo-honest-reds for the operator to merge.

**Viability triage — LAND (not superseded).** I checked whether the org-tier work here was already landed by the recent org-pack missions:
- #3520 (closed #3516) threaded the org tier through the *runtime doctrine consumers* (step-contract executor, review-gate bindings, mission-type governance, dispatch, mission-load validation) via a shared `resolve_org_dirs` helper; #3525 handled *multi-pack chains*.
- **This PR (#3524, closes #3523) is the complementary, still-unlanded slice:** the org tier in *template resolution* and *FSM mission discovery*. On current `upstream/main`, `specify_cli/runtime/resolver.py` still has no ORG tier in its asset chain, and the tier-1 mission-scoped-override drift between `doctrine/resolver.py` (has the probe) and `specify_cli/runtime/resolver.py` (does not) is still present. So the work is real and needed.

There is expected overlap with #3520 in `src/specify_cli/mission_loader/command.py` (it already imports `resolve_org_dirs`); the rebase will reconcile that. Evidence to follow.

stijn-dejongh:
## Maintainer landing pass — remediation summary

**Verdict: LAND.** Substantive, not superseded, complementary to the recent org-pack runtime work. Rebased, history-cleaned, verified green modulo one pre-existing honest red. Handing off for the operator to merge.

### Triage
Checked whether this was already landed by the recent org-pack missions: **it is not.** #3520 threaded the org tier through the *runtime doctrine consumers*; #3525 handled *multi-pack chains*. This PR (closes #3523) is the still-unlanded *template-resolution + FSM-discovery* slice — on current `main`, `specify_cli/runtime/resolver.py` has no `ORG` tier and the tier-1 mission-scoped-override drift between the two resolvers is still present.

### What I did
- **Rebased** onto current `upstream/main` (was 62 commits behind). Clean auto-merge; the overlap with #3520 in `mission_loader/command.py` is coherent — `resolve_org_dirs` (step contracts) and `resolve_org_roots` (discovery roots) are complementary facades in different functions, no double-wiring.
- **Compressed history 68 → 17 commits:** one mission-planning artifact commit + six reviewable code commits (WP01–WP06) + ten single-purpose `(landing)` folds. Tree-parity verified byte-identical to the pre-cleanup tip.

### Folds added this pass
1. `docs(landing): changelog entry for org-pack template + FSM tier` — the PR shipped no changelog entry for a user-facing change; added a consumer-focused `### ✨ Added` entry to the canonical `docs/changelog/CHANGELOG.md`.
2. `chore(landing): drop no-op PLC0415 noqa in mission_loader command` — `PLC0415` isn't in the ruff select set, so the suppression silenced nothing and diverged from the adjacent #3520 `resolve_org_dirs` lazy import (no noqa). Kept the explanatory comment, dropped the dead directive.

(The eight `(landing)` folds already on the branch are the contributor session's own remediation trail; kept as separate single-purpose commits per the runbook.)

### Local verification (rebased tip)
- **PR's own tests: 323 passed** (resolver / bridge / FSM-discovery / `charter list` / mission-loader), incl. the four-site tier-position parity test.
- **ruff:** clean on all changed src + test files.
- **mypy (CI command, whole-tree `--strict`):** **33 errors on both PR tip and main — identical; zero new.** (Single-file runs show spurious "Returning Any" only because the config uses `follow_imports = "skip"`; the whole-tree run CI uses resolves them away. mypy is advisory on CI regardless.)
- **terminology guard:** pass. **golden-count ban:** pass. **docs-freshness `--ci`:** `errors=0` (only external-URL link-health warnings). **commitlint:** all 17 commits pass. **markdownlint:** changelog clean.

### The one red — pre-existing, not this PR
**`Clean install verification (FR-010)` → NFR-003 latency gate**, and **`quality-gate`** (which is red only because it aggregates the clean-install job).

I measured the PR's actual latency contribution — same interpreter, only `src` swapped, exact gate command from the fixture:

| | median | runs |
|---|---|---|
| **main** | **1.795s** | 1.706–1.833 |
| **this PR** | **1.783s** | 1.750–1.812 |

Statistically identical (PR 12ms *faster* at median) — the org-tier wiring adds **zero** latency, exactly as designed (`resolve_org_roots()` → `[]` with no org pack, `quiet=True`, no-op loops). Main is already far over the 1.05s CI ceiling on this hardware.

This is the known-flaky toll gate **#2749** — the baseline file itself documents it as flaky at its ceiling and explicitly says *"NOT caused by unrelated PRs. Do NOT recalibrate `ci_target_median_seconds` to mask it from within an unrelated change"* (the real fix is #2853's dedicated toll-draining mission). So I left it as an **honest red** per ADR 2026-07-17-1 — no recalibration folded here.

### State
Fresh CI running on the new head (`c6a65f863`); PR is `MERGEABLE`. **Merge-ready on green modulo the NFR-003 toll red**, which is documented above and owned by #2749 / #2853. Operator merges — I don't.

stijn-dejongh:
### Follow-up fold — NFR-003 latency ceiling recalibrated (operator-adjudicated)

Per operator direction, added one more `(landing)` fold: **`test(landing): recalibrate NFR-003 latency ceiling 1.05s→1.60s (#2749)`**.

The 1.05s `ci_target_median_seconds` had been false-redding on CI-runner environment variance (CI medians drifting to ~1.25–1.35s with single-run spikes to ~1.78s), while the code under it is latency-neutral — re-confirmed this landing: `upstream/main` median **1.795s** vs this PR **1.783s** with the interpreter held constant and only `src` swapped. The new ceiling is 1.60s (~20% headroom over the observed CI median, matching the documented `tolerance_pct=20`), which still catches a genuine step-change regression. This is exactly the "deliberate, documented recalibration once #2749 is adjudicated" the baseline note reserved; the deeper step-projection caching remains #2853's territory. Evidence is recorded in the baseline JSON's `_ci_target_note` / `_known_regression_note` and the gate script's header note.

Net effect: the sole red (`Clean install verification (FR-010)` → NFR-003, and the derivative `quality-gate`) should now go green on CI. Watching the fresh run on `1ddee2102`.

PR #3554, opened 2026-08-18
fix(review): print a rejection feedback path move-task will accept

### LINKED ISSUE #3430: bug: the rejection command printed by 'agent action review' names its own output artifact and is refused when run verbatim
## Summary

`spec-kitty agent action review` prints the exact command to run if the review rejects the work package. Running that command verbatim always fails, because the `--review-feedback-file` path it names is the artifact the command is about to create. The guard that refuses it is correct — `review-cycle-N.md` is tool-authored from the reviewer's feedback, not a file the reviewer writes — but the printed instruction cannot be followed. Every rejection costs a cycle to discover this, and the error text does not say where the feedback should live instead.

## Prerequisites

- `spec-kitty-cli 3.2.6rc2 (main build, SHA 4a2367539)`
- A work package in review that the reviewer intends to reject.

## Steps

```bash
spec-kitty agent action review WP01 --mission <slug> --agent <name>
# The command prints, verbatim:
#   ❌ spec-kitty agent tasks move-task WP01 --to planned \
#        --review-feedback-file <feature_dir>/tasks/WP01-<slug>/review-cycle-1.md \
#        --mission <slug>

# Author the reviewer's feedback at exactly that path, then run exactly that command:
spec-kitty agent tasks move-task WP01 --to planned \
  --review-feedback-file <feature_dir>/tasks/WP01-<slug>/review-cycle-1.md \
  --mission <slug>
```

## Expected Behavior

The command the tool prints succeeds when run as printed.

## Actual Behavior

```text
Error: feedback_source is this WP's own review-cycle artifact (review-cycle-1.md);
pass the underlying reviewer feedback instead of a prior review-cycle artifact.
```

## Root Cause

The guard keys on the **filename**, not the directory: any file named `review-cycle-N.md` is treated as a prior review-cycle artifact. The printed command names the tool's own output path as its input, so the two are guaranteed to collide. Writing the identical content to any other path — we used one outside the mission tree — is accepted, and the CLI then generates `review-cycle-1.md` itself with its own frontmatter (`cycle_number`, `reviewed_at`, `reviewer_agent`), which confirms the file is meant to be tool-authored.

## Suggested Fix

Print a path the command accepts. Since the artifact is tool-authored, the suggested command wants to point at a reviewer-feedback location — a scratch path, or a conventional `feedback-cycle-N.md` distinct from the generated artifact. Naming the constraint in the error text would also help: the message says what is wrong with the input but not that the filename is what disqualifies it, so the natural next attempt is to move the same filename to a different directory, which also fails.

## Workaround Applied

Write the reviewer feedback to a path outside the mission tree and pass that. The CLI generates the `review-cycle-N.md` artifact from it.

## Environment

- OS: macOS Darwin 25.5.0
- Python: 3.13
- `spec-kitty-cli 3.2.6rc2 (main build, SHA 4a2367539)`

---

Reviewed before filing.

Fezaru:
Fixes #3430.

## What changes for you

When a reviewer rejects a work package, `spec-kitty agent action review` prints
a ready-to-run rejection command. **That command used to fail every time.** It
told you to write your feedback to `tasks/<wp>/review-cycle-N.md` and then pass
that same file to `move-task --review-feedback-file` — but inside the WP's own
directory that exact filename is the *tool-authored verdict artifact*, which the
feedback-provenance guard refuses as a feedback source. So the command printed
right in front of you could never be run as printed, and you only found that out
after burning a review cycle to hit the error.

Now the prompt advertises `tasks/<wp>/review-feedback-N.md` — a path
`move-task` accepts. Copy-paste the printed command and it works.

## The fix

- The advertised feedback path moves to `review-feedback-N.md`, still committed
  in the WP's own in-repo directory so the "feedback lives with the project"
  promise holds. It no longer matches `_REVIEW_CYCLE_FILE_RE`, so the provenance
  guard admits it.
- The name is owned by a single helper, `review_feedback_source_path()`, in
  `src/specify_cli/review/cycle.py` — **beside the guard that decides what's
  acceptable.** That co-location is what stops the advertised path and the
  accepted path drifting apart again; `agent/workflow.py` now calls the helper
  instead of hard-coding the string.
- The same wrong path in the `spk-run-verdict-capture` skill (which repeats the
  reject command to agents) is corrected too.

## Tests

`tests/review/test_cycle.py::test_review_prompt_feedback_path_is_accepted_as_a_feedback_source`
drives the advertised path through `create_rejected_review_cycle` and asserts it
survives the provenance guard. **Red-first verified during landing:** against
`upstream/main`'s code the old advertised `review-cycle-1.md` raises the exact
#3430 error (`feedback_source is this WP's own review-cycle artifact …`); the
fix's `review-feedback-N.md` passes.

## Landing notes

- Rebased clean onto current `upstream/main`; three-dot diff is the 4 intended
  files only.
- Gates green on the rebased tip: `ruff`, `mypy`, `tests/review/test_cycle.py`
  (27 passed), terminology guard, docs-freshness (`errors=0`), markdownlint on
  the changelog.
- Changelog entry added under `[Unreleased] › Fixed` as a separate
  `docs(landing)` fold.

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -304,6 +304,20 @@ _The 3.2.6rc2 candidate cycle is open (rc1 shipped 2026-08-12). Entries land her
 
 ### 🐛 Fixed
 
+- **Rejecting a work package no longer costs a wasted cycle: the feedback-file
+  path `agent action review` prints in its rejection command is now one
+  `move-task` will actually accept (`#3554`; closes `#3430`).** Before, the
+  review prompt told the reviewer to write feedback to
+  `tasks/<wp>/review-cycle-N.md` and then pass that same file to `move-task
+  --review-feedback-file` — but inside the WP's own directory that exact
+  filename is the tool-authored verdict artifact, which the provenance guard
+  refuses as a feedback source. So the rejection command printed verbatim always
+  failed, and the reviewer only discovered it after burning a cycle. The
+  advertised path is now `tasks/<wp>/review-feedback-N.md` — still committed in
+  the WP's own in-repo directory — and the name is owned in `review/cycle.py`
+  beside the guard that decides what is acceptable, so the printed path and the
+  accepted path cannot drift apart again.
+
 - **The two `kitty-specs/` lane guards no longer disagree about a bulk-edit
   mission's own occurrence map, so it can be kept current from the implementing
   lane without a manual unwind (`#2980`).** The pre-commit ownership guard warned
--- a/src/doctrine/skills/spk-run-verdict-capture/SKILL.md
+++ b/src/doctrine/skills/spk-run-verdict-capture/SKILL.md
@@ -26,7 +26,7 @@ append-only, deterministic, and identical across every agent.
 - **Reject** a WP (send it back with structured feedback):
   ```bash
   spec-kitty agent tasks move-task <WP> --to planned \
-    --review-feedback-file <path/to/review-cycle-N.md> --mission <slug>
+    --review-feedback-file <path/to/review-feedback-N.md> --mission <slug>
   ```
 - **Terminal (done) with an explicit review triple** (reviewer / verdict / reference):
   ```bash
--- a/src/specify_cli/cli/commands/agent/workflow.py
+++ b/src/specify_cli/cli/commands/agent/workflow.py
@@ -30,10 +30,13 @@ Sites in this module that **mention** ``review-cycle-*`` artifacts but do
 * ``_has_prior_rejection``, which performs a read-only ``glob`` check.
 * fix-mode prompt rendering, which reads the latest artifact via
   ``ReviewCycleArtifact.from_file`` / ``.latest``; no write.
-* review-prompt rendering, which computes a *placeholder* path
-  ``review-cycle-{next_cycle}.md`` for inclusion in instructional output to
-  the human reviewer. Nothing is written; the file only materialises when
-  the reviewer subsequently runs ``move-task --to planned``.
+* review-prompt rendering, which reads the counter to compute a *placeholder*
+  path ``review-feedback-{next_cycle}.md``
+  (:func:`specify_cli.review.cycle.review_feedback_source_path`) for
+  inclusion in instructional output to the human reviewer. Nothing is
+  written; the reviewer authors that file and the ``review-cycle-N.md``
+  artifact only materialises when they subsequently run ``move-task --to
+  planned``.
 
 Re-running ``spec-kitty agent action implement WPNN`` is therefore a
 counter-no-op by construction: this module never calls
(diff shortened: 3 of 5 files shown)


stijn-dejongh:
**Maintainer landing pass** — claiming #3554 for today's landing queue.

Plan: isolated worktree, rebase onto current `upstream/main`, adjudicate any red checks into the four bins, red-first verify the bugfix, then deliverable hygiene (PR body BLUF + changelog + terminology/docs-freshness guards). I merge nothing — hand off green-modulo-honest-reds for the operator.

**Early call: LAND.** Triaged against current main:
- **Not superseded** — main still advertises `review-cycle-{next_cycle}.md` (workflow.py) and `review_feedback_source_path` does not exist yet.
- **Compatible** — the seams the fix leans on are intact: `_guard_feedback_source_provenance` refuses exactly `review-cycle-N.md` in the WP dir (cycle.py), which is what made the printed reject command un-runnable.
- **Still needed** — the bug (#3430) reproduces on current main.

Evidence to follow.

stijn-dejongh:
## Maintainer landing summary — verdict: **LAND**

Rebased onto current `upstream/main` and adjudicated. This is a fresh but
substantive contributor bugfix; triaged **not superseded** (main still hard-codes
`review-cycle-{next_cycle}.md`), **compatible** (the provenance-guard seam it
leans on is intact), and **still needed** (#3430 reproduces on current main).

### Red-first verification (the fix witnesses the bug)
On `upstream/main`'s code, feeding the *old advertised path*
`tasks/<wp>/review-cycle-1.md` to `create_rejected_review_cycle` raises the exact
#3430 error:

```
ReviewCycleError: feedback_source is this WP's own review-cycle artifact
(review-cycle-1.md); pass the underlying reviewer feedback instead of a prior
review-cycle artifact.
```

The fix's `review-feedback-N.md` passes. So the guard test genuinely captures
the defect, not a tautology.

### Folds (2 commits on the branch)
1. `fix(review): print a rejection feedback path move-task will accept` — the
   contributor's fix, unchanged.
2. `docs(landing): changelog entry for review-rejection feedback path (#3554)` —
   consumer-focused entry under `[Unreleased] › Fixed`. (The original PR carried
   no changelog entry; a user-facing fix needs one.)

### Local gate evidence (rebased tip)
- `ruff` — clean; `mypy` — clean (2 source files).
- `tests/review/test_cycle.py` — **27 passed**, incl. the new #3430 guard.
- Terminology guard — 10 passed (a doctrine `SKILL.md` was touched).
- docs-freshness `--ci` — `errors=0` (only external-URL link-health warnings).
- markdownlint on the changelog — 0 errors.
- commitlint — both commits pass.

### Notes
- No adversarial squad: this is an advertised-string correctness fix, not a
  versioned-contract or shared-seam change.
- The edited `spk-run-verdict-capture` SKILL.md is the canonical doctrine source
  (`src/doctrine/skills/…`), not a generated mirror; no skill-sync gate applies.

**State:** MERGEABLE; CI approved and running. Hand-off for the operator to
merge on green — I don't merge.

---
@Fezaru — thanks for the clean, well-scoped fix and the red-first test. The only
thing I added was the changelog entry: for user-facing changes we record them in
`docs/changelog/CHANGELOG.md` (the root `CHANGELOG.md` is a symlink to it) under
`[Unreleased]`, matching the bold-lead impact style of the surrounding entries.

Issue #3563, opened 2026-08-18
Deferred (WP08): review-cycle write-side kind-flip — review/cycle.py writes kind=WORK_PACKAGE_TASK; migrate resolve_review_verdict_facts + retire _review_cycle_wp_dir


stijn-dejongh:
Deferred follow-up from mission `review-cycle-verdict-seam-rebuild-01KZ2W7W` (surfaced during WP08; recorded as finding **I4** in its `analysis-report.md`, and in code at `src/specify_cli/review/cycle.py:145-160`). No tracking issue existed — filing it so the deferral is visible.

## What is deferred
The **read side** of review-cycle verdict placement was migrated (WP07 `407ea376c4`), but the **write side** still writes review-cycle artifacts with the wrong artifact kind:

- `src/specify_cli/review/cycle.py:145-160` still writes review-cycle artifacts as `kind=WORK_PACKAGE_TASK`.
- Flipping the write-side default requires migrating `resolve_review_verdict_facts` and re-verifying `tests/.../test_analysis_report_rehome.py` **in the same change**.
- Before the default can flip, three currently-unrouted sites must be routed through the canonical placement resolver:
  - `cli/commands/agent/workflow.py::review`
  - `cli/commands/agent/workflow_cores.py::has_prior_rejection`
  - `cli/commands/agent/workflow_executor.py::implement_try_render_fix_mode_prompt`
- Tail cleanup (gated behind this): retire `review/cycle.py::_review_cycle_wp_dir` and its COORD→PRIMARY exception-absorption fallback once all consumers route through the canonical resolver (FR-007 WP06 note at `review/cycle.py:161-165`).

## Current exposure
Nil today (the read side tolerates the current write kind), but structurally deferred — the write/read kinds disagree until this lands.

## Suggested acceptance
- `review/cycle.py` writes review-cycle artifacts with the canonical kind; `resolve_review_verdict_facts` migrated; `test_analysis_report_rehome.py` re-verified; the three sites routed through the placement resolver; `_review_cycle_wp_dir` retired.

## Related
- Landed read side: WP07 `407ea376c4` (mission `review-cycle-verdict-seam-rebuild-01KZ2W7W`, ADR `2026-08-03-1`).
- Sibling review-verdict CLI-surface gap: #3547. Shape-registry residual: #3543.

Issue #3549, opened 2026-08-18
Epic: Event-log integrity — one canonical event store, honestly reported, delivered without silent loss


stijn-dejongh:
## Scope

Owns the end-to-end correctness of the status/sync **event pipeline**: that an emitted transition lands in the *one* canonical event store, is reported truthfully, and is delivered to the teamspace without silent loss or deadlock.

### Why

Multiple defects show the event pipeline reporting success while canonical/committed state diverges or delivery stalls — the "silent false success on mission state" class:

- **#3307** *(root value bug)* — the CLI emits `force=False` on review-rejection rollbacks that `spec-kitty-events` 6.1.0 declares contract-invalid; the events are created/queued and then permanently rejected at SaaS ingestion.
- **#3323** — `agent status emit` writes only the local journal, **not** the repo event log; three stores disagree for one mission (log 48 / `status.json` 35 / journal 36). The tool prints `OK` + a rising completion % while canonical `origin` still shows the WP `blocked`.
- **#3485** — the TeamSpace mission-state gate reports green while the CLI's own `--teamspace-dry-run` rejects the same repo; the gate consults the wrong subsystem, rooted in the #3307 emitter value.
- **#3491** — hosted sync deadlocks once a 2nd project is consented: selection is project-plural, batching is by count, and the cross-project guard refuses every window — a permanent refusal classified `TRANSIENT` that retries forever.

Related prior art (closed): **#3311**, **#3320** — silent state corruption of an active mission reported as success.

### For whom

Agents handing missions between machines; teamspace operators bringing several repositories into a teamspace.

### Intended effect

`emit` writes the canonical event log; count/materialize report the same number the committed log holds; delivery either succeeds or fails **legibly** (no permanent-refusal-as-transient); the mission-state gate consults the store that actually gates sharing.

### Provenance

Created 2026-08-18 during an issue-triage pass to home the event-integrity cluster (#3307 / #3323 / #3485 / #3491), which previously rolled up under the broader SaaS-sync (#1800) and metadata-authority (#2400) epics with no dedicated home.

stijn-dejongh:
**Re-scoping (not closing).** The Convergence Event (#3824) removed the delivery/teamspace half of this epic's original scope. Splitting accordingly:

**RETAINED** — this epic's surviving scope is one canonical **local** event store, honestly reported:
[...]

Issue #3461, opened 2026-08-15
Tech-debt: audit shape_registry coordination/identity meta keys are a hardcoded copy guarded by a tautological drift test


stijn-dejongh:
**BLUF:** The audit shape registry's coordination/identity meta-key sets are hardcoded literal copies, and their only "drift guard" test compares those literals against a *third* re-typed copy of the same literals — a tautology that verifies nothing about the real writers. If a writer grows a fifth coordination key, it is silently re-reported as `UNKNOWN_SHAPE`: the exact #2696 bug, just for the coordination subset that #2696's fix didn't reach.

**Origin:** Found by the adversarial review squad landing PR #3456 (https://github.com/Priivacy-ai/spec-kitty/pull/3456, "fix: coord/primary partition-authority residuals — out-of-loop read+write routing (#2160, #2720)"). Deliberately NOT folded into that PR — closing it requires introducing canonical coordination-key constants and migrating every writer to them, a write-path refactor with its own blast radius, not a landing-pass fold.

## Impact

`spec-kitty doctor mission-state` / `doctor coordination` (the #2696 lineage this repo is actively hardening under epic #2720) can misreport a legitimate, intentional coordination-meta key as `UNKNOWN_SHAPE` the moment a writer adds one, because nothing actually checks the constant against the writers — only against a second literal set inside the test itself.

## Evidence

`src/specify_cli/audit/shape_registry.py:53-59`:
```python
META_COORDINATION_KEYS: frozenset[str] = frozenset(
    {"coordination_branch", "topology", "flattened", "pr_bound"}
)

_META_IDENTITY_KEYS: frozenset[str] = frozenset({"mission_id", "mission_number"})
```
Both are hand-typed literal sets, not derived from any writer. Contrast with `_META_WRITER_KEYS` in the same module, which #2696 correctly derives from the `MissionMetaRequired`/`MissionMetaOptional` TypedDicts — the coordination/identity subsets never got the same treatment.

The parity test, `tests/audit/test_shape_registry_writer_parity.py::test_coordination_write_path_keys_are_known_audit_keys`:
```python
def test_coordination_write_path_keys_are_known_audit_keys() -> None:
    ...
    assert expected == META_COORDINATION_KEYS
    assert audit_keys >= META_COORDINATION_KEYS
```
[...]

Issue #3449, opened 2026-08-15
--owned-checkout mission create is unreachable by the runtime loop (status/setup-plan/next resolve to primary, not the worktree)


stijn-dejongh:
## Summary

`spec-kitty agent mission create --owned-checkout <linked-worktree>` **writes** the mission into that worktree's `kitty-specs/`, but every **read** command (`agent tasks status`, `agent mission setup-plan`, `next`) resolves `canonical_repo_root` by following the worktree's `.git` back to the **primary checkout** — where the mission does not exist. So a worktree-created mission is **unreachable by the runtime loop** by any handle. This makes the `--owned-checkout` create path (#3346) only half-work: create succeeds, but the mission can't be planned/advanced.

This is the **read-side of the owned-checkout story** — the same class as #3439 / #2160 (out-of-loop readers resolving the wrong dir), specifically on the `--owned-checkout` surface.

## Reproduction (verified 2026-08-15, v3.2.6rc2)

```
# In a linked worktree off main:
$ git worktree add -b kitty/mission-x .worktrees/wt upstream/main
$ cd .worktrees/wt
$ spec-kitty agent mission create "demo" --owned-checkout "$(pwd)" --json
  → result: success
  → feature_dir: .../.worktrees/wt/kitty-specs/demo-01M026N0     # written into the WORKTREE
  → canonical_repo_root: /home/.../spec-kitty                      # but root = PRIMARY checkout

$ spec-kitty agent tasks status --mission demo-01M026N0
  → Error: No mission found for handle "demo-01M026N0". … the mission exists in kitty-specs/.
$ spec-kitty agent mission setup-plan --mission demo-01M026N0
  → Error: 372 missions found, pass --mission <slug> to disambiguate   # handle did not resolve

# All handle forms fail: the slug, the mid8 (01M026N0), and the full mission_id.
# Recreating the SAME mission in the PRIMARY checkout (no --owned-checkout) → resolver works immediately.
```

## Root cause (hypothesis)

`create` honors `--owned-checkout` and writes to the caller's worktree, but the read resolvers (`resolve_canonical_root` / the mission-handle resolver behind `tasks status` / `setup-plan` / `next`) have **no `--owned-checkout` equivalent** and anchor to the primary checkout via the worktree's `.git` pointer. #3346 shipped owned-checkout for create (and claims `next`), but the status/setup-plan/next read path does not resolve a worktree-owned mission.

## Impact

[...]

PR #3567, opened 2026-08-18
fix: worktree/clone-aware root resolution + review-verdict CLI parity

### LINKED ISSUE #1734: in_review→approved guard blocks all non-move-task command surfaces, forcing --force on standard review flows
## Summary

Every outbound `in_review` transition requires a `ReviewResult` struct (guard `review_result_required`, `src/specify_cli/status/transitions.py` lines 68–73). The guard is correct and intentional. However, two production command surfaces that agents commonly use to complete reviews do not provide `ReviewResult` and have no mechanism to do so:

1. **`spec-kitty agent status emit WP## --to approved --actor <name>`** — documented in `src/doctrine/missions/mission-steps/software-dev/review/prompt.md` line 257 and `src/doctrine/missions/built_in_step_contracts/review.step-contract.yaml` line 56. This path calls `emit_status_transition` with `review_result=None`, causing `TransitionError` on any WP in `in_review`.

2. **`spec-kitty orchestrator-api transition --to approved`** — `src/specify_cli/orchestrator_api/commands.py` lines 934–953 builds a `TransitionRequest` with no `review_result` field. Fails with `TRANSITION_REJECTED` JSON.

The canonical `move-task` path (`src/specify_cli/cli/commands/agent/tasks.py` lines 2122–2141) correctly auto-constructs `ReviewResult` from `evidence_dict`. But agents following the `review/prompt.md` doctrine template use the `agent status emit` form—which has no such auto-construction—and hit the guard wall.

In this mission (`merge-done-surface-resolver-01KTDVHZ`), WP02, WP03, and WP04 all required `--force` to reach `approved` because of this gap. Every `--force` approval also increments `force_count` in the status snapshot, which triggers the hollow-review warning gate at merge time, creating false positives that obscure genuine review-quality signals.

## Root cause

The `review_result_required` guard was added to all six outbound `in_review` transitions without updating the `agent status emit` and `orchestrator-api transition` command surfaces to satisfy it. The guard was designed for the `move-task` hop-loop code path, but doctrine review templates still prescribe the older `agent status emit` form.

## Affected paths

| Entry point | File | review_result provided? |
|---|---|---|
| `move-task --to approved` (from in_review) | `cli/commands/agent/tasks.py:2134` | Yes — auto-constructed from evidence_dict |
| `move-task --to planned` (rejection from in_review) | `cli/commands/agent/tasks.py:2121` | Yes — from rejected_review_result |
| `agent status emit --to approved` (from in_review) | doctrine template path | **No — TransitionError** |
| `orchestrator-api transition --to approved` (from in_review) | `orchestrator_api/commands.py:934` | **No — TRANSITION_REJECTED** |
| `orchestrator-api transition --to done/in_progress/planned/blocked/canceled` (from in_review) | same | **No — dormant** |

## Dormant masks

All five non-approved outbound `in_review` transitions (`done`, `in_progress`, `planned`, `blocked`, `canceled`) via `orchestrator-api transition` will hit the same failure. Currently dormant because no orchestrator workflow routes them directly, but any new caller that naively passes `to_lane` without `review_result` will fail.

## Fix plan

1. **Update `agent status emit`** — add optional `--reviewer` / `--approval-ref` CLI options; auto-construct `ReviewResult` when `from_lane=in_review` using the same logic as `move-task` lines 2134–2141.
2. **Update `orchestrator-api transition`** — add optional `--review-result-json` option; add auto-construction fallback for the in_review outbound case.
3. **Update doctrine templates** — replace `agent status emit WPxx --to approved` in `review/prompt.md` and `review.step-contract.yaml` with `agent tasks move-task WPxx --to approved --note 'Review passed: <summary>'`.
4. **Add contract tests** — validate that both old-path (`agent status emit`) and new-path satisfy the guard without `--force`.

## Related issues

- #1622 (`coordination.status_service` dead-symbol debt)
- #1623 (`doctor.py` god-module split)

### LINKED ISSUE #2610: Bug: cwd ancestor .kittify discovery crosses an explicit containment boundary
## Summary

`spec-kitty-cli 3.2.5` walks every parent of the current working directory and accepts an ambient `.kittify` directory as the project root. Isolating `HOME` and all XDG roots does not constrain this read-discovery boundary. A command launched from a nested sandbox/work directory can therefore consume project assets planted above the intended cwd boundary.

This report is deliberately bounded to **read discovery/root selection**. It does not claim a cwd-ancestor write or a passwd-derived HOME bypass. Separate deny-by-default validation falsified those write hypotheses.

## Exact reproduction

CLI:

```text
/private/tmp/sk691v/venv325/bin/spec-kitty
spec-kitty-cli version 3.2.5
SHA-256 0cd1e69c649f0d9cd901c9b1c131c6e60b91105845faee5bf3b331f60d589b15
```

Create two otherwise-isolated trees under `/private/tmp`. In the treatment tree only, plant:

```text
/private/tmp/sk691v/treatment/.kittify/scripts/issue691-treatment-marker.txt
/private/tmp/sk691v/treatment/nested/work
/private/tmp/sk691v/control/nested/work
```

Run each leg with separate fixture-only `HOME`, `XDG_CONFIG_HOME`, `XDG_CACHE_HOME`, `XDG_DATA_HOME`, and `XDG_STATE_HOME`, plus `SPEC_KITTY_NO_UPGRADE_CHECK=1`:

```bash
cd /private/tmp/sk691v/control/nested/work
spec-kitty config --show-origin
# exit 1: Unable to locate repository root (missing .git or .kittify).

cd /private/tmp/sk691v/treatment/nested/work
spec-kitty config --show-origin
# exit 0; reports project script issue691-treatment-marker.txt
```

Explicit-root precedence works and further isolates the cause:

```bash
cd /private/tmp/sk691v/treatment/nested/work
SPECIFY_REPO_ROOT=/private/tmp/sk691v/explicit spec-kitty config --show-origin
# exit 0; reports the explicit-root marker, not the treatment marker
```

Marker hashes were unchanged before/after, and all CLI bootstrap writes stayed inside the per-leg fake HOME/XDG roots.

## Actual mechanism

`specify_cli.core.paths.locate_project_root()` resolves `current = (start or Path.cwd()).resolve()` and iterates `[current, *current.parents]`, accepting any parent `.kittify` directory. There is no operator-supplied containment ceiling distinct from `SPECIFY_REPO_ROOT`.

## Expected behavior

Provide a way for sandbox/dogfood callers to set a trusted discovery ceiling, or require an explicit root in containment mode. Discovery must not silently cross that ceiling even when a valid `.kittify` exists above it. Existing default walk-up behavior can remain for ordinary interactive use.

Suggested acceptance:

- A caller can name a discovery ceiling or explicit containment mode.
- With the treatment fixture above, discovery below the ceiling returns no project root.
- `SPECIFY_REPO_ROOT` remains authoritative and deterministic.
- Worktree/main-repo resolution inside the allowed boundary remains compatible.
- Add a production-level rejecting fixture, not only a monkeypatched unit test.

## Related issue

#1591 observed the same ambient-ancestor mechanism as a test-environment problem and classified production behavior as correct. This report is distinct: the parent walk is itself unsafe when the caller has declared a narrower containment boundary.

Originating validation: spec-kitty/spec-kitty-go#691 under incident spec-kitty/spec-kitty-go#577. No machine-specific real-home paths or incident evidence are required for this reproduction.

### LINKED ISSUE #2613: Bug: `doctor tool-surfaces --fix` is worktree-unaware — run from a lane worktree, `locate_project_root()` resolves to PRIMARY and silently mutates the primary checkout's manifest
## Summary

`spec-kitty doctor tool-surfaces --fix` is **worktree-unaware**. Run from *inside* a lane git-worktree, `locate_project_root()` resolves `project_root` to the **primary checkout** (the main working tree of the repo), not the enclosing worktree. As a result, `--fix` silently writes its repairs into the *primary* checkout instead of the worktree the operator invoked it from.

Witnessed live 2026-07-12/13 on current `upstream/main` (`027c60c90`): running the repair from a lane worktree to regenerate the agent-profiles manifest silently modified the **primary** repo's `.kittify/agent_profiles_manifest.json` (152 insertions / 34 deletions) — a checkout the operator was not working in. Recovery required `git checkout -- .kittify/agent_profiles_manifest.json` in the primary to revert, then regenerating via a scoped direct-API script that passed an explicit lane `project_root`.

This is a **cross-checkout mutation landmine** for any `doctor`/`upgrade`-family command run from a worktree: the same class as the "one mission per checkout / worktree HEAD race" hazard. An agent or operator running a routine repair from a lane worktree can silently corrupt the primary checkout — high-surprise, and invisible until someone notices the primary's working tree is dirty.

## Reproduction

1. From a repo with a lane git-worktree (e.g. `.worktrees/<mission>-lane-<id>`), `cd` into the worktree.
2. Run `spec-kitty doctor tool-surfaces --fix` from inside that worktree.
3. Observe that any repair (e.g. manifest regeneration) lands in the **primary** checkout's working tree, not the worktree's — check `git -C <primary> status`.

Expected: `--fix` operates on the checkout it was invoked from (the enclosing worktree), or it refuses/warns when `project_root` resolves outside the invoking worktree.
Actual: `project_root` silently resolves to the primary checkout and repairs are written there, leaving the primary dirty and the worktree unrepaired.

## Secondary defect (same command, doctor repair path)

`doctor tool-surfaces --fix` will **not self-heal a manifest whose `output_path` entries point at a foreign machine's absolute path**. `probe()` never finds those entries "present" locally, yet `_write_all` only `save()`s an entry when it is `actionable` — so stale foreign-absolute entries survive a plain `--fix` unless the manifest is deleted first. This is the doctor-repair-path gap that complements the serialization fix in #2589 (WP04 / PR #2608 fixes the write-side serialization; the repair path still can't converge a foreign-absolute manifest in place).

## Likely surface

- `locate_project_root()` — worktree-unaware root resolution (resolves to primary instead of the enclosing worktree). See also the split-brain consolidation in #1971 and the historical worktree-direction reports #539 / #237.
- `spec-kitty doctor tool-surfaces --fix` write path (`probe()` / `_write_all` actionable gate) — foreign-absolute-path non-convergence.

## Suggested fix

1. Make `locate_project_root()` (and the doctor/upgrade write path built on it) honor the **enclosing git-worktree** when the command is invoked from inside one — resolve `project_root` to the invoking worktree, not the primary checkout.
2. Failing that, **refuse + warn** when `project_root` resolves outside the invoking checkout, so a repair cannot silently mutate a foreign checkout.
3. For the secondary gap: let `--fix` treat a foreign-absolute `output_path` entry as actionable (re-target to the local checkout) so a plain `--fix` converges without a manual manifest delete.

## Related

- #2527 — same command (`doctor tool-surfaces --fix`), different defect (managed doctrine-skill drift non-convergence + stray `dist/`).
- #2589 / PR #2608 — foreign-absolute manifest at the serialization layer (this issue is the doctor-repair-path complement).
- #2412 — per-machine `skills-manifest.json` tracked / absolute-path hazard (adjacent).
- #1971, #539, #237 — prior `locate_project_root` / worktree root-resolution reports.

## Provenance

Witnessed live during mission `loop-friction-quickwins-2` (logged in its tooling-friction tracer; not previously filed) while regenerating the agent-profiles manifest for WP04 (#2589). Current `upstream/main` at `027c60c90`.

### LINKED ISSUE #3049: cutover-guard false pass: migrate backfill-runtime-state writes outside a linked worktree, and the guard reads the same redirected path
## Summary

`spec-kitty migrate backfill-runtime-state` — the remedy `cutover-guard` prints — silently writes its seed events **outside the current worktree** when run from a linked worktree. `cutover-guard` then reads that same redirected path, so a verification run from a linked worktree reports **green while the branch under review is still un-cut-over**.

The two behaviours are individually defensible. Composed, they produce a false pass on exactly the workflow the guard's own error message tells you to run.

## Reproduction

In a repo with a linked worktree (`git worktree add`) checked out to a branch whose mission is not yet cut over:

```bash
cd <linked-worktree>
spec-kitty cutover-guard --base-ref upstream/main
#   Un-cut-over mission(s) block this diff:
#     <slug>: WP01: tracker_refs mismatch (deterministic seed missing); …
#     remedy: spec-kitty migrate backfill-runtime-state --mission <slug>

spec-kitty migrate backfill-runtime-state --mission <slug>     # reports success

git -C <linked-worktree> status --short          # <-- clean; nothing written here
wc -l <linked-worktree>/kitty-specs/<slug>/status.events.jsonl   # <-- unchanged

git -C <main-checkout> status --short            # <-- kitty-specs/<slug>/status.events.jsonl modified

spec-kitty cutover-guard --base-ref upstream/main   # <-- now PASSES, from the same linked worktree
```

The guard passes because it resolves through the same canonicalization and reads the main checkout's file. The branch you are about to push is still un-cut-over, and CI will still reject it.

## Mechanism

`canonicalize_feature_dir` (`src/specify_cli/workspace/root_resolver.py:83`) deliberately rewrites a worktree-rooted `kitty-specs/<slug>` back to the canonical repo. Its docstring is explicit about why:

> Many emit callers construct `feature_dir = repo_root / KITTY_SPECS_DIR / <slug>` from a *worktree* repo_root. When that happens, status emit, charter writes, and config writes would land inside the worktree's stale copy of the mission artifacts instead of the canonical repo.

There is a carve-out for a **registered coordination worktree** (`root_resolver.py:122-130`) — the #1589/#1821 split-brain guard — but an ordinary linked worktree is not a coord worktree, so it falls through to canonicalization at `:132`. `resolve_canonical_root` (`src/specify_cli/core/paths.py:392`, rule 2) sees `.git`-as-a-file and follows the pointer to the main repo.

So the redirect is working as designed for *emit* callers. The problem is that a **migration writing seed events for the branch under review** is not an emit call — the artifacts it writes belong to the checked-out branch, not to whatever the main checkout happens to have open. And `cutover-guard`, being a read, follows the same redirect, so it cannot detect the mismatch it was built to catch.

CI does not hit this: `actions/checkout` produces a standalone clone where `.git` is a directory and the canonical root *is* the checkout (`.github/workflows/release-readiness.yml:211` runs `spec-kitty cutover-guard --base-ref "origin/${{ github.base_ref }}"`). The divergence is local-only, which is what makes it dangerous — the local check disagrees with CI in the passing direction.

## Impact

- The documented remedy appears to work, verifies green locally, and the push still fails CI.
- Worse, it mutates a checkout the operator did not name. In our case five seed events landed in an unrelated checkout's working tree and had to be reverted by hand.
- Anyone using worktrees for CI-fix work — which the lane/worktree workflow actively encourages — is exposed.

Encountered while fixing the `cutover-guard` failure on #3040. The first migration run wrote into the main checkout, the worktree's own verification reported success, and the PR branch's `status.events.jsonl` was still at its original line count. Caught only because the branch's event log was checked directly rather than trusting the guard. The fix had to be done in a throwaway `git clone --local` to reproduce CI's topology.

## Suggested resolution

Any one of these closes the false-pass; the first two are the substantive ones.

1. **Do not canonicalize for branch-scoped mission writes.** A migration seeding events for a specific `--mission` on the checked-out branch should write to the invoking worktree. If canonicalization is genuinely required for consistency, refuse instead of silently redirecting: fail with "mission artifacts for `<slug>` are canonical at `<path>`; re-run there".
2. **Make `cutover-guard` verify the working tree it was invoked from.** Its entire job is to answer "is *this diff* cut over" — resolving away from the diff's own checkout defeats the check. At minimum it should compare the canonical path against the invoking root and report loudly when they differ.
3. **Report the redirect.** If a command writes outside the invoking worktree, print the absolute destination. Silence is what turned this into a false green.
4. **Make the printed remedy worktree-safe**, or have it name the directory to run in.

## Environment

`spec-kitty` 3.2.6 (also reproduced against `upstream/main` @ `481c29eaf`), Linux, git worktrees created by `spec-kitty implement` and by hand — both affected.

## Related

- #1589 / #1821 — the coord-worktree split-brain the carve-out at `root_resolver.py:122-130` exists for. This report is the *inverse* case: a non-coord linked worktree where canonicalization is the wrong answer.
- #2624 — root & worktree-path detection epic. Related but distinct: that epic covers resolution *mechanism*; this is a write-destination/verification-scope mismatch that produces a false pass.

### LINKED ISSUE #3051: doctor mission-state --audit/--fix is worktree-unaware — same locate_project_root() collapse-to-primary that already false-greens cutover-guard (#3049)
## Summary

`doctor mission-state --audit/--fix --mission <slug>` cannot see an in-flight
mission when invoked from a linked git worktree, because the same
by-design worktree→primary-checkout collapse that already has a tracked
false-pass consequence for `cutover-guard` (#3049) also breaks mission
lookup for `doctor mission-state`. This is the **third** confirmed command
family hitting this exact mechanism (after #3049 and #2613), all rooted in
`locate_project_root()` / `resolve_canonical_root()`.

Both reproductions below were hit independently, in the same PR #3007
landing pass, by two different agents doing unrelated tasks — which is
itself evidence of how easy this is to trip over under the documented
worktree workflow (`spec-kitty implement WP##` → `.worktrees/<slug>-lane-<id>`;
`docs/development/pr-landing.md` instructs maintainers to give every PR its
own worktree).

## By-design vs. defect — read this first

`locate_project_root()` and `resolve_canonical_root()` are **not bugged**.
Both are explicit, documented single-authority resolvers whose entire
purpose is to collapse a worktree-rooted `cwd` back to the main checkout:

- `src/specify_cli/core/paths.py:182` (`locate_project_root`) — docstring:
  "Locate the MAIN spec-kitty project root directory, even from within
  worktrees, ... Returns: Path to MAIN project root (not worktree)".
- `src/specify_cli/core/paths.py:392` (`resolve_canonical_root`) — docstring:
  "the single worktree-pointer parser (IC-04): every consumer that asks
  'given some CWD (which may be a worktree), what is the canonical
  main-repo root?' resolves through here."

That collapse is legitimate for the callers it was built for: status
`emit`, config writes, and other "there is exactly one canonical copy of
this artifact" cases (see #3049's citation of
`canonicalize_feature_dir`'s docstring in `root_resolver.py:83`, and the
project's own coord/primary partition model in CLAUDE.md).

**The defect is that read/diff/lookup commands scoped to "the branch under
review" call the *same* collapsing resolver with no override, so they
silently answer about the wrong tree instead of the one the operator
invoked them from.** `cutover-guard` (#3049) needs "is *this diff* (HEAD
vs. base) cut over" — collapsing to the primary checkout answers a
question about an unrelated branch. `doctor mission-state --mission
<slug>` needs "does `<slug>` exist under *this checkout's* `kitty-specs/`"
— collapsing to primary means an in-flight mission that only exists on the
worktree's branch is invisible.

## Reproduction 1 — `cutover-guard` false green (already tracked as #3049)

Filed in more depth as #3049 ("cutover-guard false pass: migrate
backfill-runtime-state writes outside a linked worktree, and the guard
reads the same redirected path") — cross-referencing rather than
duplicating. I independently re-verified the resolver-level mechanism it
depends on:

```
$ python3 -c "
from specify_cli.core.paths import locate_project_root
from pathlib import Path
print(Path.cwd())
print(locate_project_root())
"
# cwd:    .../spec-kitty/.worktrees/pr-3007-landing   (branch pr-3007-local)
# result: .../spec-kitty                               (branch main)
```

`cutover_guard()` (`src/specify_cli/cli/commands/cutover_guard.py`) calls
`locate_project_root()` with no `start=`, so from inside the worktree it
diffs the **primary checkout** (on an unrelated branch) against
`upstream/main`, not the worktree's own HEAD:

```python
repo_root = locate_project_root()
...
changed_paths = list(changed_paths_from_git(repo_root, base_ref))
verdict = evaluate_touched_missions(repo_root, changed_paths)
```

Invoked as `spec-kitty cutover-guard --base-ref upstream/main` from inside
the worktree, this reported `Missions touched by diff: 0 / Un-cut-over: 0
/ passed`. Calling `changed_paths_from_git` / `evaluate_touched_missions`
in-process with the **worktree's own path** as `repo_root` instead gives
the true answer:

```
wrong (main) root used by command: /home/.../spec-kitty
FALSE result using wrong root -> touched: 0 un-cut-over: 0 passed: True

CORRECT result using worktree root -> touched: 5 un-cut-over: 0 passed: True files: 28
```

The verdict happens to agree here (both `True`), but the reported figures
(0 touched / 0 un-cut-over) were meaningless — a reviewer trusting that
output would have no idea it never looked at the diff under review. A
guard that can silently report the wrong scope is the severe half of this
mechanism (see #3049 for the sharper case where the scopes *disagree* and
it produces an actual false pass on an un-cut-over mission).

## Reproduction 2 — `doctor mission-state` MISSION_NOT_FOUND (new)

Not previously filed. Same mechanism, different command family
(`doctor`, not `cutover-guard`/migration), and a different failure mode
(hard error, not silent wrong-scope pass).

`_run_mission_audit` in
`src/specify_cli/cli/commands/_mission_state_doctor.py:341-359`:

```python
try:
    report = run_audit(options)
except MissionNotFoundError as exc:
    if json_output:
        _emit_json_error("MISSION_NOT_FOUND", handle=mission)
    ...
    raise typer.Exit(1) from exc
```

`options.repo_root` comes from the command shell
(`src/specify_cli/cli/commands/doctor.py:1051`):

```python
try:
    resolved_root = locate_project_root()
except Exception as exc:
    ...
```

— again no `start=`, so it always resolves to the primary checkout.

Reproduced live: a mission
(`runtime-state-birth-cutover-all-paths-01KYH654`) exists under
`.worktrees/pr-3007-landing/kitty-specs/` (checked out on the PR branch)
but not under the primary checkout's `kitty-specs/` (unrelated branch).
Running the real CLI entrypoint with cwd inside the worktree:

```
$ PYTHONPATH=src python3 -m specify_cli.__init__ doctor mission-state \
    --audit --mission runtime-state-birth-cutover-all-paths-01KYH654 --json
{"error": "MISSION_NOT_FOUND", "handle": "runtime-state-birth-cutover-all-paths-01KYH654"}
```

Tried with and without `--fixture-dir kitty-specs` (the escape hatch
`_resolve_audit_root` offers only overrides `fixture_dir`, not
`repo_root`, so it does not help here — the audit still resolves
`repo_root` via `locate_project_root()` and looks for the mission there).

This forced a hand-edit of `status.events.jsonl` in the worktree, where
the canonical `doctor mission-state --fix` should have been usable — a
direct hit against CLAUDE.md's "Use Canonical Sources, Never Improvise"
rule, and exactly the kind of workaround #3049 also had to resort to
(a throwaway `git clone --local` to reproduce CI's topology).

## Why this is worth tracking as its own issue rather than folding into #3049

#3049 is scoped tightly to the `cutover-guard` + `migrate
backfill-runtime-state` write/read pair. This issue's reproduction 2 shows
the identical resolver-choice mistake recurring in a structurally
different command (`doctor mission-state`, a pure read/audit path with no
write-redirect angle at all) — evidence that the fix belongs at the
mechanism level (#2624), not as a patch local to `cutover-guard`.

## Suggested direction

Same shape as #3049's suggested resolution, generalized:

1. Any command whose job is "answer about the checkout/branch I was
   invoked from" (diff-scoped guards, `--mission` lookups against the
   invoking tree's `kitty-specs/`) must resolve its working root from the
   **invoking `cwd`** (or an explicit `start=`/injected root), not through
   `locate_project_root()` / `resolve_canonical_root()`'s worktree→primary
   collapse. Reserve that collapse for genuinely single-canonical-copy
   writes (status emit, config).
2. Where a command must operate on the canonical/primary root by design,
   detect the mismatch and say so out loud (`invoking root differs from
   canonical root: <path>`) rather than resolving silently — silence is
   what turns this into a false green or a confusing not-found.
3. `doctor mission-state`'s `_resolve_audit_root` already has a
   `repo_root` override seam (`_ROOT_UNSET` sentinel) for tests/fixtures;
   extend the command shell to pass the invoking `cwd`-derived root through
   that seam instead of always calling `locate_project_root()` unscoped.

## Severity

`priority:P1`, `Bug`. Not `P0`: reproduction 1's false-green case is
already tracked at P1 under #3049 (established precedent for this exact
mechanism), and reproduction 2 has a workaround, if a doctrine-violating
one (manual `status.events.jsonl` edit). Still P1: it recurs across a
third independent command family, it directly undermines a *guard* meant
to catch exactly this class of gap, and it exposes every agent/maintainer
following the documented worktree workflow.

## Environment

`spec-kitty` (this repo) at `upstream/main` / PR #3007 landing branch,
Linux, git worktree created by hand under `.worktrees/pr-3007-landing`.

## Related

- #3049 — the `cutover-guard` false-pass, filed with substantially more
  mechanism detail on the write side (`migrate backfill-runtime-state`
  redirect via `canonicalize_feature_dir`). This issue's reproduction 1 is
  a lighter, independent re-confirmation of the same read-side mechanism;
  reproduction 2 is the new, previously-untracked consequence.
- #2613 — `doctor tool-surfaces --fix`, same `locate_project_root()`
  worktree→primary collapse, different command family (a *write* landmine
  rather than a lookup failure).
- #2624 — Epic: Root & worktree-path detection. Both this issue and #3049
  are instances of the "worktree-aware command routing" bullet in that
  epic's Intended Effect section; suggest parenting this issue under it
  alongside #2613.
- #1971 (closed) — the 3-way `locate_project_root` split-brain
  consolidation; that consolidation did not change the collapse-to-primary
  *behavior*, only unified where it lives, so it does not cover this.

### LINKED ISSUE #3124: setup-plan resolves branch context from the primary checkout, overrides the mission meta.json target, and reports branch_matches_target: true
## What happens

`spec-kitty agent mission setup-plan` resolves branch context from the repository's **primary checkout** rather than from the mission's own `meta.json`, and then reports the two as agreeing.

Observed 2026-07-31 on Spec Kitty 3.2.6 in `Priivacy-ai/spec-kitty-saas`. Mission `wayfinding-consolidation-01KYTEGN` declares in `meta.json`:

```json
"target_branch": "develop"
```

The primary checkout was on an unrelated mission's branch, `feat/499-billing-wp01-03`. Running `setup-plan --mission wayfinding-consolidation-01KYTEGN --json` from a `develop` worktree returned:

```json
"current_branch": "feat/499-billing-wp01-03",
"target_branch": "feat/499-billing-wp01-03",
"base_branch": "feat/499-billing-wp01-03",
"planning_base_branch": "feat/499-billing-wp01-03",
"merge_target_branch": "feat/499-billing-wp01-03",
"branch_matches_target": true,
"branch_strategy_summary": "Current branch at workflow start: feat/499-billing-wp01-03. Planning/base branch for this feature: feat/499-billing-wp01-03. Completed changes must merge into feat/499-billing-wp01-03."
```

Every branch field reports the *other* mission's branch. `meta.json` says `develop`.

## Why this is worse than staleness

`branch_matches_target: true` is the harmful part. The flag exists to tell the caller whether the checkout agrees with the mission's declared target, and here it reports agreement **because both sides were derived from the same wrong source**. The check cannot fail, so it cannot warn.

The command prompt instructs the agent to relay these values to the operator and to stop if the target is not the intended landing branch. Following that instruction faithfully produces a confident, wrong statement — the operator is told the mission will merge into a branch belonging to a different mission, framed as verified.

Had I not been comparing against `meta.json` by hand, `setup-plan` would have recorded `feat/499-billing-wp01-03` as the wayfinding mission's merge target.

## Expected

1. Branch context for a mission resolves from that mission's `meta.json`, not from whatever branch the primary checkout happens to be on.
2. When the checkout and the mission's declared target disagree, `branch_matches_target` is `false` and the disagreement is stated — naming both branches.
3. `branch_matches_target` should be structurally incapable of reporting `true` when the two values come from one source. A comparison of a value with itself is not a check.

## Workaround

Do not run `setup-plan` while the primary checkout is on another mission's branch. I authored and committed the plan directly from a `develop` worktree instead.

## Related, not duplicate

- #2938 — `finalize-tasks` targets the protected final branch instead of the feature planning branch. Same family (branch resolution ignoring mission intent), different command and different mechanism.
- #3051, #2613 — `locate_project_root()` collapsing to PRIMARY makes worktree-unaware commands mutate or misreport the primary checkout. This issue is the read-side analogue: not a wrong write target, a wrong *reported* target that asserts its own correctness.

Item 23 of 23 runtime findings from `billing-foundation-catalog-capacity-cutover-01KYARSP` and the concurrent wayfinding mission; the other 22 are in `kitty-specs/billing-foundation-catalog-capacity-cutover-01KYARSP/spec-kitty-errors.md` on Priivacy-ai/spec-kitty-saas#653.

### LINKED ISSUE #3540: Bug: mission intake from a linked worktree writes to the primary checkout and can destroy another agent's mission brief
Same mechanism as #237 and #2613, new command family, and a consequence neither of those had: **one agent destroying another agent's mission state.**

Cross-referenced to #2624 for mechanism kinship, filed separately per that epic's stated convention ("tracked as its own bug ... cross-referenced here for mechanism kinship, not moved").

## What happened

Three Claude instances were dispatched this morning to run three concurrent missions in `spec-kitty-saas`, each in its own linked worktree under `.worktrees/`. The setup instruction was mine and it was wrong.

The first instance ran `spec-kitty intake` from inside `spec-kitty-saas/.worktrees/first-run`. It wrote to **`spec-kitty-saas/.kittify/`** — the primary checkout — and overwrote `.kittify/mission-brief.md`, which belonged to a different, unrelated mission (`say-whose-account`, ingested 2026-08-17T10:52). The agent noticed and restored it.

Restore verified independently: the file is tracked, `git status` on the primary checkout is clean, and the content is the original brief with its original intake header and `brief_hash`. No loss in this instance.

**But the only reason nothing was lost is that the agent happened to look.** Two of three instances doing this concurrently would have clobbered each other, and the second clobber has no signal at all — the write succeeds, the file is valid, and only its provenance header betrays that it belongs to someone else's mission.

## Why it happens

`find_repo_root` does this deliberately. From its own docstring in `src/specify_cli/task_utils/support.py`:

> Find the **MAIN** repository root, even when inside a worktree.
>
> This function correctly handles git worktrees by detecting when .git is a file (worktree pointer) vs a directory (main repo), and following the pointer back to the main repository.

That is correct behaviour for some callers and exactly wrong for mission scaffolding. It is the same "worktree-unaware routing" the epic describes, now in the `intake` / mission-state family rather than `doctor tool-surfaces --fix` (#2613) or `create-feature`/`specify` (#237).

## Why this instance is worse than its siblings

#237 and #2613 both mis-target a checkout. This one mis-targets a checkout **that another agent is concurrently authoring**, and `.kittify/mission-brief.md` is a single fixed path — not namespaced per mission. So concurrent missions do not merely race, they collide on one filename by construction.

That matters more now than when the epic was written, because parallel agent missions in worktrees have gone from unusual to routine.

## Second, separate defect found alongside it

`activate-spk.sh` — the per-worktree isolation wrapper — **cannot activate in a `spec-kitty-saas` tree at all.** Its guard requires a venv that ships the spec-kitty CLI, and saas venvs do not have one. It fails silently enough that an operator reads it as a no-op.

The practical consequence is that the documented mitigation for concurrent-mission isolation does not apply to the repository where concurrent missions are actually being run. If that is intended, the wrapper should say so rather than appear to succeed.

## Suggested acceptance

- A mission-scaffolding command run from a linked worktree either writes into that worktree's `.kittify`, or refuses with a message naming the checkout it would have written to. Silently choosing the primary is the failure.
- `.kittify/mission-brief.md` collisions are impossible or detected: either namespace it per mission, or make intake refuse to overwrite a brief whose `brief_hash` does not match the one it is ingesting.
- `activate-spk.sh` either works in a saas tree or fails loudly with the reason.

That second point is worth doing even before the resolver work lands. The header already carries `brief_hash` and an ingest path; refusing a mismatched overwrite is a small guard that turns silent cross-agent loss into an error.

## Workaround in use

Hardlinked clones instead of linked worktrees — `spec-kitty-saas-first-run`, and similar per mission. This is apparently what the pre-existing `spec-kitty-saas-observability` clone is for, which suggests the problem was already known to somebody and never written down.

### LINKED ISSUE #3541: Critical: doctor mission-state --fix destroys recorded review verdicts and reports errors=0
`doctor mission-state --fix` **destroys recorded review verdicts and reports success.** Found in an isolated sandbox by an agent investigating Priivacy-ai/spec-kitty-saas#995, before running it anywhere real.

## Observed

In a throwaway tree outside any git repo, with `SPECIFY_REPO_ROOT` pinned to the sandbox:

```
repair complete (updated=56, unchanged=0, errors=0)
```

That run **destroyed all 102 recorded review verdicts across 29 missions.** `errors=0`. Nothing was quarantined. The manifest it writes does not mention the affected field at all, so the loss is not recoverable from the repair's own record.

## Why it is worse than ordinary data loss

Three properties compound:

1. **It reports success.** `updated=56, unchanged=0, errors=0` is indistinguishable from a clean repair. An operator has no signal to look further.
2. **The manifest omits the lost field**, so the repair's own audit trail cannot reconstruct what it removed.
3. **It quarantines nothing**, despite quarantine being the established pattern for state a repair cannot safely reconcile — see the seven missions quarantined rather than force-fixed in spec-kitty-saas#813, which was the correct judgement.

## Second, independent hazard: the write target re-anchors

`migration/mission_state.py:_anchor_repair_root` → `resolve_canonical_root` deliberately re-anchors the repair to the **primary** checkout, so a repair invoked from a linked worktree does not land in the worktree.

That is sensible on its own. It becomes dangerous with the workaround currently being used for #3540: **in a hardlinked clone the clone is its own primary**, so a repair run there rewrites the clone's `kitty-specs/` and reports "repair complete" while the real corpus is untouched. A silent false green on the one operation the command exists to perform.

So the two mitigations available for #3540 — worktree or clone — are each unsafe for a different reason with this command. It needs to target a real primary checkout deliberately.

## Suggested acceptance

- A repair that cannot preserve a field **refuses or quarantines**, and never reports `errors=0` while dropping data.
- The manifest enumerates every field the repair touched, including fields it removed.
- `--fix` states the absolute path it will write to and requires that to be the intended primary, rather than inferring it and proceeding.
- A dry-run that reports what would be lost, not only what would be updated.

## Related

- Priivacy-ai/spec-kitty-saas#995 — the audit finding that prompted this. **Note that its framing is wrong and is being corrected**; see below.
- #3540 — worktree/clone resolution, whose workarounds interact badly with this.
- Priivacy-ai/spec-kitty-saas#813 — closed; established the quarantine-rather-than-force-fix precedent this violates.

## Context worth carrying: the audit that prompted this was partly measuring itself

Same investigation established that the alarming numbers in spec-kitty-saas#995 are substantially an artifact:

- **45 of 101 findings are the measuring tool.** The `spec-kitty` on PATH is an editable install on branch `main` that predates annotation partitioning; it raises on `kind: "annotation"` rows, and the audit classifier converts any reducer exception into a `SNAPSHOT_DRIFT` **error**. Those missions are intact.
- **The "36 → 102 in a week" rate is a one-commit step change**, not a rate: `6d82c9e5d` (the 3.2.0rc18 → 3.2.6rc2 upgrade, 2026-08-13) backfilled annotation events into 66 missions at once. Measured across the commit boundary: 23 blockers before, 95 after.
- **The 56 genuine drifts are all one field**, `review_result`, which `status/reducer.py:_RUNTIME_SLOTS` omits while `StatusEvent` carries it. All recoverable from the event log; none needs quarantine.

Which makes the `--fix` behaviour worse in context: the field it destroys is `review_result`, and `review_result` is precisely what the genuine drift is about. Running the repair on this corpus would have deleted the evidence needed to fix it correctly.

An audit whose classifier turns "I cannot parse this" into "your data is corrupt" is its own bug, and probably belongs in this family too.

### LINKED ISSUE #3543: review_result is carried on StatusEvent but never projected: _RUNTIME_SLOTS omits it, so replaying a log drops recorded verdicts
`status.json` cannot round-trip a recorded review verdict. `StatusEvent` carries `review_result` on the wire, the WP snapshot has a slot for it, but the reducer never projects it, so replaying an event log drops it.

This is the producer behind Priivacy-ai/spec-kitty-saas#995 (56 of 124 missions drifting) and the reason Priivacy-ai/spec-kitty#3541 is destructive: `--fix` overwrites `status.json` with a replay that structurally cannot contain the field.

## Named producer

- **`src/specify_cli/status/reducer.py:46`** — `_RUNTIME_SLOTS` lists `"review"` but not `"review_result"`. It is the table `_apply_transition` (same file, ~:120) iterates to carry runtime slots across a transition, so the field is neither written on the review transition nor preserved afterwards. `grep -c review_result src/specify_cli/status/reducer.py` → **0**.
- **`src/specify_cli/status/models.py:328`** — `StatusEvent.review_result` exists, is serialized (`:353`), and is parsed back (`:371`). The event log has the data; only the projection is missing.
- **`src/specify_cli/audit/shape_registry.py`** — `review_result` is not registered for `artifact_type='status_event_row'`, so every review transition additionally emits `UNKNOWN_SHAPE: unknown key: 'review_result'`. On the saas corpus `UNKNOWN_SHAPE` is 8,610 INFO findings across all 124 missions; this is part of that noise.

One field was added to the event model and three consumers were never updated: the projection, the shape registry, and therefore the audit.

## Reproduction

Two events, no database, no fixtures. A `planned → in_review` transition followed by an `in_review → approved` transition carrying `review_result`, then:

```python
snapshot = materialize_snapshot(mission_dir)
snapshot.work_packages["WP01"].get("review_result")   # -> None, verdict is in the log
```

`spec-kitty-saas` carries this as a committed guard at `tests/scripts/test_mission_state_snapshot_integrity.py::test_recorded_review_verdict_survives_rematerialization`. It fails today against 3.2.6 and should pass once this lands.

## Proposed fix, validated against the real corpus

Project `review_result` last-wins per work package from the transition log, **omitted when absent**, matching the pervasive omit-when-`None` convention in `to_dict` throughout `status/models.py`. Simulated over all 56 drifting missions in `spec-kitty-saas`:

| outcome | missions |
|---|---|
| byte-exact match to persisted `status.json` | 27 |
| differs only by explicit `"review_result": null` keys | 29 |
| differs for any other reason | **0** |

All 102 affected verdicts are reconstructable from the event log; **none needs quarantine**. Worth noting for the fix shape: the legacy persisted files are internally inconsistent — within a single `status.json`, some work packages carry an explicit `"review_result": null` while others omit the key entirely — so no deterministic reducer rule reproduces them byte-for-byte. The residual 29 need a one-time null-key normalization, which loses nothing. Emitting `null` unconditionally instead would match those 29 and break the 27, so omit-when-absent is the right half of the trade.

## Suggested acceptance

- A `review_result` recorded on a transition survives `materialize_snapshot` → `materialize_to_json` → reload.
- `review_result` is registered in the audit shape registry for `status_event_row`, so review transitions stop emitting `UNKNOWN_SHAPE`.
- A round-trip property test over the bundled mission-state survey fixtures: no field present in a persisted snapshot is absent from a replay of the same log.

That last one generalizes past this field. The class of bug is "snapshot slot with no projection behind it", and `review_result` is unlikely to be the only member — Priivacy-ai/spec-kitty#3531 looks like the same shape against a 5.0.0 event log.

## Related

- Priivacy-ai/spec-kitty-saas#995 — the saas-side investigation and the corpus measurements
- Priivacy-ai/spec-kitty#3541 — `--fix` destroys this field and reports `errors=0`; blocked on this
- Priivacy-ai/spec-kitty#3531 — possibly the same projection-gap class

### LINKED ISSUE #3547: agent status emit cannot exit in_review: no --review-result-json, and its --help documents an example that cannot work
`agent status emit` cannot move a work package out of `in_review`, and its own `--help` documents an example that implies it can. Found while completing a mission whose work packages were reviewed through a repository PR process rather than through `spec-kitty review`.

## The wall

Exiting `in_review` requires a structured `review_result` carrying `reviewer`, `verdict` and `reference` — `status/wp_state.py:575-591`:

```
"Transition from in_review requires review_result (structured review outcome)"
"Transition from in_review requires review_result with reviewer"
"Transition from in_review requires review_result with verdict"
"Transition from in_review requires review_result with reference"
```

`TransitionRequest` carries `review_result` as its own field (`status/models.py:411`), separate from `evidence` (`:409`), and `aggregate.py:652` builds the guard context from `review_result=request.review_result`.

**`agent status emit` never sets it.** `cli/commands/agent/status.py:313-330` constructs `TransitionRequest(... evidence=evidence, review_ref=review_ref ...)` with no `review_result`, and the command exposes no option that could supply one — confirmed against `--help`, which offers `--evidence-json`, `--review-ref`, `--workspace-context` and `--subtasks-complete`, and nothing for a structured review outcome.

So a WP driven through `agent status emit` reaches `in_review` and stops there permanently. `retrospect create` then refuses with `MISSION_NOT_COMPLETED` even though every work package is merged.

## The misleading part

`agent status emit --help` prints this example:

```
spec-kitty agent status emit WP01 --to done --actor claude \
  --evidence-json '{"review": {"reviewer": "alice", "verdict": "approved", "reference": "PR#1"}}'
```

That is exactly the shape the guard wants, `reference: "PR#1"` is literally the PR-review case, and **it cannot work on this path** — `--evidence-json` populates `evidence`, which goes through `_build_done_evidence`, and nothing bridges `evidence` into `review_result`. The same example text appears at `cli/commands/agent/status.py:273` and in the error message at `:301`.

An operator reading that help reasonably concludes the capability exists on this command. It does not.

## It is a surface divergence, not a lane-model limitation

The lane model supports this fine. `orchestrator-api transition` exposes `--review-result-json`, validates it (`orchestrator_api/commands.py:1297-1318`, requiring non-empty reviewer/verdict/reference and a verdict of `approved` or `changes_requested`), and passes it through at `:1431`.

So there are two implementations of the same transition and only one can leave `in_review`:

| | `agent status emit` | `orchestrator-api transition` |
|---|---|---|
| `--review-result-json` | absent | present |
| sets `request.review_result` | never | yes |
| can exit `in_review` | **no** | yes |
| documents a review example in `--help` | **yes, non-functional** | yes |

**Workaround, for anyone hitting this now:**

```
spec-kitty orchestrator-api transition --mission <slug> --wp WP01 --to approved --actor <name> \
  --review-result-json '{"reviewer": "<name>", "verdict": "approved", "reference": "<PR url>"}'
```

## Suggested fix

Add `--review-result-json` to `agent status emit` and pass it into `TransitionRequest`, reusing `_parse_review_result_json` so both surfaces validate identically. Until then the `--help` example should stop showing a `review` payload under `--evidence-json`, because it currently documents a path that silently fails the guard.

Bridging `evidence["review"]` into `review_result` implicitly would also work and would make the existing documented example true — but I would not do that: it makes the same JSON mean two different things depending on which key it lands under, and the explicit option is clearer.

## Why this matters beyond one mission

The `review` command is not the only way work gets reviewed. On any repository where missions merge through pull requests, the PR *is* the review, with a reviewer, a verdict and a durable reference already in hand. That case is well served by the state machine and blocked only by the CLI surface most agents actually use.

Note the operator who hit this **declined to use `--force`** — writing a state the machine specifically refused is the wrong tool — and hand-authored the retrospective instead, leaving two work packages parked at `in_review` with the reason recorded. That is the right call, and it is worth the fix so the next person is not choosing between a forced state and an inaccurate record.

## Related

Priivacy-ai/spec-kitty#3543 is the *other* `review_result` defect, found independently: the reducer's `_RUNTIME_SLOTS` omits the field, so a recorded verdict is dropped on projection. Same field, two failure modes — one cannot be written through the common CLI surface, the other is not preserved when replayed. Worth looking at together, since a fix to either alone still leaves review verdicts unreliable end to end.

stijn-dejongh:
## Impact

A Spec Kitty command run from a **linked lane worktree or a nested clone** used to silently re-anchor its write, guard-check, or review verdict to the **primary checkout** — the operator saw cross-checkout writes land in the wrong place and guards pass falsely green. This makes root/workspace resolution **honest**: an owned invocation is distinguished from a foreign lane worktree, and in-scope commands **fail closed** (naming the target checkout) instead of redirecting. It also **unifies the review-verdict CLI path** so a work package can be walked to `done` through `agent status emit` alone with a structured verdict.

Umbrella design **#3129**; parent Epics **#2624** (root & worktree-path detection), **#3549** (event-log integrity), **#3044** (review-artifact integrity).

Closes #1734
Closes #2610
Closes #2613
Closes #3049
Closes #3051
Closes #3124
Closes #3540
Closes #3541
Closes #3543
Closes #3547

_(Epics #2624/#3044/#3549/#3129 stay open — this is a slice. #3449 is dropped as already-correct — tracker wontfix; #3563 tracks the deferred global write-side flip; #3461's coordination-key subject was fixed by #2696, not here.)_

## What changed

**Resolver / write-path (`fix(workspace)`):** one checkout-identity guard (`core/checkout_identity.py`) distinguishes an owned invocation from a foreign lane worktree — parsing `.git` directly and carrying read/write **intent** so deliberate primary-read anchors are preserved. In-scope commands adopt the **#3128 fail-closed refusal** (refuse naming the target checkout, never redirect):

| Command | Issue |
|---|---|
| `intake` shared-slot identity check (+ `--force`) | #3540 |
| `doctor tool-surfaces --fix` fail-closed from a lane | #2613 |
| `doctor mission-state` (preserve #2320 status-home + identity awareness + manifest honesty) | #3051 / #3541 |
| `migrate backfill-runtime-state` cutover-guard honesty | #3049 |
| `setup-plan` `branch_matches_target` from the invoking checkout | #3124 |
| `find_repo_root` nested-clone `.git`-dir boundary | #2610 |

**Verdict / audit (`fix(status)`):** `_parse_review_result_json` + the `for_review` commit-gate are hoisted to a shared, topology-aware surface (`status/review_result_parse.py`, `lanes/for_review_gate.py` with a surface-neutral `GateDecision`); `agent status emit` gains `--review-result-json` (walk a WP to `done` via `emit` alone) and its misleading `--help` is corrected; `review_result` is registered in the `status_event_row` audit shape with a value-equality snapshot round-trip. (#3547 / #1734 / #3543 / #3461)

## Grounding corrections (verified against `upstream/main`)

A post-plan adversarial squad (including an empirical resolver run) established that the brief's headline framing was partly a **phantom**, and the scope was reframed accordingly:

- **Standalone-clone re-anchor does not exist** — clones already resolve to self; the split is undecidable from local state. The real, decidable defect is the **linked-worktree / nested-clone invoking location**.
- The **worktree→primary re-anchor is deliberate** for the status-home (#2320) and primary-read anchors (#3328) — preserved behind a must-not-flip inventory with green characterization tests.
- **Already fixed, not re-opened:** the `review_result` reducer projection (`407ea376c4`) and repair-row preservation (`bec7c25273`) — each pinned by a green sentinel.
- **Dropped as already-correct** (recommend tracker wontfix): `#3449` (`--owned-checkout`), the `#3461` coordination-key writer half (fixed by `#2696`). `#3563` ships the safe narrow REVIEW_CYCLE kind opt-in (regression-locked); the global write-side default flip stays deferred (breaks `test_analysis_report_rehome`).

## Testing

- Each release-blocking slice ships an issue-pinned `@pytest.mark.regression` reproduction, **red on `upstream/main`**, green after.
- 68 mission tests green; broader surrounding suites green (507 passed in the touched areas).
- `ruff` clean; **zero new `mypy` errors** (5 remaining `no-any-return` are pre-existing on `upstream/main`, verified — outside this diff's hunks).

## Remaining reviewer / Sonar notes

- WP06's coord-husk fallback path (setup-plan with an unreadable `meta.json`) is implemented but not directly test-covered — a Sonar new-code-coverage follow-up, not a correctness gap.
- The pre-review regression gate reports `unverified` in this fork's layout (no DRG graph under `packs/internal`) — an environment/config quirk, not a code failure.

## Landing notes (maintainer pass)

- **Rebased twice onto current `upstream/main`** — most recently over 3565's mission-state repair, the `#2320` primary-anchor fix, a sync Wave-4 degod, and a clock-import ban. Both rebases were textually clean (one changelog resolution, keeping both `[Unreleased]` → Fixed entries); the second was verified **semantically** clean by the squad below.
- **Autoclose:** `Closes` for the **10** issues this PR resolves (GitHub-confirmed): #1734, #2610, #2613, #3049, #3051, #3124, #3540, #3541, #3543, #3547. Epics/umbrella, already-closed/preserved anchors, the wontfix (#3449), the deferred flip (#3563), and #3461 (fixed by #2696) deliberately excluded.
- **⚠️ Operator-facing behavior change (WP04, intended):** after this PR, `doctor mission-state --fix` invoked **from a linked lane worktree** now **fails closed** (refuses and names the primary) instead of silently converging-to-primary as `#2320` shipped. This is deliberate — it closes the silent cross-checkout write `#3051`/`#3541` describe; run `--fix` from the primary/a clone. The `#2320` converge-on-write mechanism (repair always targets the primary) is preserved for owner invocations.
- **Post-rebase adversarial squad (3 profile-loaded lenses):** integration-vs-merged-main = **SOUND** (the two anchor resolvers provably converge; no duplicate `#2320` anchor — `mission_state.py` is additive); architecture-scout = **sound for ratified scope, ship**; design-fit = flagged the WP04 behavior change (above) + one coherence gap, folded below. 21 identity/anchor + 35 refusal/clock + 26 verdict tests green post-rebase.
- **Maintainer folds** (on top of the contributor commits):
  - `test(landing)` — untrusted-path inventory (arch_shard_2)
  - `test(landing)` — width-invariant `emit --help` assertion (fast-tests-cli, Rich-15 wrap)
  - `fix(landing)` — FR-011 `for_review` commit-gate on `agent status emit`
  - `docs(landing)` — changelog: gate enforced on both surfaces
  - `fix(landing)` — **guard `repair_repo` against foreign-lane invocation** (the coherence fold): WP04's ownership guard was wired only into the CLI shell, leaving `repair_repo` and its TeamSpace-gate caller ungated. `repair_repo` is now self-protecting (red-first pinned), so the foreign-lane refusal can't be bypassed and `#2320`'s test can't false-green.
- **`locate_project_root` broadened (WP07, NOTE):** it now stops at any nested-clone `.git` **directory** (`.kittify` no longer required), aligning with `resolve_canonical_root`. Widely-called resolver — flagged for awareness; internally consistent with the ownership model.
- **Follow-ups filed (non-blocking):** **#3583** (weak must-not-flip anchor test + exotic-worktree `_read_worktree_gitdir` edge cases) and **#3586** (twin `checkout_identity` seams both claiming "the ONE authority"; the unforced single-channel invariant; parser/anchor drift notes). None blocks this PR.
- **Advisory gates:** `diff-coverage` / `quality-gate` red (merge state `UNSTABLE`, not required) — cross-shard coverage attribution + a god-module; surfaced not padded. Sonar hotspot/coverage stays a UI follow-up.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3567"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789685444&installation_model_id=427282&pr_number=3567&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3567&signature=b084216ba6177e9976983662623f2dfabf7f19b96bfab4611680f111c0dc9731"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


stijn-dejongh:

Picking this up: rebase onto current `upstream/main`, resolve the conflicts (PR is `CONFLICTING`), adjudicate CI, add the autoclose statements, and hand off green for merge.

**Triage — LAND.** Fresh (1-day) but substantive (6.2k lines, `src/` behavior change across resolver/verdict surfaces). Not superseded: every referenced defect issue below is still OPEN, and each release-blocking slice ships an issue-pinned `@pytest.mark.regression` reproduction that is red on `upstream/main` — I'll spot-verify a couple on the rebased tip.

**Autoclose plan** (the ask): the body references 21 issues but carries no autoclose keywords. Grounding each reference against its issue title + state, I'll add `Closes #` for the **10 issues this PR actually resolves**:

`#1734` `#2610` `#2613` `#3049` `#3051` `#3124` `#3540` `#3541` `#3543` `#3547`

**Deliberately NOT autoclosed**, with reason:
- Epics / umbrella (parents, this is a slice): `#2624`, `#3044`, `#3549`, `#3129`
- Already closed / preserved-behavior anchors: `#2320`, `#2696`, `#3128`, `#3328`
- `#3449` — PR drops it as already-correct, recommends tracker *wontfix* (autoclose would mislabel it "completed")
- `#3563` — only the narrow REVIEW_CYCLE opt-in ships here; the global write-side flip is deferred (this issue tracks that deferral)
- `#3461` — its coordination/identity-key subject was fixed by `#2696`, not this PR

Shout if you want `#3461` or any epic force-closed too. Evidence to follow.

stijn-dejongh:
## Remediation summary — landing pass complete, ready for operator merge

**Rebased** `30cffb08` → `upstream/main` `6ec05086f5` (was 19 behind); one changelog conflict resolved keeping both `[Unreleased]` → Fixed entries. Head is now `0a98c549db`.

### Red-check adjudication (four-bin)

| Check | Bin | Action |
|---|---|---|
| `arch_shard_2` (untrusted-path containment) | **Contract crossed** | Folded: inventoried the two new `_compare_checkout_mission_state` path-sinks → green |
| `fast-tests-cli` (`emit --help` substring) | **PR defect (test)** | Folded: Rich-15 soft-wraps `--review-result-json`; made the assertion render-independent → green |
| **FR-011 gap** (adversarial review, below) | **PR defect (behavior)** | Folded: wired the `for_review` commit-gate into `agent status emit` → green |
| `arch_shard_3` (`test_gate_runs_under_fast_tier_budget`) | **Perf-budget flake** | Wall-clock `<30s` scan tripped at **30.65s (2% over)** on a busy runner; **599/0 locally**, green on main + prior run, diff adds no inline meta-read. Re-ran the single job → **green** (confirms transient flake). |
| `diff-coverage` / `quality-gate` | **Advisory** | Merge state `UNSTABLE` (not required). ~70% diff coverage is mostly cross-shard attribution artifacts + a 994-line god-module; per doctrine, surfaced not padded. Sonar hotspot/coverage stays a UI follow-up. |

### Adversarial review (reviewer-renata, profile-loaded)

The checkout-identity guard and verdict/audit parity verified **sound**. One **MAJOR**, folded before hand-off:

- **FR-011 not delivered on `agent status emit`.** FR-010 newly made `emit` a walk-to-`done` door, but the `for_review` commit-gate was wired only into `orchestrator-api transition` — so a WP with **no commit on its lane** could reach `done` via `emit` alone (exit 0) while the orchestrator surface rejected it. Fold `fix(landing): enforce for_review commit-gate on agent status emit (FR-011)` wires the shared, envelope-free leaf into `emit` at the same placement the orchestrator uses (red-first proof; `--force` bypass + commit-satisfied allow pinned). Both surfaces now enforce identical, topology-aware, commit-state semantics (SC-004).
- MINOR (a weak must-not-flip anchor test) + NOTES (exotic `--separate-git-dir` / `--relative-paths` / bare-repo worktree edge cases, all safe-direction) → filed as **#3583**.

### Maintainer folds (on top of the 4 contributor commits)
- `test(landing)` — untrusted-path inventory (arch_shard_2)
- `test(landing)` — width-invariant `emit --help` assertion (fast-tests-cli)
- `fix(landing)` — FR-011 emit commit-gate
- `docs(landing)` — changelog: gate now enforced on both surfaces

### Autoclose
`Closes` added for the **10** issues this PR resolves (GitHub confirms all 10 in the sidebar): #1734, #2610, #2613, #3049, #3051, #3124, #3540, #3541, #3543, #3547. Epics/umbrella, already-closed/preserved anchors, the wontfix (#3449), the deferred flip (#3563), and #3461 (fixed by #2696) deliberately excluded.

### Verification
Local on the folded tip: full `arch_shard_3` selector **599/0**; new behaviour/sentinel/gate tests pass; agent-command suite 1430/0 (no emit hot-path regression); `ruff`/`mypy` clean on touched files; terminology guard 10/10; docs-freshness `errors=0`. Required CI checks green except the one perf-budget flake above.

### Hand-off
**All required checks are green** (the `arch_shard_3` perf flake cleared on re-run). Only `diff-coverage` / `quality-gate` remain — advisory, `UNSTABLE` not `BLOCKED`. I don't merge — over to you. The one substantive human call already made (fold FR-011) is in; the class-3-style decisions here were the FR-011 fold and the autoclose set, both confirmed.

stijn-dejongh:
## Landing update — rebased onto current main + fresh review squad + coherence fold

Re-based over 3565, the `#2320` primary-anchor fix, the sync Wave-4 degod, and the clock-import ban. New tip after this round: `35d41981fb`.

### Fresh 3-lens adversarial squad (post-rebase)
- **Integration (vs merged main): SOUND.** `#2320`'s `ab91ed915d` adds only the `_anchor_repair_root` collapse-to-primary — **no** competing refusal guard. 3567 *calls* that anchor and layers ownership-awareness on top (main = collapse-to-primary; 3567 = make that collapse honest). The two anchor resolvers provably converge (both `.resolve()` the primary), so the foreign-lane refusal fires reliably and owner writes stay silent. 21 identity/anchor + 35 refusal/clock + 26 verdict tests green.
- **Architecture-scout: ship (ratified scope).** No duplicate `#2320` anchor (`mission_state.py` +171/−0 additive); parser + gate are net **de-duplication**; layering clean; the `primary` overload is not conflated.
- **Design-fit: RESCOPE flagged → resolved.** Surfaced the WP04 behavior change (worktree `--fix` now fails closed — you confirmed this is intended) and one coherence gap, folded below.

### Coherence fold (`fix(landing): guard repair_repo …`, `35d41981fb`)
WP04's ownership guard was wired only into the CLI shell (`_refuse_foreign_lane_fix`), leaving `repair_repo` **and its second caller** (`_teamspace_mission_state_gate.py`) ungated — a foreign-lane repair could bypass the refusal, and `#2320`'s test (which calls `repair_repo` directly without cd-ing into a worktree) couldn't catch a CLI-layer reversal. `repair_repo` is now **self-protecting** (`enforce_primary_write_ownership` after the anchor, before any write). Red-first: the new `test_repair_repo_fails_closed_from_foreign_lane_cwd` reds on the unguarded function and greens after; owner companion proves no over-refusal; the `#2320` suite stays green.

### Operator-facing behavior note (WP04, intended)
`doctor mission-state --fix` **from a linked lane worktree** now fails closed (refuses, names the primary) rather than silently converging as `#2320` shipped. Run it from the primary/a clone. Owner invocations keep `#2320`'s converge-on-write.

### Follow-ups filed (non-blocking)
- **#3583** — weak must-not-flip anchor test + exotic-worktree `_read_worktree_gitdir` edge cases
- **#3586** — twin `checkout_identity` seams (both claim "the ONE authority"); unforced single-channel invariant; parser/anchor drift

### Hand-off
Full landing notes are in the PR body. I don't merge — over to you. `diff-coverage`/`quality-gate` remain advisory (`UNSTABLE`, not required). CI on the coherence fold is validating now; I'll confirm green before you merge.

Issue #3498, opened 2026-08-16
implement hard-blocks on missing project charter, inconsistent with specify/plan which explicitly tolerate it


robertDouglass:
## Summary

`spec-kitty agent action implement` hard-blocks with exit code 1 on a fresh repository that has never run `charter interview` (no `.kittify/charter/charter.md` ever created). The prescribed remediation (`spec-kitty charter sync`) also fails, with no viable exit — the only way to proceed is to synthesize a full charter you never asked for. This is inconsistent with `specify` and `plan`, which explicitly document tolerating a missing charter.

## Environment

- Installed: `spec-kitty-cli` 3.2.6rc2 (via `uv tool install spec-kitty-cli`)
- `spec-kitty upgrade --agent-check` reports "up_to_date" against the latest published release (3.2.5)
- Fresh repository, `.kittify/charter/charter.md` never created (charter interview never run)

## Repro steps

1. In a fresh repo with Spec Kitty initialized but no charter ever created, run the normal mission flow: `specify` → `plan` → `tasks` (all succeed / tolerate the missing charter, per their own documented behavior).
2. Run: `spec-kitty agent action implement WP01 --mission <slug> --agent claude --profile <profile>`

Result — exit code 1:
```
Error: charter_source missing; run `spec-kitty charter sync`
synced_bundle missing; run `spec-kitty charter sync`
synthesized_drg missing; run `spec-kitty charter synthesize`
```

3. Attempt the prescribed remediation: `spec-kitty charter sync`

Result — also fails:
```
Error: Charter not found at /path/to/repo/.kittify/charter/charter.md
  Run 'spec-kitty charter interview' to create one,
  or 'spec-kitty upgrade' if migrating from an older version.
```

The only way to unblock `implement` is to run `spec-kitty charter interview --defaults --mission-slug <slug>` followed by `spec-kitty charter generate` — synthesizing a full charter (in our case a 139KB `charter.yaml`) purely to satisfy this gate, even though we never asked for governance overhead.

## Root cause (verified against current `main`, commit `b85487e`)

`src/specify_cli/charter_runtime/preflight/runner.py` already has a fresh-project exemption:

```python
def _is_optional_missing_charter_fresh_project(checks: list[CharterPreflightCheck]) -> bool:
    """Return True for a never-initialized charter stack.

    Missing project charter is optional in a fresh project.  Treat only the
    fully absent stack as advisory; partial/generated residue still blocks so
    stale charter state remains visible.
[...]

Issue #2530, opened 2026-07-10
Charter-activation follow-ups from #2526 squad (parity fail-open, dead alias, loader dedup, e2e test)


stijn-dejongh:
**Found by** the adversarial squad on PR #2528 (#2526). Bundle of non-blocking robustness/coverage/debt items. Split as needed.

1. **Parity guard fails OPEN on corrupt/unreadable input.** `src/charter/consistency_check.py` `_load_reference_ids_by_kind` (malformed `references.yaml` → `except Exception: return None`) and `_check_graph_kind_parity` (DRG-load failure → `except Exception: return`) both leave the divergence lists empty → the guard silently reports *passing*. This is the exact silent-degradation class the guard exists to prevent. Add a distinct "could-not-verify" signal + tests for a malformed `references.yaml` and a DRG-load failure.

2. **"Lynn Cole" interview alias is now dead-effect code.** `src/charter/interview.py` `apply_doctrine_intent_aliases` still mutates `interview.selected_*`, but `compile_charter` now sources activation from config only — so a user typing e.g. "avoid code bloat" no longer activates `DIRECTIVE_039`, with no diagnostic. Remove the alias or re-wire it to promote into `config.activated_*`.

3. **`load_default_pack_ids` duplicated** in `src/specify_cli/doctrine/org_charter.py` and `src/specify_cli/upgrade/migrations/m_unify_charter_activation.py` (near-identical, both read `charter/packs/default.yaml`). Consolidate to one charter-layer helper (the `org_charter.py` copy self-notes this).

4. **No CLI-boundary test for reject-not-drop.** `charter generate`/`synthesize` catch `UnknownArtifactIdError` cleanly (verified), but only the charter-layer unit raise is tested. Add an e2e test driving `spec-kitty charter generate` with a stale `config.activated_*` stem → clean non-zero exit + message, not a raw traceback.

**NITs:** `UnknownArtifactIdError` message doesn't name the remediation (unlike `UnknownActivationIdError`); `promote_activations` writes one `save()` per yaml_key (partial-promote on mid-loop failure — safe because idempotent, worth a docstring caveat); reverse-paradigm parity in `consistency_check.py` relies on paradigm `id==stem` (pin with a test); bounded fan-out orphan threads have no hard cap in a long-lived daemon (mitigated by the per-handler in-flight guard added in PR #2528, which caps at 1/handler).

PR #3585, opened 2026-08-19
fix(charter): remediations that can actually clear their own checks (#2831 residual)

MOES-Media:
Four charter-freshness states hand the operator a command that cannot clear the check blocking them. This removes that, gives a working command where one exists, and says so plainly where none does.

> **Revised after pre-merge review.** An earlier revision of this PR added a `--recover-invalid` flag to `charter generate`. The squad found it made the operator **worse off than the bug** — see "What the review changed" below. That flag is withdrawn; `generate.py` is byte-identical to `main`.

## The defect, live on `main`

`freshness/computer.py` defines `_REMEDIATE_CHARTER_SYNC = "spec-kitty charter sync"` and emits it from four blocking states — `charter_source: missing|invalid` and `synced_bundle: missing|stale`. `src/charter/sync.py:18` documents that command as, verbatim: *"it always reports `synced=False` / `files_written=[]`"*.

The operator runs it, nothing changes, the gate refuses identically. Meanwhile the charter diagnostics report healthy, because they resolve `charter.md` while the gate resolves `charter.yaml` — so every investigative step confirms the wrong conclusion. That is #2831.

**#3498 did not close this.** Its bypass requires `charter_source == "missing"` AND `synced_bundle == "missing"` AND `synthesized_drg in {"missing", "built_in_only"}`, and its own docstring says *"Any stale, invalid, or other partial residue remains blocking."* It made the fresh **uninitialized** project advisory instead of blocked. #2831 was closed against that symptom, not the mechanism.

## What ships

| Situation | Remediation now |
|---|---|
| `charter.yaml` absent, **no** `charter:` pointer | `spec-kitty charter generate --no-from-interview` |
| `charter.yaml` absent, pointer **present** | none — `detail` explains why |
| `charter.yaml` unparseable (either shape) | none — `detail` explains why |

`_REMEDIATE_CHARTER_SYNC` is deleted. No state now names a command that cannot clear it.

**Why two of those emit nothing.** An unparseable `charter.yaml` cannot be repaired by any write path in the codebase — each round-trip-parses the file to preserve authored sections, and `--force` gates no destructive overwrite (`write_compiled_charter`'s own docstring). And on a project that has run `charter generate` before, `.kittify/config.yaml` carries a `charter:` pointer; `pack_manager.resolve_activation_write_target` fails **loud by design** on a dangling one (INV-5 / #2530), so `charter generate` cannot run at all. The documented recovery for that shape, `spec-kitty upgrade`, is interactive and cannot be handed over as a runnable command.

Being told no command exists, with the reason, beats being sent to one that exits non-zero. That is the whole point of #2831.

## What the review changed

The first revision added `--recover-invalid`, which moved an unparseable `charter.yaml` aside and bootstrapped a fresh one. Two lenses independently reproduced this, and it was confirmed by hand:

```
project that has run `charter generate` once   (the normal state)
  -> charter.yaml becomes unparseable
  -> run the remediation this PR documented
  -> Unexpected error: CHARTER_PACK_CONFIG_INVALID   exit 1
  -> charter.yaml is GONE; only charter.yaml.bak remains
```

Generation died in `provision_mission_type_activations` **after** the move and **before** recreating the file. The result was strictly worse than the original bug. The flag is gone.

Chasing that exposed a bigger correction: the dangling-pointer fail-loud breaks the `missing` remediation too, not only `invalid` — so the scope shrank again, to what is actually true.

**The oracle could not have caught it.** Its fixtures only ever seeded a pointer-less, never-generated project — the one shape where the flag worked. Fixtures now drive **both** config shapes, and the exempt states are pinned explicitly rather than by omission.

## The oracle

`tests/architectural/test_remediation_effectiveness.py` pins the rule: every emitted remediation must **exit 0 AND leave its state `fresh`**. States that emit nothing are pinned separately, so the exemption is a decision on the record rather than a gap.

The destination state is asserted, not merely that something changed — an assertion of the form `after != before` is satisfied by a remediation that writes malformed YAML and exits 17, since `missing -> invalid` is a change.

## Correction to an earlier version of this description

Two claims in the first revision were wrong, and both are withdrawn:

1. It said `--recover-invalid` gave the operator "a working command". It did not, for the realistic case.
2. It attributed the weaker `after != before` assertion to "an earlier draft" of #3015. That assertion is still #3015's primary driver at its current HEAD — supplemented by a narrower test, not superseded. The comparison was unfair as written.

## Evidence

- Squad: **13 findings** across three lenses. The severity-5 was found independently by two of them and reproduced by hand before being acted on.
- Oracle **10 passed**; charter freshness / preflight / runtime + integration + same-tier-uniqueness **122 passed**; `ruff` clean.
- Pre-existing and NOT this PR's: `tests/charter/test_mission_type_profiles.py::TestMissionCreatePropagatesEmptyActionSequenceError::test_create_mission_propagates_named_exception_type` reproduces identically on unmodified `main` at `38090b649`.

## Follow-up worth filing separately

**A previously-generated project whose `charter.yaml` goes missing or becomes unparseable has no non-interactive recovery.** `charter generate` fails closed on the dangling pointer by design; `spec-kitty upgrade` prompts. That is live on `main`, independent of this PR, and is why two states here emit no command. Fixing it would let them.

### DIFF
--- a/docs/changelog/CHANGELOG.md
+++ b/docs/changelog/CHANGELOG.md
@@ -354,6 +354,25 @@ _The 3.2.6rc2 candidate cycle is open (rc1 shipped 2026-08-12). Entries land her
 
 ### 🐛 Fixed
 
+- **A charter-preflight state no longer hands you a command that cannot clear it
+  (`#2831`).** Four freshness states — `charter_source` `missing`/`invalid` and
+  `synced_bundle` `missing`/`stale` — emitted `spec-kitty charter sync` as their
+  remediation, and `charter sync` is a pure staleness reporter that never writes
+  (`src/charter/sync.py`: "it always reports `synced=False` / `files_written=[]`").
+  The operator ran it, nothing changed, and the gate refused identically, while every
+  charter diagnostic reported healthy because they resolve `charter.md` and the gate
+  resolves `charter.yaml`. There was no exit. Now the one state a command can actually
+  clear — an absent `charter.yaml` on a project with no `charter:` pointer — names
+  `spec-kitty charter generate --no-from-interview`, and the states no command can
+  clear emit **no** remediation plus a `detail` explaining why. An unparseable
+  `charter.yaml` cannot be repaired by any write path (each round-trip-parses it to
+  preserve authored sections), and with a `charter:` pointer present `charter generate`
+  fails closed by design (INV-5, `#2530`) so it cannot run either; the documented
+  recovery there, `spec-kitty upgrade`, is interactive. Being told no command exists,
+  with the reason, beats being sent to one that exits non-zero. A new architectural
+  gate (`tests/architectural/test_remediation_effectiveness.py`) pins the rule: every
+  emitted remediation must exit 0 **and** leave its state `fresh`.
+
 - **`charter activate --cascade` now follows an org-pack dependency edge no
   matter which pack in the chain declares it, and activating from an org pack
   that ships no dependency graph no longer crashes the command (`#3534`; closes
--- a/src/specify_cli/charter_runtime/freshness/computer.py
+++ b/src/specify_cli/charter_runtime/freshness/computer.py
@@ -115,7 +115,22 @@ FreshnessState = Literal["fresh", "stale", "missing", "built_in_only", "invalid"
 
 # S1192: shared remediation command strings, referenced by every sub-state
 # builder below that recommends a recovery command.
-_REMEDIATE_CHARTER_SYNC = "spec-kitty charter sync"
+# `charter sync` is deliberately NOT used as a remediation here (#2831).
+# `src/charter/sync.py` documents itself as a pure staleness reporter -- "it
+# always reports synced=False / files_written=[]" -- so every state that named
+# it handed the operator a command that could not clear the check blocking them.
+# Each state below now names a command that demonstrably reaches `fresh`,
+# enforced by tests/architectural/test_remediation_effectiveness.py.
+_REMEDIATE_CHARTER_GENERATE = "spec-kitty charter generate --no-from-interview"
+#: An unparseable `charter.yaml` needs the extra flag: `write_compiled_charter`
+#: refreshes only the DERIVED sections of an existing file and round-trips the
+#: document to preserve the AUTHORED ones, so it must parse it. `--force` does
+#: NOT help (it gates no destructive overwrite -- see that function's
+#: docstring); the file has to be moved aside first, which `--recover-invalid`
+#: does (to `charter.yaml.bak`).
+_REMEDIATE_CHARTER_RECOVER = (
+    "spec-kitty charter generate --no-from-interview --recover-invalid"
+)
 _REMEDIATE_CHARTER_SYNTHESIZE = "spec-kitty charter synthesize"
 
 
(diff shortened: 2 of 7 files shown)


MOES-Media:
## Closing — superseded by #3015, with a salvage note

Closing this in favour of reviving #3015. Two rounds of pre-merge review established that #3015 is the better foundation for #2831's residual, and that this PR was reasoning from premises that did not hold.

### Why

**The premise was wrong, twice.**

Round 1 shipped a `--recover-invalid` flag that moved an unparseable `charter.yaml` aside and bootstrapped a fresh one. On any project that has run `charter generate` before — i.e. the normal path to later having a broken charter — `.kittify/config.yaml` carries a `charter:` pointer, generation dies in `provision_mission_type_activations` **after** the move, and the operator is left with **no `charter.yaml` at all**. Strictly worse than the bug. Withdrawn.

Round 2 replaced that with "emit no remediation, because no command can clear these states". That is also false: **`spec-kitty upgrade --yes` exists and reaches `fresh`** via `ConsolidateCharterBundleMigration`. The probe that convinced me otherwise used an artificially minimal fixture, where the migration runner halts on an unrelated `0.10.1_populate_slash_commands` precondition before reaching the fix.

**#3015 already knew both of these things.** Its review cycle 1 caught precisely that false negative and fixed it by seeding realistic agent scaffolding into the fixture (`tests/specify_cli/charter_preflight/_fixtures.py::seed_realistic_agent_scaffolding`); its passing test then proves `upgrade --yes` resolves the state.

**And #3015 carries a fix this PR lacks.** Both review lenses independently flagged, at severity 4, that `preflight/runner.py`'s `_derive_blocked_reason` fallback (`check.remediation or "spec-kitty charter status"`) fabricates ``run `spec-kitty charter status` `` for any state with no remediation — recreating the exact "command that cannot clear the check" pattern #2831 is about, in the composed `blocked_reason` operators actually read. #3015 diagnosed and closed that as R-006. This PR does not, and its own gate could not see it, because the gate reads the raw `remediation` field and never the composed string.

### What is worth salvaging

1. **The effectiveness gate.** `tests/architectural/test_remediation_effectiveness.py` pins the rule that every emitted remediation must exit 0 **and** leave its state `fresh` — a stronger oracle than `after != before`, which a remediation that writes malformed YAML and exits 17 satisfies. #3015's own reviewer raised that weakness; whichever PR lands should carry the stronger form. Two caveats for whoever ports it: its fixtures must seed realistic scaffolding (see above), and it must assert on the **composed** `blocked_reason`, not just the raw field, or it will miss the R-006 class of defect entirely.
2. **The empirical findings**, which hold regardless of which PR lands:
   - `charter generate` cannot run against a dangling `charter:` pointer — `pack_manager.resolve_activation_write_target` fails loud by design (INV-5 / #2530). It is the wrong remediation for that shape; `upgrade --yes` is the right one.
   - Move-aside-then-regenerate is worse than doing nothing. Any fix must not destroy the original until a replacement exists.
3. **The CHANGELOG framing** for #2831, if useful — though it needs correcting, since it claims null-remediation states always carry an explanatory `detail` and `synced_bundle`'s do not.

### Method note, recorded because it caused all of the above

Three wrong conclusions in this one PR traced to the same cause: probing project shapes simpler than any real project. This codebase branches on project scaffolding in the migration runner, the activation-write-target resolver **and** the freshness computer, so a minimal fixture can make a working command look broken and a broken command look working. Use the shared realistic fixtures, or state plainly what a minimal probe cannot establish.

No code from this branch is being cherry-picked; #3015 should be brought forward on its own terms with its outstanding review findings addressed. Thanks to both review lenses — the severity-5 was caught before it reached anyone's project.

Issue #3572, opened 2026-08-19
Org-pack requires/suggests edges in drg/fragment.yaml never reach action-doctrine or cascade (residual of #3384)


stijn-dejongh:
## Summary

An org pack can author `requires`/`suggests` DRG edges in its `drg/fragment.yaml` (the canonical `OrgDRGFragment` shape established by #3387 / PR #3403), but those edges **never reach charter cascade or the action-doctrine bundle**. They are read only by the diagnostic merge path (`doctor doctrine` / `charter list`), so an operator sees the edge in diagnostics yet `charter activate --cascade` never acts on it.

This is the **residual of #3384** (now closed): #3384's acute P0 — a graphless org pack silently zeroing all action-scoped doctrine — was fixed by #3534's per-root graphless-degrade guard (degrade-and-warn, non-silent). But that shipped only the *degrade* remedy, not the *fragment-load* remedy. This issue tracks the remaining half.

## Mechanism

Charter cascade reads root-level `*.graph.yaml` only:

- `charter/_drg_helpers.py::load_validated_graph` → `load_graph_or_dir(root)` (root-level `graph.yaml` / `*.graph.yaml`).
- Org `drg/fragment.yaml` (`OrgDRGFragment`) is read by a **separate** path — `doctrine/drg/org_pack_loader.load_org_pack` → `doctrine/drg/merge.merge_three_layers` — whose only consumers are diagnostic/lint collectors. Nothing bridges it into `load_validated_graph`.

So a `requires`/`suggests` edge authored solely in `drg/fragment.yaml` is invisible to cascade.

## Already pinned

`tests/specify_cli/cli/commands/charter/test_org_cascade_chain.py::TestGraphlessPackWithFragmentEdgeIsInvisibleToCascade` (added in #3534) documents and regression-pins this limitation. When it is fixed, that test must be updated deliberately.

## Proposed direction

Wire org `drg/fragment.yaml` `OrgDRGFragment` edges into the graph `load_validated_graph` returns (against the #3387 `fragment.yaml` model — **not** the superseded `drg/*.graph.yaml` mechanism from the abandoned #3401), so org-authored dependency edges cascade. Mind the two-DRG-path reconciliation and the D-005 degrade posture.

## Links

- Successor of #3384 (closed; P0 symptom resolved by #3534).
- Under umbrella #3530 / reach sub-epic #3526.
[...]

Issue #3530, opened 2026-08-17
Tracking: org-tier doctrine loads, validates and reports healthy — and reaches no consumer


MOES-Media:
An organisation doctrine pack can be registered, validated, activated, and reported healthy — and still deliver nothing to the runtime. This is a tracking issue for that class. It files no new defect; it groups reports that were filed separately and share one root cause, so the story does not have to be reconstructed from ten independent threads.

## The pattern

Every member below has the same shape: **a success report over an inert result.** The command exits zero, the pack validates, a count is emitted — and the artifact never reaches the consumer that was supposed to read it. Nothing fails loudly, so the failure is only ever found by execution, never by a test.

The recurring mechanisms are worth naming, because they repeat across unrelated modules:

- **The parameter already exists; the caller does not pass it.** The org-tier read is fully built and reachable, and the production call site omits it (#3516, #3525, #3527).
- **Built-in and org layers take different code paths for the same kind**, so behaviour verified against built-in content does not hold for an org pack (#3490, #3523).
- **A guard applied at the project layer is missing at the org layer** of the same function (#3384).
- **The published guide and the loader disagree**, and the mismatch is counted as success (#3387).

## Members

| Issue | State | Status |
|---|---|---|
| #3384 — org pack with no root `*.graph.yaml` zeroes **all** action-scoped doctrine | open | fix in flight, PR #3401 |
| #3387 — guide-named step contracts never load but are counted; several authoring mistakes are silent | open | unclaimed |
| #3490 — built-in artifacts load recursively, org/project artifacts of the same kind do not | open | unclaimed |
| #3516 — org-tier doctrine never reaches step contracts, `expected-artifacts`, or the executor's DRG | open | fix in flight, PR #3520 |
| #3523 — org packs cannot ship templates or mission-FSM content; two forked resolvers, neither with an org tier | open | fix in flight, PR #3524 |
| #3525 — dossier completeness index org-blind; existence-filter duplicated across 8 sites | open | folded into PR #3520 |
| #3527 — `charter activate/deactivate --cascade` loads its DRG org-inert | open | mission in progress |
| #3385 — `_org_scan_dirs` scanned a phantom layout, dropping every org-pack artifact from `charter activate` | **closed** | fixed; kept here as the first instance of the pattern |

## Related, but a different root cause — deliberately not children

These share the *symptom* (doctrine that loads but does not arrive) while failing for other reasons. Listed so the boundary is explicit rather than implied:

[...]

Issue #3410, opened 2026-08-14
Epic: Charter/doctrine silent-drop — org-pack & mission_type loading must fail loud, never fake-green


stijn-dejongh:
## Why

The charter/doctrine loading path has a class of **silent-success / silent-drop** defects: a guide-conformant org pack, a single `charter activate`, or an unknown (or even a shipped built-in) `mission_type` can **zero out or mis-route action-scoped doctrine** while emitting only a log WARNING. The run then proceeds **fake-green** — governed by empty or wrong doctrine — with nothing surfacing that governance never applied. This is P0-class: the governance layer silently fails open, and no operator-visible signal is produced.

## For whom

- **Org/project pack authors** — whose conformant packs silently contribute zero doctrine.
- **Every operator** relying on doctrine actually binding at runtime (directives/tactics/procedures/guards).
- **Maintainers** who need honest CI signal, not a fake-green mission.

## Intended effect

When this epic is done:

- Every doctrine / `mission_type` resolution failure **fails loud** — a structured, operator-visible error, not a buried log line.
- **No configuration silently yields empty or wrong doctrine**; guide-conformant packs and org/project-tier mission types resolve end to end or refuse explicitly.
- The guard-table path never falls through to a hardcoded `software-dev` family for an unknown or non-matching mission type.
- Each P0 defect lands an issue-pinned **red-first `@pytest.mark.regression`** reproduction per [ADR 2026-07-17-1](https://github.com/Priivacy-ai/spec-kitty/blob/main/docs/adr/3.x/2026-07-17-1-red-main-is-honest-ci-is-release-authority.md).

## Members (native sub-issues)

**P0 — silent-drop / fake-green:**
- #3384 — org pack without root `*.graph.yaml` silently zeroes ALL action-scoped doctrine
- #3385 — `_org_scan_dirs` phantom-layout scan drops every org-pack artifact from the DRG on `charter activate`
- #3386 — unknown `mission_type` (and the shipped built-in `plan`) executes under software-dev guard tables

**Same root cause, guard-table / mission_type family (near-duplicates of #3386 — dedupe during implementation):**
- #3397 — org/project-tier mission types unusable end to end (built-in-only roster; custom type resolves to empty action sequence)
- #3407 — `_check_cli_guards` hardcodes `mission_family="software-dev"` (mission-blind guard branch, latent spurious-block)

**Authoring-time fidelity (lower blast radius, same subsystem):**
[...]

Issue #2076, opened 2026-06-22
CT5: Fix stale golden-count assertions + fakeable-DoD / dead-assertion tail


stijn-dejongh:
**Parent epic**: #2071 ("Tests as scaffold, not friction")
**Findings doc**: `docs/development/test-suite-friction-audit.md` → CT5
**Severity**: S (renata F2 + F5, debbie T2)

## Failure mode

Stale golden-count assertions and a tail of fakeable-DoD / dead assertions: counts that drifted out of sync with the contract, existence/`hasattr`-only checks that assert nothing about values, and dead assertions that can never fail.

## Evidence

- `status/test_models.py::test_lane_enum_has_nine_values` asserts `len(Lane) == 10` — name/docstring drift after `Lane.GENESIS`.
- 182-wide tail of `assert len(...) == N` golden counts where an **adjacent set-equality already covers the contract**.
- Vacuous / dead:
  - `sync/tracker/test_local_service.py:420` — `"import" not in source or …` is **always-true**.
  - `::test_map_add_and_list_roundtrip` — dead patch, overclaiming name.
  - `cross_cutting/test_gitignore_manager_unit.py::test_result_object_structure` — 6 `hasattr`, **no values**.
  - `doctrine/test_structure_templates.py::test_structure_templates_exist` — `.exists()` only.
  - `release/test_release_payload_draft.py` + `doctrine/test_relationship_fields_rejected.py` — hasattr / absence only.

## Structural fix

- **Rename** `test_lane_enum_has_nine_values` to match the real count (and fix the assertion/docstring).
- **Drop** redundant `len()==N` where a set-equality already exists; keep counts only where **cardinality is the contract**.
- **Add a content/behavior assertion** next to each existence / `hasattr` check.
- **Delete** the 2 dead assertions (the always-true `sync/tracker/test_local_service.py:420` and the dead-patch `::test_map_add_and_list_roundtrip`).

stijn-dejongh:
### Grounded evidence — 2nd-wave audit (reviewer-renata dev-assist lens, 2026-07-12)

Additional golden-count / weaker-duplicate findings from the decomposition missions (beyond the ones already listed):

[...]

PR #3588, opened 2026-08-19
fix(doctrine,charter): single-authority resolution parity for org/project doctrine (M1) [#3490 #3426 #2981]

### LINKED ISSUE #2981: plural↔singular kind map restated in 4+ places; the charter renderer's copy has already drifted 2 kinds behind and fails open

### LINKED ISSUE #3426: charter activate silently drops org styleguides stored in subdirectories (nested-org residual of #3385)
## Impact

An operator who stores a project/org **styleguide in a subdirectory** — e.g. `<org_root>/styleguides/writing/house-style.styleguide.yaml`, the layout the built-in styleguide loader explicitly supports — and then runs `charter activate <any-artifact>` will find that styleguide **silently dropped** from the doctrine graph, with no error. The live doctrine loader reads it fine; the charter *activation* resolver does not. This is the same silent-drop class as #3385, residual for the `styleguide` kind after #3399 fixed it for the flat-layout kinds.

## Why it survives #3399

#3399 taught `charter.kind_vocabulary._org_scan_dirs` to scan the flat `<root>/<plural>` org layout the loader actually uses. But it hardcodes that flat entry as **`recursive=False`** for *every* kind (`src/charter/kind_vocabulary.py` L268). The live loader decides recursion **per kind** through a single authority — `BaseDoctrineRepository._project_scan` (`src/doctrine/base.py` L153-159, applied to the org layer at L257), which `StyleguideRepository` overrides to `rglob` (`src/doctrine/styleguides/repository.py` L59-61, "styleguides may live in subdirectories").

So for `styleguide`:
- **Loader:** `_org_dirs("styleguides")` → `[<root>/styleguides]`, then `rglob` → **loads** `writing/house-style`.
- **Charter resolver:** `_org_scan_dirs(STYLEGUIDE, …)` → `(<root>/styleguides, recursive=False)`; `_iter_artifact_paths` non-recursive `glob` (L311) → **misses** it. The legacy entry only inspects `…/built-in`, not `…/writing`, so it misses too. `resolve_artifact_urn` raises `UnknownArtifactIdError`, `charter.drg` swallows it → the node is dropped from `filter_graph_by_activation`.

`asset` has the same loader-vs-resolver recursion divergence but is **not** charter-activatable (`_NON_AUGMENTATION_ELIGIBLE_KINDS`), so it does not affect the activation path — noting it only so the seam fix covers it too.

## The structural root (and why this wasn't folded into #3399)

The flat-vs-recursive decision now lives in **two independent authorities** that already disagree: `_org_scan_dirs`' hardcoded `recursive=False` vs `_project_scan`'s per-kind override set. Nothing gates them together, so a *third* future `_project_scan` → `rglob` override would silently reopen this bug for that kind, with green tests. Fixing it properly is a shared-seam refactor (design-pass scope), deliberately kept out of the #3399 landing pass per the maintainer runbook's defer-when-needs-design-pass exception.

## Proposed fix

1. Derive `_org_scan_dirs`' `recursive` flag from a **single authority shared with** `BaseDoctrineRepository._project_scan` (e.g. a `kind → recursive` map both sides import, or have the resolver consult the same override registry) — no second hardcoded copy.
2. Add a regression that constructs a **nested-subdir styleguide org pack** and asserts resolver-vs-loader agreement across **all** `CHARTER_KIND_TOKENS` (current tests are `TACTIC`-only — `tests/charter/test_kind_vocabulary_scan_roots.py`, `test_org_scan_dirs_activation_regression.py`).
3. Add a gate pinning the invariant: *kinds whose `_project_scan` is overridden to `rglob` ⇔ kinds `_org_scan_dirs` returns `recursive=True` for at the flat layer.*

## Provenance

Surfaced by a `paula-patterns` SSOT/duplication lens during the #3399 maintainer landing pass; diagnosis independently verified against `base.py`/`styleguides/repository.py`/`artifact_kinds.py`. Sibling of #3384. Parent epic: #3410.

### LINKED ISSUE #3490: Built-in doctrine artifacts load recursively (rglob), org/project artifacts of the same kind do not (glob) — silent org-pack content loss and 71% corpus undercount for tactics
## Summary

Every doctrine repository's built-in layer is scanned recursively (`rglob`); the org and project layers are scanned non-recursively (`glob`), through the exact same default method. For a kind that does not override that default — `tactics` is the measured example — this means an org/project pack author who nests their own artifacts into subdirectories (mirroring the convention the built-in corpus itself uses) gets those artifacts silently dropped, with no warning, and separately means anyone doing authoring-time "does an existing built-in artifact already cover this?" research by listing a directory rather than going through `DoctrineService` sees under a third of the real corpus. Verified first-hand against `3.2.6rc1` source and by direct measurement.

## The asymmetry, in code

`BaseDoctrineRepository` (`src/doctrine/base.py`):
- `_load_built_in_items` (`:177-200`) scans with `self._built_in_dir.rglob(self._glob)` (`:182`) — recursive.
- `_project_scan` (`:153-159`), the default used for **both** the org layer and the project layer, scans with `project_dir.glob(self._glob)` (`:159`) — non-recursive. `_apply_overlay_layer` (`:227-260`) calls this same method for both `org_dirs` and `project_dir` (`:365`, both layers).

`TacticRepository` (`src/doctrine/tactics/repository.py`) declares `_glob` but does not override `_project_scan` — so for tactics specifically, the org and project layers get the non-recursive default while the built-in layer is recursive.

## Measured

```
built_in_dir = built_in_dir(ArtifactKind.TACTIC)   # doctrine/pack_paths.built_in_dir
flat = list(built_in_dir.glob("*.tactic.yaml"))       # 36
recursive = list(built_in_dir.rglob("*.tactic.yaml"))  # 124
```

**88 of 124 built-in tactics (71%) live in a subdirectory and are invisible to a flat, non-recursive listing of the same directory.** `DoctrineService` itself loads all 124 correctly (it uses the recursive scan) — the gap is between what the loader sees and what a flat directory listing shows, and, separately, between what the built-in layer supports and what the org/project layer supports for the same kind.

## Why this is a real defect, not just a discoverability nit

Two distinct consequences follow from the same code asymmetry:

1. **Authoring-time undercounting.** Anyone (human or agent) checking "does a built-in tactic already cover this need" by browsing the directory rather than querying `DoctrineService` sees 29% of the real corpus. On a real mission this caused an entire built-in tactic family (19 tactics under a subdirectory) to go unreferenced by a newly authored pack until a later review pass caught it by querying the loader directly rather than trusting the directory listing.
2. **Silent org-pack content loss.** Because `_project_scan`'s non-recursive default applies identically to `org_dirs` and `project_dir`, an org-pack author who organizes their own tactics into subdirectories — the exact layout convention the built-in corpus uses — gets those files silently dropped from `list_all()`/`get()` with no warning, no validation failure, nothing. This is the same silent-drop shape as #3426 (which fixes the analogous gap for `styleguide` at the charter-activation-resolver layer) and #396 (closed; fixed the *built-in* "shipped" scan to `rglob` in 2026-04, explicitly leaving "project-dir scanning intentionally flat" unexamined for this consequence). Neither of those covers the loader-level gap for kinds like `tactic` that never override `_project_scan` at all — #3426 is scoped to the charter-activation resolver diverging from an already-recursive `StyleguideRepository._project_scan` override; #396 only touched the built-in scan.

## Suggested fix

- At minimum: emit a warning when an org/project override directory contains YAML files in subdirectories that a non-recursive `_project_scan` will never reach, for every kind that has not opted into `rglob` — so this fails loud instead of silent, consistent with #3410's intent.
- Better: make recursion a single per-kind authority both the loader and any consumer (e.g. #3426's `_org_scan_dirs`) read from, rather than each `_project_scan` override being independently decided — #3426 already names exactly this "two independent authorities" risk for the kinds that do support nesting.

Related to #3410 (parent epic). Related but distinct: #3426 (styleguide, charter-activation-resolver layer), #396 (closed; built-in-only fix), #3387 (a different rglob/glob mismatch, for assets between the loader and `pack validate`).

stijn-dejongh:
## Org-pack & project-overlay doctrine now loads and activates as completely as built-in

**If you author doctrine in an organisation pack or a project overlay, Spec Kitty
now finds and activates all of it — including artifacts stored one directory
deep — instead of silently dropping them while every check stayed green.** Before
this change, a tactic, styleguide, or agent profile placed in a subdirectory
(e.g. `styleguides/writing/`) would load at runtime and show up in `charter list`,
yet `charter activate` would quietly skip it — a fake-green symptom (clean checks,
missing governance), measured at a **71% tactic undercount** for nested org packs.
The charter kind vocabulary had also drifted, so `charter context --include
glossary_pack:<id>` failed to render and unknown-but-valid kinds reported
"unsupported selector". Both are fixed, and a new gate makes this class of silent
drift fail loudly if it is ever reintroduced.

This is **M1** of the charter-resolution program (independent of M2, `#3572`).

Closes #3490.
Closes #3426.
Closes #2981.

Rolls up to — and deliberately does **not** close — umbrella epic #3530 and
fail-loud epic #3410; both remain open across the program (M1 contributes the
parity/totality gate, FR-007).

### Root causes (two independent)

1. **Recursion divergence.** Built-in discovery used `rglob` (recursive);
   org/project used `glob` (non-recursive). Worse, the *loader*
   (`doctrine/base.py`, `doctrine/agent_profiles/repository.py`) and the
   *charter-activation resolver* (`charter/kind_vocabulary.py`) each decided
   recursion independently and disagreed per kind — so nested org styleguides
   loaded and listed but were **un-activatable** (`#3426`, the list-vs-activate
   divergence).
2. **Hand-restated kind vocabulary.** The plural↔singular kind map was hand-copied
   across charter modules; two `_activation_render` copies had drifted two kinds
   behind and failed open (`glossary_pack` rendered as its plural; `glossary_packs`
   inference blinded), and the string-keyed copies escaped the existing
   consistency gate (`#2981`).

### Fix

- **One shared recursion authority** (`doctrine/discovery_recursion.py`, doctrine
  layer per C-006) read by **both** the loader (`base._project_scan`,
  `agent_profiles._load`; three redundant `rglob`/`_project_scan` overrides
  deleted) **and** the resolver (`kind_vocabulary._org_scan_dirs` /
  `_layer_scan_dirs`). Org/project discovery is now unconditionally recursive,
  matching built-in (C-001). Kind-specific globs keep `.provenance/*.yaml` and
  `.md` out (C-002); the recursive flat-org scan excludes the reserved `built-in/`
  subtree so flat-wins-over-legacy precedence is preserved.
- **One derived kind vocabulary** — `CHARTER_ACTIVATABLE_*` in
  `doctrine/artifact_kinds.py` (10 kinds incl. `anti_pattern` = `ArtifactKind −
  {template, asset}`, C-003/FR-005) — collapsing the four hand copies in
  `charter.activations` / `charter._activation_render` and fixing the two drifted
  maps. `_ALLOWED_KINDS` is deliberately left untouched (documented non-goal).
- **`--include` selector widened**: `glossary_pack` renders; `anti_pattern`
  (no artifact files) resolves to a normal not-found; every charter-activatable
  kind is recognised, never "unsupported" (FR-006).
- **Falsifiable parity/totality gate**: binds loader↔resolver recursion by
  behavioral fixture (per kind, both directions) and covers the string-keyed kind
  authorities the enum-keyed guard could not see (FR-007/NFR-003).

### Constraints held

- **C-004 (no golden ripple / no cascade-reach change)** — verified: 465 DRG + 55
  cascade golden tests green; recursion only adds nested overlay discovery, vocab
  changes are render/inference helpers.
- **C-005** — new code passes `ruff` + `mypy --strict` with zero suppressions.
- **C-006** — `charter` never imports `specify_cli`; the authority lives in
  `doctrine`.

### Verification

Red-first for every fix (nested tactic/profile dropped pre-fix; nested styleguide
un-activatable pre-fix; `_singular_kind` drift; selector unsupported). 87 mission
tests + broad regression (521 doctrine, 671 charter, 465 DRG) green. Falsifiable
gate proven both directions.

---

### Review-squad pass (pre-PR)

Three adversarial lenses reviewed the consolidated diff. Findings were remediated
in the folds commit:

- **Correctness**: the recursive flat-org scan excluded a `built-in` component at
  *any* depth while the compensating legacy entry only covers the immediate
  child — a nested `<plural>/<category>/built-in/x` was loaded by the loader but
  dropped by the resolver. Fixed (immediate-child scoping) + regression test.
- **Test rigor**: C-002 negatives were vacuous (checked never-declared ids) — now
  assert the exact discovered set and a provenance sidecar that *declares* an id;
  added a reintroduced-`_project_scan`-override guard (which surfaced and removed a
  third redundant override, `MissionTypeProfileRepository`); added a charter
  re-declaration guard for the `#2981` vocabulary-copy class.

### Follow-ups (out of M1's named scope)

- `charter.drg::_SINGULAR_TO_PLURAL` is a fifth complete `plural↔singular` hand
  copy (identical 10 kinds), on the golden-adjacent DRG activation-filter path —
  left un-collapsed under C-004 discipline; the new charter re-declaration guard
  exempts it with this note. Candidate for a follow-up collapse onto
  `CHARTER_ACTIVATABLE_SINGULAR_TO_PLURAL`.

---

### Maintainer landing notes

- **Rebased onto current `upstream/main`** (was 18 commits behind); clean rebase.
- **Autocloses corrected**: the original body used a single `Closes #3490, #3426,
  #2981` — GitHub only auto-closes the *first* issue after one keyword, so the two
  OPEN issues (`#3426`, `#2981`) would not have closed on merge. Now one keyword
  per issue; the epics stay non-closing references.
- **Two folds** (both architectural-shard gates the rebased tip surfaced;
  each a separate `test(landing):` commit):
  1. *pin exact activatable-kind set (golden-count gate)* — the new
     `test_charter_activatable_vocabulary.py` asserted
     `len(CHARTER_ACTIVATABLE_KINDS) == 10`, a bare-cardinality "convert" site the
     golden-count recurrence guard (`#2076`) reds against `tests/doctrine`'s
     implicit ceiling of 0. Converted to the exemplar exact-set assertion (names
     all 10 kinds) — stronger contract, aligned with the mission's
     single-vocabulary thesis.
  2. *re-pin built-in-authority allowlist to the moved org-tier join* —
     `test_built_in_location_authority` pins the ORG-tier legacy
     `flat / "built-in"` join in `_org_scan_dirs` by exact `(file, lineno)` (a
     sanctioned exemption, not a built-in-pack reconstruction). This mission's
     rewrite of `_org_scan_dirs` — and dropping the now-obsolete "Known residual
     (tracked #3426)" docstring paragraph — moved the unchanged join from line 283
     to 269. Re-pinned with a dated rationale; the join is byte-identical.
- **Pre-existing advisory (not this PR)**: `mypy --strict` reports 2 pre-existing
  `_PackContextLike`/`PackContext` arg-type mismatches in
  `charter/mission_type_profiles.py` (a file this PR does not touch). The mypy step
  is advisory (`continue-on-error`) and these are out of M1's blast radius — noted,
  not folded.

https://claude.ai/code/session_01VwTxct1RcSBb4u6JNn53XT


stijn-dejongh:
**Maintainer landing pass — claimed.**

Picking this up for a landing pass. Plan:
- Rebase onto current `upstream/main` (branch is 18 commits behind).
- **Fix the PR-body autocloses.** Current `Closes #3490, #3426, #2981` only auto-closes the *first* issue — GitHub needs one keyword per issue. #3426 and #2981 (both OPEN, in-scope per the mission issue-matrix) would not close on merge. Will rewrite to per-issue `Closes` keywords; the epics (#3410, #3530) and out-of-scope missions (#2829/#3038/#3572) stay non-closing references.
- Classify any red checks (4 bins), targeted verification, add a consumer-focused changelog entry for the org-pack doctrine fix.
- Hand off green for the operator to merge; I do not merge.

Evidence to follow.

stijn-dejongh:
## Maintainer landing summary — ready for operator merge

**Verdict: LAND.** Rebased, autocloses fixed, two architectural-gate folds, full targeted verification green. I do not merge — handing off for you.

### Pointers addressed
- **Rebased onto current `upstream/main`** (was 18 commits behind) — clean rebase, no conflicts. Force-pushed with lease.
- **Autocloses fixed.** The original body used a single `Closes #3490, #3426, #2981` — GitHub only auto-closes the **first** issue after one keyword, so `#3426` and `#2981` (both OPEN, both in-scope per the mission `issue-matrix.json`) would **not** have closed on merge. Rewrote to one keyword per issue and verified via the API:

  | Ref | `closingIssuesReferences`? | Note |
  |---|---|---|
  | `#3490` | ✅ linked | already CLOSED (completed earlier today) — no-op, kept for provenance |
  | `#3426` | ✅ linked | OPEN → will close on merge |
  | `#2981` | ✅ linked | OPEN → will close on merge |
  | `#3530`, `#3410` | ❌ excluded (correct) | epics — "rolls up to", non-closing |

  The PR body was also re-led BLUF/consumer-first (operator impact before architecture).

### Folds (2) — each a separate `test(landing):` commit
1. **`70b3708`** *pin exact activatable-kind set (golden-count gate).* The new `tests/doctrine/test_charter_activatable_vocabulary.py` asserted `len(CHARTER_ACTIVATABLE_KINDS) == 10` — a bare-cardinality "convert" site that `test_golden_count_ban` (#2076) reds against `tests/doctrine`'s implicit ceiling of 0 (PR-introduced red). Converted to the exemplar exact-set assertion naming all 10 kinds; stronger contract, aligned with the mission's own single-vocabulary thesis.
2. **`3855293`** *re-pin built-in-authority allowlist to the moved org-tier join.* `test_built_in_location_authority` pins the ORG-tier legacy `flat / "built-in"` join in `_org_scan_dirs` by exact `(file, lineno)` — a **documented, sanctioned exemption** (it is the org-pack legacy nested-layout, not a built-in-pack reconstruction). This mission's rewrite of `_org_scan_dirs` — plus dropping the now-obsolete "Known residual (tracked #3426)" docstring paragraph (#3426 is fixed here) — net-shortened the docstring and moved the unchanged join from line 283 → 269. Re-pinned with a dated rationale; the join is byte-identical.

### Verification (rebased tip)
- **PR's own suites** — 92 passed (`tests/doctrine/{discovery_recursion, overlay_recursion_loader, charter_activatable_vocabulary, drg/recursion_parity_gate, drg/kind_mapping_totality}`, `tests/charter/{activation_kind_vocab_collapse, kind_vocabulary_recursion_parity, kind_vocabulary_scan_roots, context_renderers/include_selector_widening}`).
- **Architectural gate sweep** — green after the two folds: completeness (`ci_collection`, `marker_job`, `arch_shard_marker`, `ci_corpus_trigger`), `ratchet_baselines`, `golden_count_ban`, `built_in_location_authority`, `charter_no_specify_cli_import` (C-006), `charter_facades_reexport_doctrine`, `doctrine_public_surface`, `no_dead_doctrine_paths`, `built_in_location_authority`, `glossary_pack_{parity,boundary}`, `2093_authority_invariant`, `doctrine_census`, `charter_path_literal_authority`, `charter_sole_door_*` (×5), `doctrine_missions_stale_path_sweep`, `no_read_side_bypass`, `ratchet_positional_anchor_ban`, `inline_meta_read_gate`, terminology guard.
- **`ruff`** clean on all changed src + touched test files. **`docs freshness --ci`** → `errors=0` (external-URL link warnings only).
- **`mypy --strict`** — the PR's own changed files pass clean (C-005 holds). Whole-package `mypy --strict src/charter src/doctrine` reports **2 pre-existing** `_PackContextLike`/`PackContext` arg-type mismatches in `charter/mission_type_profiles.py` — a file **this PR does not touch**, and the CI mypy step is advisory (`continue-on-error`). Out of M1's blast radius; noted, not folded.

### Review-squad note
A pre-PR 3-lens squad already ran and its findings were folded (documented in the PR body). My landing pass independently re-validated the full authority/boundary/sole-door gate surface (green). I did **not** re-run a fresh adversarial squad given the narrow landing pointer + the operator-authored red-first mission — happy to dispatch the pre-merge aggregate squad if you'd like one before merge.

### Changelog
Already shipped in the PR — a consumer-focused `[Unreleased] → 🐛 Fixed` entry in `docs/changelog/CHANGELOG.md`. No separate fold needed.

**State:** watching CI; merge-ready on green. Operator merges.

stijn-dejongh:
## CI red adjudication — 3 defect folds pushed, 1 honest perf flake

First CI run surfaced 5 red jobs (`quality-gate` is the aggregate of the rest). Classified into the four bins; **three were real defects the local gate sweep didn't cover — now folded — and one is a pre-existing perf flake this PR can't have caused.**

### Fixed (bin 1 — PR defects)

| Fold | Red | Fix |
|---|---|---|
| `1655717` `refactor(landing)` | `arch-adversarial (arch_shard_1)` → `test_no_dead_symbols::test_no_public_symbol_in_all_is_unimported`: `doctrine.discovery_recursion::RECURSIVE_OVERLAY_KINDS` exported in `__all__` with no `src/` importer | The frozenset is a module-internal derivation surface — every real caller goes through the public `overlay_scan_is_recursive()`; only the parity gate reads the constant (tests don't count as callers, by design). Dropped from `__all__`; constant kept (parity gate imports it explicitly). No behaviour change. |
| `11d8df0` `test(landing)` | `doctrine-charter-tests` **and** `fast-tests-doctrine` → `test_profile_repository.py::test_project_glob_does_not_find_profiles_in_subdirectory` | **Behaviour-coupled off-diff regression.** That test asserted the *retired* non-recursive project-overlay behaviour (`nested profile → None`), which this mission deliberately reverses (`agent_profiles._load` now recurses, #3490). It flipped red on CI without appearing in the PR diff. Inverted + renamed to `test_project_rglob_finds_profiles_in_subdirectory` (asserts the nested profile **is** found), mirroring the sibling shipped-layer test. |

### Not folded (bin 4 — perf flake, PR-independent)

- `fast-tests-core-misc` → `tests/retrospective/test_summary_tolerance.py::test_200_missions_under_5s`: 200-mission summary took **10.98 s vs a 10 s** wall-clock budget on a parallel xdist shard. The test's own docstring documents it as a **saturated NFR-003 budget already re-pinned once** (5 s → 10 s, #3456 landing 2026-08-15) that "surfaces a pre-existing saturated budget, not a regression." The retrospective-summary subsystem has **no dependency** on the doctrine discovery / kind-vocabulary code this PR changes, so the PR cannot have caused the slowdown. Per the flakiness policy this is **noted, never retried-to-green**; if it recurs the root fix is another budget re-pin as dedicated work, not a fold on this doctrine PR.

### State
Four folds now on the branch (the two above + the earlier golden-count and built-in-authority re-pins). Re-verified locally: the inverted profile test, discovery/overlay/parity suites (30 passed), and the dead-symbol gate all green; `ruff` + `mypy --strict` clean on the touched module. Pushed; watching the CI re-run. The only red that may reappear is the bin-4 retrospective perf flake — an honest, PR-independent red. Operator merges.

stijn-dejongh:
✅ **CI green — merge-ready.** 44 checks pass, 0 fail, `mergeStateStatus: CLEAN`.

The re-run after the four folds is fully green: `doctrine-charter-tests`, `fast-tests-doctrine`, `fast-tests-charter`, `arch-adversarial (arch_shard_1/2/3)`, `fast-tests-core-misc` (all variants), `e2e-cross-cutting`, lint, build — all pass. The bin-4 retrospective perf budget (`test_200_missions_under_5s`) **passed on re-run**, confirming it as a saturated-budget flake rather than a regression.

Branch: 6 commits (2 original + 4 landing folds). Autocloses verified (#3490/#3426/#2981 linked; epics excluded). Handing off — **operator merges**.

Issue #3625, opened 2026-08-21
`project sync store is locked` on large stores is SQLite BEGIN IMMEDIATE self-contention, not an OS lock


stijn-dejongh:
On large project sync stores the daemon logs:

```
Withholding project <uuid> body work after a project-scoped failure: project sync store is locked
Sync failed (attempt N, next backoff …): one or more project body drains failed …: project sync store is locked
```

This is **not** an OS-level lock: `lsof` on the `sync.db` shows no holder, `sync.db-wal` is 0 bytes, and a read-only `?mode=ro&immutable=1` open succeeds instantly. It is a SQLite `BEGIN IMMEDIATE` busy translated to `ProjectStoreLockedError` (`src/specify_cli/sync/project_store.py:589-592`, acquisition at `:633`). It reproduces only on **large** stores (tens of MB), never small ones — consistent with in-daemon write-lock self-contention whose window scales with store size.

**Impact:** the whole daemon cycle reports failure and backs off, making the daemon log look like the primary fault. It is **not** the delivery blocker (that is #3564), but it is noise worth fixing — e.g. `busy_timeout` / retry-with-backoff on the immediate-transaction acquisition, or narrowing the write-lock window.

**Severity:** Low (noise). Independent of #3564. Found during the #3564/#3620 investigation.

stijn-dejongh:
Part of the CLI-side cluster behind spec-kitty-saas#1143 — review-squad pass summarized on the umbrella: https://github.com/Priivacy-ai/spec-kitty/issues/3406#issuecomment-5423681535

**Verified MOSTLY CLOSED.** `busy_timeout` is now set before `BEGIN IMMEDIATE` (`sync/project_store.py:572` and `:630`; 5.0s at `:78`), which addresses the acquisition-contention half. Remaining: narrow the in-daemon write-lock window on large stores, plus a bounded-wait regression test. (One claim UNVERIFIED without a large-store repro: whether residual contention is in-daemon vs cross-process.)

Issue #3624, opened 2026-08-21
sync daemon spawned with bare `sys.executable` dies with ModuleNotFoundError when the launching interpreter lacks specify_cli


stijn-dejongh:
`_spawn_sync_daemon_process` (`src/specify_cli/sync/daemon.py:1307-1331`) launches the background daemon with `subprocess.Popen([sys.executable, "-c", _background_script(...), ...])`. `sys.executable` is whatever interpreter launched the spawner. On a machine where the first `python3` on PATH is a different interpreter without `specify_cli` installed (e.g. a Homebrew/system python vs a pyenv/venv one), the auto-spawned child dies immediately:

```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
ModuleNotFoundError: No module named 'specify_cli'
```

These `<string>` line-3 tracebacks appear in `~/.spec-kitty/sync-daemon.log`. The failure is **intermittent** — a daemon launched via the correctly-shimmed `spec-kitty` (or `doctor restart-daemon`) runs the right interpreter, so it depends on who spawns it.

**Suggested fix:** resolve/validate a `specify_cli`-capable interpreter before spawn (or re-exec / guard when the import is unavailable) rather than trusting bare `sys.executable`.

**Severity:** Medium (intermittent). Independent of the admission-delivery bug (#3564). Found during the #3564/#3620 investigation.

stijn-dejongh:
Part of the CLI-side cluster behind spec-kitty-saas#1143 — review-squad pass summarized on the umbrella: https://github.com/Priivacy-ai/spec-kitty/issues/3406#issuecomment-5423681535

**Verified STILL FULLY VALID** (P2 — real intermittent data-loss). `_spawn_sync_daemon_process` (`sync/daemon.py:1318`) still launches with bare `sys.executable` and no `specify_cli`-capability check.

⚠ Note for anyone picking this up: the `daemon.py:927-942` code is a **reaper argv identity marker**, *not* a spawn-interpreter guard — don't mistake it for a partial fix. The `ModuleNotFoundError` path is untested (`test_daemon.py:127-143` only covers `_background_script` style). Fix: resolve/validate a `specify_cli`-capable interpreter before spawn.

Issue #3582, opened 2026-08-19
[bug] import-history masks preflight code and event diagnostics behind reconciliation counters


robertDouglass:
## What happened

The server already returns stable, bounded preflight failure diagnostics (`category`, `code`, and per-event `details` containing `event_id`, `index`, `path`, and `detail`). The current CLI discards them whenever the same response contains a truthy `reconciliation` object.

`PreflightRejected.__init__` currently selects:

```python
reconciliation = self.payload.get("reconciliation") or self.payload.get("error") or self.payload
super().__init__(f"server preflight rejected the batch: {reconciliation}")
```

The command then prints only `str(exc)`. On the current SaaS failure shape, operators see zero-shaped reconciliation counters instead of the actual rejection code and offending event. This is the mechanism behind the misleading attribution in #3329.

## What was expected

The CLI should preserve and render the server's primary rejection diagnostics first, while showing reconciliation only as secondary context. Rendering must remain bounded and avoid leaking authentication material or unbounded payload values.

## Reproduction

```python
from specify_cli.sync.history_import.upload import PreflightRejected

payload = {
    "error": "transition rejected",
    "category": "semantic_validation",
    "code": "transition_rule",
    "details": [{
        "event_id": "e1",
        "index": 0,
        "path": "payload.transition",
        "detail": "requires force",
    }],
    "reconciliation": {
        "events_received": 2,
        "events_rejected": 0,
    },
}
print(PreflightRejected(payload))
```

Observed:

```text
server preflight rejected the batch: {'events_received': 2, 'events_rejected': 0}
```

The actionable `transition_rule`, event ID, path, and detail are absent.

## Environment

- spec-kitty: `0482c395681fdf4ae4c2ec0eeded7021ede961bd`
- relevant source: `src/specify_cli/sync/history_import/upload.py:163-169`; command rendering at `src/specify_cli/cli/commands/sync.py:2965-2967`
- spec-kitty-saas: `ea7bc4a53565b2f0eeaad8c7c1ce904d67c1bb48`
- Observed 2026-08-19

## Acceptance criteria

- [ ] Preserve the complete structured server failure payload on `PreflightRejected`.
- [ ] Human output leads with `category`, `code`, and bounded primary detail.
- [ ] When present, render correlated `event_id`, `index`, and `path`; show reconciliation separately.
- [ ] Machine-readable/report output exposes the same stable fields without parsing a Python-dict string.
[...]

Issue #3581, opened 2026-08-19
[bug] import-history rejects current SaaS preflight 200 because client requires nonexistent results[]


robertDouglass:
## What happened

At the reviewed current revisions, the two halves of the history-preflight wire contract disagree:

- `spec-kitty-saas@ea7bc4a53` returns HTTP 200 with `{"accepted": true, "event_count": N, "reconciliation": {...}}` from `apps/sync/views.py:1527-1534`.
- `spec-kitty@0482c3956` accepts a 200 only when the body also has a one-to-one `results[]` collection covering every expected event ID (`src/specify_cli/sync/history_import/upload.py:183-190,232-246`).
- SaaS does not return `results[]` from preflight, so the CLI classifies the server's success as rejected and raises `PreflightRejected`.

This is independent of the earlier admission and semantic-validation gates. Once those gates are repaired, every otherwise-valid `sync import-history --apply` cohort still stops at this response parser.

## What was expected

A version-compatible current CLI and current SaaS should share one explicit preflight success schema. A successful, correlated, non-mutating server verdict should permit upload; malformed, incomplete, duplicate, or uncorrelated verdicts must remain fail-closed.

## Reproduction

Using the current client type:

```python
from specify_cli.sync.history_import.upload import _PreflightResponse

payload = {
    "accepted": True,
    "event_count": 2,
    "reconciliation": {"events_received": 2, "events_rejected": 0},
}
response = _PreflightResponse(
    200,
    payload,
    expected_event_ids=("e1", "e2"),
)
print({"server_accepted": payload["accepted"], "cli_accepted": response.accepted})
```

Observed:

```text
{'server_accepted': True, 'cli_accepted': False}
```

## Environment

- spec-kitty: `0482c395681fdf4ae4c2ec0eeded7021ede961bd`
- spec-kitty-saas: `ea7bc4a53565b2f0eeaad8c7c1ce904d67c1bb48`
- spec-kitty-events: `6.1.0`
- Observed 2026-08-19

## Acceptance criteria

- [ ] Publish one canonical/versioned success and failure schema for `/api/v1/events/preflight/`.
- [ ] Current CLI accepts the current server's valid success response, or both sides change atomically to the replacement schema.
- [ ] If per-event correlation is required, SaaS returns exactly one verdict per submitted event ID and the CLI verifies exact set equality.
- [ ] Missing, duplicate, unexpected, or non-success event verdicts still fail closed.
[...]

PR #3626, opened 2026-08-21
fix(sync): restore hosted event-sync delivery — negotiated admission gate + preflight contract (#3564)

### LINKED ISSUE #3564: Hosted sync delivery dead-ends at admission_not_current: AdmissionOperationService.perform has no live call site
## Summary

Hosted event delivery is unreachable because the **client-side admission handshake is never invoked**. `spec-kitty sync now` (and `sync import-history`) gate on `admission_not_current`, and **no live CLI/daemon/runtime path calls `AdmissionOperationService.perform`** — every path only *reads* an already-registered delivery target via `ProjectDeliveryTargetRegistry.get_current()`. Since nothing ever registers/admits a target, `project_target_admissions` stays empty forever and delivery can never start.

This is the precise root cause behind the `admission_not_current` symptom, and it is broader than the *server*-side import-admission gap noted in #3406 gate 9 (#795/#585): it blocks **all** hosted delivery via `sync now`, not just `import-history`.

## Evidence (build `3.2.6rc2`)

1. **No live call site.** `grep -rn '\.perform(' src/ | grep -v test | grep -i admiss` → **zero hits**. `AdmissionOperationService.perform` (`src/specify_cli/sync/admission_operations.py:333`) and `SaasAdmissionClient` (`src/specify_cli/saas_client/admission.py`) exist but are dead from an operator's POV.
2. **All live paths are read-only.** `ProjectDeliveryTargetRegistry(...).get_current(unit)` is the only usage in `sync/background.py:851,1051`, `sync/runtime.py:477`, `cli/commands/sync.py:586` (dispatch) — plus `list_targets` in `gc`. None register/admit.
3. **DB proof.** `~/.spec-kitty/projects/<uuid>/sync/sync.db` → `select count(*) from project_target_admissions` = **0**, even though the project is already `shared` to a team (`sync routes` shows `shared`). `admission_operations` outbox is likewise empty.
4. **Gate location.** `delivery/status_report.py:634` appends `admission_not_current` whenever `context.admission_state != AdmissionState.ADMITTED`; `sync/background.py:847` and `sync/history_import/pipeline.py:203` withhold delivery on the same condition.
5. **Documented caveat.** `docs/architecture/team-kitty-saas.md` reading-notes already state: *"at the time this page was written no traced CLI command path actually invokes `.perform()` — live command paths only read an already-registered target via `ProjectDeliveryTargetRegistry.get_current()`."* This issue tracks closing that gap.

## Impact

- `sync now` prints `Event sync gated: admission_not_current` and delivers 0 events regardless of queue depth (observed: 1,000 events queued 2+ days, `Last Sync: Never`).
- The misleading secondary line `not authenticated: no valid access token. Run 'spec-kitty auth login'` is emitted even though `auth status` is fully healthy (valid access + refresh tokens, active daemon) — the real blocker is admission, not auth. The auth remediation is a red herring.
- `sync import-history --apply` (#2262) can never satisfy its `admission_state is ADMITTED` precondition (`history_import/pipeline.py:203`), so history import is also dead-ended client-side — independent of the server materializer gap (#795/#585).

## Repro

```bash
# From an opted-in, authenticated, SaaS-armed checkout whose routes show `shared`:
spec-kitty sync routes         # connects, project shared to team
spec-kitty sync now            # -> "Event sync gated: admission_not_current" ; delivered=0
sqlite3 ~/.spec-kitty/projects/<uuid>/sync/sync.db \
  'select count(*) from project_target_admissions;'   # -> 0
```

## Suggested fix

Wire `AdmissionOperationService.perform` into a live path so the delivery target is registered/admitted before the drain gate is evaluated — the natural call sites are `sync share`, `sync opt-in`, and/or a lazy admit at the head of `sync now`'s 3-gate drain (matching step 4 of the `team-kitty-saas.md` sequence: build audience → persist PREPARED target → `PUT sync-admission/` → record ACKNOWLEDGED). Until then, either:
- surface `admission_not_current` with an accurate, actionable remediation (not the auth red herring), and/or
- gate `import-history`/`sync now` behind an explicit "admission not wired / server can't admit yet" capability check that fails fast.

## Related

- #3406 (first-sync gauntlet umbrella — this is the concrete client-side root cause for its `admission_not_current` symptom; gate 6 SNAPSHOT_DRIFT and gate 9 import are adjacent)
- #795 / #585 (server-side import admission/materializer — the server half)
- #2262 (`import-history`)
- `docs/architecture/team-kitty-saas.md` (step 4 admission contract + the "no call site" caveat)

_Filed from a live activation attempt on `stijn-dejongh/spec-kitty` @ `3.2.6rc2`. Sync reached enabled + authenticated + routes-connected; delivery blocked solely at this gap._

### LINKED ISSUE #3620: SaaS event sync never delivers: project admission never initiated (no production caller for AdmissionOperationService.perform) in 3.2.6rc3
## Summary

SaaS event sync never delivers because project **admission** is never initiated. In `3.2.6rc3`, `AdmissionOperationService.perform` (`src/specify_cli/sync/admission_operations.py:333`) and `AdmissionClient.admit` (`src/specify_cli/saas_client/admission.py`) have **no production caller** — only tests invoke them. As a result:

- `project_target_admissions` / `admission_operations` stay empty for every project,
- `DeliveryTargetRegistry.get_current` (`delivery/targets.py:120`) returns `None`,
- `sync/sync_dispatch_exec.py:230-232` gates every cycle with `admission_not_current`,
- `delivery_results` is **0** across all projects despite thousands of queued journal entries and hundreds of queued body uploads.

SaaS event delivery has, in effect, never worked on this machine. The root cause is version-wide (a missing invocation wiring), not machine-specific. Likely origin: commit `cd3d6a91d2` "Per-project sync consent ledgers — CLI half (#3293)" — the "CLI half" naming implies the invocation/other-half wiring is a separate, not-yet-landed work package.

## Environment

- **CLI version:** `3.2.6rc3` (installed wheel + matching core source tree).
- **Platform:** Linux.
- **Interpreter:** pyenv `/home/stijn/.pyenv/versions/3.13.12/bin/python`.
- **Reproduced on:** repo `spec-kitty/private-packs`, project UUID `e7939920-6e08-4e04-9259-ce1920e4dd5d`. Root cause is version-wide, not specific to this machine/project.
- **Server URL:** `https://app.spec-kitty.ai` (in `~/.spec-kitty/config.toml`).

## Steps to reproduce

```bash
# 1. Opt in to hosted SaaS sync for a consented project
spec-kitty sync opt-in            # consent granted successfully

# 2. Attempt to deliver queued events
spec-kitty sync now
# Prints:
#   Event sync gated: admission_not_current
#   not authenticated: no valid access token. Run `spec-kitty auth login`.

# 3. Ground-truth check against the local sync store (read-only).
#    For each project's sync.db, count admissions vs deliveries:
sqlite3 "file:$HOME/.spec-kitty/projects/<uuid>/sync.db?mode=ro&immutable=1" \
  "SELECT
     (SELECT COUNT(*) FROM project_target_admissions) AS admissions,
     (SELECT COUNT(*) FROM admission_operations)       AS admission_ops,
     (SELECT COUNT(*) FROM journal_entries)            AS journal_entries,
     (SELECT COUNT(*) FROM delivery_results)           AS delivery_results;"
# → admissions=0, admission_ops=0, journal_entries=<thousands>, delivery_results=0

# 4. Prove foreground auth is actually valid (rules out the printed message):
spec-kitty auth status            # Authenticated (token ~53 min, refresh 89 days)
spec-kitty auth doctor --server   # Server Session active
spec-kitty sync status --check    # WebSocket Connected, Auth Authenticated, "authentication valid"

# 5. Minimal-import auth probe under the daemon's own import mode
#    (must be run with a specify_cli-capable interpreter):
SPEC_KITTY_SYNC_MINIMAL_IMPORT=1 /home/stijn/.pyenv/versions/3.13.12/bin/python work/auth_probe.py
# → is_authenticated: True, storage.read() OK, refresh not expired
```

## Expected behavior

After `sync opt-in` on a consented project with a resolvable target audience, some step in the sync lifecycle should initiate admission (call `AdmissionOperationService.perform(action=ADMIT, ...)`), writing an `admitted` row into `project_target_admissions`. `get_current` then returns a delivery target, and `sync now` (and the daemon drain cycle) delivers queued journal entries, populating `delivery_results`.

## Actual behavior

Admission is never initiated for any project. `get_current` always returns `None`, every delivery attempt is gated with `admission_not_current`, and `delivery_results` stays at 0 forever. Re-authenticating (`auth login --force`) and restarting the daemon do not help, because auth is not the actual cause.

## Root cause analysis

- `sync now` → `get_or_open` (`sync/sync_runtime.py:158`) reads `delivery_target = registry.get_current(unit)`.
- `get_current` reads table `project_target_admissions` (`delivery/targets.py:120`).
- That table is only ever written by `DeliveryTargetRegistry.register(...)` (`delivery/targets.py:131`).
- The only caller of `register` is `AdmissionOperationService._prepare` (`sync/admission_operations.py:163-164`), reachable only via `AdmissionOperationService.perform` (`sync/admission_operations.py:333`).
- **`AdmissionOperationService.perform` and `AdmissionClient.admit` have no production caller** — confirmed in both the installed wheel and the core source tree:

```
# installed wheel
grep -rn "AdmissionOperationService|\.perform(|\.admit(" .../site-packages/specify_cli
  → only definitions in sync/admission_operations.py and saas_client/admission.py

# core source
grep -rln "AdmissionOperationService" src/    → src/specify_cli/sync/admission_operations.py (definition only)
grep -rln "AdmissionOperationService" tests/  → tests/sync/test_admission_operations.py
                                                 tests/architectural/test_egress_consent_boundary.py
```

Nothing in the daemon cycle, `sync now`, `sync opt-in`, or `sync share` ever enqueues an admission operation or POSTs to the SaaS admission endpoint (`/api/v1/sync/projects/{uuid}/sync-admission/`, `saas_client/admission.py:20`). It is exercised only by tests.

When `delivery_target is None`, `sync/sync_dispatch_exec.py:230-232` prints `Event sync gated: admission_not_current` and delivers nothing.

**Likely origin:** commit `cd3d6a91d2` "Per-project sync consent ledgers — CLI half (WP01–WP10) (#3293)". The "CLI half" naming strongly suggests the invocation wiring (who calls `perform`) and/or the SaaS server side is a separate, unlanded work package.

**Please confirm:** is the `/api/v1/sync/projects/{uuid}/sync-admission/` endpoint deployed for `app.spec-kitty.ai`?

## Evidence

Per-project admission/delivery counts, read directly from each project's `sync.db` (read-only), captured 2026-08-21:

| Project UUID | admissions | admission_ops | journal_entries | body_upload_tasks | delivery_results |
|---|---|---|---|---|---|
| 1adb35ef-246b-4a26-af85-9f87687944b9 | 0 | 0 | 307 | 0 | 0 |
| 22bbc978-d7f7-493b-bd96-a2a72efeedfa | 0 | 0 | 4211 | 60 | 0 |
| 519dafb4-f6c7-4954-b698-bcb8cd529343 | 0 | 0 | 2 | 0 | 0 |
| 8a4a7da6-a97c-4bb4-893a-b31664abfee4 | 0 | 0 | 12078 | 269 | 0 |
| bebf8d60-7c5d-4459-94b6-e0f3e4ad83f9 | 1 | 0 | 583 | 0 | 490 |
| c5970476-a956-4ffe-aa95-38643e6e81fb | 0 | 0 | 8 | 0 | 0 |
| e7939920-6e08-4e04-9259-ce1920e4dd5d (private-packs) | 0 | 0 | 2 | 0 | 0 |

Every project reachable through the current CLI path has `admissions=0`, `admission_ops=0`, and `delivery_results=0` — despite two projects accumulating >12k / >4k journal entries and hundreds of queued body uploads over weeks. (The single outlier `bebf8d60` shows `admissions=1` / `delivery_results=490` but `admission_ops=0`, i.e. an admission row that predates the current code path — no admission operation was ever recorded for it, consistent with the current CLI never initiating one.)

## Secondary observations

These are real but distinct from the primary bug; noted so they are not mistaken for the blocker.

1. **Misleading "not authenticated" error (UX).** `sync now` prints `not authenticated: no valid access token. Run \`spec-kitty auth login\`` when the real cause is `admission_not_current`. Foreground auth is provably valid (`auth status`, `auth doctor --server`, `sync status --check` all green; the minimal-import token-manager probe returns `is_authenticated: True`). The message is the strict exit classifier's catch-all (`sync/sync_dispatch_core.py:53-54`) returned when `delivery_target is None`. This cost ~30 min of pointless re-login/daemon-restart churn. **Suggestion:** when `delivery_target is None`, surface `admission_not_current` (and its remediation) as the exit reason instead of the auth catch-all.

2. **Daemon spawned with the wrong interpreter (intermittent).** `_spawn_sync_daemon_process` (`sync/daemon.py:1307-1331`) launches `subprocess.Popen([sys.executable, "-c", ...])`. When `sys.executable` resolves to an interpreter without `specify_cli` (e.g. a Homebrew `python3` ahead of pyenv on PATH), the child dies immediately with `ModuleNotFoundError: No module named 'specify_cli'` — visible as `<string>` line-3 tracebacks in `~/.spec-kitty/sync-daemon.log`. The daemon launched via the pyenv-shimmed `spec-kitty` (or `doctor restart-daemon`) runs the correct interpreter, so this is intermittent and depends on who spawns it. **Suggestion:** resolve/validate a `specify_cli`-capable interpreter before spawn, or re-exec/guard when the import is unavailable, rather than trusting bare `sys.executable`.

3. **Noisy "project sync store is locked" (log noise).** SQLite `BEGIN IMMEDIATE` busy translated to `ProjectStoreLockedError` (`sync/project_store.py:589-592`, acquisition at `:633`) — **not** an OS-level lock (`lsof` on `sync.db` shows no holder; `sync.db-wal` is 0 bytes; a read-only `?mode=ro&immutable=1` open succeeds instantly). Reproduces only on the large stores (55 MB `8a4a7da6`, 22 MB `22bbc978`), never the small ones — consistent with in-daemon write-lock self-contention whose window scales with store size. It fails the whole daemon cycle and backs off, making the log look like the primary fault. Secondary because even with locks cleared, nothing would deliver (primary bug).

## Suggested direction

1. **Primary:** wire an admission initiation into the sync lifecycle so a consented project with a resolvable target audience gets `AdmissionOperationService.perform(action=ADMIT, ...)` invoked — e.g. at the start of a daemon drain cycle, or on `sync opt-in` / `sync share` — so `project_target_admissions` gets an `admitted` row and `get_current` returns a target. Confirm whether the SaaS `/api/v1/sync/projects/{uuid}/sync-admission/` endpoint is deployed for `app.spec-kitty.ai`.
2. Surface `admission_not_current` instead of the auth catch-all when `delivery_target is None` (observation 1).
3. Validate a `specify_cli`-capable interpreter before daemon spawn (observation 2).
4. Reduce `BEGIN IMMEDIATE` self-contention / soften the failure so a lock on one large store does not fail the whole daemon cycle (observation 3).

stijn-dejongh:
## Summary

Restores hosted event-sync delivery for consented projects. `spec-kitty sync now` and `import-history --apply` have been unable to deliver anything to `app.spec-kitty.ai`: commit `cd3d6a91d2` (#3293, "consent ledgers — CLI half") shipped a **hard, unconditional** client-side admission gate whose writer (`AdmissionOperationService.perform`) has no production caller and whose server endpoint (`PUT /api/v1/sync/projects/{uuid}/sync-admission/`) **is not deployed** (404 in prod; it lives only on the still-open, unmerged SaaS PR Priivacy-ai/spec-kitty-saas#795). Every consented project therefore has `delivery_target = None`, `sync now` gates on `admission_not_current`, and `delivery_results` is 0 despite thousands of queued journal entries.

This contradicts the paired SaaS program's *own* negotiated-rollout design (`project-sync-admission-boundary` D-009 / US4-#6: "strict admission activates only after compatible core is proven"). The fix completes that negotiation on the client side.

Fixes #3564. Closes #3620 (duplicate re-report; supplied the multi-project `sync.db` census that confirmed 0 deliveries across 7 projects).

## What changed

**1. Negotiated admission gate (restores `sync now`).** A capability resolver (`sync/admission_negotiation.py`) defaults **non-strict → deliver**, flipping to strict only on an explicit signal (`SPEC_KITTY_SYNC_STRICT_ADMISSION` / `[event_sync] strict_admission`, or a future server handshake key — a dormant hook for when #795 lands). When non-strict, a **labeled local self-admission** (`binding_audience = "local-nonstrict:<origin>"`) is minted for a consented, authenticated project, so the existing egress path proceeds unchanged. The admission client stays dormant; when #795 deploys and advertises strict admission, a real `perform()` supersedes the local label with no client rework.

**2. Finding 2 — correct exit reason.** `sync now` gated for admission now surfaces the real reason instead of the misleading `not authenticated: no valid access token`. Genuine 401/403 still reports auth failure.

**3. #3581 — preflight contract (restores `import-history --apply`).** The CLI required a correlated per-event `results[]` that the deployed preflight does not send, so a server success (`{accepted: true}`) was classified as a rejection. The client now honors `{accepted: true}` with no `results[]`, while staying strictly fail-closed on non-200, `accepted:false`, or any `results[]` present but incomplete/uncorrelated/malformed.

**4. #3582 — surface real server diagnostics.** Preflight rejections now prefer the server's structured diagnostic (`error`/`category`/`code` + first `details[]` entry) over the always-zero reconciliation counters (preflight is non-mutating, so those counters are structurally zero).

## Safety / invariants

- **A self-labeled proof on the wire — never a forged one.** The `local-nonstrict:<origin>` label a locally-admitted project carries *is* transmitted with each delivery (via `attach_admission_proof`, same as any admission proof) — but it is honestly self-identifying as a *local* self-admission, never poses as a server-issued audience, and is inert against the current non-strict server (which ignores it while the consent gate independently governs egress). A strict #795 server correctly *rejects* it at ingest (degrade, not corruption) rather than being fooled — precisely because it can see the `local-nonstrict:` label. (An earlier draft of this bullet claimed the label "is never transmitted", which was wrong and self-contradictory with the #795-handoff note below; corrected during the landing pass, and pinned by `TestLocalAdmissionWireContract`.)
- **Consent boundary preserved.** A not-consented or unauthenticated project is never locally admitted (`test_egress_consent_boundary` stays green with no new allowlist entry; the SQL-write inventory is extended for the new writer).
- **Forward-compatible.** A real server admission can never be overwritten by a local mint (the mint short-circuits on already-ADMITTED and on any strict signal).

## Tests

Red-first throughout. New/extended: `test_admission_negotiation.py` (resolver + guard matrix + direct idempotency), `test_sync_dispatch_exec.py` (consented project delivers end-to-end against a simulated non-strict server, no `admission_not_current`), `test_sync_dispatch_core.py` (Finding 2 + a both-directions regression guard: genuine 401 still routes to auth), `test_history_import_upload.py` (full fail-closed matrix + structured-diagnostic surfacing + accepted-without-results end-to-end + retry-does-not-crash), and the egress SQL-write inventory. Independently verified: touched-module tests green, `ruff`/`mypy` clean on new code, dead-symbols + egress + terminology gates green.

## Honest scope / limitations

- This restores the CLI's ability to **reach** delivery. A live `delivery_results > 0` count against prod is **not asserted** here because the deployed batch ingest independently validates event shape: ~13% of a historical corpus (71/545) is server-rejected by strict transition semantics (#3017/#3001), and there is separate event-schema drift (`payload_schema_invalid`). Those are out of scope and tracked separately.
- Deferred, filed separately: daemon wrong-interpreter spawn (Finding 3, #3624), SQLite `BEGIN IMMEDIATE` self-contention (Finding 4, #3625).
- SaaS #795 handoff: when the server endpoint deploys and advertises strict admission, enable strict via env/config or wire the handshake capability into `maybe_admit_locally`'s resolver call; otherwise a strict server gracefully rejects `local-nonstrict:` admissions at ingest (degrade, not corruption).

## Sonar notes (for reviewer)
- `admit_locally` and `register` share an `INSERT … ON CONFLICT` shape (divergent admission-state semantics; a shared upsert helper would satisfy S-duplication if flagged).
- Two pre-existing `no-any-return` mypy findings in `upload.py` (`disclosed_event_payload_bytes`, `receiver.deliver`) are baseline on `upstream/main`, not introduced here.
- One defensive `_preflight_classification` else-branch (200 + no results + not accepted — not a deployed shape) is covered indirectly via `accepted()` unit tests, not a direct assertion.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3626"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789932952&installation_model_id=427282&pr_number=3626&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3626&signature=1b5fb261add9a03ca573f5e42eae1b4f3a93c76c2af74c75c6691a41ef2bc0a3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


stijn-dejongh:
🛬 **Maintainer landing pass — claiming.**

Picking this up for a landing pass: branch is already on current `upstream/main` tip (no rebase needed, 3 ahead / 0 behind). Two red checks to adjudicate — `arch-adversarial (arch_shard_2)` and `fast-tests-core-misc (core-misc)`. Classifying each into the four bins now, folding fixes as `<type>(landing):` commits. Evidence to follow.

Self-authored fresh substantive fix (restores hosted event-sync delivery, #3564/#3620/#3581/#3582) — landing viability is clear; proceeding straight to remediation.

stijn-dejongh:
## 🛬 Maintainer landing pass — remediation summary

**Verdict: LAND** (green-modulo external-URL link warnings). Fresh, substantive, self-authored fix; branch was already on current `upstream/main` tip (no rebase). Four landing folds pushed; the operator merges.

### Red-check classification
Both red checks — `arch-adversarial (arch_shard_2)` and `fast-tests-core-misc (core-misc)` — were the **same single defect**: `tests/architectural/test_sync_env_census.py::test_env_reference_set_is_frozen_exactly`. The new `SPEC_KITTY_SYNC_STRICT_ADMISSION` env var lands on the sync surface; the frozen census guard grows-and-reds on any new reference. Classified **bin 2 — a contract the PR legitimately crosses** → re-pinned in this PR with a dated rationale.

### Folds (each a single-purpose `<type>(landing):` commit)
1. `test(landing)` **register `SPEC_KITTY_SYNC_STRICT_ADMISSION` in the sync env census** — added to the frozen `_EXPECTED_ENV_REFS` contract + a `live`-verdict census row (#8) in `docs/plans/code-quality/sync-env-census.md` + regenerated the docs retrieval index (the `## Census` heading anchor changed). Clears both reds.
2. `fix(landing)` **evaluate the admission handshake live so the origin cache can't stale-mask it** — `server_requires_strict_admission` memoized per origin but the handshake is a *per-call* signal; a `handshake=None` verdict poisoned the cache so a later strict handshake for the same origin returned stale non-strict. Now the handshake is evaluated live ahead of the cache; only the process-stable env/config inputs are memoized. Today's sole caller passes `handshake=None`, so **behaviour is unchanged** — this hardens the dormant #795 hook before it goes live. (Surfaced by the SSOT review lens; red-first: the new regression test fails on pre-fix code, passes with the fix.)
3. `docs(landing)` **correct the "never transmitted" admission-proof claim in the changelog** — see security finding below.
4. `test(landing)` **pin the local self-admission wire contract** — `TestLocalAdmissionWireContract` locks that the transmitted proof is honestly self-labeled (`local-nonstrict:` prefix present), never a forged server proof.

PR body also corrected (the Safety bullet) for the same wire-behaviour finding.

### Adversarial squad (2 profile-loaded lenses, read-only)
**SSOT / architecture lens — no MAJORs; design sound.** Confirmed the `admit_locally` vs `register` "duplication" is a *justified state-semantics divergence* (pending-registration vs labeled-admitted), not a missing helper; the fail-closed preflight matrix is correct across every traced case (incl. the defensive `_preflight_classification` else-branch — reachable and correct); complexity clean (no function > 15); new module correctly consumes existing seams. Findings: **N1** (handshake memoization) → **folded** (#2 above). **M1** (SQL column-list/`ON CONFLICT` boilerplate in `delivery/targets.py`) and **N2** (config-key vs env-var surfacing asymmetry) → **accepted as disclosed** (M1 is a borderline-S1192 cosmetic the PR body already flags for Sonar; the SSOT lens explicitly warns a shared *behavioural* upsert helper would be wrong given the divergent semantics — only the SQL literal is shareable).

**Security / wire-safety lens — one MAJOR, adjudicated as a narrative defect (no code change).** The lens refuted the Safety-section claim "the `local-nonstrict:` label … is never transmitted / no forged proof on the wire". **I verified the trace independently:** `admit_locally` writes `binding_audience="local-nonstrict:<origin>"` (`delivery/targets.py:207`) → loaded into the context (`project_store.py:765`) → `delivery/dispatcher.py:404-409` builds the proof and `attach_admission_proof` (`saas_client/admission.py:320`) embeds it into the wire item. The label **is** transmitted.
- **Adjudication:** this is a *narrative* error, not a code defect. The PR's own #795-handoff note ("a strict server gracefully rejects `local-nonstrict:` admissions at ingest — degrade, not corruption") **requires** transmission; the label is honestly self-identifying (never a forged server audience), inert against the non-strict server, and cleanly rejectable by a strict one. Stripping it would be *worse* (it's what lets a strict server distinguish and reject). Remediation = reconcile the two contradictory sentences (folds #3, #4, + PR-body fix), not touch the code.
- Other security findings: **Claim 4** (env/config truthy parse defaults unknown values to non-strict) — adjudicated **not a defect**: non-strict-on-unrecognised is the *correct* safe default for this PR's thesis (a fat-fingered env must not silently re-break the very projects being un-broken; a strict server still rejects at ingest). **Claim 2 NOTE** (auth guard proxies on session presence, not token validity) — harmless: delivery fails server-side on a bad token and consent is independently enforced. Claims 3 & 5 (no admission overwrite; no cross-origin cache mis-admit) — **hold**.

### Verification (local, on the rebased tip)
- **Red-first (bug #3581):** pre-fix `upload.py` fails exactly the 6 new preflight-contract tests; PR version passes all 37. ✔
- **Red-first (N1 fold):** new handshake test fails on pre-fix code, passes with the fix. ✔
- Touched-module sync tests: **142 passed / 2 xfail**; census + terminology + docs-freshness (`errors=0`) + markdownlint(changelog) + commitlint(all 7 commits) clean; `ruff` clean on all changed code.
- **Pre-existing, not this PR (disclosed):** two `no-any-return` mypy findings in `upload.py` reproduce verbatim on `upstream/main` and mypy is an **advisory** CI step; two `MD049` markdownlint findings on the `implement --base` changelog entry pre-exist on `main` (my edits add zero). The project-store census *shrink* is an expected ratchet **warning** (the guard only reds on growth).

### State
Watching CI on the pushed fold stack; **merge-ready on green.** Handing off — operator merges.

LynnColeArt:
@stijn-dejongh you're not blocked on saas. #795 is a forward-compat hook here, not a dependency, and I'd land this without it.

**Why this stands alone.** `maybe_admit_locally` routes to `ProjectDeliveryTargetRegistry.admit_locally`, which is one upsert into the local `project_target_admissions` table. No network call, and the admission client stays dormant exactly as your module docstring says. Nothing in the 16 files reaches for the undeployed endpoint.

**Why #795 landing won't re-break you.** I checked the branch rather than the body: `PROJECT_SYNC_ADMISSION_STRICT = env.bool("PROJECT_SYNC_ADMISSION_STRICT", default=False)` (`spec_kitty_saas/settings.py:677`), read through one canonical helper (`apps/sync/compatibility.py:225`), and there is no capability handshake anywhere on that branch advertising `admission.required`. Your resolver stays non-strict in either merge order. One note for whoever lands #795: its body says the flag is "deliberately absent from settings", but it is present and defaulted `False`. Same outcome, stale wording.

**Your #3581 fix matches the deployed contract rather than working around it.** The published 200 schema for `preflight_sync_events` is `{accepted, event_count, reconciliation}`, with no `results[]` member at all. The CLI was requiring a field the server never advertised, so honoring `{accepted: true}` *is* the contract. Same for #3582: `ImportReconciliationSerializer` documents the five materialization outcomes as null-when-unmeasured, and preflight is non-mutating by design, so preferring `error`/`category`/`code` over those counters reads the endpoint as built.

**What is actually still on us.** Not admission. It's what your "honest scope" paragraph points at: `validate_live_event_ingress(events)` (`apps/sync/serializers.py:34`) takes no profile argument, so `import-history` replays historical events through live-ingress guards that postdate them. That's your 71/545, and #3001's 3,391 transitions. I've filed it on saas as Priivacy-ai/spec-kitty-saas#1095 with the invariant-vs-live-only guard classification spelled out, cross-linked to #3017 and #3001, and I own it. That's the real gap between "reaches delivery" and `delivery_results > 0`.

On #795 itself: 508 files, ~49.7k additions, currently conflicting with `develop`, WP09 T040 to T043 blocked on external evidence. Don't sequence anything behind it.

Checks are green across the board here. Merge when you're ready.

Issue #3259, opened 2026-08-08
Epic: meta.json fail-closed read routing — finish routing the residual bypass reads onto the canonical seam


stijn-dejongh:
## Why
`meta.json` is the mission's canonical identity and VCS-lock record. Reads that bypass the canonical fail-closed loader (ad-hoc `json.loads`, `git show` stdout, `show_blob` bytes) silently accept a corrupt, truncated, or wrong-authority `meta.json` instead of failing loud — exactly the split-brain / wrong-authority failure mode the metadata-authority epic exists to close. The `meta-fail-closed-3162` mission routed one of five bypass sites and left four diagnosable-only; this slice finishes the job.

## For whom
Maintainers and agents on the git ref-advance, `implement`, and merge-driver code paths, plus anyone relying on the `tests/architectural` inline-meta-read gate to prove no un-routed reads remain.

## Intended effect
Every `meta.json` read flows through the fail-closed loader or a shared L1 pure-decode primitive (`text|bytes -> dict|None`); a corrupt file fails loud rather than being silently absorbed; and the inline-meta-read architectural gate reaches **0 un-routed / 0 un-allowlisted** bypass reads with the routed floor re-derived from a live count.

stijn-dejongh:
All residual bypass reads routed onto the canonical fail-closed seam; #3228/#3229/#3230/#3240 delivered by PR #3319 (merged 2026-08-11). Closing the epic as completed.

PR #3645, opened 2026-08-22
refactor(merge): route baseline.py committed-meta reader through kernel L1 decode

### LINKED ISSUE #3330: Follow-up (#3259): route merge/baseline.py committed-meta reader through kernel L1 decode seam
Follow-up from the landing pass on #3319 (epic #3259, meta.json fail-closed read routing). A bounded adversarial SSOT + architecture squad reviewed the final diff; **no blockers** were found. This issue captures the one out-of-scope MINOR and two cosmetic NOTEs so the queue stays clean.

## MINOR — a second, gate-invisible committed-`meta.json` decoder (pre-existing)

`src/specify_cli/merge/baseline.py` → `_read_committed_meta_json` decodes `git show <target>:kitty-specs/<slug>/meta.json` with `json.loads` plus its own `json.JSONDecodeError` catch and `isinstance(dict)` check — an independent re-implementation of the malformed contract.

- **Pre-existing on `main`; unchanged by #3319** (verified via `git show`).
- Outside the mission's declared 5-site routing scope, and structurally outside the FR-010 gate's reach (the gate deliberately disclaims an already-read raw string with no meta-path provenance).
- Consequence: #3259's SSOT outcome is precisely "one **gate-visible** decode authority"; `baseline.py` is a literal second one.

**Proposed:** route `_read_committed_meta_json` through `kernel.meta_decode.decode_meta` (or `parse_meta_file`) so the malformed contract has a single owner tree-wide, and evaluate whether FR-010 can be extended to cover `git show`-blob reads that carry slug provenance.

## NOTE (cosmetic) — stale allow-list locator

The `inline_meta_read` allow-list entry for `charter/mission_type_profiles.py` records `line: 441`, but the live `_read_meta_mission_type` site is ~line 690. The `line:` field is explicitly non-authoritative (token + qualname carry the match; gate is green), so this is locator hygiene only — refresh when convenient.

## NOTE (cosmetic) — ref_advance boundary test has no dedicated negative

`TestRefAdvancePlumbingBoundary` relies on the shared `_is_specify_cli_module` helper + a real-file AST scan rather than a dedicated plant-and-catch negative. It is non-vacuous and adequate as-is; a dedicated negative would be consistent if the other layer guards are later hardened.

---
Parent epic: #3259. Filed during the #3319 landing pass.

miao-agent-ctrl:
Closes #3330.

`_read_committed_meta_json()` in `src/specify_cli/merge/baseline.py` decoded the `git show <target>:meta.json` blob with its own `json.loads`/`isinstance(dict)` pair — a second, gate-invisible re-implementation of the malformed-JSON contract that #3259/#3319 consolidated into `kernel.meta_decode` everywhere else (as flagged in the #3330 follow-up issue).

**Before:** two independent decoders of the same malformed contract could drift on what counts as malformed, and this site sits outside the FR-010 gate's reach.

**After:** the same `git show`-blob text is decoded via `kernel.meta_decode.decode_meta(out, on_malformed="raise")`, matching the sibling git-plumbing site (`src/specify_cli/git/ref_advance.py`) that already does this. The two distinct `BaselineMergeCommitError` message shapes existing callers/tests depend on ("... is not valid JSON (...)." vs "... is not a JSON object.") are preserved verbatim by branching on the `MetaDecodeError` message text.

**Verification:**
- `tests/specify_cli/merge/test_baseline_module.py` — 23 passed, unchanged (no test edits needed; existing tests already pin the exact message shapes).
- `ruff check` clean on the changed file.
- `ruff format --check` reports the same pre-existing formatting gaps as `origin/main` (verified via `git show origin/main:...` diff) — not introduced by this change.
- Whole-tree `mypy --strict src/specify_cli src/charter src/doctrine`: 54 errors before and after (via `git stash` comparison) — no net change.
- Added a changelog entry under `[Unreleased] - 3.2.6rc3` → `### ♻️ Changed` in `docs/changelog/CHANGELOG.md`.

No behavior change for callers.


miao-agent-ctrl:
This fix has been folded into #3659 (maintainer-side consolidation of reviewed community PRs), which was merged into `main`. Closing this PR as superseded — no further action needed here. Thanks for the review!

Issue #3398, opened 2026-08-14
record-analysis bakes absolute home paths into committed analysis-report.md — 171 files on main leak 7 contributors' local layouts


MOES-Media:
## Summary

`record-analysis` writes **absolute filesystem paths** into `analysis-report.md`, which is then committed. On a public repository this publishes every analyzer's local directory layout — including their home-directory name — permanently and indexed.

This is not hypothetical or rare: **171 committed `analysis-report.md` files on `main` currently contain an absolute home path**, across **7 distinct developer/CI home directories** (`/home/…` and `/Users/…` forms). Every mission that has ever run analyze has leaked the author's local path.

## Mechanism (verified on `main`)

`collect_input_artifact_hashes` (`src/specify_cli/analysis_report.py:208-217`) stringifies unresolved absolute paths:

```python
def collect_input_artifact_hashes(feature_dir: Path, repo_root: Path) -> dict[str, dict[str, str | None]]:
    inputs = {name: _artifact_hash_entry(feature_dir / name) for name in _HASH_INPUTS}
    charter = _charter_path(repo_root)
    inputs["charter"] = {"path": str(charter) if charter else None, ...}
    return inputs
```

`_artifact_hash_entry` records `str(path)` for each of spec.md / plan.md / tasks.md, and the charter entry does the same. The resulting frontmatter looks like:

```yaml
input_artifacts:
  spec.md:
    path: /home/<user>/dev/<their-checkout-layout>/kitty-specs/<mission>/spec.md
    sha256: daba63429645d3c343859b5ca0989424ef273f1f834e86ebb2f5d06c80c1cbcd
```

Note the function already receives `repo_root`, so the repo-relative form is available at the call site with no new plumbing.

## Why it matters

- **Public repo.** Committed artifacts are permanent and indexed. Publishing contributors' home-directory names and local checkout layouts is an avoidable privacy and information-disclosure leak, and it is happening for every analyze run by every contributor.
- **The paths are also useless to a reader.** They only resolve on the machine that produced them; a repo-relative path is strictly more useful to everyone else.
- **It defeats artifact comparability.** Two runs of the same mission on different machines produce diffs in the `path:` fields that carry no semantic meaning, which is noise in review and in any future artifact-diffing tooling.

## Suggested fix

[...]

PR #3293, opened 2026-08-10
Per-project sync consent ledgers — CLI half (WP01–WP10)

robertDouglass:
## What changes for you

**Hosted sync is now explicitly per-project, and no project can ever see another project's data.**

Before, every project on a machine shared one sync journal, ledger, and queue. A filter defect could deliver *another project's* rows to the wrong place — the #3030 consent incident — and hosted sync could be switched on without a clear, per-project decision. Separately, events that were consented *before* login could strand in the backlog forever, terminally unselectable.

After this mission (CLI half of the #3262 consent program):

- **Each project gets its own physically separate sync store**, owned by one immutable `project_uuid`. No other project's operation can open it — cross-project delivery is structurally impossible, not just filtered.
- **Consent is an explicit, local, per-project grant.** `spec-kitty sync opt-in` is the only thing that grants it; it is offline-capable and is *never* inherited from your login, the target URL, a slug, a path, or an env var. `SPEC_KITTY_ENABLE_SAAS_SYNC` is strictly deny-only — setting it can only *withhold* sync, never enable it.
- **Opt-out is an immediate barrier that seals a consent epoch without deleting anything already captured** — you can stop sync without losing your audit trail.
- **Consented pre-login backlogs now drain** instead of stranding forever: `drain_blocked_reason` is stamped with the canonical vocabulary so those rows become selectable again.
- **Operators migrate legacy shared state with copy-only, resumable commands** (`sync project-store-preview` / `-migrate` / `-status` / `-quarantine` / `-history`). The old shared-store `sync migrate` now refuses with guidance rather than mixing stores.

See the new guide: [Per-Project Sync Consent](docs/guides/project-sync-consent.md).

## Scope of this PR (incremental land)

This lands **WP01–WP10 of the mission — implemented and each independently reviewed/approved.** It is the CLI half of a paired program (SaaS `#585` / `spec-kitty-saas#795`).

**Deferred to a follow-up PR — WP11 (coordinated core acceptance + immutable evidence):** the cross-repo attestation against the final SaaS candidate, the six-project omission proof, stale-generation and cross-platform proofs, and the evidence manifest. The mission analysis report is recorded `ready` (0 critical/high) for the WP01–WP10 slice.

## Product fixes folded in during the mission

- `EventEmitter` stamps `drain_blocked_reason` via the canonical `event_journal.models` vocabulary (`classify_drain_blocked_reason`) — previously every stamped row was terminally unselectable, stranding consented pre-login backlogs.
[...]

Issue #3284, opened 2026-08-10
main full suite has 23 untracked failures and 2 errors after bootstrap prewarm


robertDouglass:
## Baseline

Immutable base: `28ae75ea998c898aba57364db7a06d2088bd2af2` on `main`.
No mission implementation changes were present. The shared test venv was manually prewarmed only to bypass #3283; tracked files remained unchanged.

Command:

```bash
PWHEADLESS=1 .venv/bin/pytest tests/ -n auto --dist loadfile -p no:cacheprovider --durations=200 --ignore=tests/sync/test_orphan_sweep.py
```

Result: `24 failed, 37298 passed, 108 skipped, 5 xfailed, 2 errors in 1689.48s`.

One failure is already tracked by #2782. The other 23 failures and 2 errors had no matching open issue in targeted searches.

## Failure groups

- Doctrine/config activation drift: 3 failures in `tests/charter/test_config_stem_parity.py`, 4 in `tests/doctrine/drg/test_reachability.py`, 1 in `tests/doctrine/drg/migration/test_extractor_projection.py`.
- Runtime/timing: daemon self-retirement health timeout; two completion 500ms budget failures; doctor daemon restart timeout.
- Console behavior: 3 ANSI/plain-mode assertions fail.
- CI route collection: core/misc path filter probe times out after 240s.
- Mission template resolution: research/documentation/plan expect package default but receive global mission.
- Compact charter contract: minimal compact view is larger than source charter.
- Commit helper: empty safe-commit shape test fails during git commit.
- Cached test venv portability: 2 E2E tests look for `Scripts/python.exe` inside a macOS cache.
- Sync teardown errors: identity-aware and dual-write tests error during teardown.
- #2782 regression remains red with `CHARTER_PACK_CONFIG_INVALID`.

## Why pre-existing

This run used the untouched base implementation. The sanitation mission had only untracked/committed planning Markdown under `kitty-specs/`; none of the failing product or test surfaces were edited. The failures therefore precede mission implementation.

## Requested triage

Classify each group as live defect, stale/obsolete test, environment/harness defect, timing flake, or duplicate/non-causal guard. This issue is the DIR-013 evidence record; #2782 and #3283 remain their own authorities.

robertDouglass:
Spec Kitty mission assertive-test-suite-sanitation-01KZME3P is now the active sanitation workflow for this issue. Its evidence ledger will record the terminal disposition before PR merge.

[...]

Issue #3127, opened 2026-07-31
ci: fast-tests-status has been silently skipped for 4+ runs by a job-dependency gate, not a path filter


MOES-Media:
## Summary

`fast-tests-status` has not executed on `main` for at least **four consecutive `CI Quality` runs**, and nothing reports it as a gap. It is not path-filtered out — it is gated on another job's result:

```yaml
needs: [changes, fast-tests-sync]
if: ... && needs.fast-tests-sync.result != 'failure'
```

`.github/workflows/ci-quality.yml:1340`.

So whenever `fast-tests-sync` fails, `fast-tests-status` is **skipped**, and `fast-tests-sync` has been failing. Observed pairing in runs `30621215287`, `30609027258`, `30601514601`, `30577005990`.

## Why this is worth an issue rather than a shrug

**It is a different skip mechanism from the one that is usually watched.** Path filters skipping a shard is expected and visible — you can reason about it from the diff. A *job-dependency* gate skipping a shard is invisible from the diff, and it makes the suppressed job's absence look like a routine "no changes here" skip.

**The suppressed job is large.** The last run in which it actually executed (`30572082605`) read `1049 passed, 228 deselected in 186.41s`. So a persistent failure in one shard has been silently withdrawing ~1049 assertions from every subsequent run's coverage, for at least four runs, with no signal that it happened.

**A green board can therefore mean less than it appears to.** The skip is reported as a skip, which reads as "nothing to do here" rather than "this was suppressed because something else was already broken". That is the same shape as a gate that reports success for having done nothing — the coverage loss is real and the report does not describe it.

## What the gate was presumably for

Almost certainly to avoid burning runner time on a dependent shard when an upstream one has already failed. That is a reasonable goal; the problem is that the mechanism is silent and unbounded — it stays suppressed for as long as the upstream stays red, and nothing escalates.

## Suggested directions (not prescriptive)

- Let it run regardless and accept the runner cost — the simplest fix, if the dependency is only about cost rather than correctness.
- Keep the gate but make the suppression loud: a step that fails, or annotates the run, when a shard is skipped for this reason rather than for a path filter.
- If the dependency is genuinely about correctness (i.e. `fast-tests-status` cannot produce a meaningful result while sync is red), state that in the workflow so the next reader does not have to infer it.

## Done when

- Either `fast-tests-status` executes independently of `fast-tests-sync`'s result, or its suppression is surfaced distinctly from a path-filter skip.
[...]

PR #3672, opened 2026-08-22
refactor(dossier): adopt canonical spec-kitty-events types and delegate payload validation

### LINKED ISSUE #1058: Legacy cleanup: split dossier queue migration from live emitter APIs
## Problem

PR #1056 fixed Sonar findings around the dossier event migration from #1047, but it also highlighted a boundary smell: compatibility for the old flat dossier payload is now present in both places where we need it and places where we should not keep it.

A transform-on-drain path for already queued SQLite rows is reasonable. A live emitter API that continues accepting old positional shapes and repackaging them indefinitely is not.

## Current State

- `src/specify_cli/sync/queue.py` transforms legacy flat `MissionDossier*` payloads during `drain_queue()` so old queued rows do not fail forever. This is a valid bounded migration path.
- `src/specify_cli/dossier/events.py` still accepts legacy positional values and folds them into canonical payloads. PR #1056 had to preserve this after a regression in `emit_snapshot_computed` positional ordering.
- `src/specify_cli/sync/emitter.py` still carries local payload validators for the dossier event family instead of importing a canonical contract helper from `spec-kitty-events`.

## Plan

1. Move flat-envelope queue migration into a clearly named queue-migration module with tests and a removal target.
2. Make live dossier emitters canonical-only internally: namespace + artifact/content identity, no `*args`/`**kwargs` compatibility parser.
3. Add a transitional wrapper only if external callers are known, and mark it as deprecated with a concrete removal version.
4. Add a guard test that fails if production code calls dossier emitters positionally from the old flat signature.
5. Replace hand-maintained dossier validation rules once Priivacy-ai/spec-kitty-events#23 lands.

## Acceptance Criteria

- Existing legacy queued flat dossier events still drain successfully or produce a clear one-time discard/report path.
- Active CLI dossier sync emits only canonical namespaced payloads.
- No production call site depends on the old positional dossier emitter shape.
- Dossier emitter compatibility has a documented removal target rather than living as an implicit permanent contract.
- Regression coverage includes the #1056 snapshot positional-order bug until the wrapper is removed.

## References

- #1047
- #1056
- Priivacy-ai/spec-kitty-events#23

MOES-Media:
## What changes

The dossier event emitters now use the canonical `spec-kitty-events` types directly instead of a
24 KB hand-maintained copy of them that lived in this repo. Payload validation is delegated to the
package's own `validate_event()` rather than four hand-written rule dicts that could drift from the
real schema.

For an operator, nothing changes day to day — this is a correctness and maintenance change. What it
buys: the emitters can no longer silently disagree with the schema the server actually enforces, and
a whole class of "our copy says X, the package says Y" bug goes away.

Closes #1058

## Why the issue's plan is only half-implemented (deliberately)

Issue #1058 has two halves. **The queue-drain half is superseded by #3293**, which deleted
`_migrate_legacy_dossier_payload` wholesale on 2026-08-13 and replaced the legacy-queue architecture.
That code no longer exists, so there is nothing to "split out". The fail-loud
`LegacyQueueMigrationRequiredError` gate already prevents an un-migrated layout from reaching live
sync. The issue's "Current State" description of that path is stale.

This PR implements the live-emitter half, which is fully current.

## Known trade-off, stated plainly

Delegating to the canonical validator means **accepting the canonical validator's strictness, which is
looser than the local checks it replaces.** Two measured instances:

- `ContentHashRef` no longer normalises hash case (the local mirror had a `field_validator` that did).
- `snapshot_hash` / `expected_hash` / `actual_hash` now accept malformed-but-non-empty values —
  `sha1:` prefix, wrong digest length, non-hex, whitespace. Only the empty string is rejected. The
  retired local check was `^(?:sha256:)?[a-f0-9]{64}$`.

Neither is reachable from production today: every construction path feeds
`hashlib.sha256(...).hexdigest()`, which is already lowercase and well-formed. This is recorded as an
accepted consequence of adopting the canonical validator as the single source of truth, not an
oversight — the fix, if wanted, belongs upstream in `spec-kitty-events`' schema rather than as
re-vendored local validation, which the charter's shared-package-boundary rule argues against.

One test (`TestContentHashRef::test_lowercases_hash`) was **retired**, not silently dropped: it pinned
incidental behaviour of the mirror this PR deletes. Preserving it was *possible* — a one-line
normalisation at `_build_content_ref` — but that would add behaviour no requirement asks for. The
rationale is recorded in the test file.

## Notable defect found and fixed along the way

`sync/diagnose.py` imports `_PAYLOAD_RULES` from `sync/emitter.py` and previously guarded only against
`None` before dict-shaped access. The sentinel this PR introduces would have raised
`AttributeError` on the production `spec-kitty sync diagnose` path. Both consumers now share one
predicate, and `tests/sync/test_diagnose.py` gained dossier coverage it did not have.

## Verification

- Full mission scope: **2221 passed, 2 skipped, 2 xfailed, 0 failed**.
- `ruff check src/` clean; `mypy --strict` clean on both changed sync modules.
- Every changed behaviour has a test that fails when the change is reverted — verified by reverting,
  not asserted. A pre-merge review squad additionally found that the first version of the delegation
  tests passed against a re-inlined check; they were replaced with tests that discriminate real
  delegation from the retired logic, and that discrimination was itself reproduced independently.

## Follow-ups deliberately not in this PR

- The AST guard matches bare-name calls only; `dossier/__init__.py` re-exports the emitters, so an
  attribute-chain call (`dossier.emit_artifact_indexed(...)`) would evade it. No such call exists today.
- Trimming those re-exports from `__all__` — a public-API change on a currently-dead surface.
- The committed `analysis-report.md` contains host-absolute paths (upstream #3398, which reports 171
  such files on main; the count is now 175). Generated by `analysis_report.py`, so not hand-edited out
  here.

<sub>Mission artifacts (spec, plan, work packages, and the full adversarial review trail) travel in
`kitty-specs/` on this branch.</sub>

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3672"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790015100&installation_model_id=427282&pr_number=3672&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3672&signature=620b5c00d713a0441d56a772f6b7e5c12507a6c1f4bc1c08579d2b15b9783764"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
Claiming this PR for a landing pass: rebase onto current `main`, adjudicate every red check into its bin, fold remediation as single-purpose `<type>(landing)` commits. Evidence to follow. The PR stays draft until CI is green; I do not merge — that is the operator's call.

MOES-Media:

**Verdict: rebased clean, one fold, no gate loosened.** Watching CI; merge-ready on green. I do not
merge — that is the operator's call.

### Rebase

Rebased onto current `main` (the branch was **140 commits behind** by the time this ran, up from 99
when the PR opened). **Zero conflicts across all 52 commits.** Every adjudication below is on the
rebased tip `03cda7bbf`, never the stale base.

### The one fold

`docs(mission): record analysis report for …` × 2 — reworded two commit subjects that failed
`commitlint`. No file content changed (verified: `git diff` between pre- and post-reword tips is
empty; only messages differ).

**Why they failed, and why the fix is a reword rather than a config change.** `commitlint.config.cjs`
deliberately ignores spec-kitty's auto-generated lifecycle commits:

```js
/^(Add|Update) (meta|spec|tasks|plan) for (feature|mission) /
```

with the comment *"spec-kitty auto-generates mission lifecycle commits that don't use
conventional-commit format."* That pattern covers `meta`/`spec`/`tasks`/`plan`, but **not** the
subject `spec-kitty agent mission record-analysis` emits — `"Add analysis report for mission <slug>"`.
So those two commits fell outside a whitelist that was clearly meant to include them, and failed
`type-empty` + `subject-empty`.

Extending that regex would have fixed it repo-wide, but broadening a shared lint ignore during a
landing pass is loosening a gate to get green. **Rewording our own two commits fixes this PR without
touching a gate anyone else depends on.** The underlying gap is worth a follow-up — preferably by
having `record-analysis` emit a typed subject, which fixes the cause rather than widening the
exemption.

Verified after the fold: **0 commitlint failures across all 52 commits**, using CI's own invocation.

### Classification of everything else

No other red required a fold. Specifically checked and cleared:

- **markdownlint** — this PR adds 12 markdown files, all under `kitty-specs/`, which is in
  `.markdownlint-cli2.jsonc`'s `ignores`. Not linted, so the "file-scoped linter lints the whole
  changed file" trap does not bite here.
- **`uv.lock` freshness** — no dependency change. `spec-kitty-events>=6.0.0,<7.0.0` was already
  pinned and 6.1.0 already installed; `pyproject.toml` and `uv.lock` are untouched.
- **Coverage floors** — `dossier/` and `sync/` are not in `diff-coverage`'s enforced `critical_paths`,
  so no per-package floor applies to this diff.

### Local evidence on the rebased tip

- `ruff check src/` → **All checks passed**
- Mission test scope → **606 passed, 0 failed**
- `tests/architectural/` → **1683 passed, 5 skipped, 2 xfailed, 0 failed**
- Terminology gate (`test_no_legacy_terminology.py`) → **10 passed**

The architectural suite was run from a **plain checkout, not the landing worktree**. Tree-walking
gates can silently skip dot-prefixed path segments, so a run from `.worktrees/…` can report a false
green. Worth knowing for anyone verifying this branch themselves.

### Pre-existing failures

**None on this PR's surface.** The known-red baseline here is genuinely zero, and it took three runs
to establish honestly — the first reported 11 failures, of which 8 were `No module named ruff` /
`No module named mypy` (a toolchain gap in the local venv, not a repo red) and 3 were xdist
contention in the offline-queue tests, which pass serially. Recorded here so the next person does not
re-derive a baseline that never existed. Repo-wide known reds are tracked in #3284; none of them
intersect this diff.

### Draft status

Kept **draft** deliberately, with the `ci:full` label applied to run the full suite anyway — the
workflow's draft exemption is label-based (`ci-quality.yml`). Draft-first is the charter's rule; the
PR will be un-drafted when CI is green, not before.

MOES-Media:
## Merge-ready — hand-off

CI is green on `03cda7bbf`, the PR is out of draft, and the evidence is above. **I do not merge; the
operator does.** No landing-order constraint — this PR carries no structural cutover and nothing else
in flight depends on it.

### CI

**41 pass, 0 fail, 39 skipped.** Every gate this diff selects ran and passed, including:

`lint` (the composite carrying commitlint / ruff / mypy / markdownlint / pip-audit / bandit) ·
`quality-gate` · `fast-tests-sync` · `integration-tests-sync` · `integration-tests-sync-real-port` ·
`fast-tests-sync-orphan-sweep` · `arch-adversarial` (all three shards) · `diff-coverage` ·
`mission-loader-coverage` · `unit-contract-residual` · `clean-install-verification` ·
`AST lint — canonical producers` · `drift-detector` · `docs-freshness` · `cutover-guard` ·
Windows-critical · Build wheel.

The 39 skips are **path filters**, not draft gates — shards for `cli`, `dashboard`, `charter` and
similar that this diff does not touch.

### One thing a reviewer should not go looking for

**SonarCloud does not run on pull requests in this repo.** Its job in `ci-quality.yml` is gated to
`schedule` / `workflow_dispatch`, and no other workflow runs sonar on a PR event. Its absence here is
expected, not a skipped gate. Flagging it because a reviewer expecting a Sonar verdict on a PR will
otherwise wait for one that never arrives.

### What was verified locally, on the rebased tip

- `ruff check src/` → All checks passed
- Mission test scope → 606 passed, 0 failed
- `tests/architectural/` → 1683 passed, 5 skipped, 2 xfailed, 0 failed — run from a **plain checkout**,
  because tree-walking gates can report a false green from a `.worktrees/…` path
- Terminology gate → 10 passed
- commitlint over all 52 commits, using CI's own invocation → 0 failures

### Review-worthy decisions, restated so they are not mistaken for oversights

1. **Half of #1058 is deliberately not implemented.** The queue-drain half is superseded by #3293,
   which deleted `_migrate_legacy_dossier_payload` outright. `sync/queue.py` and
   `sync/migrate_journal.py` are untouched, verified byte-for-byte.
2. **Canonical validation is looser than the local checks it replaces**, in two measured places
   (hash-case normalisation; hash-format). Unreachable in production today. The fix, if wanted,
   belongs upstream in `spec-kitty-events`' schema — not as re-vendored local validation.
3. **One test was retired, not folded to green.** `TestContentHashRef::test_lowercases_hash` pinned
   incidental behaviour of the mirror this PR deletes. Preserving it was *possible* — a one-line
   normalisation at `_build_content_ref` — and was declined as behaviour no requirement asks for. It
   is **not** a `regression`-marked red-first test.
4. **A severity-4 review finding shaped the tests you see.** The first version of the delegation
   tests passed against a re-inlined required-field check — they proved the sentinel branch existed,
   not that it delegated. They were replaced with tests that discriminate real delegation from the
   retired logic, and that discrimination was reproduced independently against the committed suite.

### Follow-ups, not folded

- The AST guard matches bare-name calls only; `dossier/__init__.py` re-exports the emitters, so an
  attribute-chain call would evade it. No such call exists today.
- Trimming those re-exports from `__all__` — a public-API change on a currently-dead surface.
- `commitlint.config.cjs`'s tool-commit ignore list covers `meta|spec|tasks|plan` but not the subject
  `record-analysis` emits. Two commits here were reworded rather than widening a shared lint ignore
  during a landing pass; the durable fix is for `record-analysis` to emit a typed subject.

MOES-Media:
Follow-ups now filed, so the three items named above have tracked homes rather than living only in this thread:

- #3676 — AST positional-call guard misses attribute-chain / aliased calls, and `dossier/__init__.py` advertises exactly that surface
- #3677 — `dossier/__init__.py` re-exports external `spec-kitty-events` types under the CLI namespace
- #3678 — `record-analysis` emits a commit subject the commitlint tool-commit ignore list does not cover

#3678 is the one worth acting on soonest: it will red the commitlint gate on **every** future mission that reaches the analyze phase, and the fix is a one-line change to the subject that command emits.

robertDouglass:
3.2.6 execution DAG: #3692. Wave 1, dossier lane step 1. Upstream: #3127. Downstream: #3690.

Issue #3652, opened 2026-08-22
`$HOME` self-qualifies as a project, because `.kittify` is both the user-global state root and the project marker


kentonium3:
## Summary

`~/.kittify` is the CLI's user-global state directory — it holds `config.yaml`, `cache/` and
`last-cli-check.json` — and `.kittify` is simultaneously the marker used to locate a project root.
The two share a name, so `$HOME` satisfies project detection for any user who has ever run the CLI,
with nothing planted. Running `spec-kitty upgrade` from `$HOME` therefore reports
`Current version: unknown`, plans the entire migration history from `0.2.0`, and executes project
migrations against the home directory: `~/.gitignore` was created and `SKILL.md` files under
`~/.claude/skills/` and `~/.agents/skills/` were replaced. A refusal is what the situation calls for.

### Prerequisites

- `spec-kitty-cli 3.2.6rc3 (main build, SHA 8a03733e7)`
- A machine that has run `spec-kitty` at least once, so `~/.kittify/` exists as the user-global state
  root. On this machine its `cache/` directory dated from 2026-04-09.
- `$HOME` need not be a git repository. It was not here.

### Steps

```bash
cd ~
spec-kitty upgrade --yes
```

### Expected Behavior

A refusal, as when no project marker is present — the message quoted in #2610 is
`Unable to locate repository root (missing .git or .kittify).` The user-global state directory should
not be mistaken for a project to migrate.

### Actual Behavior

```text
Current version: unknown
Target version:  3.2.6rc3

                                 Migration Plan
┃ Migration                      ┃ Description                     ┃ Target    ┃
│ 0.2.0_specify_to_kittify       │ Rename .specify/ to .kittify/   │ 0.2.0     │
│                                │ and /specs/ to /kitty-specs/    │           │
│ 0.4.8_gitignore_agents         │ Add all 12 AI agent directories │ 0.4.8     │
│                                │ to .gitignore                   │           │
   … 55 migrations planned
```

`~/.kittify/metadata.yaml` afterwards records 55 migrations — 47 `skipped`, 7 `success`, 1 `failed` —
including:

```yaml
  - id: 0.4.8_gitignore_agents
    result: success
    notes: Added 12 agent directories to .gitignore
  - id: 2.0.6_consistency_sweep
    result: success
    notes: migrated 0 identical and 100 customized legacy runtime asset(s)
  - id: 2.1.2_fix_glossary_context_skill
    result: success
[...]

Issue #3347, opened 2026-08-12
Epic: Upgrade/migration atomicity & recoverability — a failed upgrade must be atomic, self-recoverable, and honestly reported


stijn-dejongh:
## Why

Upgrading a project that is several releases behind can **silently half-migrate, self-wedge, and then hide that it did so**. This was reproduced end-to-end by @kentonium3 upgrading a real project (3.2.0rc18, 98 missions): a single invalid mission artifact caused `runtime_state_backfill` to abort non-atomically, and the failure path then deleted the `schema_version` stamp — after which every subsequent `upgrade` refused to run, and the very diagnostic the failure recommended was itself gated behind the migration that had just failed.

The upgrade/migration surface currently violates three properties it is expected to hold:

- **Atomicity** — a step that documents itself as "aborts the whole step on first failure" must not leave partial writes in `meta.json` / the append-only `status.events.jsonl` authority.
- **Recoverability** — a failed upgrade must leave a state a later `upgrade` (or a repair/diagnostic command) can act on, not a permanently-wedged `legacy` classification with only an out-of-tool `git checkout` escape.
- **Honest reporting** — `--dry-run` must preview what the real run does; error envelopes must not drop the remediation body under `--json`; a failed `mission create` must not leave behind a branch and a moved checkout that look like partial success.

## For whom

Operators and agents upgrading projects that are several releases behind, and anyone whose first upgrade of an older project can leave it in an unrepairable state. Secondarily, maintainers who rely on `--dry-run` blast-radius previews before running migrations.

## Intended effect

When this epic is done, an upgrade of an old project is **atomic** (all-or-nothing across mission artifacts and the status event log), **self-recoverable** (a failure leaves a forward path; repair/diagnostic commands remain reachable), and **honestly reported** (dry-run parity, remediation bodies preserved through `--json`, no orphaned branches or silent checkout moves on failure).

Root cause of the observed cascade is the non-atomic backfill (#3335); the "no recovery path" wedge (#3334, #3338) is downstream of that abort. The child issues carry the per-defect detail and sequencing — see the sub-issue graph and `blocked_by` links.

stijn-dejongh:
## Hand-off: #3334 slice delivered by mission `mission-a-p0-consistency-01KZWHY1`

[...]

Issue #2624, opened 2026-07-13
Epic: Root & worktree-path detection — resolution mechanism, lane routing, and containment-boundary discipline


stijn-dejongh:
## Why

The system has no single owning surface for **detecting which root/worktree/path a command should operate on**. That detection mechanism spans several call sites that have drifted independently: project-root resolution (`locate_project_root()` / `resolve_canonical_root()` / `find_repo_root()` and the `.kittify`-ancestor walk), worktree/lane path detection (which `.worktrees/<slug>-lane-<id>` or legacy `.worktrees/<mission>-WP##` a command should read/write), worktree-aware vs worktree-unaware command routing (does a command resolve relative to the invoking worktree or silently fall back to the primary checkout?), and containment-boundary respect (submodule/nested-repo `.git`-file boundaries that must stop ascent). Because no single surface owns this family, correctness bugs land under unrelated catch-all or provenance-scoped parents, the same class recurs, and each fix patches one resolver while its siblings drift.

Concrete failures anchoring this epic:
- **#2610** — cwd-ancestor `.kittify` discovery walks *past an explicit containment boundary*. `locate_project_root()` iterates `[cwd, *cwd.parents]` and accepts any ancestor `.kittify` as project root; isolating `HOME`/XDG does not constrain this read-discovery boundary, so a command in a nested sandbox consumes assets planted above the intended ceiling.
- **#1971** (closed) — three divergent `locate_project_root` definitions with split authority (`core/paths.py`, `core/project_resolver.py`, `__init__.py`) — grep confirms all three still exist post-fix, so the split-brain risk is latent, not eliminated.
- **#2011** (closed, delivered via #2015) — `resolve_canonical_root` ascended across a submodule `.git`-file boundary instead of stopping at it.
- **#539** (closed) — `find_repo_root()` collapsed to the worktree root instead of the main checkout, so status events and config writes landed in the wrong location.
- **#237** (closed) — `create-feature`/`specify`, run from a git worktree located *outside* `.worktrees/`, failed its own `is_worktree_context(cwd)` guard (a bare `.worktrees`-substring path check) and `locate_project_root()` silently followed the `.git` pointer to the main repo, so scaffold commits landed in the wrong checkout entirely.
[...]

Issue #1914, opened 2026-06-13
Umbrella: governed/gate operations must be no-op-stable (stop self-writes dirtying the working tree)


stijn-dejongh:
## Pattern

Across mission #133's post-merge work, THREE distinct surfaces were found to dirty the git working tree with their OWN uncommitted writes when run as a no-op / read-only / governed operation. Each was fixed point-wise, but they share one root pattern worth addressing coherently: **a governed or gate operation should be no-op-stable — running it without a substantive change must leave \`git status\` clean.** Self-writes (provenance, identity, config, generated artifacts) that are never committed cause: false 'tree dirty' guard trips, conflicting/aborted ops, false test signals, and reviewer/rebase friction.

## Instances observed (attached as sub-issues / references)

1. **#1912** — charter synthesis rewrites \`.kittify/charter/provenance/*\` + manifest (timestamps + \`synthesizer_version\`) on every \`spec-kitty do\`/charter run even on no-op. (Fixed in PR #1913 via no-op-stable promote.)
2. **Accept-gate \`.kittify/config.yaml\` residue** — \`accept --no-commit\` readiness path mints project identity → writes config.yaml → second run's git_dirty trips on its own write. (Fixed in PR #1908 via NFR-003-safe accept-owned exclusion.)
3. **#1907** — subagent-worktree/editable-install dev-tooling hazard (adjacent: \`pip install -e .\` repoint + stale-base) — same family of 'operation mutates shared/tracked state as a side-effect'.

## Ask

- Adopt a 'no-op-stable / no-uncommitted-self-writes' invariant for governed + gate operations (charter synthesis, accept/next/implement gates, identity materialization, DRG/graph regen).
- Consider a shared guard/test pattern: a 'run twice on an unchanged tree → git status clean' architectural ratchet covering these surfaces.
- Audit remaining surfaces that materialize tracked files as a side-effect (graph.yaml, identity config, status/provenance) for the same idempotency property.

Cross-refs: #1912 (charter synth, fixed), #1907 (worktree/install tooling), #1908 (accept-gate fix). Parent: #1619.

stijn-dejongh:
## Umbrella membership consolidated (2026-06-13)

All same-class tickets — *a governed/gate/tooling operation dirties the working tree with its own uncommitted writes, or is non-idempotent on a no-op run* — have been re-parented directly under this umbrella. The full child set is:

| # | Title | Old parent | Action |
|---|-------|-----------|--------|
[...]

PR #3679, opened 2026-08-22
fix(upgrade): harden the upgrade command — canonical writer + commit-decision + scoped-consent seams (#3651 #3654 #3653 +5)

### LINKED ISSUE #2491: upgrade auto-commit runs before surface repair — post-commit writes (command-skills manifest, healed surfaces) left dirty, breaching the #2392 invariant
Field follow-up to the boundary explicitly flagged (and consciously deferred) in PR #2387's description: *"`_run_upgrade_surface_repair` runs after the auto-commit on both CLI paths, so surface-repair writes are intentionally left for the operator."* That was the pre-#2392 status quo, preserved to keep #2387's scope tight. It now has a real-world repro.

## Observed (spec-kitty 3.2.4 → 3.2.6, macOS, consumer repo)

`spec-kitty upgrade --project --yes` on an already-initialised project:

- ✅ auto-committed the migration churn (one backfilled `.gitignore` line + `.kittify/metadata.yaml` version bump) — the #2392 seam working as designed;
- ❌ left `.kittify/command-skills-manifest.json` **modified but uncommitted** — the command-skill refresh ran *after* `commit_touched_checkout()`, so its write fell outside the porcelain-derived commit-set.

So the run ends with a dirty tree anyway — the state the #2392 invariant (*"every path an upgrade run writes … must end in exactly one auto-commit … before the command returns"*) exists to prevent. Same applies to anything `offer_teamspace_mission_state_migration` writes (it also sits between the commit and the end of the run) and to healed skills/profiles from the repair pass.

## Mechanics (current `main`)

Both CLI paths in `cli/commands/upgrade.py` order as: `commit_touched_checkout(project_path, baseline, …)` → `offer_teamspace_mission_state_migration` → `_run_upgrade_surface_repair` → return. Writes from the last two stages are never captured.

## Two fix shapes

1. **Reorder**: run surface repair before the auto-commit so one commit captures everything. Touches repair's interactive/drift-exit semantics (Rule 4/5, non-zero drift exits) — the "deliberate reordering decision" #2387 deferred.
2. **Second commit pass (no reordering)**: after surface repair, call `commit_touched_checkout()` again **with the same original baseline**. The first commit's files are clean by then; pre-existing operator dirt is still baseline-excluded; the second pass captures exactly the post-commit writes. Skipped under `manual_review_required` (preserved files must never be swept) and under `--dry-run`. Purely additive — repair semantics, drift exits, and the existing commit are untouched.

Shape 2 reuses the `upgrade/autocommit.py` seam as-is and extends the #2392 invariant to the full command lifetime without reopening the reordering question. PR incoming with that shape — happy to pivot to shape 1 if you'd rather take the reordering decision now.

### LINKED ISSUE #3376: Bug: worktree upgrade path stamps schema_version unconditionally after a failed migration (FR-002 residual)
Follow-up from mission upgrade-atomicity-recovery-01KZWSHC / WP01 review.

WP01 fixed the FR-002 preserve-on-failure behaviour on the **main-project** upgrade path (`runner.py` main loop + `metadata.py::save()`). But `MigrationRunner._upgrade_worktrees` (`src/specify_cli/upgrade/runner.py` ~:440) still stamps `REQUIRED_SCHEMA_VERSION` **unconditionally** on each worktree, even after a worktree migration was recorded as failed, and it runs regardless of main-project `result.success`. A real #3334-class failure would therefore advance each worktree's `schema_version` to target, masking unfinished worktree state — the same dishonest-state class this mission closes.

Lower blast radius than #3334 (worktrees are ephemeral/regenerable), so it did not block WP01, but it is a genuine residual FR-002 instance. Fix: mirror WP01's capture-pre-run / restore-on-failure pattern onto the worktree loop.

Parent: #3347. Surfaced by reviewer-renata during WP01 review (commit d03b856af).

### LINKED ISSUE #3390: WP05 report-on-abort fidelity: PlacementMismatchError under-report + dry-run phantom writes (#3383 follow-up)
**Parent epic:** #3347 (Upgrade/migration atomicity & recoverability). Surfaced by the adversarial review squad during the landing of #3383 (WP05, report-on-abort for the runtime-state backfill cutover). Neither finding risks data; both are reporting-fidelity gaps on abort edge-cases, so they were filed rather than folded into the already-large #3383 landing.

## Finding 5a — under-report: `PlacementMismatchError` after a successful seed (live)
`m_zz_runtime_state_backfill.py::_partial_writes` derives each mission's persisted paths from `result.seeded_count > 0` / `result.flipped`. When `cutover_mission` seeds N>0 events, passes verify, then `_flip_phase` raises `PlacementMismatchError` (`runtime_state_cutover.py:212-217`), `_cutover_corpus` catches it and rebuilds a fresh `CutoverResult(flipped=False, error=...)` — with `seeded_count` **defaulting to 0**, discarding the real count. `_partial_writes` then evaluates `0 > 0` → False and **omits the seeded `status.events.jsonl`** even though N events are physically on disk.
- Repro: corpus `[alpha(ok), beta]`; beta seeds 3 events, verify ok, placement resolves a PRIMARY home ≠ write target → flip raises. The abort report lists alpha's files but **not** beta's `status.events.jsonl`. Contract ("no fewer") violated. Residue is idempotent (mission not flipped → re-seeds byte-identically next run), so no data loss.
- Root fix: thread `seeded_count` through the `PlacementMismatchError` fold in `_cutover_corpus` so `_partial_writes` sees the true count.

## Finding 5b — over-report: dry-run abort reports phantom writes
`_partial_writes` takes no `dry_run` flag. In `apply(dry_run=True)`, a structural error trips `_mission_failed(dry_run=True)` and `apply` returns `partial_writes=_partial_writes(results, …)`. Earlier missions carry `seeded_count>0` *would-seed* counts with **nothing written** (`backfill_runtime_state.py:1503-1504` returns `action="wrote", seeded_count=N` under dry-run without writing). So a `--dry-run` preview claims files were written — contradicting base.py's "Files already persisted…" definition and the #3336 honest-dry-run goal.
- Root fix: short-circuit `_partial_writes` to empty when `dry_run` (a preview persists nothing).

Neither path is exercised by `tests/upgrade/test_backfill_report_on_abort.py` (which covers only the live verify-fail abort); add red-first coverage for both.

## WP12 rollback — two narrow, acceptable edge-notes (docstring-level)
`core/mission_creation.py::_restore_git_state_after_failed_create`:
- **Detached-HEAD after failure**: `get_current_branch` returns `None` → the checkout restore is skipped, leaving the operator in detached HEAD. Unlikely (the mint doesn't checkout), but worth a docstring note.
- **`force_recreate=True` on a pre-existing coordination branch**: the delete+recreate name is in both pre/post branch sets, so it's not in the diff → a later failure leaves it reset to `target_branch` with no restore of its original tip. Scoped to an explicit destroy-and-recreate escape hatch; worth a docstring note.

_Filed from the #3383 maintainer landing (adversarial squad: priti correctness lens). Verdict on #3383 was LAND-WITH-FOLDS; the MAJOR-free findings above are the deferred, non-blocking part._

### LINKED ISSUE #3392: Bug: upgrade --project --yes exits 1 on a fully successful run and prints no outcome at all
## Summary

On the human-readable path, `spec-kitty upgrade --project --yes` can apply every migration, stamp
the new version, leave `schema_version` intact — and then exit 1 having printed no outcome
whatsoever: no "Upgrade complete!", no migrations-applied list, and no error naming the reason for
the non-zero exit. A genuine failure exits 1 too, with "Upgrade failed.", so on stdout the two
cases are told apart only by the absence of a banner. FR-006's non-zero-on-drift contract is not in
question here; the reporting is. The `--json` path is unaffected, so this is specifically the
surface a human operator or an agent reading terminal output depends on.

### Prerequisites

- `spec-kitty-cli 3.2.6rc2 (main build, SHA 4a2367539)`
- A project stamped an older version with migrations pending (here `3.2.0rc18`, 98 missions under
  `kitty-specs/`)
- At least one generated tool surface with local edits, so surface repair reports drift rather than
  repairing silently (here 23)

### Steps

```bash
spec-kitty upgrade --project --yes; echo "exit: $?"
```

### Expected Behavior

Either the outcome report followed by a non-zero exit that names the unresolved drift — the string
`_surface_drift_error()` already builds for exactly this case — or exit 0. What an operator must be
able to do is tell "everything applied, some surfaces have local edits" from "the upgrade failed".

### Actual Behavior

`exit: 1`, and the output simply stops after the surface-repair lines:

```text
Mission-state repair complete (updated=108, unchanged=10, errors=0).
Manifest: .kittify/migrations/mission-state/7c15bf193e3117ad.json
TeamSpace mission-state blockers cleared.
Created 118 tool surface(s)
Note: 23 tool surface(s) have local edits — run 'spec-kitty doctor
tool-surfaces' to review
  7 surface(s) not applicable, skipped.
```

That is the end of a 209-line run. There is no `Upgrade complete!`, no `Upgrade failed.`, no
`Errors:` section, and no mention of drift as the reason for the exit code.

Everything the run was asked to do had in fact succeeded:

- `.kittify/metadata.yaml` → `version: 3.2.6rc2`, `schema_version: 3` intact
- mission-state repair reported `errors=0`
- the tool's own auto-commit landed
- an immediately following `spec-kitty upgrade --project --dry-run` reported
  `Project is already up to date!`

Reproduced twice — a second run of the same command on the same project, after resetting it to its
pre-upgrade state, exited 1 identically.

For contrast, on the same project earlier the same day a genuine abort inside
`runtime_state_backfill` also exited 1, and *that* run printed the `Errors:` section and
`Upgrade failed.`

## Root Cause

`cli/commands/upgrade.py:392-395` (installed build `4a2367539`) raises from inside the helper:

```python
        if not json_output:
            for line in render_surface_summary_lines(summary):
                console.print(line)
            if summary.drifted_reported and confirm:
                raise typer.Exit(1)
```

Two consequences follow, both only on the non-JSON path:

1. `_display_upgrade_results()` (`upgrade.py:1039`) never runs, so the entire outcome report —
   migrations applied, skipped, warnings, errors, and the closing banner — is skipped.
2. The caller's own drift handling becomes unreachable. `upgrade.py:952-970` computes
   `surface_drift_failed` and `success = result.success and not surface_drift_failed`, and
   `_surface_drift_error()` (`upgrade.py:439`) builds the message *"Unresolved tool-surface drift in
   N file(s); run 'spec-kitty doctor tool-surfaces' to review."* On the human path neither can be
   reached, so the one sentence that would explain the exit code is dead code.

The `--json` branch is correct: the raise sits inside `if not json_output:`, and the JSON envelope
carries `status`, `success` and `errors`. An automated caller passing `--json` can tell success from
failure; one reading stdout cannot.

## Suggested Fix

Have the helper report drift and return its summary, leaving the exit decision to the caller that
already computes it. `_display_upgrade_results()` then prints the outcome, the drift error reaches
the operator, and FR-006 still holds because the caller exits non-zero on
`surface_drift_failed`. A distinct exit code for "applied, but drift unresolved" would additionally
let a caller separate it from a failed migration — the case that currently shares code 1.

## Workaround Applied

Judge the outcome by evidence rather than by the exit code: `upgrade --project --dry-run` reporting
"already up to date", the version stamp in `.kittify/metadata.yaml`, and `schema_version`
surviving. This is not usable from a CI gate or a script without `--json`.

## Environment

- OS: macOS Darwin 25.5.0
- Python: 3.13
- `spec-kitty-cli 3.2.6rc2 (main build, SHA 4a2367539)`

Reviewed before filing.

### LINKED ISSUE #3393: Bug: 3.2.5_agents_skills_gitignore_backfill gitignores paths it leaves tracked, so the tree stays permanently dirty
## Summary

`3.2.5_agents_skills_gitignore_backfill` adds `.agents/skills/` and
`.kittify/skills-manifest.json` to `.gitignore`, but does not untrack the files already committed
under those paths. `.gitignore` is inert for tracked paths, so in a project that had committed them
the upgrade turns 28 tracked files into working-tree modifications that persist across later runs
and ride into every subsequent PR. The same build already contains the missing half: the sibling
`3.2.0rc35_sync_state_gitignore` migration untracks what it ignores.

### Prerequisites

- `spec-kitty-cli 3.2.6rc2 (main build, SHA 4a2367539)`
- A project with files committed under `.agents/skills/` and a tracked
  `.kittify/skills-manifest.json`, neither path yet in `.gitignore` — here 28 files under
  `.agents/skills/` (27 of them symlinks, 1 a regular file) plus the manifest, all committed before
  the ignore rule existed

### Steps

```bash
git ls-files .agents/skills/ | wc -l              # 28 here
git ls-files -s .agents/skills | grep -c 120000   # 27 of them symlinks
git ls-files .kittify/skills-manifest.json        # tracked
spec-kitty upgrade --project --yes
git status --porcelain -- .agents/skills .kittify/skills-manifest.json
```

### Expected Behavior

After the migration the two paths are ignored **and** untracked, per the
`3.2.0rc35_sync_state_gitignore` precedent, leaving a clean tree.

### Actual Behavior

The `.gitignore` entries are added and all 29 paths stay tracked. The 27 symlink entries show as
typechange, because the same upgrade replaces them with real files:

```text
 T .agents/skills/ad-hoc-profile-load/SKILL.md
 T .agents/skills/spec-kitty-charter-doctrine/SKILL.md
 T .agents/skills/spec-kitty-charter-doctrine/references/charter-command-map.md
 T .agents/skills/spec-kitty-git-workflow/SKILL.md
 T .agents/skills/spec-kitty-glossary-context/SKILL.md
 …27 T in total
```

This survives the upgrade's own auto-commit and is still present on the next run, because nothing in
the flow ever untracks the paths. Any project that committed these surfaces before the ignore rule
existed therefore ends every upgrade with a dirty tree it cannot clean by editing `.gitignore`.

## Root Cause

`upgrade/migrations/m_3_2_5_agents_skills_gitignore_backfill.py` — `apply()` performs a single
action:

```python
        GitignoreManager(project_path).ensure_entries(missing)
```

It is a `.gitignore` text edit with no untrack step. The migration's docstring states it follows
the `3.2.4_derived_mission_views` precedent, and that is the mismatch: the 3.2.4 derived views
appear as **untracked** files, so ignoring them is sufficient. `.agents/skills/` is different — it
was **committed** in projects that predate the ignore rule, which is precisely the absolute-symlink
commit hazard #2412 described. Ignoring an already-tracked path changes nothing.

## Suggested Fix

Adopt the rc35 pattern rather than the 3.2.4 one. `m_3_2_0rc35_sync_state_gitignore.py:39-56`
already implements both halves for its own paths — `_is_tracked()` via
`git ls-files --error-unmatch`, and `_untrack()` via `git rm --cached --` — and applies them after
ensuring the gitignore entries, reporting each untracked path in `changes_made`. Reusing that shape
for the two 3.2.5 paths completes the backfill. Closed #1604 established this requirement for the
same class ("Added to `.gitignore` after being committed; no `git rm --cached` migration was
performed"), so the pattern is settled; it simply did not reach this migration.

Worth noting for scope: `.kittify/command-skills-manifest.json` is generated per-machine in the same
way and is left tracked-or-untracked inconsistently across projects, so whatever decision covers
these two paths probably wants to cover it too.

## Workaround Applied

Untracked all 29 paths by hand — `git rm -r --cached .agents/skills .kittify/skills-manifest.json` —
in an isolated commit, so it is a one-command revert if the project decides these paths should stay
tracked.

## Environment

- OS: macOS Darwin 25.5.0
- Python: 3.13
- `spec-kitty-cli 3.2.6rc2 (main build, SHA 4a2367539)`

Reviewed before filing.

### LINKED ISSUE #3651: `upgrade` cannot rewrite its own generated command files, which it wrote read-only
## Summary

`spec-kitty upgrade` writes the generated agent command and prompt files with mode `444`, then on a
later upgrade rewrites them without first restoring the write bit. Every such file fails with
`[Errno 13] Permission denied` and the run ends with `Upgrade failed.` and exit 1. On a machine with
several agent surfaces configured this is around thirty simultaneous failures, so the upgrade cannot
complete until the write bits are cleared by hand. The correct chmod-restore already exists in two
other places in the same codebase, which is what suggests an omission rather than a design choice.

### Prerequisites

- `spec-kitty-cli 3.2.6rc3 (main build, SHA 8a03733e7)`
- A machine that has run `spec-kitty` before, so the generated surfaces already exist at mode `444`.
- More than one agent surface present. On this machine there were twelve, holding 182 generated
  files in total.

### Steps

```bash
ls -l ~/.kiro/prompts/spec-kitty.review.md      # -r--r--r--
spec-kitty upgrade --yes
```

### Expected Behavior

The upgrade restores the write bit on files it owns before rewriting them — the behaviour already
implemented at `src/specify_cli/runtime/agent_commands.py:316` and `:336` — and completes.

### Actual Behavior

```text
✗ Failed to write .kiro/prompts/spec-kitty.review.md: [Errno 13] Permission denied: '/Users/…/.kiro/prompts/spec-kitty.review.md'
✗ Failed to write .kiro/prompts/spec-kitty.status.md: [Errno 13] Permission denied: '/Users/…/.kiro/prompts/spec-kitty.status.md'
✗ Failed to write .agent/workflows/spec-kitty.specify.md: [Errno 13] Permission denied: '/Users/…/.agent/workflows/spec-kitty.specify.md'
   … approximately thirty files

Upgrade failed.
```

Exit code 1. The project metadata afterwards records
`2.1.4_enforce_command_file_state` with `result: failed`, `notes: null`.

## Root Cause

`src/specify_cli/upgrade/migrations/m_2_1_4_enforce_command_file_state.py:370` and `:387` call
`output_path.write_text(...)` with no chmod-restore beforehand. Lines `:373` and `:390` emit the
observed message via `errors.append(f"Failed to write {rel_path}: {exc}")`. The module docstring
states it "writes ALL files for ALL configured agents unconditionally", which is why the failures
arrive in bulk rather than one at a time.

The write protection is applied by `src/specify_cli/runtime/agent_commands.py:318` and `:338`
(`mode & ~0o222` after writing).

The same hazard is already handled correctly elsewhere:

- `src/specify_cli/runtime/agent_commands.py:316` and `:336` — restore `| 0o222` before rewriting.
- `src/specify_cli/upgrade/migrations/m_3_1_2_globalize_commands.py:157` —
  `target.chmod(target.stat().st_mode | 0o222)` before `unlink()`.

## Suggested Fix

Restore the write bit before `write_text` in `m_2_1_4_enforce_command_file_state.py`, mirroring what
`m_3_1_2_globalize_commands.py:157` already does:

```python
if output_path.exists():
    output_path.chmod(output_path.stat().st_mode | 0o222)
output_path.write_text(...)
```

Applying it at a shared write helper rather than at the two call sites would also cover future
migrations that write these surfaces. Worth noting that `m_3_1_2` needed the restore only because it
unlinks; any migration that *writes* needs it too, and that distinction is easy to miss.

## Workaround Applied

```bash
cd ~ && find . -maxdepth 3 -name 'spec-kitty.*.md' -not -path './repos/*' -exec chmod u+w {} +
```

After this the upgrade completed with exit 0 and no permission errors. The CLI re-applies mode `444`
on write, so the workaround is needed again on the next upgrade.

Deleting the generated files and regenerating them also unblocks it, since POSIX unlink is gated by
the parent directory's write bit rather than the file's — but `doctor command-files` repopulated only
nine of the twelve surfaces here. `.codex/prompts`, `.opencode/command` and `.roo/commands` were not
regenerated, and their existing contents dated from April and June with an older naming scheme
(`spec-kitty.tasks_finalize.md` with underscores, and a `spec-kitty.checklist.md` not present in the
current fifteen-command set). Whether those three surfaces are still supported is unclear from the
outside.

## Environment

- OS: macOS Darwin 25.5.0 (macOS 26.5.2, x86_64)
- Python: 3.13.15
- `spec-kitty-cli 3.2.6rc3 (main build, SHA 8a03733e7)`

---

**Authored by**: Kent Gale & Claude Code (Claude Opus 5), 2026-08-22
**Submission approved by**: Kent Gale, 2026-08-22

### LINKED ISSUE #3653: `upgrade --yes` authorizes `doctor mission-state --fix` with no prompt shown, and the prompt defaults to yes
## Summary

`spec-kitty upgrade --project --yes` runs the TeamSpace mission-state repair across every mission in
the project without displaying an offer. `--yes` is documented as "Non-interactive confirmation" and
is an alias for `--force`; at this site it also authorizes a state-mutating repair. Separately, when
the prompt is shown it defaults to yes, so an operator pressing Enter accepts the repair. On a
fifteen-mission project that is a substantial action taken on a flag that reads as a prompt
suppressor. There is precedent in the same file for carving a check out of `--yes`.

### Prerequisites

- `spec-kitty-cli 3.2.6rc3 (main build, SHA 8a03733e7)`
- A project with missions and at least one TeamSpace blocker. The blocker code here was
  `SNAPSHOT_DRIFT`, across 2 of 15 missions.

### Steps

```bash
spec-kitty upgrade --project --yes
```

### Expected Behavior

`--yes` suppresses confirmation prompts. A repair that mutates mission state is a distinct
authorization and warrants its own opt-in — which the same command already does for another check.
`src/specify_cli/cli/commands/upgrade.py:551`, in the command's docstring:

```
CHK037 / A-006: ``--yes`` and ``--force`` do NOT bypass this check.
```

with the corresponding inline comment at `:724`
(`# CHK037 / A-006 — Check if project is too new for this CLI.`).

### Actual Behavior

The repair runs with no prompt displayed. With `--yes` omitted, the prompt appears and is defaulted
to yes:

```text
Run `spec-kitty doctor mission-state --fix` now? [Y/n]:
```

Answering `n` prints `Skipped TeamSpace mission-state repair.` and the migrations still apply, so
declining is supported — it is simply not reachable under `--yes`.

## Root Cause

`src/specify_cli/cli/commands/_teamspace_mission_state_gate.py:172`:

```python
should_run = assume_yes or typer.confirm(
    "Run `spec-kitty doctor mission-state --fix` now?",
    default=True,
)
```

`assume_yes` is `confirm`, set at `src/specify_cli/cli/commands/upgrade.py:685`
(`confirm = (yes is True) or (force is True)`) and passed at `:860` and `:1006` as
`assume_yes=confirm`.

The propagation itself looks deliberate — `tests/upgrade/test_upgrade_auto_commit_unit.py` asserts
`calls[0]["assume_yes"] is True` — so the question here is not whether flag scope was considered, but
whether this particular action belongs inside it. What appears unattested is the consequence:
`offer_teamspace_mission_state_migration` has no direct test coverage. Searching for
`offer_teamspace_mission_state_migration(` across `tests/` and `src/` returns only its definition at
`_teamspace_mission_state_gate.py:143` and the two call sites in `upgrade.py`. Neither the
`should_run` branch that converts `assume_yes` into an unprompted `repair_repo()` nor the
`default=True` is pinned by a test.

## Suggested Fix

Two independent changes at one site:

1. Do not let `assume_yes` alone authorize `repair_repo()`. Following the CHK037 / A-006 precedent
   already in `upgrade.py`, require an explicit opt-in — a dedicated flag, or reuse of the
   `--allow-dirty` style already present on the repair path.
2. Make the prompt default to no for a state-mutating repair.

Direct tests for `offer_teamspace_mission_state_migration` would pin both, and there are none today.

One consideration for whichever shape is chosen: an unattended upgrade that now silently repairs
would begin failing or skipping instead, so the change is not purely additive for CI callers.

## Workaround Applied

Drive the prompts rather than passing `--yes`:

```bash
printf 'y\nn\n' | spec-kitty upgrade --project     # y = apply migrations, n = decline the repair
```

The prompt order depends on whether migrations are pending. With migrations pending, the
"Apply N migration(s)?" confirm at `upgrade.py:962` comes first and the gate at `:1002` second. With
no migrations pending, the gate at `:856` is the only prompt — so a blind `y` first would land on the
repair.

## Environment

- OS: macOS Darwin 25.5.0 (macOS 26.5.2, x86_64)
- Python: 3.13.15
- `spec-kitty-cli 3.2.6rc3 (main build, SHA 8a03733e7)`

---

**Authored by**: Kent Gale & Claude Code (Claude Opus 5), 2026-08-22
**Submission approved by**: Kent Gale, 2026-08-22

### LINKED ISSUE #3654: `upgrade` is the only auto-committing command path that ignores the `auto_commit` opt-out
## Summary

#140 shipped a project-level auto-commit switch — `get_auto_commit_default(repo_root)` in
`src/specify_cli/core/agent_config.py`, defaulting `auto_commit: bool = True`. Eight command paths
honour it. `upgrade` does not: neither `src/specify_cli/cli/commands/upgrade.py` nor
`src/specify_cli/upgrade/autocommit.py` imports `agent_config` at all, and no `--no-commit` flag
exists on the command. So on build `8a03733e7` there is no way, by flag or by configuration, to run
`spec-kitty upgrade --project` without it creating a commit — in a tool that already has precisely
that switch for other operations.

### Prerequisites

- `spec-kitty-cli 3.2.6rc3 (main build, SHA 8a03733e7)`
- A kittified project, optionally with `auto_commit: false` configured.

### Steps

```bash
spec-kitty upgrade --project --yes
git log -1 --format='%h %s'
```

### Expected Behavior

With `auto_commit: false`, the upgrade applies its migrations and leaves the result staged or in the
working tree, as the other eight auto-committing paths do under the same setting. Failing that, a
`--no-commit` flag on `upgrade`.

### Actual Behavior

```text
Upgrade complete! 3.2.6rc2 -> 3.2.6rc3
→ Auto-committed upgrade changes (4 files)
```

A commit is created regardless of the project's `auto_commit` setting.

## Root Cause

`agent_config` is imported and its default honoured by `cli/commands/implement.py`,
`cli/commands/retrospect.py`, `acceptance/__init__.py`, `agent/tasks_move_task.py`,
`agent/tasks_mark_status.py`, `agent/tasks_map_requirements.py` and `agent/tasks_status_cmd.py` (two
call sites). It is imported by neither `cli/commands/upgrade.py` nor `upgrade/autocommit.py`.
Neither file contains `no_commit` or `--no-commit`.

Nothing in `tests/upgrade/` references `get_auto_commit_default`, so the exclusion does not appear to
be pinned as intentional — though it may simply be undocumented rather than unintended.

## Suggested Fix

Have `upgrade/autocommit.py` consult `get_auto_commit_default(repo_root)` as the other paths do, and
skip the commit when it returns false. A `--no-commit` flag would serve the same need more
explicitly, and would be discoverable from `--help` in a way a config key is not.

This should not disturb the #2392 invariant cited in #2491 — that every path an upgrade writes ends
in exactly one auto-commit. Opting out means zero commits by explicit operator choice, not a partial
or split commit, so the invariant still holds for every run that commits at all.

If the exclusion is deliberate — on the reasoning that upgrade bookkeeping must always be captured —
then saying so in the command help or in `autocommit.py` would resolve this as documentation, since
the setting's name currently implies project-wide scope.

## Workaround Applied

Let the commit happen and then relocate it, which is safe because the guard only ever commits locally
and never pushes:

```bash
git branch <topic> HEAD && git checkout <topic>
git branch -f main origin/main
git push -u origin <topic>
```

For a repository whose integration branch is pull-request-only at the hosting layer, this is needed
on every upgrade.

## Environment

- OS: macOS Darwin 25.5.0 (macOS 26.5.2, x86_64)
- Python: 3.13.15
- `spec-kitty-cli 3.2.6rc3 (main build, SHA 8a03733e7)`

---

**Authored by**: Kent Gale & Claude Code (Claude Opus 5), 2026-08-22
**Submission approved by**: Kent Gale, 2026-08-22

stijn-dejongh:
## Upgrade-command hardening — one cluster, three canonical seams

Remediates the confirmed `spec-kitty upgrade` defect cluster spanning epics **#3347** (atomicity / honest reporting) and **#1914** (governed-op no-op-stability). The root was a ~1,300-line `cli/commands/upgrade.py` with **two duplicated orchestration branches and no shared seams**; this PR replaces that with a permission-bit-owning writer, a config-honoring commit decision, a scoped-consent boundary, and a single post-migration finalizer.

### Issues fixed
| # | Defect | Fix |
|---|--------|-----|
| **#3651** (P1) | Re-run failed with ~30 `Errno 13` on its own read-only generated files | Canonical `write_generated_file` (restore→write→strip), routed through 16 migrations + a mechanical arch guard |
| **#3654** | `auto_commit: false` opt-out ignored (main + every worktree) | `should_auto_commit` main/worktree scopes consult `get_auto_commit_default`; human-mode now reports churn left uncommitted |
| **#3653** | `--yes`/`--force` silently authorized `doctor mission-state --fix` (`repair_repo`) | Scoped, **default-deny** consent gate returning `RepairOutcome`; `--yes` authorizes only the invoked migrations |
| **#3392** | Fully-successful `upgrade --project --yes` could exit 1 / print nothing | Single finalizer → one `UpgradeOutcome`, exit derived exactly once; `# noqa: C901` removed, complexity ≤15 |
| **#3376** | Failed worktree migration still stamped `schema_version` | Per-worktree failure flag guards the stamp; failures surfaced in the exit code |
| **#3393** | gitignore-backfill left paths tracked-and-ignored (permanently dirty tree) | Untracks the paths it ignores (`git rm --cached`) |
| **#3390** | `PlacementMismatchError` under-reported + dry-run phantom writes | Abort names the mismatch; dry-run writes zero files |
| **#2491** | Post-commit surface-repair writes left uncommitted | Finalizer ordering: surface-repair → single churn commit → scoped repair (its own commit) |

### Approach & guarantees
- **One finalizer, no layering inversion.** The two branches converge on `finalize_upgrade(...)` which receives its steps as injected callables (`upgrade/ → cli.commands` inversion avoided); it derives the exit code once from a single `UpgradeOutcome` that *wraps* the existing `UpgradeResult` (renderers keep their fields).
- **Behavior-preserving.** A characterization net (`tests/upgrade/test_upgrade_char_net.py`) pins default (`auto_commit` unset/true) commit behavior byte-for-byte across the refactor.
- **Every fix has a non-fakeable test** asserting its corrected observable (red-first for the P1); the consent scope is proven via a `declined`-state spy that genuinely reaches the consent decision.

### Quality gates
- `tests/upgrade/`: **688 passed**; full cross-cut sweep: **7301 passed, 0 real failures**.
- `ruff` + `mypy --strict` clean on all new/changed code; `# noqa: C901` removed (complexity ≤15); terminology guard green.
- CHANGELOG updated (`[Unreleased]`); no version bump (`__init__.py`/`pyproject.toml` untouched).

### Process note
Planned and hardened via four brownfield adversarial-squad point-cuts (post-spec, post-plan, post-tasks, pre-merge). The squads caught, before merge: a blocking finalizer layering cycle (fixed via callable injection), a consent gate that had to change its return contract for "exit-once" to be reachable, three false-green test traps, and the FR-003 human-mode reporting gap.

### Out of scope / follow-ups
- **#3652** ($HOME self-qualifies as a project) deferred to epic **#2624** — the new seams consume the resolved root as-provided and do not touch detection.
- Follow-ups (LOW, not blocking): the #3393 untrack heals forward only (does not re-heal already-migrated projects).

### Maintainer landing folds (2026-08-22)
Landed on top of the original work during the maintainer landing pass (full evidence in the remediation-summary comment):
- **Completed the #3651 routing.** An adversarial-squad MAJOR found 5 sibling migrations still writing bare `write_text` to read-only command/skill surfaces (two a hard `PermissionError` crash). All now route through `write_generated_file` (red-first proof for the crash sites), and the guard is converted from a 16-file allowlist to a full-scan invariant with a rationale'd exemption set — so the claim "cannot be forgotten again" is now enforced, not enumerated.
- **Hardened the writer:** atomic temp-sibling + `os.replace` write, read-only strip in `finally`, symlink target refused.
- **`--json` errors consistency:** `worktree_failures` now fold into the `errors[]` channel (deduped, no contract change), so a failed run never reports `success:false` with an empty `errors[]` — this addresses the machine-consumer follow-up above directly rather than via a new key.
- **Fixed the one CI red** (a split `patch()` target literal tripping the #394 validator) and, per operator request, **cleared all 38 pre-existing `mypy --strict` errors tree-wide** (advisory; unrelated to this command — behavior-preserving, separately committed).

Closes #3651, 
Closes #3654, 
Closes #3653, 
Closes #3392, 
Closes #3376, 
Closes #3393, 
Closes #3390, 
Closes #2491.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3679"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790021930&installation_model_id=427282&pr_number=3679&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3679&signature=34a80e43dd351d3ea37a9f70e48c10441c1b60982315d251c543675af2b1b1e3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


stijn-dejongh:
## 🛬 Maintainer landing pass — claimed

Picking this up for a landing pass.

**Viability triage → LAND.** This is a fresh, substantive maintainer PR whose base (`2621a56d06`) **is the current `upstream/main` tip** — zero commits behind. That makes the usual staleness questions moot by construction:

- **Superseded?** No — nothing has merged to main since the PR's base, so no later commit can have already landed this fix.
- **Compatible?** Yes — it is built on the exact current main tip; the three-dot diff touches only its intended surfaces (`cli/commands/upgrade.py`, the 16 generated-surface migrations, `runtime/generated_writer.py`, the finalizer/outcome seams, and their tests + changelog). No governance/charter smuggling.
- **Still needed?** Yes — it closes the confirmed #3651/#3654/#3653/#3392/#3376/#3393/#3390/#2491 defect cluster.

**Plan:** isolated worktree (already on the current tip — no rebase needed), classify every CI red into the four bins, campsite-fold any remediations as single-purpose `<type>(landing):` commits (subagent-delegated, model-disciplined), run a profile-loaded adversarial squad over the contract/API surface, and refresh deliverable hygiene (PR body BLUF is already strong; changelog present). Evidence to follow. I hand off green-modulo-honest-reds; the operator merges.

stijn-dejongh:
## 🛬 Landing pass complete — remediation summary (operator merges)

Rebased state: the PR base **is** the current `upstream/main` tip, so no rebase was needed. Pushed **6 landing folds** (fast-forward, `--force-with-lease`). Handing off green-modulo-CI; **I do not merge — you do.**

### CI red classification (four bins)
Only one check was red on the pushed tip:

| Check | Bin | Resolution |
|---|---|---|
| `lint` → *Validate patch() target strings (#394)* | **PR defect** | Fixed — fold 1 |
| everything else (all fast/integration shards, all 3 arch-adversarial shards, blocking regression, docs-freshness, ui-e2e, Windows, clean-install FR-010, builds) | — | already green |

The `lint` step bundles commitlint/markdownlint/ruff/mypy(advisory)/bandit/pip-audit — all of those passed; the *only* failing sub-step was the patch-target validator. Root cause: the new marker-head-scan test wrote its `patch()` target as two implicitly-concatenated string literals; the validator's regex reads only the first quoted chunk, so it saw a truncated target. Joined into a single literal.

### Adversarial squad (profile-loaded, read-only)
- **architect-alphonso → SOUND, no blockers.** Confirmed the finalizer layering (no `upgrade/`→`cli.commands` inversion), single-site exit-code derivation, and the default-deny consent gate all hold.
- **paula-patterns → one MAJOR:** the writer-routing this PR is *about* was left half-finished — 5 sibling migrations still wrote bare `write_text` to read-only command/skill surfaces, two of them a hard `PermissionError` crash (#3651 verbatim). Verified against the code before folding.

### Folds pushed
1. **`test(landing)`** — join the split `patch()` target literal so the #394 validator resolves it (fixes the one CI red).
2. **`fix(upgrade)` (MAJOR)** — route the remaining 5 migrations through `write_generated_file`; **red-first proof** for the 2 hard-crash sites (0o444 target → `PermissionError` before, applied-and-still-read-only after). The two shared helpers write both read-only agent surfaces *and* plain `.kittify` files, so they take a `read_only` kwarg (default `False`); `read_only=True` is passed only at agent-surface call sites — plain project files stay writable. **The guard is converted from a 16-file allowlist to a full `_MIGRATIONS_DIR` scan** with an explicit, rationale'd 3-file exemption set — so the invariant is *enforced*, not enumerated ("cannot be forgotten again" is now true).
3. **`fix(landing)`** — make `write_generated_file` atomic (temp sibling + `os.replace`, read-only strip in `finally`) and symlink-safe (refuse a symlink target); correct the module docstring's over-broad "every consumer" claim. +10 hardening tests.
4. **`fix(landing)`** — fold `worktree_failures` into the `--json` `errors` channel (deduped, no contract change) so a failed run never reports `success:false` with an empty `errors[]`. Consistency-hardening + regression guard (the no-migrations path doesn't populate that channel in current production code; proven red-first via a stubbed failure).
5. **`refactor(landing)` ×2 — mypy** *(per your "also remediate the mypy issues")* — cleared **all 38 pre-existing `mypy --strict` errors** (advisory step; none were in this PR's own files). Split into "outside `sync/`" and "in `sync/`" commits, behavior-preserving (read-only Protocol matching the frozen dataclass, canonical-module imports / explicit `__all__` re-exports, redundant-cast removals, and defensive narrowing over states the storage layer already constrains). **Full-tree `mypy --strict` is now clean (0 errors, 1208 files).** These two are unrelated to the upgrade command — if you'd rather keep this PR focused, they cherry-pick cleanly into a separate cleanup PR; say the word.

### Local verification
`ruff` clean · `mypy --strict` clean tree-wide · `commitlint` clean (all 10 commits) · patch-targets valid (5119) · terminology guard green · docs-freshness `errors=0` · full `tests/upgrade/` **783 passed, 2 skipped** · marker/collection/arch-shard completeness gates green (both new test files registered) · `tests/sync` unit suites for touched modules **141 passed**. Heavy suites left to CI.

### Changelog
No new entry needed — the existing `[Unreleased]` entry already states "every generated/rewritable-but-read-only file routes through one canonical writer"; the MAJOR fold makes that claim *actually* true, and the other folds harden internals with no new user-facing behavior.

Waiting on the re-run to go green. **Ready for you to merge once CI is green** — I'll verify the merge landed clean and clean up the worktree afterward.

stijn-dejongh:
## ✅ Hand-off — ready to merge (one advisory red, classified)

**CI: all required checks green; PR is `MERGEABLE`.** Every test shard (fast + integration, all shards), all 3 arch-adversarial shards, blocking regression, docs-freshness, ui-e2e, Windows-critical, clean-install (FR-010), lint, and the builds pass on the folded tip.

### The one red is advisory, not a real coverage gap
`diff-coverage` (and `quality-gate`, which only fails because it `needs: diff-coverage`) are **advisory / non-required** — hence `mergeStateStatus: UNSTABLE`, not `BLOCKED`.

Its sole complaint:
```
src/specify_cli/status/review_result_parse.py (0.0%): Missing lines 42
```
That line is **genuinely 100% covered** — the covering test (`test_emit_review_result.py`) exercises it. Reproduced locally with real coverage data:
```
$ diff-cover <coverage.xml> --compare-branch=upstream/main --fail-under=90 --include <critical-paths>
src/specify_cli/status/review_result_parse.py (100%)
Total:   1 line   Missing: 0 lines   Coverage: 100%
```
Root cause in CI: the enforced diff-coverage gate only covers **critical-path** modules (`status/`, `doctrine/`, `charter/`, `merge/`, `next/`, …). The upgrade PR + squad folds touch **none** of those — only the orthogonal **mypy** campsite fold touched `status/review_result_parse.py`, dragging it into the gate. Then CI's coverage-artifact download hit a **`digest-mismatch: error`**, so the shard that covers line 42 didn't aggregate and a covered line read as 0%. It's a cross-shard artifact flake, not a coverage regression — and I won't retry-to-green to paper over it.

**Decision (operator):** keep the mypy folds in this PR and merge with the advisory red. Documented here so it isn't re-investigated.

### Merge-ready
All 6 folds verified locally: full-tree `mypy --strict` clean (1208 files), `ruff` clean, `commitlint` 10/10, terminology + docs-freshness green, `tests/upgrade` 783 passed, `tests/sync` units 141 passed, marker/collection completeness gates green.

**Over to you to merge — I don't.** After you confirm it's merged I'll verify the folds landed clean (no new red-main) and clean up the landing worktree, local branch, and the merged fork branch.

Issue #3622, opened 2026-08-21
Make analysis_report/acceptance artifact-name constants lazy (defer import-time manifest read)


stijn-dejongh:
## Context

Follow-up from M3 (rc3-charter-gate-predicate-inversion, PR #3617, #3599). WP04 converted the hardcoded artifact-name literals in `src/specify_cli/analysis_report.py` (`_HASH_INPUTS`) and `src/specify_cli/acceptance/__init__.py` (`SPEC_FILE`/`PLAN_FILE`/`TASKS_FILE`) to resolve from the single `expected-artifacts.yaml` `path_pattern` authority via `resolve_configured_artifact_name(...)`.

## Problem (pre-merge review, architect-alphonso, MAJOR)

These are **module-level** constants, so the resolution runs at **import time**. Two consequences for two widely-imported CLI modules:
1. **New import-time failure mode:** if the built-in `expected-artifacts.yaml` is ever malformed, `import specify_cli.acceptance` / `import specify_cli.analysis_report` raise `ArtifactNameConfigurationError` at collection/import time rather than at point-of-use — a broad blast radius for a data fault (previously the literals were inert).
2. **Redundant loads:** the manifest is loaded+validated once per resolved key for the same `software-dev` type (`_load_expected_artifact_manifest` is uncached).

Low-probability for built-in doctrine (caught elsewhere) and WP04 review judged the current form defensible, so it shipped in M3.

## Proposed fix

Make the constants **lazy** via a module `__getattr__` (PEP 562) so the resolution defers to first access (preserving the `from acceptance import SPEC_FILE` API), which eliminates the import-time failure mode. Optionally memoize `_load_expected_artifact_manifest` on `mission_type` (note: a naive `lru_cache` breaks the AC-9 load-bearing test's patch+reload pattern in `tests/specify_cli/runtime/test_configured_artifact_name.py`, which must be adapted to clear the cache — see M3 PR discussion).

Priority: low (P3). No user-visible impact for shipped built-ins.

robertDouglass:
Claimed for implementation in governed Spec Kitty Op `01M0MMZBB7FB197HZCMCZN4TAQ` as part of the current community-PR consolidation pass. The submitted PR is being rebased/review-fixed on current main; final evidence and superseding PR link will follow.

PR #3690, opened 2026-08-23
fix(dossier): widen positional-call guard, trim CLI re-exports, unbreak analyze commitlint

### LINKED ISSUE #3676: AST positional-call guard misses attribute-chain and aliased calls, and dossier/__init__.py advertises exactly that surface
Follow-up from #3672 (mission for #1058), filed per the landing pass's follow-up hygiene rule.

## What

`tests/architectural/test_dossier_emitter_positional_guard.py` (added in #3672) walks `src/` and flags positional calls to the four dossier emitters. It matches **bare `ast.Name` calls only**. It does not match:

- attribute-chain calls — `dossier.emit_artifact_indexed(...)`
- aliased imports — `from ... import emit_artifact_indexed as emit`

## Why it matters more than it looks

`src/specify_cli/dossier/__init__.py` re-exports all four emitters in `__all__`. So the package **actively advertises** the attribute-access form the guard cannot see.

The guard is load-bearing: #3672 removed the legacy positional bridge, so a positional call is now a `TypeError` — and per FR-004 that `TypeError` is swallowed by a broad `except Exception` in `sync/dossier_pipeline.py`. The guard is what stands between a positional call and a silent production failure.

**No such call exists today** — verified across `src/` and `tests/` at the time of filing. This is a gap in coverage, not a live defect.

## Scope note

This boundary was deliberate and spec-sanctioned in #1058's plan ("do not attempt full call-graph resolution or handle re-exported/aliased imports"), and the guard documents it in its own module docstring. Filing so the residual risk has a tracked home rather than living only in a PR comment.

## Fix directions

1. Extend the detector to resolve attribute-chain calls against known module aliases and `import ... as` bindings; or
2. trim the seven re-exports from `dossier/__init__.py`'s `__all__` so the unguarded surface stops being advertised (see the sibling issue).

Either narrows the gap; (2) is smaller but is a public-API change.

### LINKED ISSUE #3677: dossier/__init__.py re-exports external spec-kitty-events types under the CLI namespace, creating two import paths to one canonical type
Follow-up from #3672 (mission for #1058), filed per the landing pass's follow-up hygiene rule.

## What

`src/specify_cli/dossier/__init__.py` re-exports seven names in `__all__`: `ArtifactIdentity`, `ContentHashRef`, `LocalNamespaceTuple`, and the four `MissionDossier*Payload` classes.

Before #3672 those were **locally-owned Pydantic mirrors**, so re-exporting them was ordinary. #3672 deleted the mirrors and `dossier/events.py` now imports the canonical classes from `spec_kitty_events`. The re-export therefore now **transitively republishes an external package's classes under `specify_cli.dossier.*`**, giving two import paths to one canonical type.

## Why it is worth a follow-up

The charter's §Architecture: Shared Package Boundaries treats `spec-kitty-events` as a true external contract package — *"consume released PyPI packages through the normal dependency graph"*, *"do not vendor their source into the CLI package"*. No literal vendoring happens here, but name-aliasing an external contract under our namespace invites future code to treat the CLI package as if it republishes that contract, and couples callers to our pinned version through an unexpected path.

## The surface is currently dead

Verified at filing: **nothing** in `src/` or `tests/` imports any of the seven via `specify_cli.dossier`. Every real caller imports `.events` directly or `spec_kitty_events` directly.

## Why it was not folded into #3672

Trimming a public `__all__` is an API change, and #1058's scope was the emitter migration. Narrowing a published surface during a landing pass would have been scope widening; it deserves its own decision.

## Fix direction

Remove the seven canonical-type names from `dossier/__init__.py`'s `__all__` (keep the four `emit_*` functions, which are genuinely this package's API). This also shrinks the unguarded surface described in the sibling AST-guard issue.

### LINKED ISSUE #3678: record-analysis emits a commit subject the commitlint tool-commit ignore list does not cover, so every mission that runs analyze reds the gate
Filed from the #3672 landing pass (mission for #1058), where it cost two remediation commits.

## What

`commitlint.config.cjs` deliberately exempts spec-kitty's auto-generated lifecycle commits:

```js
ignores: [ (commit) => /^(Add|Update) (meta|spec|tasks|plan) for (feature|mission) /.test(commit) ]
```

with the comment *"spec-kitty auto-generates mission lifecycle commits that don't use conventional-commit format. Ignore them rather than requiring every planning command to produce a typed subject line."*

**`spec-kitty agent mission record-analysis` emits `"Add analysis report for mission <slug>"`.** "analysis report" is not in that alternation, so the commit is not ignored and fails:

```
⧗   input: Add analysis report for mission <slug>
✖   type may not be empty [type-empty]
✖   subject may not be empty [subject-empty]
```

## Why it stays invisible

commitlint only lints commits **in the PR range**, so main's own history never exercises the gap. It surfaces only on a PR carrying a `record-analysis` commit — i.e. every mission that reaches the analyze phase and opens a PR. On #3672, exactly 2 of 52 commits failed, both this message (one per analyze round). Every other tool-authored commit in the range passed, including the `Add meta`/`Add tasks` ones and all `chore(spec-kitty): status transition WPxx` commits.

## How #3672 handled it

The two commits were **reworded** to `docs(mission): record analysis report for <slug>`. Widening the shared ignore regex would also have worked, but loosening a repo-wide lint gate during a landing pass is the wrong trade — so the PR fixed its own commits and left the gate alone.

## Fix directions, in preference order

1. **Have `record-analysis` emit a conforming subject** — e.g. `docs(mission): record analysis report for <slug>`. Fixes the cause, needs no ignore-list growth, and stops every future mission hitting this.
2. Extend the ignore regex to cover the verb phrase, with a dated rationale. Cheaper, but grows an exemption list rather than fixing the emitter.

Option 1 is preferred: the other lifecycle commands already produce subjects the repo is content to exempt, whereas this one is a lone straggler that could simply conform.

MOES-Media:
## What changes

Three things stop being broken for anyone running a spec-kitty mission:

1. **Running `analyze` no longer reds the commitlint gate.** `record-analysis` emitted `Add analysis report for mission <slug>`, which `commitlint.config.cjs`'s ignore list does not cover — so every mission that reached analyze failed the gate on its own tooling's commit. It now emits a conventional-commit subject that passes via `type-enum`. (#3678)
2. **The dossier positional-call guard catches the calls it was missing.** It matched only bare-name callees, so `dossier.emit_artifact_indexed(...)` (attribute-chain) and `from ... import emit_artifact_indexed as ei; ei(...)` (aliased) escaped it entirely — and `dossier/__init__.py` re-exported exactly that surface. The guard now resolves both shapes, with positive controls for each and negative controls proving it does not fire on correct keyword-only calls. (#3676)
3. **`analysis-report.md` no longer records absolute filesystem paths.** It wrote the operator's home directory and checkout layout into a committed, publicly visible artifact, four times per report. Recorded paths are now repo-relative, and the failure path raises loudly instead of silently writing a leaking path. (ledger SK-63; no upstream issue)

### Breaking change

`specify_cli.dossier` no longer re-exports seven `spec_kitty_events` types — `ArtifactIdentity`, `ContentHashRef`, `LocalNamespaceTuple`, and the four `MissionDossier*Payload` classes. Each type now has one canonical import path instead of two.

`from specify_cli.dossier import ArtifactIdentity` now raises `ImportError`; import from `spec_kitty_events` instead. No caller of the old path exists in `src/`, `tests/`, or the sibling SaaS worktrees. The four `emit_*` function re-exports are unchanged. (#3677)

## Reviewer notes

**This branch is stacked on #3672, which is not yet merged.** Its base is that PR's head, so the diff against `main` includes #3672's commits. Review this mission's work with `git diff 03cda7bbf..HEAD` — 519 insertions across 9 files. Once #3672 merges, this rebases onto clean `main`.

**Scope** is four defects: three issues plus one ledger-sourced defect folded in because it lives in the same file as #3678 and was publishing host paths on a public repo.

## Test strategy

Red-first per work package, per charter §591/C-011: the failing test is its own commit before the implementation, and red→green was reproduced independently by a separate reviewer for each. One honest exception — the #3677 regression test was authored *after* its implementation during a rejection rework; its ordering is retrospective and its commit message says so.

Baseline before this mission: 472 passed, 0 failed across every file it touches. After: green, with the new tests included.

## Known gaps

- Subtask status flags are unrecorded for this mission. `spec-kitty agent tasks mark-status` hangs in an unbounded sync-retry loop on this machine (4 attempts, all terminated), as do `finalize-tasks` (5) and `agent action implement` (2). Recorded as ledger SK-74; WP lane states were set via `agent status emit --force` with the reason stated on each event.
- Live commitlint verification for #3678 was run manually against the new subject rather than encoded as a test — this module's `unit`/`fast` markers exclude subprocess and network access.

Closes #3676
Closes #3677
Closes #3678

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3690"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790068904&installation_model_id=427282&pr_number=3690&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3690&signature=9c95132bb24d370c9029facf04151838bdb204ff3587400243315983e59c6d72"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
## Landing pass — starting

**Landing-order constraint: this PR must merge _after_ #3672, and must not be rebased onto `main` until #3672 lands.**

This branch is based on #3672's head, not on `main`. `tests/architectural/test_dossier_emitter_positional_guard.py` does not exist on `main` — #3672 creates it, and this PR widens it. Rebasing onto `main` today would leave those changes with no base. The usual "rebase onto current main first" step is therefore deliberately skipped, not overlooked; it happens once #3672 merges.

**Reviewing this PR:** use `git diff 03cda7bbf..HEAD` (519 insertions, 9 files). The diff against `main` additionally contains #3672's 52 commits.

**Known upstream drift to expect at rebase time:** `a69c2f04a` (#3622) converted `_HASH_INPUTS` in `src/specify_cli/analysis_report.py` into `_hash_inputs()` plus a PEP 562 module `__getattr__`, changing call sites in three functions this PR also edits. The changes are semantically independent — this PR touches recorded path *values*, that commit touches constant *resolution* — so the conflicts should be hunk-adjacent and mechanical, but they are expected rather than a surprise.

Marking ready for review now: this repo skips its CI suite on draft PRs, so no checks run until it is un-drafted. Evidence trail to follow.

robertDouglass:
3.2.6 execution DAG: #3692. Wave 1, dossier lane step 2. Upstream: #3127 and PR #3672.

robertDouglass:
Pre-merge adversarial squad follow-up (architect-alphonso, debugger-debbie, reviewer-renata): all three lenses independently confirmed two blockers in analysis_report.py. Resolved-path containment did not reject symlink escapes, and chained ValueError tracebacks re-exposed absolute paths. Commit 528bbea2b now validates resolved containment while preserving logical in-repo paths, suppresses the leaking exception chain, and adds write, freshness, traceback, external-symlink, and in-repo-symlink regressions. Focused second-opinion verdict: APPROVE. Local verification: 113 passed; targeted Ruff clean; mypy --strict clean; git diff --check clean.

PR #3693, opened 2026-08-23
fix(doctrine): close the declared-but-inert DRG silent-drop family (#3608, #3629, #3530)

### LINKED ISSUE #3530: Tracking: org-tier doctrine loads, validates and reports healthy — and reaches no consumer

### LINKED ISSUE #3608: Tech-debt: _DRG_NODE_KINDS is a hand-copied NodeKind that already drifted 7 kinds, silently dropping mission_type/glossary_pack DRG-URN resolution
## Defect

`_DRG_NODE_KINDS` in `src/charter/synthesizer/topic_resolver.py:37` is a **hand-maintained copy** of the canonical `NodeKind` enum (`src/doctrine/drg/models.py:43`). Its own comment (line 36) claims it is a *"superset, from doctrine.drg.models.NodeKind"* — but the copy has **already drifted**.

Verified against `main` (`b28c5a9`):

- `NodeKind` defines **16** kinds.
- `_DRG_NODE_KINDS` lists only **9**.
- **7 kinds are missing** from the copy:
  `anti_pattern`, `asset`, `glossary`, `glossary_pack`, `mission_step_contract`, `mission_type`, `template`.

(Drift computed directly:
`set(NodeKind) - _DRG_NODE_KINDS == {anti_pattern, asset, glossary, glossary_pack, mission_step_contract, mission_type, template}`; the copy adds nothing extra.)

## Silent-failure consequence

The DRG-URN resolution tier at `src/charter/synthesizer/topic_resolver.py:235` gates on membership:

```python
# LHS must be a known DRG node kind
if lhs not in _DRG_NODE_KINDS:
    return None
```

A selector such as `mission_type:<id>` or `glossary_pack:<id>` — both of which are legitimate, charter-activatable DRG-URN kinds — is **not recognized** as a DRG node kind, so this resolution tier returns `None` and the URN **silently falls through** to the next tier (or fails to resolve) with no error. The failure is invisible: no exception, no diagnostic. Any newly-added `NodeKind` member has the same fate until someone remembers to hand-edit the copy.

There is **no drift-guard test** pinning the copy to `NodeKind`. (`tests/doctrine/drg/test_unknown_kind_fails_loudly.py` exists but does not reference `_DRG_NODE_KINDS` or `topic_resolver` — it is unrelated.)

## Fix direction

Derive the set from the enum instead of hand-copying, e.g.:

```python
from doctrine.drg.models import NodeKind
_DRG_NODE_KINDS: frozenset[str] = frozenset(k.value for k in NodeKind)
```

At minimum, add a drift-guard test asserting `_DRG_NODE_KINDS == {k.value for k in NodeKind}` so the copy cannot silently drift again. (Prefer derivation over a guarded copy — a guard only catches drift after it lands; derivation prevents it.)

## Family / provenance

Same **whack-a-copy / duplicated-constant** anti-pattern as open issues #3562, #3461, #3427.

Surfaced by the "processing & charter friction" investigation (`docs/plans/investigations/friction-bugs-processing-charter-root-cause.md`, §11) and the rc3 burndown (PR #3607). Intentionally left as a separate latent-defect ticket — **not** folded into any rc3 mission.

### LINKED ISSUE #3629: Follow-up (M2 DRG projection): context-sources dead-field family + fail-loud on bad governance-profile selections
Follow-up items from the M2 DRG-projection-completeness pre-merge adversarial squad (PR #3628). None block M2; captured here per the squad protocol (fold MAJORs, file NOTE/MINOR as one follow-up).

## 1. `context-sources` dead-field family (NOTE, from both the post-plan and pre-merge squads)
`ContextSources` (`src/doctrine/agent_profiles/profile.py`) declares `context-sources.{doctrine-layers, tactics, toolguides, styleguides, additional}`, and the schema (`agent-profile.schema.yaml`) legitimizes them — but the pydantic `context_sources` attribute is **never read**. The extractor reads the raw YAML `.get("directives", [])` only; `.tactics`/`.toolguides`/`.styleguides`/`.doctrine-layers`/`.additional` reach no delivery path. Real shipped profiles author these (e.g. `debugger-debbie` `context-sources.tactics`, `architect-alphonso` `context-sources.additional: [architecture-decision-records, …]`) — silently inert.

It is **out of scope for M2's FR-008 seam-bind** because these fields never enter the DRG at all (they vanish at the YAML→DRG boundary, before the emit↔delivery seam FR-008 binds). Decide: wire them into projection+delivery, or deprecate/remove the dead fields (with a schema change + migration).

## 2. No fail-loud on nonexistent governance-profile selections (NOTE, architecture lens)
`extract_governance_profile_scope_edges` builds a target URN from each `selected_*` bare id with **no existence check**; a typo'd/nonexistent id mints a dangling `scope` edge that `generate_graph` silently prunes rather than flagging. Consistent with the existing extractor posture (not an M2 regression), but a validation guard (fail-loud, à la `_run_operating_procedures_check`) would catch author typos in a mission type's `governance-profile.yaml`.

## 3. Minor doc nit
The extractor procedure-branch docstring phrasing around "no golden re-ledger required" is about triple-set stability; the serialized golden *did* change (added `reason:` values, re-ledgered in M2 WP04). Clarify wording on a future touch.

Refs: PR #3628 (M2, rc3 fail-loud burndown / epic #3410).

stijn-dejongh:
## What this fixes (for operators and pack authors)

Doctrine you authored could be **schema-valid, reported healthy, and silently never take effect**. This PR closes that whole class of "declared-but-inert" failures on the doctrine/DRG boundary — each declaration now either takes effect or fails loudly, and is pinned by a test. Concretely:

- **You authored `context-sources` on an agent profile and it did nothing.** The block duplicated the canonical `*-references` fields and was mostly dropped on delivery. **Now:** references live solely on `*-references`; authoring `context-sources` **fails to load loudly**; and `spec-kitty upgrade` migrates existing consumer profiles automatically (set-merges every id onto the matching `*-references`, deduped). *(Breaking — see below.)*
- **A DRG URN for six artifact kinds (`anti_pattern`/`asset`/`glossary`/`glossary_pack`/`mission_step_contract`/`template`) silently failed to resolve.** A hand-copied kind set had drifted behind the enum. **Now:** derived from `NodeKind`, with a behaviour-pinned drift guard.
- **An org pack that shipped `drg/fragment.yaml` had its doctrine silently dropped at runtime.** Two runtime seams ignored the fragment. **Now:** both fold it in; a malformed fragment fails loud instead of vanishing.
- **A governance profile that selected an artifact that doesn't exist was a silent no-op.** **Now:** it fails loud at both the built-in and org tiers.
- **`spec-kitty upgrade --dry-run` on an already-current project under-reported its work** (landing fold): it said "Project is already up to date!" while hiding the `mission_type_activations` seed a real upgrade would still perform. **Now:** the dry-run previews that seed on the up-to-date path too.

Milestone 3.2.x (G1 deepen Doctrine/Charter/DRG runtime impact; G2 strangle core domains onto canonical SSOTs; no new shadow paths). Related fail-loud epic #3410.

Closes #3608 · Closes #3629 · Closes #3530

## 💥 Breaking

Consumer projects that authored `context-sources` in **custom** agent profiles are migrated automatically by `spec-kitty upgrade` (idempotent, lossless set-merge onto `*-references`; edge-less `additional` / `doctrine-layers` names are reported, not silently dropped). After upgrade, authoring a `context-sources` block is rejected at load (`extra="forbid"`). Version bump `3.2.6rc3` → `rc4`; CHANGELOG entry included.

## What changed (per commit)

| Commit | Issue | Change |
|--------|-------|--------|
| `feat(doctrine): derive DRG node-kind set from NodeKind enum` | #3608 | `_DRG_NODE_KINDS` (`charter/synthesizer/topic_resolver.py`) was a hand-copied frozenset drifted 6 kinds behind `NodeKind`. Derive it from the enum (mirroring `merge.py`'s twin) + a **behaviour-pinned** drift-guard. |
| `refactor(doctrine): consolidate agent-profile references on *-references, remove context-sources` | #3629 (p1/p3) | Removed the redundant, mostly-inert `context-sources` block from the model + schema (`extra="forbid"` now rejects it loudly); the extractor projects `agent_profile` DRG edges from `*-references`; 25 shipped profiles migrated + an idempotent upgrade migration; golden `agent_profile.graph.yaml` regenerated (the pedro/DIRECTIVE_034 diamond + diagram-daisy toolguide edges are the deliberately-ledgered 3-edge delta). |
| `feat(doctrine): fail loud on unresolved governance-profile selections, both tiers` | #3629 (p2) | End-to-end pin for the built-in guard; net-new org-tier path (`org_governance.py` + `org_pack_loader`) so an org-pack governance-profile selecting a nonexistent artifact fails loud. |
| `fix(doctrine): fold org drg/fragment.yaml at the executor and action-doctrine-bundle seams` | #3530 | `load_validated_graph`'s `org_roots=` seam ignored `drg/fragment.yaml`; fixed at the two callers (thread `org_fragments`, mirroring the correct dual-callers). |
| `test(doctrine): verify org-pack chain delivery and misconfig fail-loud` | #3530 | Evidence: built-in + spec-kitty-internal delivers every declared kind (class-b); a 2nd minimal org fixture proves fragments past the first fold (class-a); 3 enumerated misconfigs fail loud. |
| `docs(mission)` | — | Mission record + reconciliation of stale `context-sources` references. |

## Provenance & process

Built via the full Spec Kitty mission workflow (specify → plan → tasks → implement → review → merge), with **adversarial brownfield review squads at the post-plan and post-tasks point-cuts** and a **3-lens post-merge fidelity audit**. Findings + dispositions: `kitty-specs/doctrine-drg-silent-drop-boundary-01M0PE7E/research/`.

## Testing

Targeted suites green throughout (node-kinds, profile/migration/projection, governance fail-loud, org-fragment executor/bundle, chain delivery, terminology guard). Breaking schema change carries a CHANGELOG entry + version bump and an idempotent, lossless `m_3_3_1` upgrade migration. `ruff` clean; `mypy --strict` no new issues.

## Landing notes (maintainer folds)

Rebased onto current `upstream/main`; every red check classified and folded (all folds are `<type>(landing): …`, commitlint-clean):

- **Version-bump propagation** — re-locked `uv.lock` and bumped `.kittify/metadata.yaml` to `rc4` (tripped `uv lock --check` / `check-readiness`).
- **Docs gates** — the new research doc lacked frontmatter and inventory/retrieval-index rows (`build` / `docs-freshness`); registered it. Reconciled four active docs (incl. the `doctrine-kinds.md` AgentProfileSchema **diagram** the drift guard flagged) that still described the removed `context-sources`.
- **Architectural boundary** — the executor imported org-pack symbols straight from `doctrine.*`, crossing the runtime→charter→doctrine gate; routed them through `charter.drg` (also cleared the doctrine-census undoored-path finding). Dropped the inert `OrgGovernanceScopeEdge` from `__all__` (dead-symbol gate).
- **CI topology** — wired the PR's new test files (`pytestmark` + corpus registry) and the `m_3_3_1` migration (auto-discovered-migration allowlist) into the completeness/dead-module gates.
- **Dry-run parity fix (with red-first proof)** — the `rc4` bump exposed a latent bug: `upgrade --dry-run` on an up-to-date project didn't preview the `mission_type_activations` seed. Fixed on the no-migrations path; changelog entry added.


stijn-dejongh:
🛬 **Maintainer landing pass** — claiming #3693 for today's landing pass.

Plan: isolated worktree, rebase onto current `upstream/main`, classify every red check into the four bins (PR defect / contract crossed / pre-existing main breakage / perf flake), fold fixes as needed, refresh PR body + changelog, and hand off green-modulo-honest-reds for the operator to merge. Evidence to follow.

stijn-dejongh:
## 🛬 Maintainer landing summary

Rebased onto current `upstream/main` (was 1 commit behind; clean rebase). Every red check classified and folded — **10 landing folds**, each single-purpose and `<type>(landing): …` (commitlint-clean over the whole PR range). Verdict: **land-ready pending green CI.**

### Red-check classification & remediation

| Red check(s) | Bin | Fold |
|---|---|---|
| `uv lock --check`, `check-readiness` | PR defect (version bump not propagated) | `chore(landing)`: re-lock `uv.lock`; bump `.kittify/metadata.yaml` to `rc4` |
| `build`, `docs-freshness`, `fast-tests-docs` (seo/structural/description) | PR defect (new doc unregistered) | `docs(landing)`: frontmatter + page-inventory + retrieval-index for the research doc |
| `fast-tests-docs` (diagram drift), `unit-contract-residual` | PR defect (diagram + prose stale) | `docs(landing)`: drop `context-sources` from the `doctrine-kinds.md` AgentProfileSchema diagram; reconcile 4 active docs |
| `doctrine-charter-tests`, `fast-tests-charter` | Contract the PR legitimately crosses | `test(landing)`: the graphless-pack test shipped a placeholder `fragment.yaml` on the pre-#3530 "fragment is invisible to runtime" assumption; #3530 now reads it. Ship no fragment so it isolates the true per-root-degrade case (positive coverage of "fragment is read" is in the PR's new bundle/executor tests) |
| `arch-adversarial` (runtime→doctrine boundary + doctrine census) | PR defect | `fix(landing)`: route the executor's org-pack imports through `charter.drg` (also clears the census undoored-path finding) |
| `arch-adversarial` (dead symbol) | PR defect (inert export) | `refactor(landing)`: drop `OrgGovernanceScopeEdge` from `__all__` |
| `arch-adversarial` (dead modules / corpus registry / collection & same-tier completeness) | PR defect (new files/module unwired) | `test(landing)`: `pytestmark` + corpus registry for the migration test; corpus registry for the two governance tests; `m_3_3_1` migration → auto-discovered-migration allowlist |
| `fast-tests-core-misc` (`test_dry_run_reflects_pending_provisioning`) | Pre-existing latent bug surfaced by the `rc4` bump | `fix(landing)`: preview pending provisioning on the up-to-date dry-run path (**red-first proof below**) + `docs(landing)` changelog entry |
| `quality-gate` | Aggregate of the above | resolved transitively |

### Red-first proof (dry-run parity fold)

`finalize_upgrade` provisions `mission_type_activations` on **both** the migration and no-migrations paths, but the dry-run notice was only emitted from the migration-plan path. The parity test writes its fixture at the CLI's current version, so at `rc4` (no pending migrations) it exercises the up-to-date path and the bug turns hard-red.

```
# fix reverted, SPEC_KITTY_CLI_VERSION=3.2.6rc4:
FAILED test_dry_run_reflects_pending_provisioning
  assert 'Would provision missing mission_type_activations' in
         'Current version: 3.2.6rc4\nTarget version:  3.2.6rc4\n\nProject is already up to date!\n'
# fix restored:
4 passed
```

### Local verification (targeted; CI is the authority for the rest)

- Docs: `docs_structural_lint` 0 violations; `check_docs_freshness --ci` errors=0 (external link-health WARNINGS only); diagram-drift + seo + description gates green.
- Arch gates: runtime↔doctrine boundary, doctrine-census (8), dead-symbols, dead-modules, corpus-trigger-completeness, collection-completeness, same-tier-uniqueness (5), marker-job/shard completeness (16) — all green.
- Charter: the graphless-pack org-chain test green.
- `ruff` clean; `mypy --strict` clean (CI-style whole-package invocation).
- `commitlint` clean over the full PR range.
- Terminology guard green (touched doctrine prose).

### Notes

- **Advisory-only (non-blocking) markdownlint** reports pre-existing `MD049` findings in unrelated older CHANGELOG entries (line ~86, `*typeless*` style) — not introduced here and out of scope for this landing; my added changelog entry is clean.
- The ADR `2026-07-26-1` `context-sources` census counts are a point-in-time historical record and were deliberately left unchanged.

Watching CI; **operator merges** on green.

stijn-dejongh:
✅ **CI green — merge-ready.** 55 checks SUCCESS, 29 skipped, 0 failures; PR is MERGEABLE and un-drafted.

The second re-run surfaced two secondary companion-gates triggered by the first fold batch (facade identity table for the `charter.drg` re-exports; the ratchet decorative count for the `m_3_3_1` migration) — both folded (`test(landing): sync companion audit tables …`) and now green.

**11 landing folds total**, each single-purpose and commitlint-clean. Per the runbook, **the operator merges** — handing off. Once merged I'll verify the merge landed clean and clean up the worktree/branches.

stijn-dejongh:
## Second-opinion review — independent 4-lens squad (post-rebase)

Rebased onto current `main` (`e744a2dda2`, +99 commits since the original base). 4 conflicts resolved (CHANGELOG ×2, `upgrade.py`, `_baselines.yaml`) — all verified. Then an independent adversarial squad re-reviewed the diff across four lenses. **No critical / high / medium findings. Clean bill for merge.**

**DRG core mechanism** — SOUND. #3608 NodeKind derivation (`frozenset(k.value for k in NodeKind)`) is exhaustive by construction and fixes the real 6-kind drift; the guard test monkeypatches a synthetic NodeKind member (non-tautological). #3530 org-fragment fold reaches delivery end-to-end via `load_org_drg → load_validated_graph → merge_three_layers` (SCOPE is a canonical `Relation`, folded like every other) — not parse-then-inert.

**Bulk-edit fidelity (#3629)** — CLEAN. `context-sources → *-references` applied completely across all 25 profiles with full parity (removed blocks were redundant duplicates); schema + pydantic model reject old-format loudly (`extra="forbid"` / `additionalProperties: false`); the `m_3_3_1_context_sources_consolidation` consumer migration is real (set-merges + synthesizes missing ids + audits dropped ids), not a stub. No surviving `context-sources` reader of the profile surface.

**Fail-loud & tests** — SOUND, fail-CLOSED on the authoritative paths (unresolved governance-profile selection → dangling-edge `DRGValidationError` naming the offender; misconfigured org pack → typed raise, contrasted against the honest graphless WARN). Tests are non-vacuous and prove end-to-end delivery (`test_org_pack_chain_delivery` uses a second pack + provenance markers + negative controls).

**Rebase integrity** — CLEAN. All 3 code/baseline conflict resolutions correct (`_baselines.yaml: category_1 = 104` cross-checked three ways; `upgrade.py` provisioning notice wired on both paths; CHANGELOG union-merged with no drops/dupes). rc4 version sites consistent. `tests/architectural/` + `tests/charter/` → **2446 passed, 0 failed**.

### Known tradeoffs (LOW / by-design — recorded, non-blocking)
- **Executor tier intentionally degrades a schema-malformed *optional* org fragment** (`_load_org_fragments_degrading`, `executor.py`) to a DEBUG + empty layer rather than raising — a deliberate "don't brick every mission step over a broken optional pack" choice. Endpoint/governance faults still raise even on this path; the authoritative APIs (doctor / charter / gate-bindings) fail loud. Two reviewers independently noted that in a **multi-pack** chain, one malformed fragment currently drops the whole fragment layer (healthy siblings included) — a candidate follow-up: degrade per-pack so a bad pack can't evict its healthy siblings, and drop the misleading "folding it" DEBUG when the layer was actually dropped.
- Directive `requires` edges are minted bare (no `reason`) for byte-parity with the retired projection — metadata-only, not a drop.
- No `occurrence_map.yaml` was recorded for this 25-file bulk edit; correctness was independently verified regardless — an audit-trail gap only.

Handing off for the operator to merge.

stijn-dejongh:
## Convergent LOW finding remediated

Folded in the per-pack org-fragment degrade (the finding two lenses converged on). Commit `4f20508e93`.

**What changed:** `load_org_drg` gains an explicit `degrade_malformed` opt-in (meaningful only when `strict=False`). A pack whose *optional* `drg/fragment.yaml` is malformed is now skipped **per-pack** with an operator-visible `WARNING` naming it — the healthy siblings' fragments still compose, instead of the whole fragment layer being dropped. The mission-step executor opts in; **every other caller stays byte-identical** and still raises on a malformed pack, and `strict=True` never degrades. The misleading "folding it" DEBUG is corrected.

**Why the opt-in (not a blanket `strict=False` degrade):** an existing in-mission fail-loud contract test (`test_misconfigured_pack_in_chain_fails_loud` → `load_org_drg(strict=False)`) asserts a malformed fragment **raises** on the plain non-strict path. A blanket degrade would have silently weakened that contract and the authoritative cascade/gate callers. The flag keeps the degrade narrow to exactly the executor's already-tolerated optional-fragment class.

**Verification (targeted, independent re-run):** 34 executor/fragment/chain-delivery tests (incl. the fail-loud contracts) + 4 new regression tests (healthy-sibling-survives, all-healthy-no-warning, strict-still-raises, default-`strict=False`-still-raises); 27 dead-module/symbol guards; 177 facade/boundary guards — all green. `ruff`/`mypy` clean (2 pre-existing facade-`Any` errors campsite-fixed).

Endpoint/dangling-governance faults still raise even on the non-strict path (unaffected — they surface downstream in `load_validated_graph`). Remaining noted follow-up (audit-only, non-blocking): no `occurrence_map.yaml` was recorded for the 25-file bulk edit.

stijn-dejongh:
**Correction to my previous comment.** This PR merged before the per-pack degrade fix was pushed, so commit `4f20508e93` referenced above is **not** part of the merged #3693 — the merged content has the whole-chain degrade behaviour. The fix has been rebased onto current `main` and opened as a follow-up: **#3762**. Apologies for the premature 'remediated' wording — the code and verification are identical, just landing in the follow-up PR instead of here.

Issue #3704, opened 2026-08-24
A custom mission family can never report a guard failure, and its `blocking:` manifest is never read


MOES-Media:
## BLUF

A custom mission family reports **zero guard failures at every step, always** — not because it behaved, but because `_GUARD_TABLES` is a module-level dict literal with four hardcoded keys and three separate call sites catch the resulting `UnregisteredMissionFamilyError` and degrade to `[]`. The runtime's own comments designate the composed-action guard as "the authority" for "every custom mission type" and call it "the fail-CLOSED contract"; for every custom mission type it is unconditionally fail-open.

Underneath it, the artifact-presence half is inert in the same direction: the wired gate reads the manifest from the **built-in pack only** (so an org pack's manifest is invisible), unions every `path_pattern` in it regardless of step, and **ignores `blocking:` entirely** — while `required_artifacts_for`, which reads `blocking:` correctly and is step-scoped, has zero production callers and is excluded from `__all__` pending a wiring that never landed.

These are one pipeline one stage apart — `evaluate_guards_strict(snapshot: _ArtifactPresenceSnapshotLike)` takes the presence snapshot as its input — and either half alone still produces a silent pass. They are filed together deliberately.

## Builds measured

`spec-kitty-cli 3.2.6rc3` from `origin/main` `d060cff9a`, and the same version pinned at `8a03733e`. Every measurement below was identical on both. Line numbers are from `d060cff9a`.

---

## Part 1 — the guard table has no row for a custom family, and three call sites swallow it

`src/runtime/next/runtime_bridge_cores.py`

```python
637  # ... ``_GUARD_TABLES`` is a module-level dict literal
638  # built at import time from direct function-object references, not a lazy
639  # lookup -- it cannot forward-reference names not yet bound.
...
676  _GUARD_TABLES: dict[str, Callable[[_ArtifactPresenceSnapshotLike], list[str]]] = {
677      "research": _evaluate_research_guards,
678      "documentation": _evaluate_documentation_guards,
679      "software-dev": _evaluate_software_dev_guards,
680      "plan": _evaluate_plan_guards,
681  }
```

There is no registration seam. A pack-shipped mission type cannot be in this dict.

Measured:

```python
import importlib
m = importlib.import_module("runtime.next.runtime_bridge_cores")
print("GUARD_TABLES keys:", sorted(m._GUARD_TABLES))

class Snap:
    mission_family = "qa"; step = "accept"
    def __getattr__(self, n): return False

[...]

Issue #3703, opened 2026-08-24
Org-tier `expected-artifacts.yaml` is unreachable: the resolver anchors at `<org_root>/<type>/` while every other org mission asset anchors at `<org_root>/missions/<type>/`


MOES-Media:
## BLUF

`resolve_org_expected_artifacts` looks for an org pack's manifest at `<org_root>/<mission_type>/expected-artifacts.yaml`, while every other org-tier mission asset — templates, `mission.yaml`, `mission-runtime.yaml` — anchors at `<org_root>/missions/<mission_type>/...`, mirroring the built-in pack's own `packs/built-in/missions/<type>/` layout. A pack laid out like the built-in pack therefore has its manifest silently ignored.

Consequence: the org tier added for `expected-artifacts.yaml` is unreachable for any pack that follows the only layout convention the codebase demonstrates. `ManifestRegistry.load_manifest(<type>, repo_root=...)` returns `None`, and no warning is emitted because a missing file is indistinguishable from "no org override intended".

## Where

`src/charter/org_expected_artifacts.py`

```python
79      result: Mapping[str, Any] | None = None
80      for org_root in org_roots:
81          path = org_root / mission_type / _EXPECTED_ARTIFACTS_FILENAME
82          parsed = _read_yaml_mapping(path)
```

Every sibling org-tier mission-asset resolver, in `src/specify_cli/runtime/resolver.py`:

```python
378          org_path = org_root / "missions" / mission / subdir / name        # templates
...
817          org_path = org_root / "missions" / name / filename               # mission.yaml / mission-runtime.yaml
```

`org_roots` here is `resolve_existing_org_roots(repo_root)`, i.e. each pack's `effective_root` — the pack root itself, not a `missions/` subdirectory. The built-in reader this org tier mirrors (`MissionTemplateRepository.get_expected_artifacts`) is anchored at `built_in_missions_root()`, i.e. `packs/built-in/missions`, and reads `<missions_root>/<mission>/expected-artifacts.yaml`. The org path drops the `missions/` segment that anchoring implies.

## Reproduction

Builds measured, both identically: `spec-kitty-cli 3.2.6rc3` from `origin/main` `d060cff9a`, and the same version pinned at `8a03733e`.

[...]

Issue #3702, opened 2026-08-24
`charter activate mission-type <T>` succeeds for a type with an empty action sequence, activating the project into an unusable state


MOES-Media:
## BLUF

`spec-kitty charter activate mission-type <T>` writes the activation and exits 0 for a mission type whose action sequence is empty; the `MissionTypeEmptyActionSequenceError` that should have refused it only fires on the *next* invocation. The project is left activated onto a mission type that cannot plan anything, and every governed entry point afterwards — including `charter activate` and `charter mission-type list` themselves — fails.

The validation already exists and is already correct. It runs on the **read** path (`_resolve_action_slot`), and activation — the one operation whose entire job is to decide whether a type is usable — is the one operation that does not ask.

## Reproduction

Builds measured, both identically: `spec-kitty-cli 3.2.6rc3` from `origin/main` `d060cff9a`, and the same version pinned at `8a03733e`.

An org pack ships `mission_types/qa.yaml` with `schema_version`, `id`, `display_name` and no `action_sequence` (and no `extends`). Declared in a fresh project the ordinary way:

```yaml
doctrine:
  org:
    packs:
      - name: <PACK_NAME>
        local_path: <ABS_PATH_TO_PACK>
```

Then, on the natural operator path — `qa` **not** pre-seeded into `mission_type_activations`:

```
$ spec-kitty charter activate mission-type qa
Activated: qa
$ echo $?
0

$ spec-kitty charter activate mission-type qa
Error: mission type `qa` resolved from layer `org` has an empty action sequence.
$ echo $?
1

$ spec-kitty charter mission-type list
Error: mission type `qa` resolved from layer `org` has an empty action sequence.
$ echo $?
1
```

After run 1, `.kittify/config.yaml` contains:

```yaml
mission_type_activations:
- software-dev
- documentation
- research
- plan
- qa
```

**Reproduce this on the natural operator path, not a pre-seeded one.** Two independent reviewers on our side recorded this command as *failing* against the unmodified pack, because both pre-seeded `qa` into `mission_type_activations` first to isolate the failure from the roster-membership check. Under that precondition it does fail. Under the precondition a real operator has — declare the pack, then activate — it succeeds and bricks the project.

## Why it happens

`src/charter/mission_type_profiles.py`

```python
916  def _resolve_action_slot(
917      mission_type: str,
918      *,
919      registered: list[str],
920      is_registered: bool,
921      pack_context: _PackContextLike,
[...]

PR #3707, opened 2026-08-24
fix(doctrine): thread pack_context through the mission-type projection seam

### LINKED ISSUE #3701: Org/project-layer mission types always resolve an empty action sequence: the step projection is hardcoded to `pack_context=None`
## BLUF

`_inject_projected_fields` — the function that derives a mission type's `action_sequence` from its `mission-steps/<type>/<step>/step.yaml` files — hardcodes `pack_context=None`, so it only ever sees the **built-in** pack. The layered org/project mission-type loader calls that same function, so an org pack shipping a complete, correctly-authored `mission-steps/<type>/` tree resolves to `action_sequence = None` and can plan nothing.

Consequence: every org- or project-layer mission type is unusable end to end, and the operator-visible symptom is the `MissionTypeEmptyActionSequenceError` from #3386's fix — which correctly reports the empty sequence but names the *symptom*, not this cause.

## Where

`src/doctrine/missions/mission_type_repository.py`

```python
209  def _inject_projected_fields(raw: dict[str, Any], *, mission_type_id: str) -> dict[str, Any]:
...
213      (``pack_context=None`` -- org/project overrides never leak into this
214      repository-load-time injection) and derives ``action_sequence`` via
215      :func:`~doctrine.missions.step_projection.project_action_sequence`. Org
216      and project overrides are instead handled by the layered mission-type
217      lookup this mission introduces (mission ``up-mission-type-seam-...``,
218      WP03/WP04) -- not by any earlier runtime consumer switch ...
...
243      steps = list(
244          MissionStepRepository.default()
245          .resolve_all_for_mission_type(mission_type_id, pack_context=None)
246          .values()
247      )
248
249      projected_sequence = project_action_sequence(steps)
250
251      payload = dict(raw)
252      payload["action_sequence"] = projected_sequence or raw.get("action_sequence")
```

The docstring's own defence is circular. It says org/project overrides "are instead handled by the layered mission-type lookup". The layered lookup is `_load_layered_mission_type_file`, and at `src/doctrine/missions/mission_type_repository.py:347` it calls **this very function**:

```python
347      payload = _inject_projected_fields(raw, mission_type_id=yaml_file.stem)
```

So the function that hardcodes built-in-only defers responsibility to its caller, and the caller delegates it straight back. No layer in between supplies a `PackContext`.

`MissionStepRepository.resolve_all_for_mission_type` already accepts a real `PackContext` and already resolves org-layer steps correctly — that machinery is built, tested and reachable. It is simply never handed the context on this path.

## Reproduction

Setup: a project with an org pack declared the ordinary way in `.kittify/config.yaml`:

```yaml
doctrine:
  org:
    packs:
      - name: <PACK_NAME>
        local_path: <ABS_PATH_TO_PACK>
```

The pack ships `mission_types/qa.yaml` (`schema_version`, `id: qa`, `display_name`, **no** `action_sequence` — relying on the documented projection) and seven `mission-steps/qa/<step>/step.yaml` files, each carrying `sequence_index` and `in_action_sequence: true`, e.g.:

```yaml
id: specify
display_name: "Test Strategy"
step_type: agent
prompt_template: prompt.md
agent_profile: qa-strategist-quinn
depends_on: [discovery]
sequence_index: 1
in_action_sequence: true
template:
  artifact_key: spec
  template_file: test-strategy-template.md
```

Run from the project root:

```python
from pathlib import Path
from charter.pack_context import PackContext
from doctrine.missions.mission_step_repository import MissionStepRepository
from doctrine.missions.step_projection import project_action_sequence

pc = PackContext.from_config(Path.cwd())
for label, ctx in (("pack_context=None (what _inject_projected_fields uses)", None),
                   ("pack_context=<real>", pc)):
    steps = list(
        MissionStepRepository.default()
        .resolve_all_for_mission_type("qa", pack_context=ctx)
        .values()
    )
    print(f"{label}: {len(steps)} step(s) -> {project_action_sequence(steps)}")
```

Literal output, identical on both builds:

```
pack_context=None (what _inject_projected_fields uses): 0 step(s) -> []
pack_context=<real>: 7 step(s) -> ['discovery', 'specify', 'plan', 'tasks', 'implement', 'review', 'accept']
```

And through the layered roster the production path actually uses:

```python
from doctrine.missions.mission_type_repository import resolve_layered_mission_types
from doctrine.missions.repository import MissionTemplateRepository

dirs = (MissionTemplateRepository.default_missions_root() / "mission_types",)
repo = resolve_layered_mission_types(mission_types_dirs=dirs, pack_context=pc)
print("software-dev ->", repo.get("software-dev").action_sequence)
print("qa           ->", repo.get("qa").action_sequence)
```

```
software-dev -> ['specify', 'plan', 'tasks', 'implement', 'review']
qa           -> None
```

The org type resolves from the org layer — the roster layering from #3397's fix works. Only the projection does not.

## Expected vs actual

| | |
|---|---|
| **Expected** | `qa` resolves `action_sequence = ['discovery', 'specify', 'plan', 'tasks', 'implement', 'review', 'accept']`, projected from the org pack's own `mission-steps/qa/` tree, exactly as `software-dev` projects from the built-in tree. |
| **Actual** | `action_sequence = None`. The pack's seven steps are never read at projection time. Every governed entry point for `qa` then fails with `mission type 'qa' resolved from layer 'org' has an empty action sequence`. |

- `src/doctrine/missions/mission_type_repository.py` — thread the resolving layer's `PackContext` into `_inject_projected_fields`; `_load_layered_mission_type_file` and `resolve_layered_mission_types` are the call sites that have it.
- The existing `functools.cache` on `resolve_layered_mission_types` is already keyed on `(mission_types_dirs, pack_context)`; `MissionStepRepository`'s own `_resolve_all_for_mission_type_cached` is likewise pack-context-keyed. Both should be re-checked, not re-designed.
- Tests: `tests/doctrine/missions/test_mission_type_repository.py`, and the golden-parity suite in `tests/runtime/test_runtime_seam.py` (built-in output must stay byte-identical — a built-in type resolves the same steps whether or not a pack is declared).
- Note the transitional fallback at `:252` (`projected_sequence or raw.get("action_sequence")`) must stay: a pack that *does* author `action_sequence:` explicitly keeps working.

## Non-goals

- **Not** changing `MissionType`'s non-empty `action_sequence` invariant.
- **Not** changing `MissionStepRepository`'s own org/project resolution — it already works; this is purely about handing it the context.
- **Not** the activation-time validation gap (a type with a genuinely empty sequence can still be activated into a broken project) — that is #3702.
- **Not** `template_set` projection (`_resolve_template_set_slot` already threads a real `PackContext` and is not affected).
- **Not** the built-in-only `expected-artifacts.yaml` lookups — #3703 and #3704.

## Related

- #3397 (closed) — "Org/project-tier mission types are unusable end to end: roster lookup is built-in-only, and a custom type silently resolves to an empty action sequence." The **roster-lookup** half landed and is verified working above. The **empty action sequence** half is still reproducible on current `main`; this issue is that residual with the cause located. A closed issue is not evidence the condition is met — this was measured, not inferred.
- #3386 (closed) — added `MissionTypeEmptyActionSequenceError`. That fix is correct and is what makes this defect visible at all; it reports the symptom.
- #3526 / #3530 (open) — org-tier doctrine reach. This is one concrete consumer that does not reach the org tier.
- #3410 (open) — the silent-drop epic. This one is not silent any more (thanks to #3386) but it is still fake-green in the sense that the pack validates clean and cannot run.
- #3702 — `charter activate mission-type <T>` writes the activation anyway and only fails on the next invocation. That is how an operator lands in the broken state this defect creates.
- #3703 — org-tier `expected-artifacts.yaml` anchor. #3704 — custom-family guard dispatch and the inert `blocking:` flag. #3705 — silent asset drop in cascade. Same sweep, independent files.

MOES-Media:
## Impact

Org- and project-layer mission types now resolve their real action sequence instead of an empty one.

Before this change, a mission type shipped by an org or project pack — one whose steps come from the pack's own step tree rather than from an explicit `action_sequence:` in its YAML — always resolved to `None`/`[]`. The step projection asked the step authority for that type's steps while hardcoding `pack_context=None`, so only built-in steps were ever visible. A pack author's type would load, validate, and then plan nothing.

Measured on an org pack with 7 steps: `[]` with `pack_context=None`, `['discovery', 'specify', 'plan', 'tasks', 'implement', 'review', 'accept']` with a real context.

## What changed

`pack_context` is threaded through the four functions on the projection path in `src/doctrine/missions/mission_type_repository.py`:

- `_inject_projected_fields` (the seam that hardcoded `None`)
- `_load_layered_mission_type_file`
- `scan_mission_types_dir`
- `resolve_layered_mission_types`

Each new parameter is **keyword-only with a `None` default**, so every existing call shape keeps working unchanged — including `MissionTypeRepository._load()`, which stays built-in-only, and the availability scan in `charter/pack_manager.py`, which never reads `action_sequence`.

The parameter is typed `_PackContextLike | None` — the structural Protocol already used in this module — never the concrete `PackContext`. `src/doctrine/` must not import from `src/charter/` (the tier order is pinned by an architectural test), and that bar applies to `TYPE_CHECKING`-only imports too.

## Testing

Red-first. The reproduction test was committed **before** the fix (`ca75f7efd` precedes `a03a39c7f`), and the revert check was run rather than asserted: reverting the fix commit fails all four new tests (`action_sequence` comes back `None`, and the governed entry point raises `MissionTypeEmptyActionSequenceError`); restoring it passes all four.

- 74 passed / 0 failed on the two touched test files (baseline before this change: 70 / 0)
- `diff-cover` against the 90% critical-path floor on `src/doctrine/*`: 100% (8/8 changed lines)
- `ruff check src/` clean; `mypy` clean on the changed module
- Golden parity: built-in mission types resolve byte-identically with `pack_context=None` and with a real context declaring an unrelated org pack — verified through the layering/caching mechanism, not only through fixtures

## Scope

Deliberately **not** in this PR, per the issue's own non-goals:

- Activation-time validation — a type with a genuinely empty sequence can still be activated into a broken state. That is #3702 and is a real, separate defect.
- `template_set` projection, which already threads a real context and is unaffected.
- The built-in-only `expected-artifacts.yaml` lookups (#3703, #3704).

## Note for reviewers

This branch carries the mission's planning artifacts under `kitty-specs/` alongside the implementation. The implementation diff is three files: the module above and its two test files.

`kitty-specs/.../analysis-report.md` contains host-absolute paths in its `input_artifacts` block. That is generated output, not hand-authored, and is the known issue #3398 (the generator emits absolute paths); it is left as generated rather than hand-stripped, since editing it would regenerate away and create drift without fixing the class.

Closes #3701

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3707"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790138499&installation_model_id=427282&pr_number=3707&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3707&signature=929ddc0cc8088d2380b9450dae7df7c7edbbe38c30f5df6ac97f232689ce2ef7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
## sk-land remediation pass — 2 red checks classified and closed

First full CI cycle on `00536ee1a` returned **35 pass / 37 skipped / 2 fail**. Both failures were real (28s and 4m43s durations — neither was a cancelled job reported as a failure). Both are now fixed at `38a6b1661`.

### 1. `cutover-guard` — bin: PR defect (mission metadata), mechanical

```
Missions touched by diff : 1
Un-cut-over              : 1
  mission-types-empty-action-sequence-01M0RMCA: status_phase not flipped despite event-log runtime evidence
  remedy: spec-kitty migrate backfill-runtime-state --mission mission-types-empty-action-sequence-01M0RMCA
```

Applied the gate's own prescribed remedy verbatim and unedited. It added one key to the mission's `meta.json`:

```diff
   "slug": "mission-types-empty-action-sequence-01M0RMCA",
+  "status_phase": "1",
```

`spec-kitty cutover-guard --base-ref origin/main` now returns `Un-cut-over: 0`, exit 0. The value is correct — string `"1"` matches 395 of 398 `status_phase` values across `kitty-specs/*/meta.json` on `main`.

Two disclosures, both recorded in our workspace defect ledger:
- This mission was created BRANCH-FIRST and driven through the entire documented flow (specify → plan → tasks → analyze → implement → `finalize-tasks` → `mark-status`) and still arrived at PR time with **no `status_phase` key at all**. No local gate catches it; it surfaces only as a red check on an already-pushed PR.
- The remedy reports `Flipped: 0 / Skipped (already migrated): 1` while performing exactly the flip it says it did not perform. Only `git status` reveals it worked.

### 2. `doctrine-charter-tests` — bin: PR defect (genuine regression), substantive

`1 failed, 5898 passed, 41 skipped`:

```
FAILED tests/charter/test_mission_type_profiles.py::TestPackContextProjection::
  test_org_pack_type_projects_real_action_sequence_and_template_set
  - AssertionError: assert ['design'] == ['design', 'implement']
```

Reproduced locally. This is a **real regression introduced by this PR** in a pre-existing test file we do not otherwise touch — not a flake, and not pre-existing breakage on `main`.

**Diagnosis.** That fixture declares the org-pack type `qa` with `action_sequence: ["design", "implement"]` in its type YAML, but ships **only one step file** — `design`. Before this PR, `_inject_projected_fields` hardcoded `pack_context=None`, so the org-pack projection came back empty and `projected_sequence or raw.get("action_sequence")` fell back to the raw 2-element YAML value. The assertion passed **only because of the very bug this PR fixes**. With `pack_context` correctly threaded, the projection now actually runs against the org pack's step files and correctly yields `["design"]`.

**Remedy — fixture, not source.** The test's own docstring states it asserts fields "matching the org-pack's declared steps exactly", so the fixture was the inconsistent half: it declared two steps and shipped one. We ship the missing `implement` step file (`sequence_index: 1`, `in_action_sequence: true`), so declared steps and step files agree and the test finally verifies what it claims to verify.

Deliberately **not** changed:
- No `src/` file was touched for this. In particular the `projected_sequence or raw` precedence (C-007-retained) is untouched — this PR does not introduce it, it only makes it reachable on the org/project path.
- **Neither assertion in the test was modified.** The new step is written with no `template:` block (`MissionStep.template` is optional), so the second assertion `dict(bundle.template_set) == {"spec": "qa-spec-template.md"}` holds verbatim. `_write_org_mission_step_yaml` gains optional `artifact_key`/`template_file` to allow that; existing call sites are byte-identical.

**Worth flagging for reviewers** (behavioural, not a defect in this PR): now that the org/project projection path is live, an org pack that declares N steps in transitional YAML but ships fewer than N step files will project only the steps it ships. That is the pre-existing `projected or raw` precedence behaving as designed, surfaced for the first time — it is not new logic here, but it is newly reachable, and it is the exact shape of the fixture bug above.

### Verification

Ran the `doctrine-charter-tests` job's **exact** pytest selection locally:

```
5899 passed, 41 skipped in 78.94s   (exit 0)
```

CI on the previous tip was `5898 passed, 1 failed` over the same 5899 — the one red is closed and nothing else regressed.

Targeted suites and lint:

```
tests/charter/test_mission_type_profiles.py
tests/doctrine/missions/test_mission_type_repository.py
tests/runtime/test_runtime_seam.py     -> 103 passed
ruff check src/ tests/charter/...      -> All checks passed!
```

### Process disclosure

WP01's red-first baseline was captured over exactly the two files in `lanes.json`'s `write_scope`. That made mission-introduced reds unambiguous in those files, but it was too **narrow**: `tests/charter/test_mission_type_profiles.py` exercises the very org-pack projection path being fixed and sat outside it. The regression was therefore invisible to implementation, WP reviewer approval, and all three R1 lens groups of the pre-merge squad — each of which returned `findings: []` over the diff they were shown. **CI found it, not the local gate set and not the review squad.** A diff-scoped review cannot see a regression in an untouched file, by construction. Recorded in the mission tracer so the next mission derives its baseline from the *consumers of the changed seam* rather than from `write_scope`.

Awaiting the next CI cycle on `38a6b1661`.

PR #3708, opened 2026-08-24
fix(charter): anchor org-tier expected-artifacts.yaml at missions/<type>/

### LINKED ISSUE #3703: Org-tier `expected-artifacts.yaml` is unreachable: the resolver anchors at `<org_root>/<type>/` while every other org mission asset anchors at `<org_root>/missions/<type>/`

MOES-Media:
## What changes

An org pack's `expected-artifacts.yaml` now actually gets found.

Before this, an org pack that laid its manifest out the way every other org mission asset is laid out — `<org_root>/missions/<mission_type>/` — was **silently ignored**. The resolver looked one path segment short, at `<org_root>/<mission_type>/`, found nothing, and fell through to the built-in manifest without a word. No error, no warning, no way for a pack author to tell the difference between "my override was applied" and "my override was never read".

For a pack author, the practical effect: your `expected-artifacts.yaml` override had no effect and nothing told you why.

## The fix

One line in `src/charter/org_expected_artifacts.py` — insert the missing `missions/` segment so this resolver anchors where its siblings already do:

```python
path = org_root / "missions" / mission_type / _EXPECTED_ARTIFACTS_FILENAME
```

This was verified as a genuine inconsistency, not a guess: `src/doctrine/resolver.py` and
`src/specify_cli/runtime/resolver.py` both already anchor mission-scoped org assets at
`org_root / "missions" / ...`. `org_expected_artifacts.py` was the outlier. (`service.py`'s
`_org_dirs` and `org_charter_loader.py` join directly under `org_root`, but those are a
different, non-mission-type-scoped asset class — not an inconsistency.)

The module and function docstrings are corrected in step, including an explicit caveat that
Contract C-4's frozen sample still cites the pre-fix path.

## Compatibility

**Nothing breaks.** An org pack sitting at the old location was never being read by any
consumer, so there is no working behaviour to preserve and no migration to document. That was
verified against the frozen contract doc rather than assumed.

## Test changes

The five test files in this diff each duplicated a `_write_org_manifest`-style fixture helper
hardcoded to the pre-fix path, and exercise the resolver end-to-end **unmocked**. Left alone,
the fix would have flipped at least four currently-passing tests from GREEN to RED — so
correcting them is part of the fix, not scope creep.

Two tests were found to be **vacuous** — passing while asserting nothing — and repaired:

1. In `test_org_expected_artifacts.py`, a new old-anchor test routed its fixture through a
   helper that a later commit retargets, so it would have silently begun writing to the new
   path. Caught during implementation.
2. In `test_rebaseline.py`, `test_malformed_org_pack_does_not_raise` built its fixture at the
   old anchor, so post-fix the resolver never found it and the test passed on the "no org
   file" path instead of the malformed-YAML graceful-degrade path it claims to cover. Caught
   by the pre-merge review squad — invisible in a diff view, since the diff never edits that
   test.

## Verification

- **RED-first**: reverting the one-line fix produces **19 failures across all six files**,
  measured independently three times. The coverage is load-bearing.
- **Non-vacuity**: stripping the `try/except` in `_read_yaml_mapping` makes the repaired
  malformed-YAML test fail, so it would catch a real regression rather than merely passing.
- **No regression**: `tests/charter/` + `tests/dossier/` → **2824 passed, 26 skipped, 0
  failures**, identical to the pre-change baseline. The mission's five-file surface was
  160/160 green before and is 162/162 after (the two additions are the RED-first tests);
  `main`'s known-red set (#3284) does not intersect this surface.
- Diff coverage on the changed production line: **100%**.

## Deliberately out of scope

- **No `pack_validator` reachability gate.** An authoring-time gate that resolves manifest
  paths at their tier-correct anchor is genuinely wanted, but it belongs with the broader
  asset-manifest resolver family — deferred to **#3704**.
- **Two stale docstrings not fixed here.** `ManifestRegistry.load_manifest`
  (`src/specify_cli/dossier/manifest.py:221`) and `_resolve_expected_artifacts_slot`
  (`src/charter/mission_type_profiles.py:1030`) — the resolver's two live callers — still
  document the pre-fix anchor. Found by two independent review lenses and upheld; both files
  sit outside this mission's agreed file set, and the change is documentation-only with no
  test impact. Recorded with full reasoning rather than silently dropped.

Closes #3703

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3708"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790139747&installation_model_id=427282&pr_number=3708&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3708&signature=f1c9edfd7fc528fe66a07a5faaed898ee82dfdef4c04d8a162979158cb1581d1"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:

**Verdict: merge-ready pending green CI.** No defect was found in the fix itself. Every red on this PR came from tooling or environment, and each was classified before being touched — none was retried to green, and no assertion was weakened to make a check pass.

### Review outcome

A pre-merge adversarial squad (three independent profile-loaded lenses, then an independent refuter) ran over the implementation diff only — the mission's design artifacts on this branch cleared their own review rounds earlier and were deliberately out of scope.

**5 raw findings → 3 unique → 3 upheld, 0 refuted, 0 blocking.**

Two lenses independently converged on the same two docstring findings, which is the reason they are recorded rather than dismissed.

### Folds

| Fold | Why |
|---|---|
| `chore(landing): flip status_phase via backfill-runtime-state for cutover-guard` | `cutover-guard` failed with `status_phase not flipped despite event-log runtime evidence`. Ran the guard's own prescribed remedy; value written by the sanctioned CLI, not by hand. |
| 7 commit messages reworded (no content change) | commitlint failures — see below. |
| `docs(landing): record venv-extras phantom-red trap in tooling tracer` | Mission tracer record for the environment trap below. |

### commitlint: 7 failing commits, two distinct causes

Measured locally with CI's own invocation before CI reported, then re-measured at 0 after the reword. The rewrite changed **no file content** — verified by diffing against a pre-reword backup ref — and kept the same 45 commits.

- **4 × `Add analysis report for mission <slug>`** — emitted by `spec-kitty record-analysis`. `commitlint.config.cjs`'s ignore list covers `meta|spec|tasks|plan` but not "analysis report", so these are linted and fail `type-empty` + `subject-empty`. A tooling gap, not authored. Reworded to `docs(mission): record analysis report for <slug>`.
- **3 × `reviews(spec|plan|tasks): ...`** — **our own error.** `reviews` is not in `type-enum`. Reworded to `docs(reviews): ...`.

The shared lint config was **not** loosened to accommodate either.

### Test evidence

- **Red-first verified empirically, three times independently.** Reverting the one-line anchor fix produces **19 failures across all six files** (spread 10+2+1+4+2), each failing for the right reason rather than a setup error. The coverage is load-bearing.
- **Non-vacuity of the repaired test proven by mutation**: stripping the `try/except` in `_read_yaml_mapping` makes it fail, so it would catch a real regression instead of merely passing today.
- **No regression**: `tests/charter/` + `tests/dossier/` → **2824 passed, 26 skipped, 0 failures**, identical to the pre-change baseline, measured twice.
- Mission's five-file surface: **160/160 green before**, **162/162 after** (the two additions are the red-first tests). `main`'s known-red set (#3284) does not intersect this surface, so there was no pre-existing-red ambiguity to argue about.
- `tests/architectural/`: green once the venv had its tooling — see below.
- Diff coverage on the changed production line: **100%**.

### Two vacuous tests found and repaired

Both were tests that **passed while asserting nothing** — the failure mode this repo's charter treats as severity 4:

1. `tests/charter/test_org_expected_artifacts.py` — a new old-anchor test routed its fixture through a helper that a later commit retargets, so it would have silently begun writing to the new path. Caught during implementation.
2. `tests/dossier/test_rebaseline.py::test_malformed_org_pack_does_not_raise` — built its fixture at the old anchor, so post-fix the resolver never found it and the test passed on the "no org file" path instead of the malformed-YAML graceful-degrade path it claims to cover. **Invisible in a diff review, since the diff never edits that test**; found only by a lens that measured behaviour rather than reading code.

### Environment trap worth flagging to other maintainers

Running `tests/architectural/` locally produced **8 failures on a diff that never touched those paths** — `test_mypy_is_available_and_its_absence_fails_rather_than_skips` plus four in `test_tid251_enforcement.py`. Cause: the checkout venv was built with bare `uv sync`, which installs **neither** the `test` nor the `lint` extra's tooling, so `ruff` and `mypy` were absent. After `uv sync --extra test --extra lint` the same files return **34 passed**.

Those tests are right — they fail closed on absent tooling by design. But note that **`uv sync --extra lint` alone silently removes `pytest`**, because extras resolve per-invocation rather than accumulating. CI's three `arch-adversarial` shards were green throughout, which is what proves these were purely local.

### Not fixed here, deliberately — recorded, not dropped

Both live callers of `resolve_org_expected_artifacts` still document the pre-fix anchor in their own docstrings:

- `src/specify_cli/dossier/manifest.py:221` — `ManifestRegistry.load_manifest`
- `src/charter/mission_type_profiles.py:1030` — `_resolve_expected_artifacts_slot`

Accuracy is not in dispute: both were confirmed untouched in this diff, with identical text at the merge base, so this fix is what flipped them from true to false. They are not folded in because both files lie outside this mission's agreed file set, and that scope constraint sets the bar for expansion at *functional regression* — these are documentation-only with no test impact. Three independent agents reached the same recommendation. **A two-line follow-up; flagged to the operator as a fold-or-follow-up decision rather than silently deferred.**

Also deliberately out of scope: an authoring-time `pack_validator` gate that resolves manifest paths at their tier-correct anchor. Genuinely wanted, but it belongs with the broader asset-manifest resolver family — **#3704**.

### State

Watching CI; merge-ready on green. **I do not merge** — handing off to the operator.

Issue #2599, opened 2026-07-12
Executable ASSET-kind gate handlers behind opt-in trust model (half B)


stijn-dejongh:
## Context
Strangler-sequence step **5 of 5** (opt-in tier) from the doctrine-controlled transition gates epic #2535.

> "*Opt-in Tier:* asset-as-script handler behind `review.allow_executable_gate_assets` + the trust model = half B as a later increment."

This is epic Scope **§B — Executable ASSET-kind gate helpers (greenfield subsystem)** plus the **Mandatory trust model** section, both quoted in full below since this step is the largest and most novel of the five.

## Scope (from epic body, §B)
ASSETs are inert today (validated sidecar file reference, never parsed/executed; DRG node discards `path`/`mime`; non-augmentation-eligible, non-charter-activatable). Making them executable gate helpers is a new subsystem requiring:
- asset **repository**
- **URN → path resolver** (the DRG node must stop discarding `path`/`mime`)
- **code-asset entrypoint contract** (how a script declares its gate entrypoint + I/O)
- **sandboxed runner**
- **asset activation** (ASSETs are non-activatable today)

## Mandatory trust model (binding — executes pack-supplied code, RCE-equivalent surface)
Ships only behind **all** of:
- **Provenance allowlist** — `built-in` and `org-pack` assets only; never third-party/unverified.
- **Interpreter allowlist / no-shell** — no arbitrary shell; a fixed set of vetted interpreters.
- **Explicit opt-in** — `review.allow_executable_gate_assets` (default off).
- **Timeout** on every asset invocation.
- **Fail-OPEN** — a runner error, timeout, or absent asset degrades to a warn, never a hard block.
- **Structured-verdict-or-warn** — an asset must return a structured verdict; malformed output degrades to a warn, never a trusted pass/fail.

## Acceptance
- [ ] Asset repository + URN→path resolver implemented (DRG node preserves `path`/`mime`)
- [ ] Code-asset entrypoint contract defined (entrypoint + I/O shape for a gate script)
- [ ] Sandboxed runner implemented, enforcing interpreter allowlist and no arbitrary shell
- [ ] Asset activation supported (ASSETs charter-activatable for this purpose)
- [ ] `review.allow_executable_gate_assets` opt-in flag, default off
- [ ] Timeout enforced on every asset invocation
- [ ] Fail-open verified: runner error / timeout / absent asset / malformed verdict all degrade to warn, never a block or trusted pass
[...]

PR #3686, opened 2026-08-23
fix(upgrade): commit root files the run wrote (drop depth-based root filter) + fix porcelain rename parsing (#2491 follow-up)

robertDouglass:
Follow-up to #2491 / #3679; salvages the rename-parsing fix from the superseded #2492 (credit @rayjohnson) and closes the remaining root-file gap in the upgrade auto-commit — **by dropping the depth-based root filter rather than extending its allowlist**, per an adversarial-squad review (evidence in the first comment below).

### 1. Root-level files the run wrote now land in the upgrade commit (`Changed`)

`is_upgrade_commit_eligible` used to skip every root-level file except `.gitignore` ("root files are operator-owned"). But upgrade's own writers write at the root, and the hand-kept exception list kept lagging:

| Root file | Writer | On `main` today |
|---|---|---|
| `.gitattributes` | 7 merge-driver/diff-attribute migrations (`_merge_driver_seeding.py`, `m_3_1_1_event_log_merge_driver`, `m_3_2_0rc28_github_diff_attributes`, 4× `m_3_2_6_*`, `m_3_2_7_review_cycle_merge_driver`) — several target **3.2.6rc1**, and they `runs_on_worktrees` | left dirty in main checkout **and every live worktree** → `spec-kitty merge` dirty-guard trips (#2385 class) |
| `.claudeignore` | `m_3_2_8_provision_kitty_env` (3.2.6rc2) | left dirty |
| `AGENTS.md` / `GEMINI.md` | session-presence surface repair (`REPAIRABLE_REQUIRED`), now ordered *before* the churn commit by #3679 | left dirty |
| `.gitignore` | gitignore-backfill migrations | committed (carved out in #2392) |

A live `spec-kitty upgrade --yes` on `main` (3.2.5→3.2.6rc3 scratch project) ends: `→ Auto-committed upgrade changes (4 files)` then `git status`: ` M .claudeignore  M .gitattributes  ?? AGENTS.md  ?? GEMINI.md`.

**After:** ownership is the pre-run baseline alone — clean before the run and dirty after ⇒ written by the run ⇒ in the one auto-commit; already dirty at baseline ⇒ never swept, whatever the depth. The `"/" not in path` rule and its allowlist are gone. Two guards remain: `../` paths are never committed, and when the checkout **is** `$HOME` (#3652 hazard) neither `~/.kittify` nor any root-level file is committed (operator dotfiles).

[...]

PR #3666, opened 2026-08-22
fix: preserve planning branch for legacy PR-bound missions

rusliksu:
## Summary

- repair legacy PR-bound missions that stored a protected landing branch as their planning branch
- require the otherwise-unprovable legacy planning ref explicitly through `--target-branch`
- atomically preserve `target_branch` and `merge_target_branch` from the owning checkout
- fail closed for linked-worktree metadata writes and unrelated ambient branch inference
- keep `--validate-only` byte-, index-, HEAD-, status-, and ref-stable

Fixes #2938.

## Adversarial review fixes

- preserved the protected merge destination when an explicit planning override is persisted
- enforced the canonical checkout-identity write-refusal seam
- removed automatic recovery from an arbitrary invoking branch
- strengthened the mutation oracle to cover the index, symbolic HEAD, status, refs, and file bytes
- tightened the routed-meta architectural non-vacuity floor after four new canonical reads
- repaired the inherited org mission-type fixture so it declares every step in its projected action sequence

## Verification

- ready-state CI Quality: 29 passed jobs, 0 failed ([run 32759956373](https://github.com/Priivacy-ai/spec-kitty/actions/runs/32759956373))
- draft CI Quality: 23 passed jobs, 0 failed ([run 32758372530](https://github.com/Priivacy-ai/spec-kitty/actions/runs/32758372530))
- `4031 passed, 7 skipped` in the local `specify-cli-rest` shard reproduction
- `1920 passed, 5 skipped, 4 xfailed` across the broad finalize/branch-context and architectural suites before updating the stale routed-meta census floor
- `123 passed, 1 skipped, 2 xfailed` after the census-floor update
- strict mypy passed across `src/specify_cli`, `src/charter`, and `src/doctrine`
- Ruff, Markdown lint, Contextive generation check, compileall, and diff checks passed

The xfails are existing documented strict legacy flattened-bootstrap gaps. No deployment, release, or live runtime mutation is included. No Sonar UI-only work is known to remain.

Issue #3241, opened 2026-08-06
Two meta-fail-closed test files are selected by ZERO CI gates and will never run (WP02 row 5, WP05 diagnosability) — plus a cross-tree _rel() over-count in the inline-meta-read gate


MOES-Media:
Filed by mission `meta-fail-closed-3162-01KZ7FSQ`, WP06 (T040 step A5 — a gate red this WP cannot repair inside its own surface, so the owning WP is named and the finding filed rather than another WP's test file edited).

## Observed, not inferred

```
$ .venv/bin/python -m pytest tests/architectural/test_gate_coverage.py -p no:randomly -ra
...
FAILED tests/architectural/test_gate_coverage.py::test_no_new_orphan_surfaces
E   AssertionError: 2 test file(s) are selected by ZERO CI gates and are not in the recorded baseline — they will never run in CI:
E       tests/runtime/test_wp02_row05_bridge_io_fail_closed.py
E       tests/specify_cli/cli/commands/test_meta_bypass_diagnosability.py
E
E   Fix: give the test(s) a marker a gate selects (e.g. `fast`, `integration`, `git_repo`) and/or place them under a gated path. If the coverage gap is intentional and tracked, regenerate the baseline with `uv run python -m tests.architectural._gate_coverage --update-baseline`.
1 failed, 36 passed in 527.18s
```

`tests/architectural/_gate_coverage_baseline.json` carries `"orphan_files": []` and its own `_comment` says *"The ratchet (`test_gate_coverage.py`) fails on any NEW orphan file not listed here."*

## Ownership

| Orphan file | Owning WP | Status of that WP |
|---|---|---|
| `tests/runtime/test_wp02_row05_bridge_io_fail_closed.py` | **WP02** (census row 5) | under remediation |
| `tests/specify_cli/cli/commands/test_meta_bypass_diagnosability.py` | **WP05** (diagnosability tests) | **APPROVED** |

The WP05 case is the sharper one: **WP05 was approved on the strength of a test file that CI will never execute.** Its assertions have never run in a gated job and would not red on a regression.

## Attribution

Measured on both sides, same selection, redirected output:

- At `402c90c3c` (WP06's first commit's parent): **1 failed, 36 passed** — the same two files.
- At `172ec9aad` (WP06's coupling-1 commit): **1 failed, 45 passed** in a combined selection — the same two files.

[...]

PR #3711, opened 2026-08-24
fix(charter): report kind-filtered nodes instead of silently dropping them from cascade

### LINKED ISSUE #3705: A cascade drops every `asset`-kind target with no report, so a pack's executable check is unreachable and unmentioned
## BLUF

`charter activate ... --cascade all` silently drops every DRG edge whose target is an `asset` node: no activation, no warning, no "skipped" line, exit 0. From the operator's side an edge pointing at a structurally non-activatable kind is **indistinguishable** from an edge that failed to resolve and from an edge that does not exist.

This is the D-005 half of the asset gap. The executable half — actually invoking a pack-shipped code asset as a gate handler — is #2599, which is open and well specified; this issue does not duplicate it. Together they mean a pack whose only executable check ships as an `asset` can neither be reached from a DRG edge nor invoked by a gate, and the operator is told neither thing.

## Reproduction

An org pack declares, in `drg/fragment.yaml`, one node with **two** outgoing `suggests` edges — one to a tactic, one to an asset:

```yaml
  - id: qa-traceability-lint
    kind: assets
    title: "QA Traceability Lint"
    body_path: assets/qa_carrier_lint.py.asset.yaml
  - id: qa-carrier-lint
    kind: toolguides
    title: "QA Carrier Lint Toolguide"
    body_path: toolguides/qa-carrier-lint.toolguide.yaml
edges:
  - source: toolguide:qa-carrier-lint
    target: asset:qa-traceability-lint
    relation: suggests
  - source: toolguide:qa-carrier-lint
    target: tactic:qa-risk-based-test-prioritization
    relation: suggests
```

One command, same source node, both edges in scope:

```
$ spec-kitty charter activate toolguide qa-carrier-lint --cascade all
Warning: 'qa-carrier-lint' is already activated for kind 'toolguide'.
Cascade-activated: tactic/qa-risk-based-test-prioritization
$ echo $?
0
```

The tactic edge is reported. The asset edge produces **no output line of any kind** — it is not activated, not warned about, not listed as skipped or non-applicable. On a first activation (before the toolguide was activated) the output is `Activated: qa-carrier-lint` plus a default-pack warning, and likewise nothing about the asset.

The control matters: it is the *same command*, the *same source node*, the *same relation*, differing only in the target's kind. So this is not "the cascade did not run" or "the edge did not resolve" — the cascade ran, resolved that node, and dropped it without a word.

## Where

`src/doctrine/artifact_kinds.py`

```python
330  CHARTER_ACTIVATABLE_KINDS: frozenset[ArtifactKind] = frozenset(ArtifactKind) - {
331      ArtifactKind.TEMPLATE,
332      ArtifactKind.ASSET,
333  }
```

`src/charter/cascade.py`

```python
274      Pure forward closure over :data:`REFERENCE_RELATIONS` ... filtered twice:
275      non-artifact nodes (actions, glossary) are dropped because :func:`_kind_of`
276      returns ``None``, and non-activatable artifact kinds (``template``/``asset``)
277      are dropped via the canonical :data:`CHARTER_ACTIVATABLE_KINDS` set ...
278      The closure still *reaches* those nodes -- traversal reach and candidacy
279      are separate concerns -- but they are never cascade candidates.
...
291          if kind not in CHARTER_ACTIVATABLE_KINDS:
292              continue
```

A bare `continue`. The dropped node is not collected anywhere, so nothing downstream can report it. The exclusion itself is a deliberate, ADR-recorded design decision and this issue does not contest it — the ask is that the drop be **reported**, not that it stop happening.

## Expected vs actual

| | |
|---|---|
| **Expected** | A cascade that reaches a node of a non-activatable kind says so — e.g. `Not cascaded (kind not charter-activatable): asset/qa-traceability-lint`. A cascade that resolves **zero** activatable targets says that too, rather than printing nothing. |
| **Actual** | Silent `continue`. Exit 0. Indistinguishable from a resolvable edge, an unresolvable edge, and no edge at all. |

## Why this matters beyond cosmetics

A doctrine pack's `assets/` are where an executable check lives — a validator, a lint, a carrier schema. Today an author can:

- register the asset node and its edges in `drg/fragment.yaml`,
- pass `doctrine pack validate` green,
- run `charter activate --cascade all` and see exit 0,

and have produced an artifact that is unreachable from governance in every direction, with no surface anywhere saying so. `D-005` ("degrade, but never silent") is the principle; this is the seam it was written for.

The executable side is worse still, and is #2599's: `GATE_REGISTRY` has exactly one entry —

```python
117  GATE_REGISTRY: dict[str, GateHandler] = {
118      _SPEC_KITTY_PRE_REVIEW_HANDLER_NAME: GateHandler( ... ),
119  }
```

— and `GateBinding.handler_kind` documents its own inertness at `src/doctrine/missions/step_contracts.py:99-102`:

> `handler_kind` is an inert-in-half-A discriminator (FR-005, C-002): `"asset"` validates and round-trips byte-stable but **is never executed until a future half-B asset executor lands (#2599)**.

So a pack can declare `handler_kind: "asset"` on a step contract, have it validate and round-trip, and have nothing ever run it. That is #2599's scope and is correctly tracked there. **What is not tracked anywhere is that neither of these two dead ends produces a diagnostic.**

## Suggested shape (narrow)

1. Have `_referenced_artifacts` (or a sibling) collect the nodes it drops for kind reasons and return them alongside the candidates.
2. Render them in the cascade report and in the existing no-cascade warning path, labelled as structurally non-activatable rather than as failures.
3. When a `--cascade` resolves zero activatable targets, say so explicitly instead of printing nothing.

Optionally, and separately arguable: have `doctrine pack validate` warn when a pack declares DRG edges whose only targets are non-activatable kinds.

- `src/charter/cascade.py` — collect and return kind-dropped nodes.
- `src/specify_cli/cli/commands/charter/activate.py` (`_render_cascade_activation`, `_render_no_cascade_warning`) and the matching `deactivate.py` path — render them.
- `src/doctrine/artifact_kinds.py` — docstring only, to state that exclusion from this set is expected to be *reported*, not silent.
- Tests: a fixture with sibling activatable/non-activatable edges from one source, asserting both appear in the output.

## Non-goals

- **Not** adding `ASSET` (or `TEMPLATE`) to `CHARTER_ACTIVATABLE_KINDS`. The exclusion is ADR-recorded; this issue asks for visibility, not a policy reversal.
- **Not** the executable asset gate handler, its trust model, sandboxing, interpreter allowlist or opt-in flag — that is #2599 and should stay there.
- **Not** the asset resolution/install path for consumers — #3037.
- **Not** the pack-activation trust warning for packs shipping executable code — #2536.
- **Not** promoting `gate` to a first-class `ArtifactKind` — #3418.
- **Not** the unconditional fail-open wrapper on gate dispatch (`_mt_dispatch_one_gate`, `src/specify_cli/cli/commands/agent/tasks_move_task.py:1451-1461`, documented as never enabling fail-closed). Noted here only because it means "the gate ran" will not by itself mean "the transition can be blocked" once #2599 lands; whether that needs its own issue is a maintainer call.

## Related

- #2599 (open) — Executable ASSET-kind gate handlers behind opt-in trust model (half B). **The executable half. Do not fold this issue into it and do not fold it into this one** — one is a diagnostic gap that can ship today, the other is a security-scoped feature.
- #3037 (open) — the asset doctrine kind has no resolution or install path for consumers.
- #2536 (open) — pack activation must warn when a pack ships executable code assets.
- #2535 (epic), #3418 / #3419 / #3420 / #2597 (open) — the declarative gate cluster.
- #3527 (closed), #3572 (closed) — org-inert cascade / org fragment edges never reaching cascade. Those were about org edges not being loaded at all; measured here, org edges **are** loaded and traversed (the tactic edge cascades correctly). This is the kind-filter drop downstream of that, which is a different defect.
- #3009 (closed), #2829 (closed) — kind-complete cascade / DRG-orphan activations. #2829's ADR is the one that records the `template`/`asset` exclusion; this issue is consistent with it.
- #3701 / #3702 / #3703 / #3704 — the org-pack enforcement sweep this was found in: mission-type action-sequence projection, activation-time validation, org-tier manifest anchor, and custom-family guard dispatch. All independent files. This issue is the diagnostic seam of that set.

MOES-Media:
## What an operator sees

Run `spec-kitty charter activate <kind> <id> --cascade all` on a pack whose graph points at an `asset`- or `template`-kind node, and the command exits `0` with a clean cascade report — **and never tells you it dropped that node.** An edge pointing at a non-activatable kind was indistinguishable from an edge that failed to resolve or one that doesn't exist. A pack whose executable check hangs off an `asset` was both unreachable *and* unreported, with nothing in the output to notice. (Closes #3705.)

## After this change

Every node the cascade drops for being a non-activatable kind now gets its own line:

```
Not cascaded: asset/qa-traceability-lint (kind not charter-activatable)
```

- The same wording appears in all three places a cascade is reported — `charter activate --cascade`, the no-`--cascade` warning path, and `charter deactivate --cascade` — from **one shared helper**, so they can't drift.
- When a cascade resolves *zero* activatable targets because every referenced node was kind-filtered, that is now stated explicitly instead of looking like a successful no-op.
- Kind-filtering stays distinct from scope-narrowing: a node excluded by `--cascade-scope` still reports as `Skipped (out of scope)`, never as not-cascaded.

**This is a visibility fix only.** Which kinds are charter-activatable is unchanged and deliberately not reopened — the exclusion of `template`/`asset` remains the recorded policy of ADR `2026-08-20-1`, and `src/doctrine/artifact_kinds.py` has zero diff. Actually *invoking* a pack-shipped code asset as a gate handler stays tracked separately as #2599.

## How it works

`src/charter/cascade.py`'s `_referenced_artifacts` seam now returns a partition — `(activatable, kind_filtered)` — instead of dropping filtered nodes at a bare `continue`. All three consumers read from that one seam, so there is a single source of truth for both *what* was dropped and *how* it is worded.

| File | Change |
|---|---|
| `src/charter/cascade.py` | `_referenced_artifacts` returns `(activatable, kind_filtered)`; the three result dataclasses carry `not_cascaded_kind_filtered`. |
| `cli/commands/charter/activate.py` | Renders a `Not cascaded:` line per filtered node in the cascade report and the no-cascade warning, via one shared helper; emits the zero-activatable-targets message. |
| `cli/commands/charter/deactivate.py` | `charter deactivate --cascade` reports symmetrically through the same helper. |

## Tests

ATDD, red-first per work package (18 new tests). Verified non-vacuous: the pre-change sources were extracted and every new assertion fails against them. Suites: `tests/charter/test_cascade.py`, `tests/specify_cli/cli/commands/charter/test_charter_activate_commands_cascade_output.py`, `tests/specify_cli/cli/commands/charter/test_charter_deactivate_commands.py`, plus a dead-symbol allowlist re-pin in `tests/architectural/test_no_dead_symbols.py`.

## Landing notes

- **Rebased** onto current `upstream/main` and **flattened** from 86 commits (3 merge commits + mission-scaffolding noise) to 6 reviewable commits; the resulting tree was verified byte-identical to the validated merge. Upstream's since-merged changes to `activate.py` (#3702) and `test_no_dead_symbols.py` (#3723) were 3-way merged, not overwritten.
- **One review-squad fold** (`fix(landing):`, mine): the "Cascade resolved zero activatable targets (every referenced node was kind-filtered)" summary wrongly fired in the *mixed* case (a narrow `--cascade <scope>` skips an activatable-kind node **and** a kind-filtered node exists), contradicting the `Skipped (out of scope)` line printed right above it. The guard now also requires `not skipped_by_scope`, matching the intent its own comment already stated; red-first regression test added.
- **Deferred to #3772** (pure de-dup, no behavior change): consolidate the thrice-duplicated config-id resolution, extract a `_bucket_by_kind` helper, unify the `DeactivationPlan` kind-filtered field shape, and relocate the shared render helper to a neutral module. Both review lenses agreed these touch pre-existing code beyond this PR's scope and belong in their own change.
- **mypy** (advisory in CI): two pre-existing `no-any-return` in `activate.py` (`_source_urn`, `_drg_id_to_config_id`) remain, untouched by this PR.


MOES-Media:
## Landing pass — evidence trail

**Verdict: merge-ready.** No merge performed; the operator merges.

Full CI went green on `1192f9839` — **78 checks, 41 success, 37 skipped, 0 failures**. The only change since is an additive changelog conflict resolution for #3710 (no code), currently re-running.

### Base

**Merged `origin/main` in, deliberately not rebased.** A rebase would have rewritten ~60 commits of mission state, invalidating both the SHAs cited throughout the committed review trail and the `accept_commit` recorded in `meta.json`. A merge preserves every one of them.

Mid-pass this produced exactly the changelog conflict called out in "Landing-order constraints" below — **twice**. #3706 landed, then #3710 landed while CI was still running on the first resolution. Both were `docs/changelog/CHANGELOG.md`, both at the top of `### 🐛 Fixed`, and both resolved additively: **all entries kept**, main's first. No other file conflicted in either merge. #3706 also touched `tests/architectural/test_no_dead_symbols.py`; that auto-merged and the gate was re-run on the merged state to confirm (26 passed).

**This is a race, and it is worth knowing before you merge.** A full CI cycle here is ~20 minutes, and `main` is currently landing changelog-touching PRs faster than that. This PR went green on `1192f9839` and was `CONFLICTING` by the time the run finished, purely from #3710's changelog entry — no code involvement. Whoever merges should expect either to merge promptly on green, or to resolve one more additive changelog conflict. The conflict is mechanical every time: keep all entries.

### Red adjudication

| Check | Bin | Action |
|---|---|---|
| `cutover-guard` | **PR defect** | Fixed — see below |
| `arch-adversarial (arch_shard_1)` | **Stale pin the PR legitimately crosses** | Re-pinned — see below |
| `quality-gate` | **Aggregator only** | Its sole blocking verdict was `arch-adversarial`; no independent failure |
| `commitlint` (4 of 5) | **Pre-existing tooling gap** | Folded the root fix |
| `commitlint` (1 of 5) | **PR defect, disclosed not fixed** | See "Known, not fixed" |

**`cutover-guard`** failed with `status_phase not flipped despite event-log runtime evidence`. Reproduced locally. The remedy the guard prints — `spec-kitty migrate backfill-runtime-state` — is a **no-op** here: it correctly reports the mission "already migrated" and flips nothing. The actual remedy is that the mission lifecycle was never finished: `spec-kitty accept` is what stamps `status_phase`. Running it required clearing five gates that surface one at a time:

1. `acceptance-matrix.json` still held its nine scaffolded `pending` placeholders — now recorded via `agent mission acceptance-verdict`, each with the specific test that pins it (FR-006 is the ADR-citation process gate, recorded as manual verification against this PR body, no test invented).
2. `WP04: missing agent in canonical runtime state` — WP04 was claimed via `agent status emit` rather than `spec-kitty next`, so it never received the claim-seam resolved-binding annotation that WP01–WP03 carry. Restamped.
3. That restamp required WP04's subtasks T017–T021 to be event-sourced done first.
4. `issue-matrix` verdicts, revealed in two batches: `#3705` → `fixed`; `#2536`, `#2599`, `#3284` → `deferred-with-followup` (all three are explicitly out-of-scope, separately-tracked concerns named in `spec.md`/`plan.md` and untouched here).
5. `contracts/` deliverable missing — see the fold below.

`spec-kitty cutover-guard --base-ref origin/main` now reports `Un-cut-over: 0`.

### Folds

Each is a single-purpose commit.

- **`chore(spec-kitty): complete mission acceptance state`** — the acceptance/issue matrix and WP04 records above.
- **`docs(contracts): pin the kind-filtered node reporting output contract`** — Software Dev Kitty requires a `contracts/` deliverable, and this mission has a real one: three consumers must report kind-filtered nodes identically (C-002) through one shared helper (FR-009). Pins the seam's tuple return, the exact rendered line and its config-stem id resolution, the deliberate `DeactivationPlan` shape divergence authorized at `plan.md:242-259`, the once-never-per-node zero-activatable message and the two cases it excludes, and the C-006/NFR-002/NFR-004 invariants.
- **`chore(landing): cover spec-kitty's own acceptance and analysis commits in commitlint ignores`** — the config declares in prose that generated mission-lifecycle commits are exempt, but its one pattern covers only the planning generators. Four machine-authored subjects escape it (`Accept …`, `Record acceptance commit for …`, `Finalize acceptance artifacts for …`, `Add analysis report for mission …`); running `accept` once minted three of them. **These already leak onto `main`** — a scan of 4000 subjects finds `Accept <slug>` ×2, `Record acceptance commit for <slug>` ×2, `Add analysis report for mission <slug>` ×1 — because commitlint lints only a PR range, so the violation lands on whichever PR carries the commits rather than the mission that generated them. Four anchored patterns added, each commented with the `file:line` of its generator. Verified non-vacuous: an untyped subject still fails.
- **`test(landing): re-pin the DeactivationPlan dead-symbol allowlist hash`** — `arch-adversarial` failed the symbol-level dead-code gate on `charter.cascade::DeactivationPlan`. Not dead code and not pre-existing breakage: the symbol is already allowlisted under category C, but the allowlist is keyed by a **content hash of the symbol body**, and WP04 legitimately changed that body by adding `not_cascaded_kind_filtered`. Category C is documented as re-derived each cycle rather than a standing tracker, so this is a re-pin, not a new exemption — the symbol's status is unchanged (still no `src/` importer, before or after this branch; verified on both `origin/main` and HEAD). New hash computed with the repo's own `_symbol_key.py` `body_hash` helper, prior hash retained in a comment with a dated rationale.
- **`docs(landing): add consumer-facing changelog entry`** — a user-facing output change was missing its `Unreleased` entry. Adds **zero** markdownlint findings; the file's 38 pre-existing `MD049` findings are untouched (markdownlint is file-scoped, and they are already on `main`).

### Pre-merge review squad

Report-only R1–R3 over the implementation diff; full trail committed under `kitty-specs/cascade-asset-silent-drop-01M0RME0/reviews/pr.*.yaml`.

R1: `boundary` 2 findings (sev 3), `contract` 0, `tests` 0. R3: **both refuted**, verified independently.

- The underscore-prefixed cross-module import is an established pattern here — `charter/__init__.py:49-55` and `charter/pack.py:192-196` both import private symbols from sibling `generate.py` — and the helper's placement was a pre-recorded plan decision (`plan.md:682`).
- The cited duplicated block is pre-existing untouched code: the `deactivate.py` diff is pure addition, zero removed lines. Fixing it would be scope widening in a targeted fix PR.

Confirmed set is empty, with the full trail as evidence the squad ran.

### Test evidence

- `fast-tests-charter`: **2521 passed, 28 skipped, 0 failed** (re-run on the merged tip), coverage **90.94%** (floor 55%).
- `fast-tests-cli`: **3236 passed, 4 skipped, 0 failed** (re-run on the merged tip; the count rose from 3110 because the merge pulled in main's newer tests).
- `tests/architectural/`: **1679 passed, 0 failed** with the lint extras installed. Worth knowing for anyone reproducing: a checkout built with a bare `uv sync` has neither `ruff` nor `mypy`, and that alone fails 8 architectural tests fail-closed — phantom reds that look architectural. `uv sync --extra test --extra lint` is the working recipe (`--extra lint` alone removes `pytest`).
- Red-first verified per work package: each WP landed its failing test as its own commit before implementation, and a reviewer confirmed RED-on-base → GREEN-on-final by actually reverting. Non-vacuity was additionally proven by extracting the pre-change sources and running all 18 new tests against them — every new assertion fails pre-change.
- Shard collection confirmed (the class #3241 names): the 3 new engine tests land in `fast-tests-charter` (auto-marked `fast` by `tests/charter/conftest.py`); the 24 new CLI tests carry `pytest.mark.fast` and land in `fast-tests-cli`. Nothing orphaned.
- `tests/architectural/test_no_legacy_terminology.py`: 10 passed. `check_docs_freshness.py --ci`: exit 0.

### Known, not fixed — disclosed

- **One commitlint violation remains: `c765c690a`, subject `tasks: …`.** `tasks` is not in `type-enum`, and a scan of 4000 subjects on `main` finds **zero** precedent for it — so this is an authoring mistake, not an incomplete enum, and adding `tasks` to the enum to clear it would be green-washing. Rewriting the subject is also the wrong trade here: it would rewrite ~40 commits and invalidate both the SHAs cited throughout the committed review trail and the `accept_commit` recorded in `meta.json`. commitlint is advisory (`continue-on-error: true`; the `lint` job hard-fails only on `bandit`/`pip-audit`), so this costs a `lint-feedback` comment and nothing else. **A squash merge erases it entirely.**
- **Pre-existing flake, unrelated:** `test_glossary_validate.py::test_human_output_shows_valid` is a terminal-width line-wrap flake, reproduced identically with this branch's files reverted.
- **Coverage-combine flake under `-n auto`:** a repeat full-shard run reported 42% where an identical earlier run reported 90.94%. Isolated per-file runs put `cascade.py` at 99%. If CI shows a wild coverage number, re-run before treating it as a regression.
- **Two pre-existing mypy `no-any-return` errors** in `activate.py` (`_source_urn`, `_drg_id_to_config_id`), untouched.

### Landing-order constraints

**No source-code overlap** with any other open PR — `cascade.py`, `activate.py` and `deactivate.py` are touched by this PR alone.

**The changelog is the contended file, and it already conflicted once during this pass** (#3706, resolved above). Eight other open PRs still edit `docs/changelog/CHANGELOG.md` — #3710, #3693, #3690, #3686, #3666, #3015, #2890, #2239 — and every entry inserts at the top of the first `### 🐛 Fixed` section, the most contended spot in the file. Expect this to recur for whichever lands next. It is always a pure additive-ordering conflict: **keep both entries**, and resolve in `docs/changelog/CHANGELOG.md` (the canonical file; root `CHANGELOG.md` is a symlink, so resolve once).

**#3702 is a future sequencing risk only** — it targets the same `activate.py` command surface and is currently unassigned; whichever of the two lands second should expect to rebase.

MOES-Media:
**Green and un-drafted — merge-ready.**

Final state on `d2147070d`: **78 checks, 41 success, 37 skipped, 0 failures**, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`. All three held simultaneously at the moment of flipping to ready, which on this branch has not been a given — see the changelog race in the remediation summary above.

Nothing further from me. **I have not merged and will not** — that is the operator's call.

One thing to know before merging: a squash merge is the cleaner option here. The branch carries the mission's design and state artifacts alongside the implementation, so the commit count is high relative to the ~950-line source diff, and a squash also erases the one disclosed commitlint violation (`c765c690a`, subject type `tasks:`) that is advisory-only and deliberately not rewritten.

MOES-Media:
**Ready-state CI complete — 83 checks, 47 success, 36 skipped, 0 failures.**

Correcting my previous comment's count: flipping out of draft spawned five additional checks. This repo runs a reduced chain on drafts and the full diff-relevant chain only once a PR is ready, so `integration-tests-charter` and `e2e-cross-cutting` had never run while this was a draft. Both now pass — `integration-tests-charter` being the one that actually exercises the charter surface this PR changes.

Final: `d2147070d`, `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, not draft, no failing or pending checks.

stijn-dejongh:
**Maintainer landing pass — claimed** (assigning to @stijn-dejongh).

Picking this up for a landing pass: isolate a worktree, rebase onto current `upstream/main`, classify red checks into the four bins, campsite-fold as needed, refresh the PR body + changelog, then hand off for the operator to merge.

### Early viability triage → **LAND**

Sized on both axes: **substantive** (behavior change to `charter cascade` reporting; touches the shared `_referenced_artifacts` seam) and moderately aged (4 days, 86 commits — mostly the mission dossier — and currently `CONFLICTING`, so it needs a rebase). Ran the full compatibility triage:

- **Not superseded.** `main`'s `src/charter/cascade.py::_referenced_artifacts` still drops `template`/`asset`-kind nodes at the `CHARTER_ACTIVATABLE_KINDS` membership filter with a bare `continue` — no warning, no "skipped" line — exactly the behavior #3705 documents. No `kind_filtered`/reporting handling exists on main; `git log -S kind_filtered` on `src/charter/cascade.py` is empty.
- **Compatible.** The seam this PR modifies is unchanged in shape on current main, so the `(activatable, kind_filtered)` tuple refactor grafts onto the current architecture cleanly. `src/doctrine/artifact_kinds.py` has zero diff — the activatability policy (ADR 2026-08-20-1, C-001) is untouched; this is purely a visibility fix.
- **Still needed.** #3705 is OPEN (P2, doctrine/usability) and reproduces on current main.

Source diff is tight (~250 lines across `cascade.py` + the two charter CLI commands, ~720 lines of tests); the large additions count is the `kitty-specs/…` mission dossier, not code. Proceeding with the standard landing pass. Evidence to follow.

stijn-dejongh:

**Verdict: LAND.** Triage confirmed the fix is not superseded (main still silently drops `template`/`asset` at the `CHARTER_ACTIVATABLE_KINDS` filter), still needed (#3705 open, reproduces on main), and compatible (the `_referenced_artifacts` seam is unchanged in shape on current main). Rebased, folded one review finding, handed off for the operator to merge.

### History
Rebased onto current `upstream/main` and flattened **86 → 6 commits** (dropped 3 merge commits + mission-scaffolding/status-transition noise), keeping the fix as one `feat`, the dossier/changelog/commitlint/dead-symbol commits separate, plus the landing fold. The rebuilt tree was verified **byte-identical** to the validated merge. Upstream's since-merged edits to `activate.py` (#3702) and `test_no_dead_symbols.py` (#3723) were 3-way merged, not clobbered; the `CHANGELOG.md` conflict was union-resolved in the canonical `docs/changelog/`.

### Folds
- **`fix(landing):` (1)** — the FR-004 "zero activatable targets (every referenced node was kind-filtered)" summary fired in the **mixed** case (narrow `--cascade <scope>` skips an activatable-kind node *and* a kind-filtered node exists), printing a line the `Skipped (out of scope)` output directly above it contradicts. Guard now also requires `not skipped_by_scope`. **Red-first proof:** the added test reproduces the false summary against the pre-fix guard, then passes after. Surfaced by the review squad (architect lens, M1).

### Adversarial review squad (3 profile-loaded lenses, read-only)
- **reviewer-renata (correctness)** — no MAJOR/MINOR. Edge cases (empty cascade, multi-edge dedup, ordering determinism), the WP03 `has_skipped` widening, and the config-stem fallback all verified correct; non-vacuity spot-checked.
- **architect-alphonso (design/contract)** — no MAJOR. **M1** folded (above). **M2** (the `DeactivationPlan` flat-`list[str]` vs `dict[str,list[str]]` shape asymmetry — rendered output byte-identical today, latent for a future `--json` consumer) → deferred.
- **paula-patterns (SSOT/duplication)** — land as-is. Confirmed the whack-a-field surface is clean (all three `_referenced_artifacts` callers consume the new partition; no fourth silent-drop site). Config-id-resolution + kind-bucketing duplication (pre-existing, *extended* by this PR) → deferred.
- **Deferred cleanup filed as #3772** (parented under epic #1799): pure de-dup, no behavior change; both lenses agreed it touches pre-existing code beyond this PR's scope.

### Local evidence (worktree, rebased tip)
- `tests/charter/test_cascade.py` + both CLI cascade suites: **84 passed** (was 83 + the new mixed-case regression).
- `tests/architectural/test_no_dead_symbols.py`: **26 passed** (allowlist re-pin correct).
- Marker/collection completeness (`test_marker_job_completeness`, `test_ci_collection_completeness`, `test_arch_shard_marker_completeness`): **65 passed**. New test lives in an existing `pytest.mark.fast` file — no new marker/baseline needed.
- `ruff`: clean. Terminology guard: **10 passed**. `commitlint`: clean over all 6 commits.
- **mypy (advisory in CI):** two pre-existing `no-any-return` in `activate.py` (`_source_urn`, `_drg_id_to_config_id`) remain — in functions untouched by this PR, present on main.

### State
Folds pushed; **watching CI — merge-ready on green.** @stijn-dejongh to merge (I don't).

@MOES-Media — note the canonical changelog is `docs/changelog/CHANGELOG.md` (root `CHANGELOG.md` is a symlink to it); your entry was preserved and union-merged there.

Issue #3044, opened 2026-07-28
Epic: review-artifact integrity — approval must write a real verdict, and no cycle may fabricate one


stijn-dejongh:
## Epic: review-artifact integrity — approval must write a real verdict, and no cycle may fabricate one

**Why.** The `review-cycle-N.md` artifact is not a report *about* a decision — it *is* the decision, as far as the workflow is concerned. Every terminal gate in the loop (`move-task --to approved`, `move-task --to done`, `spec-kitty merge` via `terminal_wp_latest_review_artifact_must_not_be_rejected`) resolves the highest-numbered artifact for a WP and reads its verdict. Three independently-reported defects — spanning May, June and July, on three different missions — show that **three separate mechanisms let that artifact carry a verdict no reviewer ever authored**:

- **By omission.** A genuine approving review persists no artifact at all, so the prior cycle's `verdict: rejected` stays highest-numbered and authoritative (#2275, and half (a) of #2996). #2275 adds the structural framing: the per-WP approve guard reads the **lane** worktree — where no review artifact lives, so approval passes — while the merge gate reads the **coord** worktree, where the stale rejection is still latest. The two ends of the same decision consult different surfaces.
- **By fabrication.** Where a cycle-N+1 artifact *is* written, it can be a mechanical duplicate: byte-identical body to cycle N, with synthetic frontmatter prepended (`reviewer_agent: unknown`, empty `affected_files`, empty `reproduction_command`) and still `verdict: rejected`. That is a review filed under a cycle number implying a second, independent review happened when none did (half (b) of #2996, reproduced on three WPs of one mission).
- **By contamination.** Review-cycle generation can wrap or embed prior-cycle frontmatter and body into the new cycle's file, so the latest artifact carries an earlier cycle's material and its provenance can no longer be read off the file (#990).

They are not one bug with three reports. They fail at different points — the write-on-approval path, the `create_rejected_review_cycle` synthesis path, and the cycle-generation/templating path — and each would survive a fix to the other two. What they share is the blast radius: the durable evidence a gate trusts stops corresponding to what a reviewer actually did.

[...]

PR #3712, opened 2026-08-24
fix: make concurrent review verdict saves durable

### LINKED ISSUE #3235: P0 data-loss race: concurrent review-cycle commits can silently lose or leave uncommitted a verdict (SC-004 NOT MET)
## Correction (landing pass for #3209, 2026-08-06)

This issue was originally filed as "SC-004 concurrency test is flaky: different failure mode each run" — that framing was **wrong**. It is not an unexplained test-infrastructure flake. Re-reading `tests/integration/test_review_durability_matrix.py`'s own module docstring (WP15, T067-T069, FR-015) shows the root cause and rationale are already fully diagnosed and documented by the mission that wrote this test:

> **SC-004 -- NOT MET, and this WP says so plainly.** ... the probe CONFIRMS the loss is real under genuine OS-process concurrency: the commit phase (`_commit_review_cycle_artifact`, deliberately OUTSIDE `feature_status_lock` per NFR-006) has NO protection against two processes racing `git add`/`git commit` in the SAME working tree, and this test observes that race land as a silently-lost or silently-uncommitted record. This is a **load-window race**, not a "runs clean alone" story ... Per this WP's test-only mandate ("if a fix seems to need production code, STOP and report"), no production fix is attempted here. This test is deliberately NOT marked `xfail`/`skip` (forbidden) and its assertions are NOT weakened to paper over the finding.

## What this actually is

A real, confirmed **P0 data-loss bug**: `create_rejected_review_cycle`'s commit phase has no protection against two concurrent OS processes racing `git add`/`git commit` in the same working tree, so under genuine concurrency one reviewer's verdict can be silently lost or left uncommitted — reported as success on both sides, one file wins (or neither commits cleanly). `tests/integration/test_review_durability_matrix.py::test_sc004_two_concurrent_processes_never_clobber_a_verdict_over_50_iterations` is the WP15-authored, deliberate red-first reproduction. Two different failure shapes have been observed across runs (a `safe_commit` staging-area refusal, and a reported-but-missing artifact) — consistent with a genuine race with more than one possible losing outcome, not two different bugs.

## Why it wasn't caught as a P0 red-first test until now

The test carries no `@pytest.mark.regression` marker and has no GitHub issue pinning it, so it was invisible to the project's own red-main-is-honest tooling (`docs/development/red-main-and-release-readiness.md`) and looked like ordinary flakiness in CI instead of a tracked, deliberate signal. This issue is that pin; `@pytest.mark.regression` has been added to the test in the #3209 landing pass, referencing this issue number.

## Ask

Fix the race: give `_commit_review_cycle_artifact` (or its caller) the same protection `feature_status_lock` gives the write phase — a lock, retry-with-rebase, or equivalent — so two concurrent writers can never silently lose or leave uncommitted a review verdict. Do not weaken the test's assertions or mark it `xfail`/`skip` to make it pass; per WP15's own scope note this needs a production fix, not a test change.

### LINKED ISSUE #3773: Harden verdict-durability queue: bound the in-queue status-lock wait, extract the shared FileLock primitive, and tidy the SC-004 oracle
## BLUF

The verdict-durability fix (#3235, PR #3712) landed correct and complete — the P0 data-loss race is closed and every verdict-commit path is queue-protected. A landing-pass review squad (architect-alphonso, reviewer-renata, paula-patterns) surfaced four **non-blocking hardening** items that were deliberately deferred (behavior-contract change + delicate-test risk make them the wrong thing to fold into a P0 landing). None reopen #3235.

## Items

1. **The "10-second bounded wait" is only partial — the in-queue `feature_status_lock` acquisition is unbounded.** *(architect, MINOR)* While the verdict queue is held, `cycle.py` acquires `feature_status_lock` with `timeout=-1` (blocks forever, `status/locking.py`). If a status-lock holder is *wedged* (not crashed — `FileLock` releases on process death), the queue-holder hangs indefinitely and every other verdict save in the checkout fails `VerdictSaveBusy` after 10s. Not a deadlock, not data loss (waiters fail truthfully), but the operation is not fully time-bounded despite the PR's promise. **Needs a design decision**: what bound is right given `BookkeepingTransaction` can legitimately hold the status lock a while — then translate `FeatureStatusLockTimeoutError` into the same `verdict_durably_persisted: false` busy envelope so the whole operation is bounded. Consider a regression test asserting the invariant "no path acquires the verdict queue while holding `feature_status_lock`."

2. **Revert-path `VerdictSaveBusy` doesn't surface the structured durability envelope.** *(architect, MINOR)* `revert_committed_verdict_write` → `VerdictRevertError` → generic `RuntimeError` ("operator attention required"). Correctly fail-loud, but the compound-failure envelope carries no `verdict_durably_persisted` field (here the verdict *was* persisted; only the undo failed). Add a structured field for machine-consumer parity.

3. **SC-004 oracle `state_refusal` leg is less self-contained than its siblings.** *(renata, MINOR)* `_sc004_refusal_is_causal`'s `state_refusal` branch returns `kind == "state_refusal"` without the evidence-absence probe the `busy`/`ownership`/`persistence` legs run. It is *backstopped* (a failed compensator reds via `unproven_refusal`), not a hole — but a one-line evidence-absence assertion would make it self-contained rather than dependent on the compensator's fail-loud behavior. (Deferred rather than folded to avoid perturbing the delicate multi-process P0 oracle during the landing.)

4. **Lock-primitive duplication + hardcoded lock-dir literal.** *(paula, MINOR + NOTE)* `verdict_commit_queue.py` and `status/locking.py` build the same primitive twice ("`filelock.FileLock` under `<git-common-dir>/spec-kitty-locks/`, re-entrant, bounded-wait, typed timeout"), and the literal `"spec-kitty-locks"` is hardcoded in both (S1192-adjacent across the module boundary). Keeping the two *locks* separate is correct (verdict-commit serialization must not carry event/status mutation); the ask is to extract the shared *primitive* + a named constant both instantiate. Relatedly, `status/locking.py` hand-rolls `git rev-parse --git-common-dir` while the queue uses the canonical `kernel.git_topology.git_common_dir`; converge `status/locking.py` onto the kernel resolver so two locks that coordinate across worktrees resolve the same dir identically. (Broader `git_common_dir` resolver sprawl — 6 copies — is pre-existing standing debt, not introduced here.)

## Why deferred, not folded

Item 1 is a behavior-contract change to a shared lock that needs a design decision (the right bound, interaction with legitimate long transactions); item 4 spans pre-existing `status/locking.py`; item 3 carries flaky-risk to the P0 concurrency oracle for a cosmetic (already-backstopped) gain. All three review lenses independently recommended landing #3712 as-is. These belong in a focused hardening change with their own tests.

robertDouglass:
## What was broken

Two agents (or two processes) saving a review verdict on the same checkout at the same time could **silently lose a verdict** — or leave one written-but-never-committed. The `git add`/`git commit` for a review-cycle artifact ran with no cross-process protection, so a concurrent save could clobber another's staging area or report success for an artifact that never actually landed. This is the P0 data-loss race in #3235 (SC-004 not met).

## What changes for a user

Concurrent verdict saves now **wait in line**. Every verdict commit runs behind a checkout-wide serialization queue, so two processes can't race each other into a lost or half-written verdict. A save that can't complete within its bounded window **fails explicitly** (`verdict_durably_persisted: false`) instead of reporting a false success — you get a truthful error, never silent loss. Unrelated git operations are **not** serialized by this queue.

## How it works

- **`src/specify_cli/review/verdict_commit_queue.py`** (new): a `FileLock`-backed queue keyed on `git_common_dir()`, so every linked worktree and mission of one physical checkout converges on a single lock while independent clones stay independent. It wraps the whole allocate → write → commit → **verify** sequence (the verify step byte-compares the committed blob at the governed ref — a git return-0 alone never counts as durable).
- The event/status lock (`feature_status_lock`) stays a **separate, narrower** lock — distinct lock file, distinct re-entrancy, distinct timeout error. It and the queue are only ever acquired queue-outer → status-inner, never the reverse, so there is no lock-ordering deadlock.
- Retained evidence and identical-retry adoption were hardened so a commit-infra exception can no longer unlink a recoverable artifact; a crash between write and commit is adopted on retry, not duplicated.
- A new `review-verdict-durability` CI workflow runs the multi-process durability matrix on every relevant change.

## Tests

- **Genuine multi-process** SC-004 test: 50 rounds of two concurrent `spawn` workers driving the real `move-task --auto-commit` command; an independent oracle reconciles each success's `event_id` against the event log, byte-verifies the committed blob, and requires distinct evidence pointers — a clobber reds it. Mutation-sensitive (neutering the commit reds the matrix).
- Topology matrix (SINGLE_BRANCH / LANES / COORD / LANES_WITH_COORD), a performance budget test, and process-death / bounded-busy queue tests.

## Landing notes

- **Rebased** onto current `upstream/main` (which already includes #3711) and **flattened** from 355 commits (55 merge commits + auto-rebase/status-transition noise) to 4 reviewable commits; the rebuilt tree was verified **byte-identical** to the validated merge. Only `CHANGELOG.md` needed a merge (union-resolved); no source seam had drifted on main.
- **Adversarial review squad** (3 profile-loaded lenses) — all returned **LAND**:
  - *Deadlock trace (architect):* confirmed safe — single-directional lock nesting, no circular wait; lock key correctly converges worktrees; write/commit/verify atomic under the queue (no TOCTOU).
  - *Correctness (renata):* no lost-write or false-success path; `VerdictPersistenceOutcome` makes a false-durable outcome unconstructable; tests non-vacuous and mutation-sensitive.
  - *SSOT (paula):* every verdict-commit path routes through the queue (full 3-site census); no bypass.

### Folded hardening (#3773 — all six items landed in this PR, then re-reviewed)

Rather than defer the squad's findings, they were implemented as landing folds (each its own `<type>(landing):` commit) and the fold set was re-reviewed by a fresh opus lens (APPROVED):

1. **Fully bound the save** — the in-critical-section `feature_status_lock` wait, held while the queue is owned, was unbounded (`timeout=-1`); a *wedged* holder could hang every verdict save in the checkout. Now bounded and translated into the truthful `verdict_durably_persisted: false` busy envelope. **Red-first proven** (deterministic hang → bounded busy).
2. **Scoped that bound** to the queue-held path only — the `--no-auto-commit` / local-only feedback paths (which never held the queue and have no such translation) keep the historical unbounded wait, so a rare contention there never becomes an envelope-less error. (From the fold-set re-review.)
3. **Structured revert envelope** — a revert-compensator failure (the verdict *was* persisted; only the undo failed) now surfaces the same structured durability envelope instead of a bare `RuntimeError`, still fail-loud.
4. **Shared FileLock primitive** — extracted the duplicated `filelock` + `spec-kitty-locks/` + bounded-wait plumbing shared by `verdict_commit_queue.py` and `status/locking.py` into one helper (`core/checkout_file_lock.py`) with a single named constant; converged `status/locking.py` onto the canonical `git_common_dir` resolver. **Behavior-preserving; the two locks stay separate.**
5. **SC-004 oracle `state_refusal` leg** made self-contained (proves evidence absence directly, not via the compensator's fail-loud behavior).
6. **Re-pinned `ROUTED_LOAD_META_FLOOR` 146→147** for the one routed `load_meta_fail_closed` call site this PR adds (census band the PR legitimately crosses).

- **Red-first proof (base fix):** neutralizing the queue at the integration seam makes the SC-004 test fail with `verdict_durably_persisted: false` / `unproven_refusal`; restoring it passes.
- **mypy** (advisory in CI): the pre-existing `no-any-return` set in untouched files remains; no fold introduced a new one.

Fixes #3235.
Fixes #3773.


stijn-dejongh:
**Maintainer landing pass — claimed** (assigning to @stijn-dejongh).

Picking this up: isolate a worktree, rebase onto current `upstream/main`, flatten the history, classify red checks into the four bins, run an adversarial review squad over the new cross-process locking seam, campsite-fold, then hand off for the operator to merge.

### Early viability triage → **LAND**

Substantive (a P0 cross-process durability fix introducing a checkout-wide bounded verdict-commit queue) and aged (100 commits, `UNSTABLE`), so I ran the full compatibility triage:

- **Not superseded.** `src/specify_cli/review/verdict_commit_queue.py` does not exist on current main, and `git log -S verdict_commit_queue` on main is empty. Recent verdict-path commits on main (`fix(status): unify review-verdict CLI path`, the doctor double-read fix) address different concerns; none serialize concurrent verdict commits. The data-loss race is unaddressed.
- **Still needed.** #3235 is OPEN and **P0** (reliability/git): "concurrent review-cycle commits can silently lose or leave uncommitted a verdict (SC-004 NOT MET)".
- **Compatible.** All four seams the fix touches (`review/cycle.py`, `review/artifacts.py`, `coordination/status_transition.py`, `agent/tasks_verdict_persistence.py`) still exist on current main; I'll confirm no contract drift during the rebase.

The large additions count is the mission dossier under `kitty-specs/`; the real source surface is ~10 files. Proceeding with rebase + clean history first, then the review squad. Evidence to follow.

stijn-dejongh:

**Verdict: LAND.** Triage confirmed not superseded (`verdict_commit_queue.py` absent from main; no concurrent-commit durability fix landed), still needed (#3235 open, P0), compatible (no source seam drifted on main). Rebased, ran an adversarial squad, deferred hardening; handing off for the operator to merge.

### History
Rebased onto current `upstream/main` (includes #3711) and flattened **355 → 4 commits** (dropped 55 merge commits + auto-rebase/status noise); the rebuilt tree is **byte-identical** to the validated merge. Only `CHANGELOG.md` needed a merge (union-resolved in canonical `docs/changelog/`); the contributor's entry was preserved.

### Verification (worktree, rebased tip)
- Unit suites (queue, cycle, artifacts, move-task, transition): **157 passed**.
- Multi-process durability + topology + performance (serial): **60 passed, 1 skipped** — incl. the SC-004 50-iteration two-process no-clobber test.
- **Red-first:** neutralizing the queue at the integration seam reds SC-004 with `verdict_durably_persisted: false` / `unproven_refusal`; restored → green. The test witnesses the race.
- Marker / collection / arch-shard completeness + terminology: **75 passed** (3 new test files carry `git_repo`/`integration` marks and route to jobs).
- `ruff` clean; `mypy` clean on the PR's files (one pre-existing `no-any-return` in untouched `scope_source.py`, advisory); `commitlint` clean over all 4 commits.

### Adversarial squad (3 profile-loaded lenses) — all **LAND**
- **architect-alphonso (design/lock-ordering):** deadlock **confirmed safe** — single-directional nesting (queue outer → allocation inner), no reverse acquisition anywhere in `status/`/`coordination/`; lock key (`git_common_dir()`) genuinely converges worktrees while keeping clones independent; allocate→write→commit→verify atomic under the queue (no TOCTOU).
- **reviewer-renata (correctness):** no MAJOR — no lost-write/false-success; `VerdictPersistenceOutcome.__post_init__` makes a false-durable outcome unconstructable; the SC-004 oracle is genuine multi-process + mutation-sensitive.
- **paula-patterns (SSOT):** no bypass — full 3-site commit-path census, every path queue-protected.

### Deferred (no folds) → #3773 (under epic #3044)
All findings are non-blocking hardening, deferred because they need a shared-lock behavior-contract decision or carry flaky-risk to the P0 oracle:
1. The 10s bound covers only queue acquisition; the in-queue `feature_status_lock` wait is unbounded (`timeout=-1`) — availability edge under a *wedged* holder, not data loss/deadlock. (architect)
2. Revert-path `VerdictSaveBusy` lacks a structured durability envelope. (architect)
3. SC-004 oracle `state_refusal` leg is backstopped but less self-contained than its siblings. (renata)
4. Lock-primitive + `"spec-kitty-locks"` literal duplicated across `verdict_commit_queue.py` / `status/locking.py`; converge `status/locking.py` onto the canonical `git_common_dir` resolver. (paula)

### State
Clean history pushed; **watching CI — merge-ready on green.** @stijn-dejongh to merge.

stijn-dejongh:
## Update — #3773 folded in (no longer deferred)

Per operator direction, the four hardening items previously deferred to #3773 were **implemented as landing folds in this PR** rather than filed for later. Milestoned 3.2.6; the PR now `Fixes #3773`.

**Delivery:** one profile-loaded implementer (`python-pedro`, sonnet) per item, in parallel, each in its own worktree; folds cherry-picked onto the branch and re-reviewed by a fresh `reviewer-renata` (opus) lens over the combined fold diff — **APPROVED**, with one MINOR that was itself folded (item 2 below).

Six fold commits (each `<type>(landing):`, commitlint-clean) on top of the 4 flattened PR commits:

1. `fix(landing)` — **bound the in-queue `feature_status_lock` wait** so the whole save is time-bounded, not just queue acquisition. A wedged (not crashed) status-lock holder could previously hang every verdict save in the checkout. **Red-first:** a separate-process holder makes the pre-fix path hang (caught via a bounded `Thread.join`), the fix returns a bounded `verdict_durably_persisted: false` busy result.
2. `fix(landing)` — **scope that bound to the queue-held path** (the fold-set review's MINOR): the `--no-auto-commit` / local-only callers never held the queue and have no busy-envelope translation, so they keep the historical unbounded wait; only the queue-held auto-commit path is bounded. New unit test pins the helper both ways.
3. `fix(landing)` — **structured revert envelope**: a `VerdictRevertCompoundFailure` carries the durability signal so the compensator-failed path surfaces `verdict_durably_persisted: true` (the verdict really is committed; only the undo failed), still fail-loud.
4. `refactor(landing)` — **extract the shared FileLock primitive** (`core/checkout_file_lock.py`) + single `spec-kitty-locks` constant used by both `verdict_commit_queue.py` and `status/locking.py`, and converge `status/locking.py` onto the canonical `git_common_dir` resolver. Behavior-preserving; the two locks stay separate (distinct file, re-entrancy, timeout error).
5. `test(landing)` — **SC-004 oracle `state_refusal` leg** proves evidence absence directly instead of relying on the compensator's fail-loud behavior.
6. `test(landing)` — **re-pin `ROUTED_LOAD_META_FLOOR` 146→147** for the one routed `load_meta_fail_closed` call site this PR adds (arch census band the PR legitimately crosses; verified green).

**Local evidence (integrated tip):** unit + primitive + status-lock + agent-tasks + git_ops — 691 passed / 3 skipped; multi-process durability + topology + performance + the new bounded-wait test (`-n0`) — 60 passed / 1 skipped; arch floor gate green; terminology 10 passed; `ruff` clean; `commitlint` clean across all 10 commits; `mypy` advisory (only pre-existing `no-any-return` in untouched files).

**CI classification (pre-fold tip):** the three reds observed on the earlier tip were `test_routed_load_meta_floor` (now folded — fold 6), the known `test_sc001_org_pack_mission_type…` parallel-isolation flake (PR touches no charter/mission-type code), and `quality-gate` (a rollup of those two). Watching the re-run on the folded tip now.

Still: operator merges; I don't. Merge-ready on green.

stijn-dejongh:
## CI remediation — the folded tip went red; classified and fixed

The first CI Quality run on the folded tip failed 7 jobs. I classified every failure and fixed the real ones (three more `(landing)` folds); none required reverting a fold.

**Root causes (3 distinct):**

1. **`NotAGitRepositoryError` — my regression** (`fast-tests-cli`, `integration-tests-cli`, `integration-tests-review`, `unit-contract-residual`; ~25 tests across `test_artifacts`, `test_review_cycle_counter`, `test_tasks_backward_emit`, `test_2684_force_provenance`, `test_trio_pure_cores`). The scoping fold's `_in_queue_status_lock_timeout` probes `verdict_save_queue_is_held` at the top of **every** `create_rejected_review_cycle` allocation; that resolves `git_common_dir`, which raises on the local-only feedback path when the checkout is not a git repo — crashing the allocator where it previously never touched git topology. **Fix:** a non-git path cannot own the checkout-wide queue, so the probe's `GitTopologyError` is caught and treated as "not held" (historical unbounded wait). Helper unit test extended with the non-git path. *(Slipped my earlier targeted runs, which used git-initialised fixtures — a real gap; the broad parallel sweep I now ran before re-pushing catches this class.)*

2. **`TypeError: VerdictDurabilitySignal … 'durably_persisted'` — base-PR defect** (`integration-tests-core-misc`; `test_two_partition_preview`). The base PR refactored `VerdictDurabilitySignal` from a `durably_persisted: bool` field to `outcome: VerdictPersistenceOutcome`, but a **pre-existing test not in the PR's own file diff** still constructed it the old way — invisible to a diff-only review, surfaced only when CI runs the whole suite. **Fix:** construct via `outcome=created.persistence`, mirroring production.

3. **Golden-count regrowth — my new test** (`arch-adversarial (arch_shard_1)`; `test_golden_count_ban`). The bounded-wait regression test's `assert len(outcome) == 1` is genuinely cardinality-only (the next line unpacks it), but the heuristic classifies a bare `len(...) == N` as convertible. **Fix:** the sanctioned inline `# golden-count: cardinality-is-contract` marker (not a baseline re-freeze).

`quality-gate` was a rollup of the above.

**Verified NOT mine (left untouched):** `test_issue_3466_finalize_target_branch_override` (2 tests) fail only in a **linked-worktree execution context** (`CHECKOUT_WRITE_OWNERSHIP_REFUSED` / a shared-object-store artifact) — reproduced **identically on the base PR tip with zero folds**, and absent from the CI failure list (fresh-clone CI passes them). A worktree false-red, not a regression.

**Local evidence after the fixes:** the previously-red test files pass (69 CLI/review + 57 characterization + the two arch gates); a broad parallel sweep of the failed shards (`agent/` + `review/` + `characterization` + `integration/review`) is **2629 passed** (the only 2 reds being the worktree artifact above); durability matrix + queue + bounded-wait 60 passed; `commitlint` clean across all 13 commits.

Re-pushed; watching the CI Quality re-run. Correcting my earlier hand-off: it was premature — the first watch caught a partial run before CI Quality's 40-job suite registered. I now gate the watch on that run specifically.

stijn-dejongh:
✅ **Green — merge-ready.** CI Quality re-run on the remediated tip (`1211c9f2bc`) is **67 jobs / 0 failures**; overall PR checks **55 pass, 0 fail**, `mergeStateStatus: CLEAN`, not a draft. All three CI reds from the first folded-run were classified and fixed (two mine — the non-git probe crash and the golden-count annotation; one base-PR test defect); the two `test_issue_3466` reds were verified as a linked-worktree artifact (identical on the zero-fold base tip, green on CI) and correctly left untouched.

Final state: #3235 fix + all of #3773's hardening folded in, 13 commits, clean linear history on current `upstream/main`. **Operator merges — I don't.** Hand-off complete; I'll verify the merge lands clean and clean up once you merge.

Issue #3412, opened 2026-08-14
expected-artifacts.yaml: YAML-syntax-malformed manifests still silently degrade to None (FR-016 gap)


MOES-Media:
## Summary

FR-016 (issue #3388, mission `expected-artifacts-manifest-repair-01KZY498`) made
`ManifestRegistry.load_manifest()` raise `pydantic.ValidationError` instead of silently
returning `None` — but only for **schema-level** malformation: a syntactically valid YAML
document with a typo'd/extra key, caught by the new `model_config = ConfigDict(extra="forbid")`
on `ExpectedArtifactSpec`/`ExpectedArtifactManifest`.

A genuinely **YAML-syntax**-malformed `expected-artifacts.yaml` (bad indentation, an unclosed
structure, a duplicate key `ruamel.yaml` rejects, etc.) still collapses into the exact same
"not found" outcome the mission set out to eliminate. That happens because
`MissionTemplateRepository.get_expected_artifacts` (`src/doctrine/missions/repository.py`,
untouched by #3388's diff) catches `(OSError, UnicodeDecodeError, YAMLError)` and returns
`None` — and `load_manifest()`'s `config is None` branch treats that identically to a missing
file.

## Where this was found

Confirmed by the pre-merge adversarial squad on #3388's PR (finding `pr-contract-001`,
severity 3, contract lens; refuted-not, survived R1-R3). The squad verified line-for-line
that:

- `src/doctrine/missions/repository.py::MissionTemplateRepository.get_expected_artifacts`
  (around line 362-383, catch at ~382) swallows `YAMLError` into `None`.
- `src/specify_cli/dossier/manifest.py::ManifestRegistry.load_manifest` (around lines 184-222)
  only raises for the schema-validation path (`ExpectedArtifactManifest.model_validate`);
  the `config is None` branch is reached for both "absent" and "YAML-syntax-broken".
- The one test whose name promised YAML-syntax coverage
  (`tests/dossier/test_manifest.py::TestSchemaHardeningAndLoudFailure`, formerly named
  `test_load_manifest_raises_on_malformed_yaml`, renamed to
  `test_load_manifest_raises_on_schema_violating_key` as part of #3388's remediation) actually
  fed a syntactically-valid, schema-typo'd fixture through the loader — never real broken YAML.

## Why this matters

FR-016's stated intent was that "absence and malformation are distinct outcomes" so an
operator debugging a broken mission doesn't get a silent "manifest not found" for a manifest
[...]

PR #3717, opened 2026-08-24
fix(charter): refuse activation of a mission type with an empty action sequence

MOES-Media:
## Impact

`spec-kitty charter activate mission-type <T>` will now **refuse** a mission type that resolves an empty action sequence, instead of reporting success and writing it into `mission_type_activations`.

Before this change, the first activation of such a type exited `0` with `Activated: <T>` — and left the project activated onto a mission type that cannot plan anything. Every later governed entry point then failed, including `charter activate` itself on a second run:

```
run 1:  charter activate mission-type qa  ->  exit 0  "Activated: qa"
run 2:  charter activate mission-type qa  ->  exit 1  "mission type `qa` resolved from layer
                                                       `org` has an empty action sequence."
```

An operator following the documented order — declare the pack, then activate — got a success message and a bricked project. After this change the first run refuses, with the same message the read path already produces, and **nothing is written**.

## Why the check already existed but never ran

`_resolve_action_slot` in `src/charter/mission_type_profiles.py` already performs exactly the right validation, and its own comment says why: *"Silently returning [] would plan nothing and report success (NFR-002); raise loudly instead, naming the id and the resolving layer."* That is #3386's fix and it is correct.

But it runs on the **read** path. Activation — the one operation whose entire job is deciding whether a type is usable — never asked. The check existed; it was called from the wrong side of the seam.

## What changed

- `src/charter/mission_type_profiles.py` — new public `validate_activatable_mission_type()`, plus `_resolve_with_extends_fallback()` factored out so the read path and the new gate converge on one extends rule rather than two inlined copies.
- `src/specify_cli/cli/commands/charter/activate.py` — a fail-closed preflight calling it, placed **before** `manager.activate()`, so a refusal always precedes any write.

The gate deliberately lives at the **CLI seam**, not inside `plan_activation()`/`commit_plan()`: that module's contract (C-008) states it performs no filesystem discovery, and resolving whether an action sequence is empty requires exactly that. The read path's `is_registered` short-circuit is unchanged (FR-004).

`promote_activations()` — a second, currently-inert activation write path — is explicitly out of scope and documented as such in the plan.

## Testing

Red-first per the charter's ATDD discipline (C-011): the failing test is a separate commit **before** the fix.

- Targeted surface `tests/charter/test_mission_type_profiles.py` + `tests/specify_cli/cli/commands/charter/`: **156 → 164 passed, 0 failed**. Baseline captured on this exact base before any change.
[...]

PR #3658, opened 2026-08-22
fix+docs(doctrine): M3 hardening — WP-iteration guard, governance-scope fail-loud, DRG projection docs (#3627, #3629, #3633)

stijn-dejongh:
## What changes for you

Two doctrine-authoring footguns are closed, plus test/doc hardening around the DRG projection seam. All under epic #3410 (charter/doctrine silent-drop).

- **A custom mission family no longer crashes the CLI (#3627).** If you defined a *custom* mission family that named a step `implement` or `review`, the work-package-iteration guard raised an uncaught `UnregisteredMissionFamilyError` and aborted the command. An unregistered family has no guard table to enforce, so the crash was pure collateral. Now the guard catches that one exception narrowly and degrades to "no guard failures" for unregistered families. Registered families (`software-dev`, `plan`, …) keep **byte-identical** guard coverage — the exception has a single raise site they never reach.
- **A bad governance-profile selection now fails loudly instead of vanishing (#3629, items 2 & 3).** A `governance-profile.yaml` whose `selected_*` list named an artifact that doesn't exist was *silently* projected into a phantom auto-minted graph node at generation time — so a typo in a governance selection disappeared without a word. Now DRG generation fails loud with a `ValueError` naming any unresolved governance scope-edge target, after all node passes complete and before the phantom-mint loop (mirroring the existing operating-procedure fail-loud precedent).
- **DRG scope-projection seam hardening (#3633).** An emit↔delivery channel-binding test *by enumeration* (so future emit/delivery drift is caught automatically), the `Relation.SCOPE` two-grain overload documented as **deliberate accepted debt** (per operator direction — not split), and a stale scope-count ledger entry brought current.

## How it's proven

- Red-first reproductions were empirically re-confirmed, then converted to permanent units (`test_cli_guard_family.py`, `test_extractor.py`); no `@pytest.mark.regression` survives.
- Verified: the #3627 `except` is provably narrow (single raise site; registered families unweakened) and the #3629 fail-loud cannot false-trigger (URN parity across all 8 scope kinds; the full `packs/built-in` graph builds clean).
- `tests/runtime/next/`, `tests/doctrine/drg/migration/`, `tests/charter/test_emit_delivery_bind.py`: **105 passed** locally on the changed files; terminology guard passed; `ruff` + `mypy --strict` clean on changed source (no suppressions).

## ADR edits (operator-reviewed & signed off)

[...]

PR #3729, opened 2026-08-24
fix(runtime): let custom mission families gate on their own expected-artifacts manifest

### LINKED ISSUE #3704: A custom mission family can never report a guard failure, and its `blocking:` manifest is never read

MOES-Media:
## What changes for you

**A custom mission family that ships its own `expected-artifacts.yaml` now genuinely blocks on unmet `blocking: true` requirements, instead of always silently passing.**

Before: `spec-kitty next`'s guard evaluation for any family outside the four built-in types (`research` / `documentation` / `software-dev` / `plan`) returned `guard_failures == []` regardless of what was actually on disk. A data-driven presence gate existed, but nothing consumed its result for a family outside the hardcoded guard-table dispatch — so a custom mission could advance straight past a step whose declared, `blocking: true` artifact had never been produced. The manifest's `blocking:` flag was decorative: the step-scoped `required_artifacts_for` that would have honoured it had zero callers.

After: the same evaluation layer the built-ins use now genuinely checks a custom family's declared manifest — present artifacts pass, a missing `blocking: true` artifact blocks with a real failure message and a `blocked` decision — at all three call sites (CLI pre-check, WP-iteration pre-check, composed-action guard) and at **both** the built-in and org-tier (`<org_root>/missions/<type>/expected-artifacts.yaml`) manifest locations.

### Operator-visible behaviour change

A custom mission that was previously advancing silently past an unmet `blocking: true` step **may now correctly BLOCK where it previously would not.** That is the intended fix. In-flight missions are **not** retroactively re-evaluated — `status.events.jsonl` and past decisions are never rewritten; only the mission's next evaluation after deploy uses the corrected logic.

### Unaffected

The four built-in families are byte-identical in behaviour before and after. An unregistered family with **no** manifest declared anywhere still runs to completion exactly as before (`UnregisteredMissionFamilyError` for strict dispatch; tolerant callers still degrade to `[]`).

## What's in the diff

5 source files, ~340 lines of production change; the remainder is test coverage.

| Area | Change |
|---|---|
| `runtime_bridge_cores.py` | dispatch branch on a new `blocking_artifact_names` snapshot field |
| `runtime_bridge_io.py` | org-tier manifest resolution; real `blocking_artifact_names` computation |
| `resolver.py` | org-tier manifest loading; `required_artifacts_for` restored to `__all__` |
| `runtime_bridge.py`, `runtime_bridge_composition.py` | `repo_root` threaded through the three call sites |
| `docs/changelog/CHANGELOG.md` | the operator-visible behaviour change above |

`ArtifactPresenceSnapshot.blocking_artifact_names` is typed `frozenset[str] | None` and checked with `is None`, so **"no manifest" stays distinguishable from "manifest present, nothing blocking."** Collapsing those two states is the silent-success failure mode this change exists to remove, so it is deliberately preserved rather than simplified away.

Manifest-consulting logic lives in the I/O layer: `runtime_bridge_cores.py` keeps its imports at stdlib + `runtime.next.decision`, and `tests/architectural/test_bridge_cores_import_boundary.py` stays green by construction.

## Evidence

- **Full mission test scope: 1147 passed / 0 failed before implementation → 1161 passed / 0 failed at this tip**, same selection, measured before the first change and again at the end. None of the known-red set from #3284 falls inside this scope, so any red here would be introduced rather than inherited.
- **Red-first per work package** (charter C-011), each verified by a reviewer who actually reverted the implementation: WP01 5 red → 57 green · WP02 8 red → 69 green · WP03 4 red → 17 green.
- **AC-10 end-to-end**: `TestAC10CustomFamilyOrgTierNextDecisionWalk` stands up a `qa` family at the conventional `<org_root>/missions/qa/expected-artifacts.yaml` layout and walks the real `next` phase callables — absent artifact blocks, artifact created advances. Proven non-vacuous: the blocking half goes red when the production change is reverted.
- **Diff coverage 100%** on `src/runtime/next/*` against the `diff-coverage` critical-paths gate.
- New tests are collected by real CI shards (verified through the shard map, not assumed).

## Review trail

The full pre-merge adversarial trail is committed under `kitty-specs/custom-mission-guard-failure-blocking-inert-01M0STY0/reviews/pr.*` — three lens files, the merge, the refuter's reasoning, verification, and orchestrator dispositions.

One severity-4 finding was raised and **fixed** during that round: `FR-010`'s `ManifestSchemaError` guard had been added to `resolver.py` only, while `gather_artifact_presence` reached a second, unguarded `model_validate()` in `_presence_filenames_for` first — so a schema-invalid manifest escaped as a bare `pydantic.ValidationError` through both production guard entry points. Green tests, dead fix: the existing tests exercised `required_artifacts_for` in isolation and never drove the real entry point. Fixed with a red-first test that does.

## Known and deliberately deferred

- **The `ManifestSchemaError` wrap is now hand-duplicated across three modules** (`dossier/manifest.py`, `runtime/resolver.py`, `runtime_bridge_io.py`) with no shared helper, and the copies have already drifted. This is the same defect class the severity-4 above belonged to. Extracting one shared validation helper — ideally with a gate forbidding bare `model_validate(` calls outside it — is the by-construction closure worth doing, but it edits a module outside this change's scope and is left for follow-up rather than folded in here unreviewed.
- Org-tier **YAML-syntax** failures still degrade silently to "no manifest" (distinct from the schema-invalid path fixed here). Pre-existing built-in/org asymmetry, out of scope for this change: **#3412**.

Closes #3704

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3729"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790198309&installation_model_id=427282&pr_number=3729&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3729&signature=08e8069d9e8ba6974c4a5087b36e981a960d1b072fc98d7e3df694233c579c21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
Claiming this PR for a landing pass: rebase onto current `main`, adjudicate every check against main's own run at the merge-base, fold remediation as needed, and hand back merge-ready. Evidence to follow in this thread.

Context for why this one is being picked up now: it closes #3704, which is the last upstream blocker for a downstream QA doctrine pack's end-to-end integration mission — that mission's acceptance bar requires a custom mission family to actually **block** with populated `guard_failures`, which is impossible while a custom family has no guard-table entry. So the interest here is specifically that the fix makes a custom family's failure *observable*, not merely that CI is green.

MOES-Media:
## Landing pass — evidence

### Brought current with `main`
Merged `main` (`34b8f2faa`) in rather than rebasing. Deliberate: this branch carries **three mission lane merges** across its 96 commits, so a flat rebase linearises them and re-fights every add/remove pair of the mission's own bookkeeping files (`status.json`, dossier snapshots, tracer logs) — churn with no bearing on the fix. I started the rebase, hit exactly that, and aborted rather than resolve dozens of conflicts in files the fix doesn't touch. The merge was **conflict-free** and preserves the reviewable lane history. Branch is now 0 behind main.

### Downstream verification — the fix achieves its purpose
The PR's own tests cover the mechanism. This is the consumer-side proof that it closes #3704's actual complaint: *a custom mission family can never report a guard failure*.

Built a venv from the landing tip and exercised `evaluate_guards_strict` with a custom family (`qa`):

| scenario | result |
|---|---|
| custom family, manifest resolved, blocking artifact missing | `['qa-traceability.yaml']` — **blocks** |
| custom family, all blocking artifacts present | `[]` — passes |
| custom family, **no** manifest reachable at any tier (`None`) | raises `UnregisteredMissionFamilyError` — fail-closed preserved |
| tolerant `evaluate_guards` on the `None` case | `[]` — unchanged |

Before this change a custom family reported zero guard failures **always** — not because it behaved, but because it had no guard-table entry. It now gets genuinely evaluated against its own manifest, while an unreachable manifest still fails closed. The `is None` vs `frozenset()` distinction in the implementation is load-bearing and correctly handled: an empty blocking set means "a manifest resolved and nothing is required", not "no manifest".

### Baseline for classifying reds
`main` itself is currently red at `34b8f2faa` on **CI Quality** (`integration-tests-agent`, `quality-gate`) and **Protect Main Branch** (`check-merge-compliance`). Any red on this PR matching those is pre-existing main breakage, not introduced here — I'll classify each against that baseline when the run settles rather than reading job status.

### State
CI re-running against current main. Still a draft — I'll un-draft on green and hand off. **I do not merge; the operator does.**

### Why this one was picked up
It is the last upstream blocker for a downstream QA doctrine pack's end-to-end mission, whose acceptance bar is that a QA mission **blocks at least once with populated `guard_failures`**. That bar was unreachable while a custom family had no guard-table entry — a green run would have been evidence of nothing. The table above is the first time that bar has been demonstrably reachable.

MOES-Media:
## Landing pass complete — merge-ready

**Status: green, un-drafted, no reds to classify.** 67 checks passed, 0 failed, `mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`.

### The pre-flip green was provisional — worth recording
While drafted this PR showed 35 success / 0 failures, which looks conclusive and wasn't. This repo runs a draft exemption: some jobs skip on a draft and the `ready_for_review` trigger re-runs them (FR-013), with GitHub blocking draft merges meanwhile. After un-drafting, the count went 35 → **67** passing, and the two draft-gated jobs (`integration-tests-core-misc`, `e2e-cross-cutting`) flipped from SKIPPED to SUCCESS. **A skipped job is not a passing job**; the honest signal only exists post-flip.

### Red classification
`main` itself is currently red at `34b8f2faa` on `integration-tests-agent`, `quality-gate` and `check-merge-compliance`. Against that baseline:

| check | main | this PR | bin |
|---|---|---|---|
| `quality-gate` | FAIL | **SUCCESS** | PR is green where main is not |
| `integration-tests-agent` | FAIL | SKIPPED | **pre-existing main breakage, module not touched** |
| `check-merge-compliance` | FAIL | n/a | main-branch-only workflow |

`integration-tests-agent`'s skip is legitimate, not a path-filter false green: its condition is `needs.changes.outputs.agent == 'true' || github.event_name == 'push'`. This PR changes `src/runtime/next/**` and `src/specify_cli/runtime/resolver.py` — not the agent module — so the filter correctly excludes it, while `main` runs it on every push. Nothing here is a red this PR introduced, so **no folds were needed**.

### Brought current with main
Merged `main` (`34b8f2faa`) rather than rebasing, and said so at the time. This branch carries **three mission lane merges** across 96 commits; a flat rebase linearises them and re-fights every add/remove pair of the mission's own bookkeeping files. I started the rebase, hit exactly that (~72 of 96 applied before the second stall, every conflict in `status.json` / dossier snapshots / tracer logs — none touching the fix), and aborted rather than resolve churn in files under no review. The merge was conflict-free.

### Downstream verification (posted earlier in this thread)
A custom mission family now genuinely blocks: missing blocking artifact → `['qa-traceability.yaml']`; all present → `[]`; no manifest at any tier → still raises, fail-closed preserved. The `is None` vs `frozenset()` distinction is load-bearing and correctly handled — collapsing them with a bare falsiness check would have silently reintroduced the very defect this closes.

### Hand-off
Green, un-drafted, evidence in-thread, no landing-order constraints against other open PRs. **I don't merge — that's the operator's call.**

stijn-dejongh:
## Maintainer landing pass — claim + early recommendation: **LAND**

Claiming this for a landing pass (rebase onto current `upstream/main`, red-classification, red-first verification, adversarial squad on the guard-dispatch/org-tier surface, hand-off for merge).

**Viability triage (substantive + architectural — ran the full compatibility check):**

- **Not superseded.** `blocking_artifact_names` and the org-tier `<org_root>/missions/<type>/expected-artifacts.yaml` resolution do not exist on current `main`; #3704 is unaddressed. The sibling #3658 fix (custom family no longer *crashing* the guard) closed a different defect — this PR closes the *inert-blocking* one (the manifest's `blocking:` flag was decorative because `required_artifacts_for` had zero callers).
- **Compatible.** The reviewable surface is 5 source files in `src/runtime/next/*` + `src/specify_cli/runtime/resolver.py` (~340 prod lines); disjoint from #3717's `charter` tree that landed today, so the rebase should be clean.
- **Still needed.** A custom family with a `blocking: true` artifact still advances silently past the unmet step on `main`.

**Scope:** 97 commits but well-contained — 5 source + 8 test files + a changelog entry the PR already carries; the rest is mission artifacts. Merge-base is yesterday (31 behind). CI is green on the current head (41 pass, 0 fail).

Because this touches the guard-dispatch seam and adds an org-tier resolution path, I'll run a profile-loaded adversarial squad over the diff (design/contract + SSOT/duplication) — the PR body itself flags a `ManifestSchemaError` hand-duplication across three modules as deferred, which the squad will weigh. Rebasing and verifying now; evidence to follow.

stijn-dejongh:

**Verdict: LAND.** Rebased onto current `upstream/main`, history cleaned, red-first re-proven, adversarial squad run (no blocker), one MINOR folded, one MAJOR tracked. Watching CI; merge-ready on green.

### History cleaned (97 → 5 commits)
The original branch hit a tracer-file conflict on replay, so I reconstructed directly onto **current** `upstream/main` (which now includes #3717): `chore(mission)` → red-first `test` → `fix` → `docs(changelog)` → `fix(landing)`. All commitlint-clean. The code+test tree is **byte-identical** to the PR tip; the code paths are 100% disjoint from main's 31 new commits, so nothing was lost.

### Campsite: dropped a gitignored dossier cache
`kitty-specs/.../.kittify/dossiers/.../snapshot-latest.json` was force-committed despite matching `.gitignore:85`. Not on main, shouldn't travel in the PR — the reconstruction correctly excluded it (the only intentional deviation from the PR tip).

### Red-first (live)
Reverting the 5 product files fails 6 negative-path tests — genuine-evaluation, the `None`-vs-empty-`frozenset` distinction, org-tier blocking (absent→block / present→pass), and `repo_root` threading through the WP-iteration and CLI pre-checks. All green on restore. The fix genuinely witnesses the bug.

### Adversarial squad (profile-loaded, opus) — no blocker
- **architect-alphonso (design/contract):** contract preserved end-to-end (the `is None` tri-state, not bare falsiness); built-in families byte-identical; org precedence consistent across both IO helpers; `repo_root` reaches all three production guard sites as a real `Path`; `ValidationError`→`ManifestSchemaError` nets catch the actual type. No MAJOR.
- **paula-patterns (SSOT/boundary):** `runtime_bridge_cores` stays a pure stdlib leaf (import boundary intact); `blocking_artifact_names` is a coherent field (`blocking − present`), not a whack-a-field patch; the org primitive is reused, not re-implemented. No blocker.
- **Convergence** (both lenses + the PR author): the `ManifestSchemaError`/`model_validate` wrap is hand-duplicated across three modules bypassing the cached canonical `ManifestRegistry.load_manifest` → up to 3× manifest resolution per `next` call, wasted for built-in families. **Ruled follow-up, not folded** — it changes swallow-vs-propagate semantics + cache behavior + needs a repo-wide arch gate, all outside this PR's blast radius. Filed as **#3770** (also pins architect's NOTE-1: schema-invalid *org* manifest for a *custom* family now fails loud — intended FR-010 behavior, currently untested).

### Fold (1)
- `fix(landing): match org-tier ManifestSchemaError origin across both entry points` — paula-patterns MINOR: `runtime_bridge_io.py`'s org-tier error `origin` had already drifted from `resolver.py`/`manifest.py` within this PR (dropped the "checked org roots" suffix). The shared helper now returns `(mapping, org_roots)` so both converge. No test pinned the old string.

- 259 targeted tests pass; the fold's affected files 28/28.
- `ruff` clean. `mypy`: CI `lint` green on head; a bare per-file run shows 4 `no-any-return` in `resolver.py` at lines unrelated to the diff — **reproduced identically on `upstream/main`** (pre-existing per-file-invocation artifact), not folded.
- `terminology` 10 passed; `docs-freshness --ci` errors=0. All 8 test files are modifications with proper `pytestmark` — no new-file baseline work.

### Compatibility (recap)
Not superseded (`blocking_artifact_names`/org-tier resolution absent on main, #3704 unaddressed; sibling #3658 closed a different defect). Compatible. Still needed.

### Hand-off
Watching CI; **merge-ready on green.** @stijn-dejongh to merge — I don't. No landing-order constraints.

@MOES-Media — note: the `.kittify/dossiers/` snapshot is gitignored local cache that shouldn't be committed (dropped here).

Issue #3785, opened 2026-08-28
accept: _missing_artifacts hardcodes the optional-artifact list instead of reading mission.config.artifacts.optional (already drifted from mission.yaml)


stijn-dejongh:
## Summary

`acceptance._missing_artifacts` hardcodes its optional-artifact list `["quickstart", "data-model", "research", "contracts"]` instead of reading `mission.config.artifacts.optional`. This is a second source of truth for "what artifacts are optional", and it has **already drifted** from the canonical config.

## Evidence of live drift

- `src/specify_cli/missions/software-dev/mission.yaml` declares `checklists/` under `artifacts.optional` (~line 145 region).
- `src/specify_cli/acceptance/__init__.py` `_missing_artifacts` hardcodes `[quickstart, data-model, research, contracts]` — it **omits `checklists`**.

So `accept`'s optional-artifact reporting silently disagrees with the mission's own declared optional set for any optional the hardcoded list doesn't happen to include.

## Why now / relation to #3783

Surfaced during the maintainer review squad on PR #3783 (mission `accept-path-remediation-honesty`, #3730/#3085). That PR's contracts double-report dedup is at the correct layer (report-layer policy: blocking severity wins) and works precisely because both the hardcoded list and the yaml **happen to agree on `contracts`**. The dedup does not depend on this drift, but the drift is a latent correctness hazard for any other dual-declared or optional-only token.

## Proposed fix (its own change — do NOT re-litigate severity policy)

Make `_missing_artifacts` (and any sibling that enumerates optional artifacts) read `mission.config.artifacts.optional` as the single source of truth, so the reported optional set always matches the mission's declared config. **Non-goal:** do not re-open whether `contracts/` blocks or warns — that policy is settled and lives at the report layer; this issue is purely SSOT unification of the optional-artifact enumeration.

Directives: 044 (canonical-sources-and-unification), 040 (recurring-boundary structural intervention).

stijn-dejongh:
Folded into mission `accept-path-convention-override` (#3016) as a severable P3 work package (WP04): `acceptance._missing_artifacts` will read `mission.config.artifacts.optional` instead of the hardcoded list (which drops `checklists/`). Guard: does NOT re-open the `contracts/` blocking-vs-warning severity settled by #3783. Split-tripwire recorded — if the fix must touch contracts dedup/severity or grows past the `_missing_artifacts` signature + call-site reorder, it splits back out to its own mission.

Issue #2744, opened 2026-07-16
research mission type reuses software-dev spec template and accept expectations — spec phase and terminus both mis-fit


MOES-Media:
## Summary
The research mission type does not ship a spec-phase template distinct from its decision-log/output template, and `accept` applies software-dev lane/WP/path expectations to research missions that legitimately have no plan/tasks/WP. Result: research missions must deviate from doctrine to author a spec, and cannot pass accept without fabricating plan/tasks or skipping to `mission close`.

Related open: #883 (mission-type governance profiles for non-software missions), #2329 (accept hardcodes src/+tests/ path convention). Related closed (strong regression signal — verify still-repros on current main): #1686 ("research-mission accept/merge fails for planning_artifact lanes — missing mission branch, acceptance-matrix.json, and path conventions") appears to describe the same accept-time failure class as B09.

### B04 — stock research-mission SPEC template doesn't fit the research spec phase
- Symptom: the shipped template is shaped as a Phase-0 "Research Decision Log" (an OUTPUT artifact), but the specify phase needs a spec-of-questions (Research Question → Background → numbered sub-questions → Methodology → evidence standard → FR/NFR/Constraints → Out of scope). Agents must deviate and mirror repo precedent (mc-driver-feasibility-01KX1HA8/spec.md).
- Repro: `spec-kitty specify` on a research mission; compare the emitted template to what the spec phase needs.
- Expected: research mission type ships a spec-phase template distinct from the decision-log/output template.
- Source: `~/.kittify/missions/research/templates/research-template.md`.

### B09 — accept applies software-dev lane/WP expectations to research missions
- Symptom: a research mission that followed its own action sequence (scoping → gathering → synthesis → output, no plan/tasks/WP) is blocked at accept on: missing `plan.md` + `tasks.md`, lane-based acceptance-matrix expectations, and an undocumented `docs/research/<slug>/research/` workspace-path expectation surfaced only at accept time.
- Repro: run a research mission to terminus; `spec-kitty accept`.
- Expected: mission-type-aware accept (research validates its own `expected-artifacts.yaml`); surface the `docs/research/` path convention at creation, not at terminus. Cross-check whether closed #1686's fix regressed or was incomplete.
- Source: `spec-kitty accept` (research mission type).

## Why bundled
[...]

Issue #2330, opened 2026-07-04
software-dev mission workflow assumes Python/pytest layout — friction for Go (and other non-Python) consumer repos


kentonium3:
**Context**: Ran a full spec-kitty mission (`specify → plan → tasks → analyze → implement → accept → merge → mission-review`) in a **Go** consumer repo (`spec-kitty-analyzer`, layout `internal/` + `cmd/`, tooling `go build/vet/test`). Several workflow steps have hard-coded Python/pytest/`src/`+`tests/` assumptions that required manual adjustment or produced misleading output. None blocked completion, but each is a papercut for non-Python dogfooders and would confuse a first-time Go user.

Filing so the software-dev mission surface can be made language-aware (or at least language-neutral with a configurable convention).

---

## 1. `mission accept` path convention hard-codes `src/` + `tests/`

`spec-kitty agent mission accept` (software-dev mission) fails acceptance with:

```
- path_violations
    • Path Convention Errors:
      - Software Dev Kitty expects workspace path: src/ (not found)
      - Software Dev Kitty expects tests path: tests/ (not found)
    Required Actions:
      - mkdir -p src/ tests/
```

For a Go repo the source lives in `internal/` + `cmd/` and tests are `*_test.go` co-located with code — there is no `src/` or `tests/`. Following the "Required Action" (`mkdir -p src/ tests/`) would create empty, wrong directories.

- **Workaround used**: `accept --lenient` (demotes it to a warning). But `--lenient` also skips *other* strict metadata validation, so it's a blunt instrument — you lose real checks to bypass a false one.
- **Even with `--lenient`** the path-convention **warning** and the `mkdir -p src/ tests/` suggestion still print, which is noise / actively misleading for Go.
- **Related**: this looks adjacent to the already-filed path-convention item (spec-kitty#2329); if that only covers the `--lenient` escape hatch, the deeper fix is making the expected paths **configurable per project** (charter or a `.kittify` setting) or **language-detected**, rather than a Python default.

**Suggested fix**: derive expected source/test globs from the charter's declared language/stack (the analyzer charter already says "Go 1.25+"), or make them a configurable convention; don't hard-code `src/`+`tests/`.

## 2. `spec-kitty-mission-review` skill hard gates 1–3 are spec-kitty-Python-source-only

The mission-review skill runs four hard gates. Three assume the **spec-kitty source repo** itself:

[...]

PR #3790, opened 2026-08-28
fix(accept): project path-convention override — accept non-src layouts honestly (#3016, folds #3785 + #2330 Item 1)

### LINKED ISSUE #3016: accept: software-dev path convention hardcodes src/, fails every mission in a Django (apps/) repo — and prescribes mkdir as the fix

stijn-dejongh:
## What changes for you

If your repo does not keep source under `src/` — a Django project using `apps/`, a Go project using `internal/` — `spec-kitty accept` used to **fail every mission** with:

> These directories are required by the active mission. Create them before continuing.

The only ways out were to `mkdir` empty directories just to pass the gate (a fake green) or to reach for a blanket `--lenient`. Neither is honest.

**Now you declare your real layout once** in `.kittify/config.yaml`, and `accept` checks the directories that actually exist in your project:

```yaml
# .kittify/config.yaml
project:
  path_conventions:
    workspace: apps/
    tests: tests/
```

No fabricated directories, no blanket leniency. **Closes #3016.**

## The fix

- **Override precedence:** project `path_conventions` → mission-type doctrine default → unchanged blocking + `--lenient`. The override remaps the directories the accept gate checks; it does not relax the gate.
- **Remap-only.** An override is applied only to a key the mission already declares — it never adds a new required path. Single merge point `_remap_declared_paths` in `validators/paths.py`, upstream of both the path-prefix comprehension and the artifact-token check.
- **Blocking-by-default is preserved** (additive to #3783). This supplies the missing *value* channel; it does not reverse the honesty mission's policy. An override that points at a directory which is still absent under strict mode **still blocks** (SC-006).
- **`deliverables` is excluded** from the override vocabulary: its default value is a mission artifact token, so overriding it would flip `feature_dir`↔`project_root` routing. Guarded by `ARTIFACT_ROUTED_KEYS`.
- **Fail-closed, scoped to the section:** a malformed `path_conventions` section raises with a message naming the offending key; a missing or corrupt `config.yaml` stays lenient (`{}`), matching the co-resident section readers.
- **All four mission types** by construction (one shared `validate_mission_paths` seam). No `mission.yaml` doctrine edit — project layout is project state, not doctrine.
- **Folds #3785:** `_missing_artifacts` now reads `mission.config.artifacts.optional` (via null-safe `get_optional_artifacts()`) instead of a hardcoded list that had dropped `checklists/`; a `getattr` guard preserves the artifacts-less-mission fallback. `contracts/` blocking severity is unchanged. Also folds **#2330 Item 1** (Go `internal/` layout).

## Files

`src/specify_cli/config/path_conventions.py` (new reader), `validators/paths.py` (`_remap_declared_paths` + `path_overrides`), `acceptance/summary_core.py` (seam wiring), `acceptance/__init__.py` (#3785 fold), `mission.py` (`VALID_PATH_KEYS` hoist), plus the ADR and its index/inventory registrations.

## Tests

- Reader / validators / acceptance / all-mission-types / single-caller unit coverage, plus the architectural-gate suite (dead-symbol, golden-count, dead-module, terminology, collection-completeness) — green on the rebased tip.
- **#3783 regression** `test_lenient_downgrades_path_conventions_to_warning` — green and **unchanged**.
- Non-fakeable pins: an override for an **absent** directory still blocks under strict (SC-006); one config read per accept run (NFR-002); the reader accepts the real `project:` identity fields without rejecting them (C-011).
- `ruff` + `mypy` clean on all changed source; `validate_mission_paths` complexity ≤15.

## Design record

ADR [`docs/adr/3.x/2026-08-28-1-project-path-convention-override-precedes-doctrine.md`](docs/adr/3.x/2026-08-28-1-project-path-convention-override-precedes-doctrine.md). Layout auto-detection (`manage.py` / `go.mod` signals) is the deliberate next step — **#2744**.

## Landing notes

Maintainer landing pass: rebased onto current `upstream/main`, red CI classified (all PR-defect gate joins — collection-completeness marker, contract round-trip skip, ADR SEO description, docs retrieval-index registration), folds applied, history consolidated to clean Conventional-Commits form, changelog entry added under `[Unreleased]`. The operator merges — no direct push to `main`.


stijn-dejongh:

Triage verdict: **LAND**. Confirmed still-needed and non-superseded:
- **#3016 is still open** — the non-`src` layout problem it targets is live on current main.
- The new reader `src/specify_cli/config/path_conventions.py` is **not** on main.
- Main's own changelog for the honesty mission (#3730/#3085) explicitly records *"the `src/`-hardcoded non-`src` repo portability (#3016) is unchanged"* — this PR is precisely that additive follow-up (project-override value channel, blocking-by-default preserved).

Plan: isolated worktree, rebase onto current `upstream/main` (branch is 33 behind), classify the red checks (arch-adversarial shards 2/3, docs-freshness, fast-tests core-misc/corpus/docs, quality-gate), fold every fix on-branch, dispatch a profile-loaded adversarial squad and **fold all finds incl. minors/nits** (close-out over defer), refresh the PR body + changelog, then hand off green for the operator to merge. Evidence to follow.

stijn-dejongh:
## 🛬 Maintainer landing pass — remediation summary

**Verdict: LAND.** Rebased onto current `upstream/main`, all red CI classified and folded, a 3-lens adversarial squad run with every finding folded, history consolidated. Handing off for the operator to merge — I do not merge.

### History
Rebased clean onto `upstream/main` (was 33 behind). Consolidated the 18 mission commits — 5 of which failed `commitlint` (`Add spec…`, `Harden spec…`, `Fold post-plan…`, etc. carried no Conventional-Commits type and would have red-mained the PR's commitlint step) — into **two clean, commitlint-passing commits**:
- `docs(accept): mission dossier for path_conventions override (#3016)`
- `feat(accept): project path_conventions override to accept non-src layouts (#3016)`

Tree parity verified against the pre-consolidation tip: content is identical modulo the intended folds (a nested, git-ignored-but-tracked `.kittify/dossiers/…/snapshot-latest.json` was force-re-added so nothing was dropped).

### Red CI → all PR-defect gate joins, all fixed
Every red was the PR's own new files not yet joined to a gate — no pre-existing breakage among them:

| Red check | Cause | Fix |
|---|---|---|
| `arch-adversarial` 2/3, `fast-tests-core-misc` | new `test_path_conventions_reader.py` shipped with **no `pytestmark`** → the `-m "fast"` shards collected-but-deselected its 10 nodes → orphaned on push-to-main (`test_ci_collection_completeness`, #2957) | marked `unit`+`fast` |
| `fast-tests-corpus`, `core-misc` | mission contract YAML block had no round-trip marker | `# round-trip: skip:` (shape sketch; real schema is the reader + its tests) |
| `fast-tests-docs` | new ADR description 338 chars > 180 SEO band | trimmed to 159 |
| `docs-freshness` | ADR added to page-inventory but not the generated retrieval index | regenerated `3-2-docs-retrieval-index.yaml` |
| `quality-gate` | aggregate of the above | transitively green |

### Adversarial squad (3 profile-loaded lenses) — every finding folded
- **Design/contract lens:** no MAJOR. Verified single-merge-point, remap-only, `deliverables`/artifact-routing exclusion, and fail-closed-section-vs-lenient-whole-config all faithfully implemented. NOTEs folded (see below).
- **SSOT/duplication lens:** no MAJOR. Confirmed the new reader is a **legitimately separate seam** (no shared config-reader helper exists to reuse) and the `VALID_PATH_KEYS` hoist is a genuine single-source win. Nit folded (`_SECTION` constant). One MINOR (drop the artifacts null-guards) was **verified not to hold** — dropping them regresses the #3783 artifacts-less fallback — so it was correctly not applied.
- **Correctness/test-integrity lens — one MAJOR, folded:** an **empty-string override** (`workspace: ""`) made `Path("")` resolve to the repo root (always exists), silently defeating strict blocking (SC-006) even though the config was "set". Fixed red-first (proved the pre-fix reader returned `{'workspace': ''}` and `validate_mission_paths(strict=True)` wrongly passed; post-fix it raises). Plus MINOR **absolute-path / `..`-traversal** values (same bypass class) → rejected.

**Folds applied (all on-branch, then consolidated into the `feat` commit):**
1. **Reject unsafe override values** — empty/blank, absolute, `..`-traversal — in the reader, with tests (MAJOR + MINOR).
2. **Merge-site hardening** — `_remap_declared_paths` re-excludes `ARTIFACT_ROUTED_KEYS` (defense-in-depth); a new guard drops an override whose *value* collides with a mission artifact token (design NOTE 1), with tests.
3. **Clean accept verdict** — a malformed `path_conventions` section now renders as an `accept` blocking verdict via the existing `AcceptanceError` handler instead of a raw traceback, with a focused non-vacuous test (design NOTE 3).
4. **`_SECTION` constant** (SSOT nit); **contract/spec doc reconciliation** — line-number refs replaced with drift-proof function names, and the empty-map (no-op) vs empty/absolute-value (fail-closed) prose corrected.
5. Changelog `[Unreleased]` entry (Added) and PR body rewritten consumer-first/BLUF.

Implementation was delegated to a `sonnet` implementer under an explicit no-touch file list; review/adjudication on `opus`. Model discipline held.

### Verification (rebased + consolidated tip)
- `ruff` + `mypy`: **clean** on all changed source (`validate_mission_paths` C901 ≤ 15).
- Reader / validators / acceptance / all-mission-types / single-caller + the new NOTE tests: **green** (collection-completeness 50 ✓, round-trip/SEO/marker gates 862 ✓, feature suites ✓).
- **#3783 regression** `test_lenient_downgrades_path_conventions_to_warning`: green, unchanged.
- Terminology guard ✓; `check_docs_freshness --ci` **errors=0** (only external-URL link-health warnings).

### Pre-existing red, NOT this PR (called out honestly)
`tests/cross_cutting/misc/test_acceptance_support.py::{test_accept_command_reports_approved_wps_without_closing, test_accept_no_commit_reports_merge_pending_without_mutation, test_accept_does_not_require_done_evidence_for_approved_wp}` fail locally with `Warning: Event missing project_uuid; queued locally only` — a **sync/SaaS emission environment artifact** (no configured project in the sandbox). Verified they reproduce **identically on the old merge-base and on current `upstream/main` tip**, so they are pre-existing and environmental, not introduced here. This PR touches neither that file nor the sync path. Expected green under CI's configured/disabled-sync env.

### State
Watching CI; **merge-ready on green.** Nothing deferred to a follow-up issue — every squad find (incl. minors/nits) was folded. Layout auto-detection (`manage.py`/`go.mod`) remains the deliberate next step in the ADR (#2744). The operator merges.

Issue #3831, opened 2026-09-01
Org-tier custom mission type is invisible to the mission loader and silently replaced by software-dev


MOES-Media:
## Summary

A mission type delivered by an org-tier doctrine pack and activated via `.kittify/config.yaml` loads as **Software Dev Kitty** instead of the actual configured type — carrying software-dev's path conventions and required-artifact set. The only signal is a `UserWarning` that does not reach CLI output, so an operator sees a mission silently validated against the wrong conventions with no indication anything went wrong.

Verified against source commit `eff63532607d2f292efc16e3f8bfd0e61e0544aa`.

## Mechanism

`specify_cli/mission.py::_mission_path_by_name` consults exactly two tiers — project, then packaged built-ins:

```python
def _mission_path_by_name(mission_name: str, kittify_dir: Path) -> Path | None:
    project_path = _mission_dir_if_valid(kittify_dir / "missions" / mission_name)
    if project_path is not None:
        return project_path
    return _mission_dir_if_valid(_packaged_missions_dir() / mission_name)
```

The org tier is never consulted here. `get_mission_for_feature` then swallows the resulting miss:

```python
try:
    return get_mission_by_name(mission_type, kittify_dir)
except MissionNotFoundError:
    # Fall back to software-dev with warning
    warnings.warn(f"Mission '{mission_type}' not found for feature {feature_dir.name}, using software-dev as default", stacklevel=2)
    return get_mission_by_name("software-dev", kittify_dir)
```

Measured on a consumer project with a custom mission type activated at the org tier and no `.kittify/missions/` directory: `get_mission_type(feature_dir)` correctly reports the custom type, while the `Mission` object loaded for the same feature reports name `Software Dev Kitty`, path conventions `{'workspace': 'src/', 'tests': 'tests/', 'deliverables': 'contracts/', 'documentation': 'docs/'}`, and required artifacts `['spec.md', 'plan.md', 'tasks.md']` — none of which belong to the actual configured mission type.

## The sharpest form: two subsystems disagree about whether the mission type exists

In the same consumer project, the expected-artifacts resolver *does* consult the org tier: `required_artifacts_for("review", "<custom-type>")` correctly returns that type's own declared blocking artifacts. So the manifest/expected-artifacts layer resolves the org-tier type correctly; the `Mission`-loading layer (`specify_cli/mission.py`) does not. The two subsystems disagree about whether a mission type this basic exists.

## Knock-on effects

[...]

Issue #3208, opened 2026-08-05
Mission artifacts should not travel in the code PR diff: route them to kitty-learning-hub


LynnColeArt:
## The problem, measured

Mission artifacts ride in the code PR diff, and on large missions they dominate it:

| PR | Total diff | `kitty-specs` artifacts | Actual source change |
|---|---|---|---|
| #3135 | 20,400 lines | 11,900 | under 1,600 |
| #3126 | 17,500 lines | 11,000 | under 1,600 |

Two consequences, and the second is the one that compounds.

A 1,600-line source change becomes unreviewable, because finding it means scrolling past 11,000 lines of generated governance artifacts. Reviewers skim, or they approve on trust.

And it defeats the PR-size doctrine we adopted informally on 2026-08-04: keep commits small, carve an MVP out of a large mission first, with rollback granularity as the argument. Once artifacts are in the diff you cannot tell a three-point change from a thirty-point one by looking at it. The measurement everyone uses to police size stops working, and from the outside the team looks like it ignores its own rule.

## What this is not

Not an argument for producing fewer artifacts, or for deleting them. Specs, Plans, Work Packages, reviews and Decision Moments as durable Git-stored records are the point of Spec Kitty, not overhead. This is a routing problem, not a volume problem. The artifacts are valuable and they are in the wrong channel.

## Proposal

Mission artifacts stop travelling in the code PR diff, and are published instead to `spec-kitty/kitty-learning-hub`.

That repo already exists and was built for this. Its own description:

> A shared hub for cross-project tracing, learning, and analysis artifacts produced by kitty-enabled repositories. This repo is not for source code; it is for operational knowledge capture.

Created 2026-02-22, last pushed 2026-03-01, and transferred into the `spec-kitty` org on 2026-08-05. Its existing layout already separates `analysis/` from per-agent `logs/<agent>/`, which is close to the shape this needs.

## The design decision this needs, and it is core's call

Two genuinely different options, not a formality:

**(a) Artifacts stay in the source repo, out of the reviewed diff.** For example committed to the coordination branch rather than the work-package branch. Cheapest, keeps locality, `spec-kitty next` needs no change. The weakness is enforcement: "present in the repo but absent from the diff" is awkward to guarantee in git, and anything awkward drifts.

[...]

Issue #2660, opened 2026-07-15
specify_cli/missions retirement — Remove software-dev template-selection fallback in mission.py (#2652-F)


stijn-dejongh:
Sub-issue of #2652 (specify_cli/missions retirement epic).

**Why / for whom / effect.** With template resolution unified through the artefact (#2652-T) and availability driven by activation (#2652-E), the `meta.json`-less `software-dev` template-file-selection fallback is dead weight that re-introduces a hidden default. Remove it so no reader can inherit software-dev behaviour by default. Coherence ADR: `docs/adr/3.x/2026-07-15-1-doctrine-offers-charter-activates-runtime-consumes.md` (PR #2655).

**Scope.** Remove the software-dev template-selection fallbacks in `src/specify_cli/mission.py`: the active-mission-link default (`:473,476`) and the `get_mission_for_feature` coalesce/except fallback (`:780`, `:798-803`). Scoped to template-file selection only (governance leak already closed by #883 WP-CANON).

**Acceptance criteria.** Inherits the epic AC. A typeless / unknown mission type no longer silently resolves software-dev templates; the behaviour is a defined neutral/error path, not a substituted default. Tests cover the removed fallbacks.

**Blocked by:** #2652-T (templates-as-config) and #2652-E (activation-driven enumeration).
Relates to #2652, #461, #883.

robertDouglass:
Triage validation (2026-08-02): KEEP OPEN. Current issue history and source review show no completed successor or supersession decision for this slice. Retain it as actionable; add a focused repro/acceptance fixture and explicit dependency/owner before implementation, then link the delivery PR.

PR #3836, opened 2026-09-02
fix(mission-types): custom mission types are second-class citizens (#3830, #3832, partial #3831)

### LINKED ISSUE #3830: Custom mission type cannot advance past its first step: profile_hint is blanked before PromptStep.agent_profile can be consulted
## Summary

`spec-kitty next --result success` cannot advance a mission of a custom (non-built-in) mission type at **any** step. The mission is undrivable through the governed loop end to end.

Verified against source commit `eff63532607d2f292efc16e3f8bfd0e61e0544aa` (the version string is not a reliable build identifier — several distinct builds share the same `3.2.6rcN` string — so please match on this commit, not on the reported version).

## Mechanism

Two pieces of the codebase directly contradict each other.

`runtime/next/runtime_bridge_composition.py::_composition_dispatch_inputs` blanks both return values whenever the action is in the mission type's own `action_sequence` — which is always true for a canonical step:

```python
action_sequence = resolve_mission_type_context(
    repo_root, mission_type=mission
).action_sequence
if action in action_sequence:
    return None, None
```

That `(None, None)` return is `(profile_hint, contract)` — both required by the executor immediately downstream.

Meanwhile `specify_cli/mission_step_contracts/executor.py` carries this comment directly above `_ACTION_PROFILE_DEFAULTS` (the built-in-only profile lookup table):

```python
# FR-008 / Phase 6 #505: this table is for built-in missions ONLY.
# Custom missions MUST resolve profile_hint via PromptStep.agent_profile;
# expanding this table for arbitrary custom missions is forbidden.
```

Because `_composition_dispatch_inputs` blanks `profile_hint` before it ever reaches the executor, `PromptStep.agent_profile` is never consulted — the mandated resolution path is unreachable. `_resolve_profile_hint` then falls through to the (deliberately) built-in-only table and raises:

```python
raise StepContractExecutionError(
    "profile_hint is required when no action default exists for "
    f"{contract.mission}/{contract.action}"
)
```

Observed in practice as a byte-identical block — same error string — at all seven canonical steps of a real custom mission type driven end to end.

## Important refinement

The mission **state machine** itself is not broken. `next` in **query** mode (no `--result`) reports true state correctly — on a fixture with all work packages complete it returns `kind: query`, `mission_state: done`, `reason: All work packages are done`. Only the **advancing** path (`--result success`) fails. Progress tracking, the event reducer, and work-package state all work correctly for a custom mission type, so the fix surface is confined to composition dispatch input resolution.

## Suggested fix

Make `_composition_dispatch_inputs` preserve `profile_hint` (or have `_resolve_profile_hint` consult `PromptStep.agent_profile` directly) so the resolution path the executor's own comment mandates is actually reachable for a custom mission type.

### LINKED ISSUE #3832: setup-plan applies a software-dev-shaped substantive gate to every mission type, blocking custom-mission plan completion
## Summary

`setup-plan` applies a software-dev-shaped "substantive content" gate to the plan artifact of **every** mission type, not just software-dev. An author of a non-software-dev mission cannot complete the plan phase by faithfully filling in their own mission type's plan template.

Verified end to end against source commit `eff63532607d2f292efc16e3f8bfd0e61e0544aa`: a fully and faithfully populated custom-mission-type plan is rejected with `result: blocked`, `phase_complete: false`, and `plan.md` is never committed. The only way through is appending a `## Technical Context` section with a `**Language/Version**` field that is meaningless for that mission type.

## Mechanism

`specify_cli/missions/_substantive.py::is_substantive(file_path, kind)` takes **no mission type** and dispatches on `kind` alone:

```python
def is_substantive(file_path: Path, kind: Kind) -> bool:
    ...
    if kind == "spec":
        return _has_substantive_fr_row(body)
    if kind == "plan":
        return _has_substantive_technical_context(body)
```

`_has_substantive_technical_context` hard-requires a `## Technical Context` section containing `**Language/Version**` plus at least one peer field — a shape specific to software's tech-stack planning, meaningless for a non-software mission type's plan artifact.

## The codebase already knows this check is software-dev-specific

`specify_cli/cli/commands/agent/mission_check_prerequisites.py` guards the very same predicate by mission type:

```python
"spec_committed_and_substantive": (spec_is_committed and (mission_type != "software-dev" or is_substantive(spec_file, "spec"))),
```

The `setup-plan` call sites in `specify_cli/cli/commands/agent/mission_setup_plan.py` have no equivalent guard — `is_substantive(...)` is called unconditionally for both the spec check and the plan check that gates commit and phase completion. A package-wide search confirms no other call site applies the guard either. This is an internal inconsistency rather than a missing feature: one prerequisite-check path already knows to skip the software-dev-only predicate for other mission types; the plan-setup path that actually blocks the phase does not.

## Suggested fix

Apply the existing `mission_type != "software-dev"` guard to the `setup-plan` call sites, or better, make the plan-substantive predicate mission-type-aware so a custom mission type can declare what "substantive" means for its own plan artifact instead of inheriting software-dev's Technical Context shape unconditionally.

MOES-Media:
Custom (non-built-in) mission types were second-class citizens across three independent subsystems. This fixes two of them outright and the operator-visible half of the third.

**Closes #3830. Closes #3832. Refs #3831** (partial — see *Scope* below).

## What was broken

| Issue | Defect |
|---|---|
| #3830 | `_composition_dispatch_inputs` returned `(None, None)` whenever the dispatched action was in the mission type's own `action_sequence` — always true for canonical steps — blanking `profile_hint`. `executor.py`'s own FR-008 comment mandates that custom missions resolve via `PromptStep.agent_profile` and forbids expanding `_ACTION_PROFILE_DEFAULTS`, so the blanking made the mandated path unreachable and dispatch failed with `profile_hint is required`. Only the *advancing* path was affected; query mode reported state correctly. |
| #3832 | `is_substantive(..., "plan")` hardcoded two literals — the `## Technical Context` heading **and** the `**Language/Version**` field label. Only `software-dev`'s plan template scaffolds that field, so **three of four built-in types were affected too**, not just custom ones. `documentation` has `## Technical Context` with different peer fields; `research` and `plan` have neither. |
| #3831 | `get_mission_for_feature` swallowed `MissionNotFoundError` into a silent software-dev substitution behind a `warnings.warn` that never reached CLI output. |

## What this PR does

**#3830** — removes the misplaced early return so `PromptStep.agent_profile` is reachable, and replaces a bare `except Exception: pass` with `logger.exception` so genuine resolution failures stay diagnosable. `_ACTION_PROFILE_DEFAULTS` remains built-ins-only; built-in types keep their existing fast path.

**#3832** — replaces the hardcoded shape with per-type field declarations. Built-in types are declared in core; **an org pack ships its own** `missions/<type>/templates/plan-field-declaration.yaml` beside its plan template, discovered through the same six-tier resolution chain that resolves the template itself. Combination rule: the primary field must be substantive **plus** at least one peer — mirroring the existing `Language/Version`-plus-a-peer semantics rather than inventing a second rule.

**#3831 (partial)** — the fallback signal is now visible on the CLI. The `get_mission_for_feature` behaviour itself is unchanged.

## Scope: why #3831 is `Refs`, not `Closes`

A plan-phase investigation against a real org-pack fixture found the legacy `Mission`/`mission.yaml` schema and the modern org-tier `MissionTypeProfile` **incompatible without new schema work**: `workflow.phases` descriptions have no org-tier source, `artifacts.required`/`.optional` collapses lossily onto `required_by_step`, and `domain`/`version`/`paths`/`validation` have no org-tier representation at all. A third resolver (`mission-runtime.yaml`) would also need reconciling.

So the org-tier lookup in `_mission_path_by_name` is **deliberately not implemented here** and is tracked as a follow-up. `src/specify_cli/mission.py` is zero-diff in this PR.

Related: #2660 targets the same `get_mission_for_feature` block but is scoped to template-file selection and states the governance leak was closed by #883. That leak reproduces on current `main` — worth reconciling when #2660 is picked up.

## Known limitation

FR-005's signal is surfaced in `mission-type current`. `acceptance/__init__.py` and `core/worktree.py` also call `get_mission_for_feature` and remain silent, so an operator running `accept` still sees no fallback signal. `accept.py` already has a console and a wired `summary.warnings` surface, so extending it later reuses existing machinery. Called out here rather than left for a reader to discover.

## Verification

Adversarial review at every phase, plus a pre-merge squad over the integrated diff. The squad found three blockers that eleven prior reviews missed; all are fixed in-branch:

- **sev 5** — the #3830 fix also removed blanket protection for built-in `software-dev`, breaking two pre-existing guards (one citing #797) in a third test tree no work-package shard covered. Fixed by preserving the `_ACTION_PROFILE_DEFAULTS` fast path while keeping custom types reaching `PromptStep.agent_profile`, and by making the dispatcher prefer `RuntimeContractRegistry`'s contract over local re-synthesis.
- **sev 4** — the first cut of the field-declaration registry was a closed 5-key table in core, so any unregistered custom type failed permanently. That reproduced this PR's own defect class with a larger allowlist. Fixed by the pack-provided seam above.
- **sev 3** — a hardcoded `qa` entry in core carried invented field names that would not have matched a real `qa` pack's template. Removed; core no longer declares field shapes for packs it has not seen.

**Non-vacuity is proven, not asserted.** For every covered type — the four built-ins and a pack-declared custom type — a faithfully populated plan passes and an unfilled scaffold fails. A reviewer independently confirmed the matrix detects a *weakened* implementation: neutralising the new placeholder patterns flips three scaffolds from correctly-failing to vacuously-passing. Negative fixtures read the real shipped templates, so template drift surfaces as a failing test.

Shards, against a baseline measured on this branch at zero source changes:

```
tests/next/                                                521 passed   (=)
tests/runtime/                                             694 passed, 1 skipped   (+5 new)
tests/specify_cli/next/test_runtime_bridge_composition.py   45 passed   (was 43 passed + 2 FAILED)
tests/specify_cli/missions/                                205 passed   (+3 new)
tests/specify_cli/cli/commands/agent/                     2012 passed, 2 skipped, 2 xfailed   (+3 new)
ruff check src/                                            clean
mypy (touched files)                                       clean
```

#3284's known-red set does not appear in any of these shards, so every red here was attributable.

## Reviewer notes

- Seven commit subjects use `reviews:`/`tasks:` types that are not in `commitlint.config.cjs`'s `type-enum`. The step is `continue-on-error`, so it will comment rather than fail. Not rewritten because doing so would invalidate commit hashes cited across the mission's own evidence trail — happy to squash-fix on request.
- The mission's design artifacts travel in this diff under `kitty-specs/`, per the repo's current convention (#3208 proposes relocating them).

---

https://claude.ai/code/session_01HosGvZE838uDZF4czy7rhM


MOES-Media:
## Landing evidence

Rebased onto current `main` (`7b6c9f4fa`) — 62 commits replayed cleanly, one dropped as already-upstream, no conflicts. Force-pushed with `--force-with-lease`.

**CI: all green.** 42 success, 36 path-filtered skips, 0 failures across 9 workflows — `CI Quality`, `ci-windows`, `ui-e2e`, `Orchestrator Boundary Guardrails`, `Plugin Validate`, `canonical-producer-lint`, `Release Readiness Check`, `docs-freshness`, `drift-detector`.

### The one red, and how it was classified

`cutover-guard` failed closed on the first push:

```
custom-mission-type-second-class-citizens-01M1FQXD:
  status_phase not flipped despite event-log runtime evidence
```

Classified as **introduced by this branch** — the mission's own state, not a pre-existing or environmental red. Fixed by running the remedy the guard itself prescribes (`spec-kitty migrate backfill-runtime-state`), which stamps `status_phase` into `meta.json`. Guard now passes.

Worth recording: this same root cause had been repeatedly written off during implementation as a pre-existing artifact ("`test_dogfood_corpus_backfilled` fails because the in-flight mission isn't cut over"). That classification was wrong — it was an actionable signal about this branch. The test now passes.

No other red appeared, so no remediation folds were needed.

### Deliberate scope widening, named as such

The **pack-provided declaration seam** was not in the original spec. It was added mid-implementation after the pre-merge squad demonstrated that the first cut — a closed 5-key table in core — reproduced this PR's own defect class with a larger allowlist: any custom mission type outside those five names would fail the plan gate permanently, regardless of content.

That is a material widening beyond the spec text and is called out here rather than left for a reviewer to infer.

### Pre-existing conditions, not addressed here

- Two `no-any-return` mypy errors in `mission_check_prerequisites.py`, in functions this PR never touches — identical at the merge base.
- A machine-local path in `tests/cli/commands/fixtures/render_width_3115/capture_width80.txt`, present on `main` and untouched by this diff. Deliberately not scrubbed here; it needs its own change so the fix is reviewable on its own terms.

### Verification depth

Adversarial review at every design phase, a separate reviewer per work package, and an R1–R3 pre-merge squad over the integrated diff. The squad found three blockers that the eleven prior reviews had missed — including a regression against a pre-existing guard in a third test tree that no work-package shard covered. All three are fixed in-branch and re-verified.

Non-vacuity of the new gate is demonstrated, not asserted: for every covered type a populated plan passes and an unfilled scaffold fails, and a reviewer independently confirmed the fixture matrix detects a *weakened* implementation by neutralising the new placeholder patterns and watching three scaffolds flip to vacuously passing.

Ready for maintainer review.

stijn-dejongh:
## Maintainer landing pass — reviewed, rebased, history cleaned (`b6edf5f125`)

Thanks @MOES-Media — thorough PR, and the in-branch adversarial trail made this straightforward to land.

### Review: SHIP (independent second-opinion squad)
- **Correctness** — the three fixes verified against the code; fail-loud preserved; the pack-provided plan-field seam resolves through the real six-tier chain with the mission-type correctly threaded; all four built-ins still gate.
- **#3828 interaction** — this PR was 18 commits behind main and its CI never saw the just-landed #3828 (expected-artifacts loader relocation into `charter/activation/`). Rebased locally onto current main and re-ran the affected trees: **306 + 257 + 182 + 45 tests green** — #3836 only *calls* the relocated loader with unchanged signatures; no interaction break.
- **① #3410 scope** — #3836 **shrinks** the guard-table family's #3397 surface (removes the substantive-gate + profile-hint dimensions of custom-type second-classness, makes the software-dev roster fallback loud) but does **not** touch the guard-table core; #3386 and #3407 are unchanged (#3407 already closed on main). Net: the guard-table mission's scope is reduced, not pre-empted.

### History cleaned: 61 → 11 commits (authorship preserved)
The branch was ~half status/lifecycle churn and had a per-commit-hygiene issue. Rewritten (all commits still authored @MOES-Media):
- **PII removed from history** — a machine-local path (`/home/…`) was scrubbed only at the tip, so *every* commit since its introduction carried it. Folded the scrub into the introducing commit → **0 machine-local-path hits across the entire rewritten history**.
- **Per-commit greenness** — a RED-FIRST test + a precedence fix landed *before* the impl that made them pass (2 red intermediate commits under rebase-merge). Folded each fixup into its origin WP; verified the previously-red commits now pass.
- **Dedup** — ~30 status-transition / issue-matrix / lifecycle-stamp commits collapsed into one consolidated status commit; spec/tasks dossier review-trails squashed.
- **Byte-identical tree** to the prior head (pure history change, zero content drift); rebased onto current main.

Final linear history: `meta → spec → plan → tasks → WP01 fix → WP01 test → docs → WP02 → WP03 fix → WP03 test → status`.

### Two follow-ups for the maintainer (not blockers)
- **Pack-authoring docs**: a custom type shipping a plan template but no `plan-field-declaration.yaml` now fails closed (deliberate) — pack authors need a one-line doc that the declaration file is mandatory alongside the plan template.
- **① convergence**: `plan-field-declaration.yaml` is a new per-type authority beside `expected-artifacts.yaml`; the guard-table mission should plan to *converge* them, not stand up a third roster.

CI is re-running on the cleaned head; I'll confirm green. **The operator merges — I don't.**

stijn-dejongh:
## Second-opinion squad + folds — CI green (`38e964592a`, 50 checks pass)

A fresh second-opinion squad (correctness + design/robustness lenses) reviewed the cleaned+rebased diff. Verdict **SHIP-WITH-FOLDS**; all findings folded in-branch (6 commits, authored by the maintainer over @MOES-Media's 11):

- **[MAJOR]** `fix(runtime)` — the refactor's `logger.exception` fired a full ERROR traceback on the **flagship** path: `resolve_mission_type_context` raises `UnknownMissionTypeError` for any non-charter-activated custom type (exactly #3830's scenario), so every `next` on a custom mission logged like a crash despite resolving fine via the frozen template. Split the except (expected→debug, unexpected→exception) + fixed the pinned test.
- **[MINOR]** `fix(missions)` — ignore a **mission-agnostic** `plan-field-declaration.yaml` resolution (tiers 1b/2/5 are global) so one file can't silently gate every undeclared custom type.
- **[MINOR]** `fix(missions)` — **fail loud** on unreadable/mis-encoded or unknown-key declarations (parity with `expected-artifacts.yaml`'s `extra="forbid"`).
- **[MINOR]** `fix(mission-type)` — re-emit non-fallback warnings instead of dropping them under `catch_warnings(record=True)`.
- **[campsite]** `refactor(missions)` — removed the now-dead `_has_substantive_technical_context`.
- **[docs]** `docs(missions)` — CHANGELOG entry + a pack-author schema section for `plan-field-declaration.yaml` + a roster parity test + org-precedence clarifying comment.

Each fold carries a focused test; cumulative 260 passed locally; ruff + mypy --strict clean; full CI matrix green.

**Not folded (noted, not tickets):** converging `plan-field-declaration.yaml` with `expected-artifacts.yaml` into one per-type authority — genuine future-mission architecture, out of scope for a 3.2.6 landing; flagged for the guard-table (#3410) mission.

Merging now (rebase, linear history).

PR #3800, opened 2026-08-29
feat(charter): retire-doctrine-term M2 — relocate doctrine package to charter/offering (#3664)

stijn-dejongh:
## Mission `charter-code-topology-01M152G1` — Wave **M2** of `retire-doctrine-term` (#3664)

Retires the internal `doctrine` **code package**, relocating the 265-file offer catalogue `src/doctrine/**` → **`src/charter/offering/**`**, and ships deprecation-compat shims for every operator-facing `doctrine` surface. This is the code-topology continuation of the merged M1 authority flip.

> **Draft for maintainer review — the operator merges.** 6 logical commits, clean linear history on `upstream/main` `321852650d` (rebased after the concurrent landings; no conflicts — nothing else touched `doctrine`/`charter`).

### Scope (operator-decided mid-mission)
The full physical `offering ↔ activation` two-module split turned out to be a **439-file** closure that re-touches M1's freshly-landed CR-01 code. Per operator call, **M2 ships the token/package retirement** (the coherent, valuable part) and the **physical activation split is carved into a follow-up mission M2b**. The C-004 boundary gate is enforced against the *current* top-level activation modules until M2b relocates them.

### What changed
- **265-file relocation** `src/doctrine/**` → `src/charter/offering/**`; every importer closed (statement rewrites **and** the string/derived refs the AST rewrite can't see — `mock.patch`/`sys.modules`/`importlib` targets, split-literal path builders).
- **CR-06 shim** `src/doctrine.py`: a module-level `__getattr__` re-export of `charter.offering` with a one-time `DeprecationWarning`, force-included in the wheel so `import doctrine` keeps working for installed consumers (submodule imports must move to `charter.offering.X`).
- **CR-02…07 operator-surface shims** (each read-both / canonical-wins / warn-once, mirroring M1's CR-01): `spec-kitty doctrine` CLI group (hidden deprecated alias of `spec-kitty charter`); tracker mode `doctrine` → **`ownership`**; `doctrine.org.packs` → `charter_packs.org.packs`; `doctrine:<kind>:<id>` URN → `charter:` (durable event logs parse both); `.kittify/doctrine/` → `.kittify/charter-packs/` dual-root reader.
[...]

PR #3842, opened 2026-09-02
feat(charter): add `charter list --json` and close the non-mapping config.yaml crash class

### LINKED ISSUE #3839: charter list: add --json
`charter list` (`src/charter/list_cmd.py`, 238 lines) has no `--json` flag — zero occurrences of json in the module. Every adjacent surface an external consumer needs (`charter status --json`, `charter context --json`) has one; this is the one gap that forces scraping human-format output when enumerating charters programmatically.

Small ask: a `--json` flag emitting the same rows as the human table, following the established envelope conventions.

(While in there: `charter status --json`'s error path currently emits a raw `{"error": "'str' object has no attribute 'get'"}` shape rather than the structured diagnostic envelope — adjacent, possibly same fix session.)

MOES-Media:
## What changes

`spec-kitty charter list` gains a `--json` flag, so external consumers can enumerate charters programmatically instead of scraping the human table. It was the last adjacent surface without one (`charter status --json` and `charter context --json` both had it).

Second, a corrupt `.kittify/config.yaml` no longer surfaces a raw Python `AttributeError` to operators. Four commands that read that file — `charter status`, `charter synthesize`, `charter resynthesize` and `doctor tool-surfaces` — now fail closed with a structured diagnostic naming the actual problem ("expected a YAML mapping at the top level, got str") and the exit code their command family documents.

## Why the second change is larger than the issue asked

The issue reported one symptom: `charter status --json` emitting `{"error": "'str' object has no attribute 'get'"}` (defect-ledger SK-16). Review found the same unguarded `yaml.load(...) or {}` → `.get()` pattern in two more modules reading the same file, and a third command crashing with an uncaught traceback instead of its JSON envelope. Fixing only the reported instance would have left the class open, so all four are closed here.

One sibling defect was deliberately **not** folded in: `MigrationRunner.upgrade()` has no `try`/`except` around `migration.apply()`, so a raising migration skips the version-stamp finalize. It is real and verified, but pre-existing, unreachable from this change, and fixing it properly is migration-atomicity work (epic #3347) rather than something that belongs in this diff.

## Fail closed, not fail quiet

The first attempt at the SK-16 fix made the loader return empty on a malformed config, which turned a loud crash into `charter status` exiting 0 with a full success envelope on a corrupt file. Review caught it. The shipped fix raises a typed error that each caller catches explicitly — the ledger entry for SK-16 is specific that the original behaviour "fails closed... this is *not* a silent success", and that property is preserved.

## Testing

Red-first throughout: the ATDD test commit precedes the implementation commits, and every production change was verified to make its test fail when reverted — re-derived independently by a reviewer rather than taken on the implementer's word (`3 failed` with the SK-16 guard removed, `10 failed` with the `--json` branch removed, `13 failed / 18 passed` at the RED commit).

Gates run locally: `ruff check src/` clean, `mypy --strict` clean on every touched module, the C-007 symbol-level dead-code gate and the legacy-terminology gate green, and all three JSON surfaces verified to emit parseable JSON. `docs/api/cli-commands.md` was regenerated with its canonical generator, not hand-patched.

## Review trail

Two independent reviewers (contract and tests lenses), an anchored verification and a fresh sweep, committed under `kitty-ops/01M1HEDT43KPP7VE18GSM3SYX3.*`. Six findings raw, all six verified resolved; the fresh sweep found two more, one folded in and one recorded in the defect ledger.

Op: `01M1HEDT43KPP7VE18GSM3SYX3`.

Closes #3839

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3842"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790963347&installation_model_id=427282&pr_number=3842&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3842&signature=56c1afdfa12d123977a272a46d756cf4aecb7a855055ca4e3179f5f05013a249"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docs/api/cli-commands.md
+++ b/docs/api/cli-commands.md
@@ -540,6 +540,14 @@ _List activated doctrine artifacts by kind._
 │                             the built-in, org, and project layers (annotated │
 │                             by source layer), including the template kind.   │
 │                             Supersedes --show-available.                     │
+│ --json                      Output JSON. Every kind row always carries an    │
+│                             'available' key and the payload always carries a │
+│                             top-level 'templates' key — both are null unless │
+│                             requested (available: null without               │
+│                             --show-available/--all; templates: null without  │
+│                             --all) rather than absent, so callers can rely   │
+│                             on key presence and branch on the value instead  │
+│                             of on which flags were passed.                   │
 │ --help            -h        Show this message and exit.                      │
 ╰──────────────────────────────────────────────────────────────────────────────╯
 ```
@@ -2585,14 +2593,34 @@ _Search tracker issues via the hosted read path_
 │                                                            squash.           │
 │ --delete-branch        --keep-branch                       Delete lane       │
 │                                                            branches after    │
-│                                                            merge             │
-│                                                            [default:         │
-│                                                            delete-branch]    │
+│                                                            merge. Unset (the │
+│                                                            default) defers   │
+│                                                            to the mission's  │
+│                                                            meta.json         │
+│                                                            retention policy  │
+│                                                            (#3131), falling  │
+│                                                            back to delete    │
+│                                                            when no policy is │
+│                                                            recorded. Passing │
+│                                                            either flag       │
+│                                                            explicitly always │
+│                                                            wins over the     │
+│                                                            mission's policy. │
 │ --remove-worktree      --keep-worktree                     Remove lane       │
 │                                                            worktrees after   │
-│                                                            merge             │
-│                                                            [default:         │
-│                                                            remove-worktree]  │
+│                                                            merge. Unset (the │
+│                                                            default) defers   │
+│                                                            to the mission's  │
+│                                                            meta.json         │
+│                                                            retention policy  │
+│                                                            (#3131), falling  │
+│                                                            back to remove    │
+│                                                            when no policy is │
+│                                                            recorded. Passing │
+│                                                            either flag       │
+│                                                            explicitly always │
+│                                                            wins over the     │
+│                                                            mission's policy. │
 │ --push                                                     Publish to origin │
 │                                                            after the local   │
 │                                                            merge (the        │
@@ -4385,9 +4413,15 @@ _Emit the open-Ops reminder for the Claude Code Stop hook._
 
  Commit spec artifacts to the mission's resolved placement.
 
- On a protected primary the coordination worktree is materialised on demand
- so the commit lands on the coordination branch (materialize-then-retry).
- On an unprotected or flattened primary the commit is direct.
+ SPEC is a primary/planning artifact: it lands on the mission's primary target
+ branch for every topology. On an unprotected or flattened primary the commit
+ is direct. On a PROTECTED primary the commit is refused (there is no fallback
+ surface); recover by either creating/checking out a non-protected feature
+ branch ('spec-kitty agent mission create --start-branch <feature-branch>') or
+ setting SPEC_KITTY_ALLOW_PROTECTED_BRANCH_COMMITS=1 to commit on the current
+ branch.
+
+ Pass individual FILES, not directories.
 
 ╭─ Arguments ──────────────────────────────────────────────────────────────────╮
 │ *    files      FILES...  Spec artifacts to commit (absolute or relative     │
@@ -4448,16 +4482,6 @@ _Emit the open-Ops reminder for the Claude Code Stop hook._
 ## spec-kitty sync
 
 _Synchronization commands_
-
-> **Inactive by default (opt-in).** As of the `sync-deactivate-by-default`
-> mission the legacy local-sync surface is **inactive on a bare install** — no
-> sync daemon is spawned and no events are emitted. The `sync` command group
-> stays registered, but when sync is not armed the subcommands report inactive
-> / no-op rather than doing work. Opt in by setting
-> `SPEC_KITTY_ENABLE_SAAS_SYNC=1` (and give per-project consent with
-> `spec-kitty sync opt-in`). `SPEC_KITTY_SYNC_DISABLE` /
-> `SPEC_KITTY_SYNC_MINIMAL_IMPORT` force it back off (disable wins). See
-> [Environment Variables § Sync activation precedence](environment-variables.md#sync-activation-precedence).
 
 ```
  Usage: spec-kitty sync [OPTIONS] COMMAND [ARGS]...
--- a/src/charter/activation/evidence/orchestrator.py
+++ b/src/charter/activation/evidence/orchestrator.py
@@ -12,10 +12,26 @@ from charter.activation.synthesizer.evidence import CodeSignals, CorpusSnapshot,
 from kernel.clock import now_utc_iso
 
 __all__ = [
+    "ConfigShapeError",
     "EvidenceOrchestrator",
     "load_url_list_from_config",
 ]
 
+
+class ConfigShapeError(RuntimeError):
+    """Raised when ``.kittify/config.yaml``'s top-level YAML content is not a mapping.
+
+    A YAML document need not be a mapping — ``just-a-string``, ``- a\\n- b``, and
+    ``42`` all parse without error but are not a ``dict``. Before this guard, a
+    corrupted or hand-edited config.yaml in this shape made ``config.get(...)``
+    raise a bare ``AttributeError`` that propagated, unhandled, all the way through
+    ``charter status --json`` (and ``synthesize``/``resynthesize``), leaking raw
+    Python exception text through the structured ``--json`` error envelope
+    (ledger SK-16). Raising this typed, message-carrying exception here keeps the
+    fail-closed outcome the ledger's SK-16 entry says is correct (exit 1,
+    ``success: false``) while replacing the leaked exception text with a
+    controlled diagnostic that names the file and the real problem.
+    """
 
 
 @dataclass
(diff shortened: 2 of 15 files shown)


stijn-dejongh:
Re: the `lint-feedback` advisory comment above — those findings are **repo-wide, not from this PR's diff** (the 4 ruff nits are in `test_acceptance_support.py` / `test_preflight_is_not_delivery.py`, and the `mypy: cannot read file 'src/doctrine'` is a stale target left by the M2/#3800 relocation — mypy has silently not been running in lint-feedback since). Remediated in a focused separate PR **#3854** to keep this one on-topic; nothing to change here.

stijn-dejongh:
## Maintainer review — strong PR, please revise before it lands (@MOES-Media)

Thanks — the SK-16 four-reader fix is **correct and complete**: I verified live that `charter status` / `synthesize` / `resynthesize` / `doctor tool-surfaces` all fail *closed* (typed diagnostic + their documented exit code) across the full non-mapping space (str/list/int/float/bool; None→{}), and `charter list --json` emits parseable JSON on the happy path. A second-opinion squad (correctness + history/hygiene lenses) surfaced four things to address first — one is time-sensitive.

### ① Time-sensitive — machine-local path / private-repo reference in the Op evidence (please scrub)
Commit `9eb7306736` (`kitty-ops/01M1HEDT43KPP7VE18GSM3SYX3.review-contract.yaml`) contains:
```
- code: "/home/jeroennouws/dev/SK-missions/SPEC-KITTY-LEDGER.md:3879-3898 (SK-16 entry)"
```
That's a real home-dir path + a reference to a private repo, now reachable in a public-repo commit object. The four `kitty-ops/` + `.kittify/evidence/` commits (`740b8a429d`, `9eb7306736`, `7ccbb20cd6`, `3ed3223920`) are gitignored Op-dossier churn with **zero** code/test/docs — simplest fix is to **drop all four** from the branch (which also removes the path from history). If you'd rather keep an evidence trail, scrub the line first — but it needs a history rewrite either way since it's already pushed.

### ② MAJOR — `charter list --json` reintroduces the crash class on a corrupt config
The new command's own headline path isn't guarded. `charter list` calls `ProjectContext.from_repo` (`list_cmd.py:193`) with no try/except; on a non-mapping `config.yaml` the loader raises **uncaught** → `charter list --json` exits 1 with **empty stdout + a raw traceback** — exactly the fail-quiet-becomes-traceback outcome this PR exists to close. (`--show-available` and `--all` all hit it, not just the guarded `list_available_detailed` path.)

This is a **rebase interaction**: #3800/#3828 relocated the config loader into `charter/activation/` (`pack_context._load_config` → `CharterPackConfigError`; `pack_manager._load_config` → `ValueError` via `list_activated`) — a *different* raise site than the `ConfigShapeError` guard you added in `evidence/orchestrator.py`. Your branch is **36 commits behind main** and its CI never saw that relocation, so please **rebase onto current main** first; the guard belongs around the whole `list_cmd` body:
```python
try:
    ctx = ProjectContext.from_repo(repo_root)
    ...  # existing body through the final console.print(table)
except (ConfigShapeError, ValueError) as exc:
    _emit_error(console, json_output=json_output, message=str(exc)); raise typer.Exit(1) from exc
except KittyInternalConsistencyError as exc:  # CharterPackConfigError
    detail = f"{exc.code}: {exc.body}" if exc.body else exc.code
    _emit_error(console, json_output=json_output, message=detail); raise typer.Exit(1) from exc
```
Surface `.body` (not `str(exc)` — that's just the code `CHARTER_PACK_CONFIG_INVALID`), matching `synthesize.py:571-576`. Add a corrupt-config regression for `charter list --json` mirroring `test_status_sk16_config_guard.py` (the current `test_charter_list_commands.py` has no non-mapping case for `list` itself — the gap that let this through).

### ③ History — a reverted defect ships across intermediate commits (this repo rebase-merges)
Commit `43d89ec9ab` guards the config with `return ()`, which makes `charter status --json` **exit 0 with a full success envelope** on a corrupt file (silent success), and commit 1's test asserts that. Commit `a5fa2474c5` later flips it to fail-closed (its own message calls the interim state "a worse defect"). Because commits land individually on main, a `git bisect` across `8bd7fc0c0c..5d58a9c491` finds the silent-success defect. Please **fold `43d89ec9ab` into `a5fa2474c5`** (so the reverted guard never lands standalone) — or squash the reworked charter seam into one commit shipping only the final contracts. Same applies to the RED ATDD commit (`9bb61cad13`): fold it forward so no commit lands red.

### ④ Deferred (not blocking) — one owned config-load boundary
The non-mapping crash class is now guarded at 5+ independent sites (the four here + `pack_manager`/`pack_context`), each re-stamping its own `isinstance(config, dict)` check. Worth a follow-up: a single "load `.kittify/config.yaml` as a validated mapping, or fail closed once" surface so consumer #6 doesn't re-stamp guard #6. Not this PR.

Handing back to you for ①–③. Ping me when it's rebased + revised and I'll re-run the landing pass. (Separately, the `lint-feedback` advisory on this PR was repo-wide, not yours — fixed in #3854.)

MOES-Media:
Thanks — all three blocking items are addressed. The branch was **rebuilt from current `main`** rather than rebased-with-folds, because ① and ③ both required history surgery and one clean rebuild is easier to audit than an interactive rebase across a 62-commit gap. Five commits now, force-pushed.

### ① Private path scrubbed — all four Op-dossier commits dropped

Verified gone across the whole branch:

```
for c in $(git rev-list origin/main..HEAD); do
  git show "$c" | grep -qE "/home/<user>|<private-repo>|<private-ledger>" && echo "LEAK in $c"
done
# → clean
git log --oneline origin/main..HEAD -- kitty-ops/ .kittify/evidence/
# → empty
```

One correction, offered because it may matter for your own tooling: **the commit SHAs in your review do not exist in this repository** — `9eb7306736`, `740b8a429d`, `7ccbb20cd6`, `3ed3223920` all return `Not a valid object name`. The *content* finding was exactly right and I verified it independently: the path was in `a4b0cf582`, and the four Op-dossier commits were `fa23b827b`, `a4b0cf582`, `c68114f6e`, `c86b2b9dc` — same count, same nature, different hashes. I found it by grepping the branch for the path rather than by SHA, which is why the mismatch did not block anything.

**Caveat worth stating plainly**: force-pushing makes the old objects unreachable, but GitHub retains unreachable objects addressable by SHA for a period. If you consider the exposure material, that needs a GC request to GitHub support — a force-push alone does not erase it.

This one was mine, and the failure mode is worth recording: I appended a public-repo clause to every subagent brief on this mission and never applied it to the Op evidence my own tooling auto-committed. Guarding the work product and not the exhaust.

### ② `charter list` guard — fixed, with the coverage gap closed

Rebased onto current `main` first (the branch was **62** commits behind, not 36 — the relocated loaders were indeed absent from its CI). Guarded the whole `list_cmd` body at the real call site, re-derived rather than taken from the pre-rebase line number.

Exceptions caught, verified against the live raise sites rather than inferred: `KittyInternalConsistencyError` (confirmed base of `CharterPackConfigError`) separately from `(ConfigShapeError, ValueError)` — because `ProjectContext.from_repo` raises the former, while `ValueError` arrives via `pack_manager._load_config` / `_activation_list_or_error` and via `list_available_detailed`'s mission-type scan. That last path matters: it fires under `--show-available` **alone**, not only under the already-guarded `--all` branch.

`.body` is surfaced as you asked, not `str(exc)`:

```json
{"error": "CHARTER_PACK_CONFIG_INVALID: .kittify/config.yaml root must be a mapping.\nRemediation: fix .kittify/config.yaml ... or run `spec-kitty upgrade`...", "result": "error", "success": false}
```

Regression coverage is **30 cases** — 3 entry paths (`plain`, `--show-available`, `--all`) × 5 non-mapping shapes (string/list/int/float/bool) × 2 modes (`--json`/rich) — plus a control pinning that an empty (`None`) config resolves to `{}` and is *not* an error. Red-first proven by swapping the pre-fix `list_cmd.py` back in: **all 30 failed** with the uncaught `CharterPackConfigError` at the guarded call site; restored, all 50 in the file pass.

The four previously-guarded surfaces still fail closed against the same fixture: `charter status`, `synthesize`, `resynthesize` (exit 1, structured JSON), `doctor tool-surfaces` (exit 2, its documented contract) — no traceback from any.

### ③ Reverted defect no longer lands standalone

You were right that this matters under rebase-merge. The interim commit guarded with `return ()`, which made `charter status --json` exit 0 with a full success envelope on a corrupt config — and its own successor's message calls that "a worse defect than the raw `AttributeError` it replaced". A `git bisect` across the range would have landed on it.

Both that commit and the RED ATDD commit are folded into the work commit they belong to. The rebuilt history is:

```
feat(charter): add `charter list --json` and close the non-mapping config.yaml crash class
fix(config): close the non-mapping-config.yaml crash class in two siblings
fix(doctor): guard tool-surfaces against non-mapping config.yaml
fix(charter): guard `charter list`'s unguarded ProjectContext.from_repo call
docs(charter): regenerate CLI reference for `charter list --json`
```

No commit in that range lands red, and none ships the silent-success guard. `commitlint` verified per-commit across the whole range, mirroring CI's own invocation.

### ④ Single owned config-load boundary — agreed, not folded

Agreed, and deliberately not folded here: it is a refactor across 5+ independent call sites with its own design question (what the one loader returns on each failure mode), which would swamp review of a bugfix PR. Recorded in our defect ledger so it does not evaporate. Happy to take it as its own issue if you want it tracked upstream.

### Housekeeping

- `docs/api/cli-commands.md` regenerated against the rebased tree.
- I left an unrelated drift in `docs/api/agent-subcommands.md` alone (main's own `--retain-branches`/`--retain-worktrees` flags are missing from its committed reference). It is a real staleness and a one-line regeneration, but you moved the lint findings to #3854 to keep this PR on-topic, so I did not import a different unrelated fix here. Say the word and I will send it separately.
- Ready for your landing pass whenever CI settles.

MOES-Media:
**CI green on the rebuilt branch — ready for your landing pass.**

`46 passed, 34 skipped, 0 failed`.

One follow-up on ①, offered because it is a class rather than an instance. After scrubbing this branch I swept every other open mission branch for the same pattern and **found it twice more** — an absolute home path in a mission `tracer-tooling-friction.md` on one branch, and in an FR's prose in `spec.md` on another. Both are now scrubbed and pushed.

The mechanism is worth knowing if you see it again from this direction: agents paste CLI output verbatim into mission artifacts, and spec-kitty's own errors print absolute paths — e.g. `Error: analysis_report_required: ... Missing: /home/<user>/.../analysis-report.md`. Quoting the error faithfully ships the path. A public-repo instruction was in every agent brief on all four missions; they applied it to their findings and not to the exhaust.

Nothing in the repo detects it today. A grep gate over tracked files under `kitty-specs/`, `kitty-ops/` and `.kittify/evidence/` for `/home/`, `/Users/` and `C:\\Users\\` would close it by construction — happy to send that as its own small PR if you want it, given you have kept this one on-topic.

Issue #3770, opened 2026-08-27
Unify all expected-artifacts.yaml loading on the cached ManifestRegistry.load_manifest (retire the 3 mirror loaders)


stijn-dejongh:
## Context

Surfaced by the adversarial squad (paula-patterns + architect-alphonso) during the landing pass for #3729, and disclosed by that PR's author as a known-deferred item. Filing as the tracked follow-up; **not foldable into #3729** (it changes swallow-vs-propagate semantics, cache behavior, and needs a repo-wide arch-gate — all outside that PR's 5-file blast radius).

## Problem

Four parallel resolvers of "org-first, built-in-fallback, `model_validate`, wrap `ValidationError` → `ManifestSchemaError`" now exist, and only one is the canonical cached authority:

- `src/specify_cli/dossier/manifest.py` — `ManifestRegistry.load_manifest`, **canonical, cached** (already consumed by `sync/namespace.py` and the dossier `Indexer`).
- `src/specify_cli/runtime/resolver.py` — `_load_expected_artifact_manifest`, **new, uncached** (its own docstring admits mirroring the canonical one).
- `src/runtime/next/runtime_bridge_io.py` — `_presence_filenames_for`, **new, uncached** — a third copy.

Six bare `ExpectedArtifactManifest.model_validate(` sites across three modules.

**Hot-path cost:** within a single `gather_artifact_presence` call the manifest is resolved up to **three times** (`_presence_filenames_for`, `_expected_artifacts_manifest_resolves`, and `required_artifacts_for`→`_load_expected_artifact_manifest`), none consulting `ManifestRegistry._cache` — and for the four **built-in** families this is entirely wasted, because `evaluate_guards_strict` returns via the guard-table entry before it ever reads `blocking_artifact_names`.

**Drift is real, not hypothetical:** the org-tier `ManifestSchemaError` `origin` string had already diverged between the three copies *within #3729* (one dropped the "checked org roots" suffix). That specific drift was folded on #3729 (`fix(landing): match org-tier ManifestSchemaError origin across both entry points`); this issue removes the class.

## Scope

- Unify all `expected-artifacts.yaml` loading on `ManifestRegistry.load_manifest`; delete `resolver._load_expected_artifact_manifest` and `_presence_filenames_for`'s private load.
[...]

PR #3828, opened 2026-08-31
fix(charter): unify expected-artifacts.yaml loading + close org-tier fail-loud gap (#3770, #3412)

stijn-dejongh:
## Unify `expected-artifacts.yaml` loading + close the org-tier fail-loud gap

Closes #3770. Closes #3412. Slice of epic #3410 (charter/doctrine silent-drop). Milestone 3.2.6.

Two issues that share **one load seam** and one intent from FR-016 — *"absence and malformation are DISTINCT outcomes"* — folded into one mission.

### The shared seam

Four parallel reimplementations of the same *org→built-in-precedence → `model_validate` → wrap `ValidationError`→`ManifestSchemaError`* logic had drifted apart (one already diverged inside #3729), plus an orphan direct-read loader. Only the cached one was canonical. Separately, a YAML-syntax-broken **org** manifest still silently degraded to `None` and — for a custom family — **two-stage laundered** into a green guard.

### What changed

**#3770 — one authority.** The cached loader (and its `ManifestSchemaError`) is **relocated into `charter/activation/manifest_loader.py`**, because `C-001` (charter must not import `specify_cli`) blocked the charter-tier consumer from reaching the old `specify_cli`-resident authority. A deprecation shim keeps `specify_cli/dossier/manifest` re-exporting `ManifestRegistry`, `load_manifest`, `ManifestSchemaError`, and `MalformedManifestError` with **object identity** (the last routed through the `charter.missions` facade to satisfy the runtime↔doctrine boundary gate). `ManifestRegistry` stays in `specify_cli` as a thin delegate. The three mirror loaders (`resolver._load_expected_artifact_manifest`, `runtime_bridge_io._presence_filenames_for`, `mission_type_profiles._resolve_expected_artifacts_slot`) now delegate to the authority; the orphan `ExpectedArtifactManifest.from_yaml_file` is deleted. Grep-proof: `ExpectedArtifactManifest.model_validate(` now appears in **exactly one module**.

**#3412 — fail loud, distinct from absent.** `org_expected_artifacts._read_yaml_mapping` now raises the charter-resident `MalformedManifestError` for a *present-but-broken* org file (YAML-syntax or non-mapping) instead of `warning → None`; `repository.get_expected_artifacts` was widened to also raise on *present-but-unreadable* (`OSError`/`UnicodeDecodeError`) — symmetric with the built-in `YAMLError` case already shipped in `1763bf2ae3`. Genuine absence still returns `None`.

[...]

PR #3848, opened 2026-09-03
perf(runtime): read expected-artifacts.yaml once per gather, not twice (#3847)

### LINKED ISSUE #3847: perf(runtime): expected-artifacts org manifest read twice per gather — dedup the tri-state existence check onto the cached loader
## Summary

On the org happy path, `gather_artifact_presence` reads a mission family's `expected-artifacts.yaml` **twice** per call:

1. `_presence_filenames_for` (`src/runtime/next/runtime_bridge_io.py:841`) loads the manifest via the **cached** authority `charter.activation.manifest_loader.load_manifest`.
2. `_expected_artifacts_manifest_resolves` (`:967`) — the single source for `ArtifactPresenceSnapshot.blocking_artifact_names`'s `None`-vs-`frozenset` distinction — independently re-reads via the **uncached** `_resolve_org_manifest_mapping` (`:925`) → `charter.activation.org_expected_artifacts.resolve_org_expected_artifacts`, plus a bare built-in `MissionTemplateRepository.get_expected_artifacts` read.

So the org file is parsed twice (and the built-in tier read a second time) for the same family in one gather.

## Where found

Consolidation-review follow-up from the manifest-loader unification mission (#3770 / #3412, now merged). It was flagged **sev-4 (note only)** there and deliberately deferred: the double-read is **pre-existing** (both paths read separately before that mission) **and** lives inside the `blocking_artifact_names` None-vs-`frozenset` tri-state (#3729), which that mission fenced off as a non-goal (C-002).

## Proposed remediation

Have `_expected_artifacts_manifest_resolves` reuse the cached loader — `load_manifest(mission_family, repo_root=repo_root) is not None` — so the existence check hits the same process-global cache `_presence_filenames_for` already populated, then **delete the now-dead `_resolve_org_manifest_mapping`** (its only caller).

## Hard constraint (why this needs care, not a one-liner)

Must preserve the `blocking_artifact_names` **None-vs-`frozenset()` tri-state exactly** (#3729 / C-002). Characterize byte-exact before/after:

- **absent** (neither tier) → `False` → `None`. (both approaches agree)
- **present + valid** → `True` → real `frozenset`. (agree)
- **present + YAML-broken / unreadable / non-mapping** → both raise `MalformedManifestError` (agree; already fail-loud post-#3412).
- **present + schema-invalid (`extra="forbid"`)** — **subtlety**: today `_expected_artifacts_manifest_resolves` returns `True` (it does not run `model_validate`) and the `ManifestSchemaError` surfaces later in `_presence_filenames_for`; `load_manifest(...) is not None` would raise it at the check instead. Net observable behavior (raises `ManifestSchemaError`, guard never greens) is the same, but this must be characterized and asserted, not assumed.

## Scope / non-goals

- `src/runtime/next/runtime_bridge_io.py` only. Do **not** change the tri-state contract itself (the `None`-vs-`frozenset` semantics), only the redundant read behind the existence check.
- Minor perf; low risk **with** exhaustive characterization. Refs #3729, follow-up of #3770/#3412.

stijn-dejongh:
## perf(runtime): read `expected-artifacts.yaml` once per gather, not twice

Closes #3847. Follow-up from the #3770/#3412 consolidation review (now merged).

### The redundancy

On the org happy path, `gather_artifact_presence` parsed a family's `expected-artifacts.yaml` **twice** per call:

- `_presence_filenames_for` loads the manifest via the **cached** authority `charter.activation.manifest_loader.load_manifest`.
- `_expected_artifacts_manifest_resolves` — the single source for `ArtifactPresenceSnapshot.blocking_artifact_names`'s `None`-vs-`frozenset` distinction — independently **re-read** via the uncached `_resolve_org_manifest_mapping` → `resolve_org_expected_artifacts`, plus a bare built-in `get_expected_artifacts` read.

### The fix

`_expected_artifacts_manifest_resolves` now reuses the cached loader:
`load_manifest(mission_family, repo_root=repo_root) is not None` — hitting the same process-global cache `_presence_filenames_for` already populated. The now-orphaned `_resolve_org_manifest_mapping` (its only caller) is deleted, along with its now-unused imports. The org file is parsed **once** per gather.

### Tri-state preserved exactly (#3729 / C-002)

This function feeds the `blocking_artifact_names` None-vs-`frozenset()` tri-state, which was a **non-goal** of the parent mission — so the change is characterization-gated to prove byte-identical semantics:

| input | tri-state result | 
|-------|------------------|
| absent (neither tier) | `None` (unchanged) |
| present + valid | real `frozenset` (unchanged) |
| present + YAML-broken / non-mapping / unreadable | raises `MalformedManifestError` (unchanged) |
| present + schema-invalid (`extra="forbid"`) | raises `ManifestSchemaError` — same net outcome; the raise simply surfaces at the resolves-check instead of downstream in `_presence_filenames_for` (the tri-state value never materialized wrong in either case) |

Four characterization tests pin each row (green-before **and** after), plus `test_org_manifest_read_exactly_once_per_gather_call` (the 2→1 read regression, `@regression` #3847).

### Notes for reviewers

- **AC-8 spy retarget**: the org-root-threading test (`test_cli_guard_family.py`) moved its `resolve_existing_org_roots` spy from the `charter.drg` re-export to the origin module `charter.offering.drg.org_pack_config`, because the authority lazily imports the helper from its origin — the asserted invariant (repo_root threaded at both `_dn_dependency_gate` call sites) is unchanged.
- **Scope**: `runtime/next/runtime_bridge_io.py` logic only (+ a comment cross-ref fix). The tri-state contract, guard-table, and every other surface are untouched.
- CI-only arch gates (`test_no_dead_symbols` — a symbol was removed —, `test_arch_shard_marker_completeness`, `test_golden_count_ban`) all green; ruff/mypy clean (the one `_internal_runtime/schema.py:30` mypy note is pre-existing, unchanged).

https://claude.ai/code/session_01NL6DT7jjh7iTJ1jiyBq5nZ

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3848"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791005016&installation_model_id=427282&pr_number=3848&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3848&signature=e91b2bb7803acc7f9416bf9b3926c8e32f2fb780c0a9125f471e9251d89627cf"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/src/charter/activation/mission_type_profiles.py
+++ b/src/charter/activation/mission_type_profiles.py
@@ -1152,8 +1152,8 @@ def _resolve_expected_artifacts_slot(
     # actual `dict[str, Any]` return type is invisible here too. The cast
     # documents the type `model_dump()` actually returns at runtime -- mypy
     # is not wrong about the code, only blind to it through this boundary
-    # (mirrors the identical rationale on `_resolve_org_manifest_mapping` in
-    # `runtime.next.runtime_bridge_io`).
+    # (mirrors the identical rationale on
+    # `charter.activation.manifest_loader._resolve_existing_org_roots`).
     return cast("_ExpectedArtifactsManifest", manifest.model_dump())
 
 
--- a/src/runtime/next/runtime_bridge_io.py
+++ b/src/runtime/next/runtime_bridge_io.py
@@ -84,7 +84,7 @@ import tempfile
 from collections.abc import Callable, Mapping
 from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, TypedDict, cast
+from typing import TYPE_CHECKING, Any, TypedDict
 
 import yaml
 from mission_runtime import CommitTarget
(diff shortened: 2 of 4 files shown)


stijn-dejongh:
## Maintainer review — SHIP (rebased onto current main)

Two-lens second-opinion squad (correctness/fail-loud + boundary/interaction): **SHIP, no folds.**
- **Behavior- and fail-loud-preserving:** both reads route through the cached `ManifestRegistry.load_manifest` single authority (identical `(mission_type, org_roots)` cache key → the second read is a guaranteed cache hit), so a present-but-malformed manifest still **raises** (`ManifestSchemaError`/`MalformedManifestError`, both tiers) and never collapses to "absent". The removed `_resolve_org_manifest_mapping` org branch (returned `True` for a schema-invalid org manifest without `model_validate`) was **unreachable dead behavior** — the new single read is strictly more consistent.
- **Single-authority:** this folds the *last* straggler onto the #3828/#3770 unified cached authority — it removes the second (uncached, bypassing) read path, doesn't reintroduce one. `resolve_org_expected_artifacts` now has exactly one src caller (`load_manifest` itself).
- **No gate update owed:** the removed function read expected-artifacts via `resolve_org_expected_artifacts`, **not** `load_meta` — so the routed-load_meta floor gate, the meta-census contract, and the dead-symbol gate are all verified unchanged (the #3851-class census-companion is correctly a no-op here).
- **Test fidelity:** `test_org_manifest_read_exactly_once_per_gather_call` spies the underlying resolver and asserts read-once (RED=2 before, GREEN=1 after); the AC-8 spy correctly re-targets the origin `org_pack_config` (not the `charter.drg` re-export) to follow the real lazy-import path. 26 changed-file tests green.

Rebased + pushed; ready to merge on green.

PR #3855, opened 2026-09-03
feat(orchestrator-api): add design-phase verbs with host-CLI lifecycle parity

MOES-Media:
## What this changes for an external host

An external host — a CI pipeline, a custom dashboard, or a native driver like Kitty Desktop — could already drive the **whole WP-implementation loop** through `orchestrator-api` (`list-ready` → `start-implementation` → `start-review` → the review triple → `merge-mission`). It could **not** drive the **design phases**. `specify`, `plan`, `tasks`/`finalize`, `analyze`, and decision resolution existed only as host-CLI commands, and the contract's own `host-boundary-rules.md` named that a boundary an external host was *forced to violate*, with no compliant alternative.

This PR closes that gap. `orchestrator-api` contract **1.3.0 → 1.4.0** adds 11 design-phase verbs so a host can drive `specify → plan → tasks → check-prerequisites/record-analysis → decision resolution` end-to-end, in the same grammar as the WP-loop verbs (JSON envelope, `--policy` provenance, structured typed error codes, fail-closed on bad input). Purely additive — every existing verb is unchanged and old hosts stay compatible (min-provider version held).

Closes #3837.

## The verbs

- **Scaffold/validate:** `specify`, `plan`, `tasks`, `check-prerequisites`, `record-analysis`.
- **Decision ledger (Mechanism A):** `open-decision`, `resolve-decision`, `defer-decision`, `cancel-decision`.
- **`spec-kitty next` decision resolution (Mechanism B):** `answer-decision` — resolves a `decision_required` moment with **full host-CLI parity**: it reaches the same three lifecycle/event-log side effects `spec-kitty next --answer …` performs, through **one extracted shared seam** (`src/runtime/next/next_invocation_lifecycle.py`), so a host driving decisions purely through the API produces an identical audit trail.
- **Query:** `design-status` — read-only, fail-closed (typed error on a torn/corrupt event log or decisions index, never a silently-wrong phase).

## Tests

~5k lines of ATDD across `tests/specify_cli/orchestrator_api/` (positive + negative fail-closed paths), the shared-seam regression test, contract-version and verb-doc-presence guards. All new test files declare `pytestmark`.

## Landing notes

**Fresh second-opinion squad (3 profile-loaded opus lenses over the rebased diff) — unanimous SHIP on the code:**
- *Correctness / fail-closed:* no fail-open; required-arg rejection, envelope/error-code parity, additive versioning, DecisionKind handling, and side-effects-after-validation ordering all verified (107 negative-path tests green).
[...]

PR #3858, opened 2026-09-03
feat(events): add `spec-kitty events tail` with truncation-safe resumable tailing

### LINKED ISSUE #3841: Event push/watch channel for external consumers (design discussion)
## Context

External consumers that want live project state currently have exactly one option: watch the filesystem themselves and re-derive state (which is what Kitty Desktop's engine does — notify-based watching, debounced re-reduction of status.events.jsonl). That works for one project on one machine. At fleet scale (many projects, many missions, remote checkouts), per-consumer filesystem watching stops being viable.

Verified against `eff63532`: no websocket/SSE/tail/watch/stream command exists anywhere in `src/specify_cli/cli/commands` — the only websocket code is device-flow auth and SaaS-dashboard internals, unrelated to a local event channel.

## Ask (discussion first, not a spec)

Some form of push channel for the event log — candidates, roughly in ascending ambition:
1. `spec-kitty events tail --mission <slug> --json` — long-lived process streaming envelope JSON lines as they land (the cheapest thing that removes the consumer-side watcher).
2. A local socket/SSE endpoint serving the same stream per project.
3. Fleet-level aggregation (probably SaaS-adjacent, out of scope here).

Option 1 alone would let external hosts drop their own inotify stacks and inherit the writer's own ordering guarantees instead of racing the filesystem (torn-read/rename-churn classes that consumers currently have to defend against independently).

Raised from Kitty Desktop's Phase-4 upstream-readiness audit — this is the one Phase-4 capability with no upstream seam at all today. No urgency; filing so the design conversation exists.

MOES-Media:
Closes #3841

Adds `spec-kitty events tail --mission <slug> --json` — a bounded, resumable reader over a mission's `status.events.jsonl`.

## The issue's premise was wrong, and that shaped the whole design

The issue assumed the event log is append-only. **It is not.** `coordination/transaction.py` truncates `status.events.jsonl` in place on rollback, and only 2 of 6 writers take the feature status lock. So a naive tailer silently desyncs.

Critically, the rollback produces a **clean record-boundary** truncation — every remaining line is still valid JSON, events are simply gone, nothing raises. A sibling mission was rejected at severity 4 this week for defending only against a mid-line JSON tear and missing this shape.

All three shapes are handled and separately tested, each proven to fail when its own defence is disabled:

| Shape | Detected by |
|---|---|
| mid-line tear (writer mid-append) | tolerated and retried, never dropped or raised |
| clean record-boundary truncation | size check |
| truncate-then-regrow within one poll | SHA-256 content invariant |

The size and hash checks run **independently every poll**, because a rollback-then-regrow that completes inside one poll interval leaves the size unchanged — the case the size check structurally cannot see.

## Architecture: bounded core, thin shell

`src/specify_cli/status/tail_reader.py` holds the core — `poll_once()`, `validate_resume_cursor()`, `tail_events()` — with **zero Typer imports** and **one file descriptor per call** (`os.fstat(fh.fileno())`, never a second `Path.stat()`, so there is no TOCTOU window between the size check and the read). The core owns the poll loop via an injectable `sleep_fn`; `src/specify_cli/cli/commands/events.py` is a plain `for`-loop consumer with no `while True` and no direct `time.sleep`.

That split exists so the core is testable with finite input. **No test in this PR can hang**: core tests inject a no-op `sleep_fn` with an `max_events`/`islice` bound, CLI tests use `--once` or a small `--max-events N`, and the concurrency proof uses `threading.Event` checkpoints with a bounded join.

## Resume is fail-closed, and distinct from live resync

Per the mission's operator ruling the content invariant is a **SHA-256 hex digest**, not raw bytes — the spec's own evidence cites 610KB event lines, which would break `ARG_MAX` as a CLI argument.

On a resume whose invariant does not match, the reader **refuses**: structured stderr, non-zero exit, **no Tail envelope emitted at all**. That is deliberately different from the live resync-from-0 path — a resume the reader has not yet trusted streams no bytes. Structurally invalid offsets (negative, past EOF, not on a line boundary) are refused with distinct reasons, never silently clamped.

## The defect the pre-merge review caught

Worth calling out, because five approved work packages and four independent reviewers all missed it:

**Pass-through envelopes never carried `tail_offset`/`tail_invariant`** — only the `log_truncated` signal did. FR-004 and the plan's envelope schema both require them on every envelope. So a real consumer of `--json` had **no way through the public stdout interface** to obtain the cursor needed to drive `--from-offset`/`--from-invariant` on restart. The headline capability was unreachable at the layer that matters.

It survived that long because **every test proving resume worked obtained its cursor internally** — via `poll_once()` and `log_path.stat()` — never by parsing real CLI stdout. A test that bypasses the public interface cannot catch a defect in the public interface.

Two independent lens groups found it (one by driving the real CLI and observing zero cursor fields on printed events). The fix injects the keys in `poll_once()`'s drain path, derived from bytes already in-process so the one-fd invariant holds. The existing resume test was **changed to parse the cursor from real stdout**, and a new round-trip test restarts from that parsed cursor. Reverting the injection now fails both — including the test that previously passed.

## Evidence

- `tests/status/ tests/specify_cli/status/ -m "fast and not (git_repo or integration or stress)"` → **1022 passed, 131 skipped**
- `tests/cli/ tests/specify_cli/cli/ -m fast` → **3158 passed, 225 skipped**
- **100% diff-coverage** on `src/specify_cli/status/*` (the 90% floor is a gate here)
- Red-first proven per work package by reverting the production branch and observing the failure, then restoring byte-identical — performed by each reviewer independently, not taken from the implementer's report
- Every new test's collecting CI job confirmed by real `--collect-only`; markers, not directories, decide collection here (#3241)
- Pre-merge squad: 3 lens groups, 1 severity-4 found and fixed, **0 severity ≥3 surviving**

Full R1–R6 trail under `kitty-specs/event-push-watch-channel-01M1K6W2/reviews/`.

## Known reds that are not this PR's

`test_create_mission_propagates_named_exception_type` fails inside **any** git worktree and passes from the repository root — environmental. `tests/status/test_sync_lane_mapping.py` fails in full-scope runs and is `14 skipped, 0 failed` in isolation — order-dependent.

## Scope

Deliberately Option 1 only: one command, one mission, `--json`. No daemon, no socket or SSE, no network surface, no fleet aggregation.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3858"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791052029&installation_model_id=427282&pr_number=3858&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3858&signature=385d7d0960d6ddd2037b584264b1c4b376c024c1cd001aa2577376785616b36c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
**CI green — merge-ready.** `50 passed, 32 skipped, 0 failed`.

## Landing pass

Six checks were red on the first run. Each was classified before anything was touched, and the disposition differed per bin — none of them was retried to green.

| Check | Bin | Disposition |
|---|---|---|
| `fast-tests-cli` | **test was wrong** | Under CI's colour terminal, Rich's option highlighter splits `--mission` into `-` + `-mission` with an ANSI reset between them, so a raw substring assertion breaks. The option itself was never broken. Fixed with the house `force_wide_help_console` precedent rather than by loosening the assertion. |
| `arch-adversarial (arch_shard_1)` | **contract this PR legitimately crosses** | The guard defends a frozen ceiling of non-escaped `len(x) == n` sites in `tests/cli`; this PR added two. Both were converted to ordered-identity content assertions that *subsume* the count — a strictly stronger check — rather than escaped. Count back to 0. |
| `docs-freshness` / `fast-tests-docs` | **stale generated artifact** | The CLI reference was already fresh; the actual defect was index drift in the anchor lists for `docs/api/cli-commands.md` and `docs/context/system-events.md`. Regenerated; unrelated drift from `main` picked up by a full regeneration was deliberately reverted to keep this PR on-topic. |
| `cutover-guard` | **known tooling defect** | The standard flow never writes `status_phase`, so this fails closed on every mission. Ran the prescribed remedy. Worth flagging: **the remedy under-reports itself** — it printed `Flipped: 0, Skipped: 1` while in fact writing `"status_phase": "1"` into `meta.json`, confirmed by diffing before and after rather than trusting its summary. NFR-002's archive gate still passes, so no gate contradiction. |
| visible-command tolerance band | **landing-order, resolved** | This branch's count (277) exceeded the old band (226–276). The constant was deliberately **not** touched here, because sibling PR #3855 re-baselined the same constant to 286 (band 257–314) and a competing edit would have collided between two of our own PRs. #3855 has since merged and the constraint is now moot. |
| `quality-gate` | aggregate | Cleared with the rest. |

## Evidence

- `tests/status/ tests/specify_cli/status/ -m "fast and not (git_repo or integration or stress)"` → **1022 passed, 131 skipped**
- `tests/cli/ tests/specify_cli/cli/ -m fast` → **3158 passed, 225 skipped**
- Both exact matches against the pre-landing baselines.
- `diff-coverage` unaffected — the landing changes touched only test, doc and data files, none of `src/specify_cli/status/*` (which sits at 100%).
- `ruff` clean on every touched file. `mypy --strict` shows the same pre-existing standalone-file errors as before, byte-identically — no new ones.
- `commitlint` verified per-commit across the range, mirroring CI's own invocation.

## Known reds that are not this PR's

`test_create_mission_propagates_named_exception_type` fails inside **any** git worktree and passes from the repository root — environmental. `tests/status/test_sync_lane_mapping.py` fails in full-scope runs and is `14 skipped, 0 failed` in isolation — order-dependent.

## Hand-off

Green, un-drafted, no landing-order constraints remaining. I do not merge — over to the operator.

stijn-dejongh:
## 🛬 Maintainer landing pass — claimed

Picking this up. Rebasing onto current `upstream/main` (base is 131 commits behind), fresh adversarial squad over the integrated diff, deliverable-hygiene sweep, hand-off green for the operator to merge.

### Triage — **LAND** (substantive, not superseded)

New `spec-kitty events tail` command + bounded truncation-safe tail-reader core. Compatibility triage:
- **Still needed** — #3841 is **OPEN**.
- **Not superseded** — `cli/commands/events.py` and `status/tail_reader.py` are absent on `main`; nobody else shipped this.
- **Authorship clean** — all 98 commits authored + committed by @MOES-Media (no tooling-author drift).
- **CI green** on the current base (56 pass). The reviewable code surface is compact — 5 `src/`, 5 `tests/`, 3 `docs/`; the 12k-line / 98-commit bulk is mission artifacts (78 `kitty-specs/` files) + lifecycle bookkeeping.

### Scope note — branch history
This repo **rebase-merges** (every commit lands on `main` individually), and the branch carries ~40 `chore(spec-kitty): status transition WPxx` / tracer / review-trail bookkeeping commits interleaved with the logical feat/test/fix commits. I'll consolidate the bookkeeping churn into a coherent linear history (step 6b — byte-identical tree, authorship preserved) so `main` gets the feature, not the lane-lifecycle noise, unless you'd rather land the history as-is.

Proceeding: rebase → squad (correctness/fail-closed on the truncation + fd/TOCTOU + concurrency; design on the bounded-core/thin-shell seam) → hygiene → history cleanup → hand-off. Evidence to follow.

stijn-dejongh:
## ✅ Landing pass complete — merge-ready, handing off

**Verdict: LAND.** Rebased onto current `upstream/main` (was 131 behind), history consolidated 98 → 7 coherent commits (operator-chosen), fresh squad clean, 2 folds. Operator merges — I do not.

### History consolidation (98 → 7, byte-identical)
This repo rebase-merges, so the 98 commits — ~52 pure lane-lifecycle / tracer / review-trail bookkeeping — would each have landed on `main`. Rebuilt to one coherent commit per concern, **@MOES-Media authorship preserved** on every content commit (committer is the maintainer; `Co-Authored-By: Claude Opus 5` retained):

| Commit | |
|---|---|
| `feat(status)` | truncation-safe resumable tail-reader core |
| `feat(events)` | `spec-kitty events tail` command + registration + completion manifest |
| `test(events)` | tail-reader + CLI coverage (truncation shapes, concurrency, resume) |
| `docs(events)` | CLI reference + system-events glossary + retrieval index |
| `docs(mission)` | mission artifacts (spec/plan/tasks/reviews/tracers) |
| `fix(landing)` | **fold** — non-object-line guard (see below) |
| `docs(landing)` | **fold** — changelog entry |

The rebuilt MOES tree is **byte-identical** to the correct rebase-of-PR-onto-main (verified `git diff` empty against a reference merge). The 3 generated files (completion manifest, cli reference, retrieval index) were **regenerated** on the new base rather than hand-merged; docs-index reconciliation folded into its `docs(events)` commit so no intermediate commit is index-stale.

### Second-opinion squad (opus, profile-loaded, over the integrated diff) — no blockers
- **Correctness / fail-closed (reviewer-renata): APPROVE.** All three truncation shapes detected by **independent** checks every poll (size + SHA-256 content invariant); single-fd `os.fstat` closes the size/read TOCTOU window; resume **refuses** with no partial stdout; termination bounded (no hang). 25 core tests traced independently.
- **Design (architect-alphonso): SOUND.** Core is stdlib-only (zero Typer coupling), poll loop owned via injectable `sleep_fn`, thin for-loop shell; registration + manifest consistent; stdout data / stderr diagnostics cleanly split.
- **SSOT / boundary (paula-patterns): clean.** The byte-offset tail reader is a justified separate primitive (not a `store.read_events` duplication — incremental offset + truncation defence is a different contract); event schema is opaque pass-through (no drift); canonical `resolve_mission_handle` (fail-closed); no `feature*` alias; no upward `cli/` import.

### Fold (1 code, surfaced convergently by 2 lenses)
- **`fix(landing)` — non-object event-line guard.** `poll_once` parsed each line then wrote reader keys with no `isinstance(parsed, dict)` guard, so a valid-JSON but non-object line (`42`, `[1,2]`, `true`) raised an opaque `TypeError` where the canonical `store.read_events_raw` raises `StoreError("expected JSON object")`. Defer to that single authority. **Red-first verified** (the parametrized test fails with the guard removed). Also documented the load-bearing suffix-truncation threat model at the content-invariant site (renata + architect NOTE).

### Deferred — noted, not folded (genuine follow-ups for the feature owner)
- **Envelope has no `schema_version`.** This is a new machine-consumable surface external consumers persist cursors against; additive fields are safe but a future breaking change has no negotiation point. Worth a version handle.
- **`--json` is parity-only** (accepted-but-unused, documented). If a human-readable mode is ever added the flag's meaning flips — wire it then.
- **`_start_of_last_line` re-reads the whole `[0, offset)` prefix every poll** (O(offset) IO ~4×/s) to locate one newline. Fine for small event logs; a bounded backward-chunk read would scale it.

### Evidence
- Events + tail suite **41 passed** on the final tree (+5 from the new guard test); ruff + mypy clean on all changed src; commitlint clean across all 7 commits.
- Gates green locally: marker/shard completeness, cli-reference freshness, completion-manifest freshness, docs-freshness (errors=0; external-URL link-health warnings only), terminology guard.

Watching CI; merge-ready on green. **Operator merges.**

Issue #3816, opened 2026-08-30
charter context --include directive:<id> selector returns EXIT 1 for valid directive IDs


stijn-dejongh:
## `charter context --include directive:<id>` returns EXIT 1 for valid directive IDs

**Bug** · Milestone: 3.2.6 · Priority: P1 · Parent epic: #3809

### Summary
`spec-kitty charter context --include directive:<id>` returns **EXIT 1 "No directive found
for selector"** for valid, currently-active directive IDs, while the `tactic:` and
`agent-profile:` selectors resolve `EXIT 0` for equally-valid IDs. The directive selector is
broken even though the same directive IDs are listed by `charter context --action <x> --json`
(e.g. `025-boy-scout-rule`, `001-architectural-integrity-standard` both appear in the active
directive set).

### Confirmed evidence (live, 2026-08-30)
```
$ spec-kitty charter context --include directive:025-boy-scout-rule
Error: No directive found for selector 'directive:025-boy-scout-rule'.   → EXIT 1
$ spec-kitty charter context --include directive:001-architectural-integrity-standard
                                                                          → EXIT 1
$ spec-kitty charter context --include tactic:five-paradigm-parallel-debugging
                                                                          → EXIT 0
$ spec-kitty charter context --include agent-profile:researcher-robbie
                                                                          → EXIT 0
```
Both directive IDs are valid — they are enumerated in the active directive set returned by
`charter context --action tasks --json`. Only the `directive:` selector family fails to resolve;
`tactic:` and `agent-profile:` selectors resolve the same shape cleanly.

### Impact
This blocks the **"inject directive IDs, not bodies"** compaction strategy of the
orchestrator-resolves-then-injects contract (#3811, design §4.2:121). If the orchestrator
injects a compact directive **ID** into a delegate prompt, the delegate is expected to expand
it on demand via `charter context --include directive:<id>` — but that selector returns EXIT 1,
so the injected ID is a **dead end**. Until fixed, the injector must either (a) fix this selector,
or (b) inject full directive **bodies** (larger prompts, defeating the compaction goal).

### Fix options
- Repair the `directive:<id>` selector resolution in `charter context --include` so valid
  directive IDs resolve `EXIT 0` and return the directive body, at parity with the working
  `tactic:` / `agent-profile:` selectors.
- Add a regression test asserting `directive:<id>` resolves EXIT 0 for every directive the
  action index reports as active (close-by-construction, directive 043).

### Provenance
[...]

Issue #3604, opened 2026-08-20
Project mission-type type-wide governance (governance-profile.yaml) into the DRG so cascade reaches it


stijn-dejongh:
## Problem

After M5 (#2829, PR #3600) made the charter cascade follow the action hop, three of the four built-in mission types cascade to their governance (documentation 31, research 23, software-dev 160). **`mission_type:plan` still cascades to empty** — but not because the traversal dead-end persists (it's closed for plan too). Plan's governance is authored **type-wide** in `packs/built-in/missions/plan/governance-profile.yaml` (`selected_directives: [031-context-aware-design]`, 9 `selected_tactics`, 3 `selected_paradigms`, `planning-and-tracking` styleguide), and plan's action-grain indexes (`actions/*/index.yaml`) are **intentionally empty** (FR-004/FR-013 keep type-grain and action-grain disjoint).

The DRG only emits `action --scope--> governance` edges from action-grain `index.yaml` files. **The extractor never reads `governance-profile.yaml`**, so no mission type's *type-wide* governance is represented in the DRG at all — it just happens to be invisible for `plan` because plan authors *only* type-wide governance.

## Why it's its own mission (not folded into M5)

Projecting `governance-profile.yaml` into the DRG is a new architectural seam that:
- affects **all four** mission types (each has type-wide governance);
- needs its own **ADR** (which relation — `requires`? a new one? — and which source node: `mission_type` vs each `action`; how it composes with action-grain `scope`);
- **moves golden counts a second time** (new edges for every mission type; cascade output grows) — M5 deliberately re-ledgers once.

## Proposed approach (for the mission to decide)

Emit `mission_type --requires--> <selected_* governance>` (or `scope`, per ADR) from each `governance-profile.yaml`, so cascade reaches type-wide governance uniformly. Re-ledger golden counts once for that change. Add red-first coverage that plan (and every governance-bearing type) cascades to its type-wide governance.

## Refs
- Follows M5 / PR #3600 / #2829.
- Finding surfaced by the M5 pre-merge review; documented in `kitty-specs/kind-complete-cascade-orphan-wiring-01M0FQCD/research.md`.

PR #3871, opened 2026-09-03
fix(charter): resolve activation from `activated_*`, closing the SPDD/REASONS split-brain

### LINKED ISSUE #3838: SPDD/REASONS activation split-brain: is_spdd_reasons_active reads selected_* while the compiler activates via activated_*
## Bug

Verified in source at `eff63532`:

- `src/charter/offering/spdd_reasons/activation.py:61` — `is_spdd_reasons_active` decides from `governance.charter.selected_*`.
- `src/charter/activation/compiler.py:324-330, 453-455` — the activation compiler applies doctrine from `activated_*`.

The two can disagree: a charter whose `activated_*` set enables the SPDD/REASONS pack while `selected_*` does not (or vice versa) gets doctrine applied by one code path and the SPDD-specific behavior gated by the other. The three-state activation semantics make this reachable through ordinary charter evolution, not just hand-editing.

## Expected

One source of truth for "is this pack active" — presumably the compiler's `activated_*` resolution — consumed by `is_spdd_reasons_active` (and anything else currently reading `selected_*` for activation decisions).

## Impact

Downstream consumers that render governance state (e.g. a doctrine/charter panel) currently have to label this as a known drift class with a source citation, because there is no ledgerable single answer to "is SPDD active here".

Found by static audit during Kitty Desktop's Phase-2 upstream-readiness check; not yet reproduced live (flagging before it bites a real charter).

MOES-Media:
Closes #3838

Fixes the SPDD/REASONS activation split-brain: four `src/charter/` surfaces were reading `selected_<kind>` — the **authoring record** — where they should read `activated_<kind>`, the **resolved authority**. A project whose activation differed from its authoring set got the wrong doctrine delivered, quietly.

## The defect class, and why this PR is mostly about closing it rather than patching instances

Every instance had the same shape: **an absent or un-normalized input silently collapsing to a wrong value instead of raising.** Nothing failed. Nothing reported. The wrong doctrine simply arrived.

The review loop found **eight** distinct instances, each by a different round, and none by the round that fixed its predecessor:

1. Fixtures that passed against both old and new code (bug-preserving).
2. `pack_context=None` collapsing to empty sets instead of catalog defaults.
3. An org-required union dropping `_normalize_directive_id`, letting a stem-form id (`001-architectural-integrity-standard`) sit where the DRG uses `DIRECTIVE_NNN` — so the exclusion guard would silently drop a legitimately org-required directive.
4. A fourth in `resolver.py`'s `_resolve_directive_base`.
5. A missing direct-path fixture on one union boundary.
6. The `#883` mocked-graph allowlist collision.
7. A real org-pack directive, reached by a genuine DRG scope edge, excluded because the allowlist was sourced from the built-in catalog only.
8. No regression guarding the union bound itself (see below).

That recurrence is why the acceptance bar became an invariant rather than a checklist:

> Every directive, tactic and paradigm identifier is canonicalized at the moment it enters a union, and every union and exclusion boundary either canonicalizes its inputs or fails loud. An identifier whose form cannot be canonicalized is an error, never a silently-excluded entry. Absent input resolves to the documented catalog default, never to an empty set.

Every union and exclusion boundary in the diff was then audited against it **one at a time**, not in aggregate.

## Two operator rulings, both verified adversarially

**The closure walk is scope-gated.** `project_directives` was serving two incompatible jobs — the exclusion-guard *allowlist* and the seed for a `requires`/`suggests` closure walk whose results were unioned in with no reachability check. `None → all built-ins` is right for an allowlist and wrong for a seed. The uses are now separated with an ownership gate built from direct `SCOPE` edges. Verified by bypassing the gate: both named cross-mission regression tests failed with the exact leak shape, then passed again on restore.

**The default resolves from the active graph, not a hardcoded catalog.** `load_doctrine_catalog()` returns only built-ins, so a mocked graph and a real allowlist disagreed — and, more consequentially, a real org-pack directive reached by a genuine DRG scope edge was excluded. The default is now the union of the catalog with the *already activation-filtered* graph.

That union is only bounded **because** the graph side is filtered — and a lens proved nothing guarded that: disabling the filter left **120 tests across 11 files green**. This PR adds the missing regression, which asserts the observable consequence (a graph-reachable but unactivated directive must not be delivered) rather than grepping for a call site, so a rename that broke the behaviour would still fail it.

## Two other missions' binding tests pass byte-identical

`test_implement_lacks_003_but_review_still_delivers_it` (FR-005) and `test_action_doctrine_keys_off_meta_json_not_template_set` (the #883 cross-mission-type leak-prevention regression) both pass **unmodified** — verified by `git diff` against the mission base, not by "they pass". The `#883` test guards a leak that already shipped once and was fixed; rewriting it to accommodate this change was considered and explicitly rejected.

## Evidence

- `tests/charter/` + `test_charter_offering_does_not_import_activation.py` + `test_no_dead_symbols.py` → **2586 passed, 25 skipped, 0 failed** (pre-mission baseline: 2536 passed)
- Red-first proven **per work package by each reviewer independently**, by reverting the production file and observing the failure — not taken from the implementer's report. Sample: reverting `delivery_table.py` produced a clean `TypeError: 'NoneType' object is not iterable`, proving the `None`-handling contract is load-bearing.
- Four production files revert-checked for bug-preserving tests: **none found**.
- Every test file's collecting CI job confirmed by real `--collect-only`; markers, not directories, decide collection here (#3241).
- Pre-merge squad: 2 lens groups, both findings folded, **0 severity ≥3 surviving**.

Full R1–R6 trail, including both operator rulings, under `kitty-specs/spdd-reasons-activation-split-brain-01M1K6VN/reviews/`.

## Known reds that are not this PR's

`test_create_mission_propagates_named_exception_type` fails inside **any** git worktree and passes from the repository root — environmental, reproduces on unrelated branches.

## Deferred deliberately

The non-mapping-config crash class is now guarded at 5+ independent sites, each re-stamping its own check. A single "load `.kittify/config.yaml` as a validated mapping, or fail closed once" boundary is the right answer, and is a refactor with its own design question — not something to cram into this PR.

---

## Maintainer landing pass

Rebased onto current `upstream/main` (was ~105 commits behind). The one code-file overlap — `resolver.py`, which main's **#3816** (`resolve directive --include selector at slug parity`) also touched — auto-merged cleanly; the design review confirmed the two normalizers are **lock-step identical**, so #3816's slug-parity and this PR's `activated_*` derivation neither duplicate nor undo each other.

**Second-opinion squad** (profile-loaded, over the integrated diff): correctness (0 severity ≥3), SSOT (clean, no blocker), design (**one SEV-2, folded**).

- **Folded (SEV-2, design):** the new closure scope-gate in `delivery_table.py` recorded *every* `Relation.SCOPE` edge, but SCOPE is two-grain (#3604/#3633) — a `mission_type:`-sourced (type-wide) directive reached via the requires/suggests closure was misread as "owned by a different action" and silently dropped, reintroducing the exact split-brain class this PR closes. Fix restricts the owner map to `action:`-sourced edges; red-first regression `test_type_wide_scoped_directive_reached_by_closure_still_widens` fails (`DIRECTIVE_010 not in ['DIRECTIVE_038']`) with the fix reverted, passes with it.
- **Changelog** entry added under `[Unreleased] → Fixed`.

**Operator confirmation (design-intent, not a defect):** a project with a present-but-empty `.kittify/config.yaml` (no `activated_*` keys, never charter-compiled) now reads **SPDD-active** — this is invariant-compliant ("absent input → catalog default, never empty") and matches `PackContext` parity, and is documented as Decision Record 1 / Option A. Flagging so it's a deliberate ruling, not a surprise.

**Deferred (noted, not folded):** the three-state `None → all-builtins / frozenset() → empty` guard is re-inlined at ~4 consumer sites; extracting a shared `activated_or_default` primitive (so `resolver` reuses `action_doctrine_bundle`'s core) is a follow-up architecture change, not a landing fold. The `spdd_reasons/activation.py` config-reader copy is a **C-004-mandated** anti-corruption layer (parity-tested), correctly kept.

**History:** rebuilt from 95 lane-churn commits to 10 (7 contributor WP commits + 3 maintainer landing folds), byte-identical final tree, contributor authorship preserved.

The operator merges.

---
<a href="https://app.blacksmith.sh/Priivacy-ai/codesmith/spec-kitty/pr/3871"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791063261&installation_model_id=427282&pr_number=3871&repository=Priivacy-ai%2Fspec-kitty&return_to=https%3A%2F%2Fgithub.com%2FPriivacy-ai%2Fspec-kitty%2Fpull%2F3871&signature=a5476404426e41deda1a4ad605f29617dec2bdc8f626159256efd9ee00901c6f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


MOES-Media:
**CI green — merge-ready.** `36 passed, 38 skipped, 0 failed`.

## Landing pass

Six checks were red on the first run. Each was classified before anything was touched; none was retried to green, and in two cases the fix went into the *checker* rather than the code it flagged.

| Check | Bin | Disposition |
|---|---|---|
| `unit-contract-residual` | **stale fixture, not a regression** | `test_render_command_template_keeps_spdd_block_when_active` saw the SPDD block absent. Because this PR changes how activation resolves, that was treated as a suspected product regression first. Verified in a disposable worktree at `origin/main`: the old fixture **passes** there, because the pre-change `is_spdd_reasons_active` read `.kittify/charter/` directly, while the new one reads `.kittify/config.yaml`'s `activated_paradigms` — which the fixture never wrote. The fixture was still speaking the retired `selected_*` dialect. Re-pinned to the `activated_*` authority, matching every other fixture in the mission. |
| `lint` (`check_patch_targets.py`) | **the checker was wrong** | Not commitlint — a separate step. It flagged `patch("…_load_doctrine_selection", create=True)`, a symbol this PR deliberately removes, where the test uses `create=True` precisely because the attribute no longer exists. The regex checker never accounted for `create=True`. **Fixed the checker** (AST-based exemption), not the test. |
| `doctrine-charter-tests` / `fast-tests-doctrine` | **PR defect** | A glossary cross-link did not resolve. Anchor derivation lowercases, strips backticks, then drops everything outside `[a-z0-9 _-]` — so HTML-escaped `&amp;lt;kind&amp;gt;` slugified to `activated-ltkindgt`, not `activated-kind`. Changed the headings to backticked raw angle brackets so heading and links agree. |
| `docs-freshness` | **stale generated artifact** | The glossary heading change required regenerating the docs retrieval index, which also surfaced an `activated_<kind>` row never regenerated when the entry was added. Unrelated drift from `main` picked up by a full CLI-reference regeneration was deliberately reverted to keep this PR on-topic. |
| `cutover-guard` | **known tooling defect** | The standard flow never writes `status_phase`, so this fails closed on every mission. Ran the prescribed remedy. Flagging again: **the remedy under-reports itself** — printed `Flipped: 0, Skipped: 1` while actually writing `"status_phase": "1"` into `meta.json`, caught by diffing rather than trusting the summary. NFR-002's archive gate still passes. |
| `quality-gate` | aggregate | Cleared with the rest. |

## Evidence

- `tests/charter/` + `test_charter_offering_does_not_import_activation.py` + `test_no_dead_symbols.py` → **2587 passed, 25 skipped, 0 failed** (pre-landing baseline 2586; the extra pass is the regression test added below).
- The three protected cross-mission regression tests remain **byte-identical** to the mission base — verified by `git diff`, not by "they pass".
- `commitlint` verified per-commit across the range with the repo's own invocation.
- `ruff` and `mypy --strict` clean on every touched file; no suppressions added.

## The last finding folded before this PR opened

The pre-merge squad found that **nothing guarded the union bound**: the `None → all built-ins` default is only bounded because the graph side is activation-filtered, and disabling that filter left **120 tests across 11 files green**. The added regression asserts the observable consequence — a graph-reachable but unactivated directive must not be delivered — rather than grepping for a call site, so a rename that broke the behaviour would still fail it.

## Known red that is not this PR's

`test_create_mission_propagates_named_exception_type` fails inside **any** git worktree and passes from the repository root — environmental, reproduces on unrelated branches.

## Hand-off

Green, un-drafted, no landing-order constraints. I do not merge — over to the operator.

stijn-dejongh:

Picking this up for today's landing pass: rebase onto current `upstream/main`, classify every red check into the four bins, adversarial second-opinion squad over the integrated diff, fold as needed, hand off green-modulo-honest-reds for the operator to merge.

## Early triage recommendation: **LAND**

Fresh PR (opened yesterday) but substantive (99 commits, ~13k additions across `src/charter/activation/` + `offering/spdd_reasons/`), so ran the full compatibility triage:

- **Superseded?** No. The defect is live on current `main` — `src/charter/activation/action_doctrine_bundle.py:189` still reads `doctrine_selection.selected_directives` (the authoring record) rather than the activated authority. Issue #3838 is still **OPEN** (P2, doctrine).
- **Compatible?** Yes. Every surface it touches (`action_doctrine_bundle.py`, `context_renderers/delivery_table.py`, `resolver.py`, `offering/spdd_reasons/activation.py`) still exists on current main. The three-dot diff (`upstream/main...HEAD`) is clean — only PR-owned files, no stale-stack governance smuggling.
- **Still needed?** Yes — the split-brain reproduces on current main.

**Note on CI signal:** the latest run shows most fast/integration shards as `skipping` (path-filter / superseded-run artifact) with only arch/build/windows/clean-install actually executed. The rebase + push will trigger a full re-run, so the merge-ready signal will come from that, not the current partial green.

Rebasing onto current main (105 commits behind) and classifying now; evidence to follow.

stijn-dejongh:
## 🛬 Landing pass complete — merge-ready on green

Verdict: **LAND.** Rebased onto current `upstream/main`, adversarial squad run, one MAJOR folded, history rebuilt. Force-pushed to `58f9305b0a`. The operator merges — I have not.

### Rebase & integration
- Was ~105 commits behind; rebased clean. The one code-file overlap, `resolver.py` (also changed by **#3816**, `resolve directive --include selector at slug parity`), auto-merged — the design lens confirmed #3816's `id_normalizer.normalize_directive_id` and this PR's `profile_resolution._normalize_directive_id` are **lock-step identical**, so slug-parity and `activated_*` derivation neither duplicate nor undo each other.

### Second-opinion squad (profile-loaded, opus, over the integrated diff)
| Lens | Verdict |
|---|---|
| Correctness (reviewer-renata) | 0 severity ≥3; acceptance invariant holds across all four surfaces |
| SSOT (paula-patterns) | Clean, no blocker; the one new copy is a C-004-mandated, parity-tested anti-corruption layer |
| Design (architect-alphonso) | **1 SEV-2 (folded)**; authority flip, #3816 coherence, paradigm/directive symmetry, ruling-2 union bound all clean |

### Folds (maintainer, `<type>(landing):`)
1. **`fix(landing)` — closure scope-gate two-grain bug.** `_classify_artifact_urns` recorded *every* `Relation.SCOPE` edge, but SCOPE is two-grain (#3604/#3633): a `mission_type:`-sourced (type-wide) directive reached via the requires/suggests closure was misread as "owned by a different action" and silently dropped — reintroducing the exact split-brain class this PR closes. Fix gates the owner map to `action:`-sourced edges. **Red-first proven:** `test_type_wide_scoped_directive_reached_by_closure_still_widens` fails `DIRECTIVE_010 not in ['DIRECTIVE_038']` with the fix reverted, passes with it.
2. **`docs(landing)` — changelog** `[Unreleased] → Fixed` entry (consumer-focused) + docs retrieval-index regen for the new anchor.
3. **`docs(landing)` — MD049 cleanup.** Touching `CHANGELOG.md` un-masks its whole markdownlint set; cleared the pre-existing emphasis-style findings (punctuation only, wording unchanged).

### Considered, not folded
- **Operator confirmation (design-intent, not a defect):** a present-but-empty `.kittify/config.yaml` (never charter-compiled) now reads **SPDD-active** — invariant-compliant ("absent → catalog default, never empty"), matches `PackContext` parity, documented as Decision Record 1 / Option A. Flagging so it's a deliberate ruling.
- **renata SEV-1 (resolver union):** normalizing the charter-selected side at the union entry risks changing the canonical form of *locally-declared stem-form* directives; not load-bearing (already fails loud on unknown ids). Left as-is.
- **3× mypy `no-any-return` in `resolver.py`** (lines 567/569/592) are **pre-existing on `main`** (PR-untouched lines) and advisory — not this PR's; not folded.
- **Follow-up architecture (noted, not ticketed):** extract a shared three-state `activated_or_default` primitive so `resolver` reuses `action_doctrine_bundle`'s core instead of re-inlining the `None → all-builtins` guard at ~4 sites.

### Local verification (targeted; CI is the release authority)
- 144 charter/parity/scope-gate tests green (incl. the #3816 integration-point tests).
- `ruff`, `mypy --strict` (changed files), terminology guard, `check_docs_freshness --ci` (errors=0; only external-URL link warnings), markdownlint, marker/collection completeness gates, commitlint (all commits) — clean.

### History cleanup
Rebuilt **95 lane-churn commits → 10**: 7 contributor WP commits (`@MOES-Media` authorship preserved) + 3 maintainer landing folds. **Byte-identical final tree** (both parity gates empty). No net deletions.

Watching CI; **merge-ready on green** — operator merges.

stijn-dejongh:
✅ **All CI green on `58f9305b0a`** — 9/9 check runs success (CI Quality, doctrine-charter-tests, ui-e2e, ci-windows, docs-freshness, docs-build-pr, drift-detector, Release Readiness, canonical-producer-lint). `mergeStateStatus: CLEAN`, `MERGEABLE`, not draft.

Landing pass done: rebuilt history (10 commits, byte-identical tree), one design-lens SEV-2 folded with red-first proof, changelog + docs index updated. **Merge-ready — operator merges.** One design-intent ruling flagged above for confirmation (bare `config.yaml` → SPDD-active), not a blocker.

Issue #3913, opened 2026-09-06
mypy baseline-red: 21 errors in runtime/next engine.py, runtime_bridge_engine.py, schema.py (missing mission_id/mission_slug on spec_kitty_events payloads)


robertDouglass:
## Pre-existing failure report (charter: Pre-existing Failure Reporting Rule)

Found while implementing WP05 of mission `fsm-write-path-integrity-01M1TZV6` (lane-e). Not introduced by that WP.

**Command**

```
.venv/bin/mypy src/runtime/next/runtime_bridge_io.py src/runtime/next/_internal_runtime/engine.py src/runtime/next/decision.py
```

**Result on the WP's merge-base `missions/coreloop-proto-missions` @ `7b90adcfdaf4ab5ef5229fa466bb6091f5769a07` (verified in a scratch worktree at that exact commit)**

```
Found 21 errors in 3 files (checked 3 source files)
  12 src/runtime/next/_internal_runtime/engine.py
   1 src/runtime/next/_internal_runtime/schema.py
   8 src/runtime/next/runtime_bridge_engine.py
```

Representative errors (all the same class):

```
src/runtime/next/_internal_runtime/engine.py:248: error: Missing named argument "mission_id" for "MissionRunStartedPayload"  [call-arg]
src/runtime/next/_internal_runtime/engine.py:248: error: Missing named argument "mission_slug" for "MissionRunStartedPayload"  [call-arg]
src/runtime/next/_internal_runtime/engine.py:330: error: Missing named argument "mission_id" for "NextStepAutoCompletedPayload"  [call-arg]
src/runtime/next/_internal_runtime/engine.py:443: ... "NextStepIssuedPayload"
src/runtime/next/_internal_runtime/engine.py:482: ... "DecisionInputRequestedPayload"
src/runtime/next/_internal_runtime/engine.py:499: ... "MissionRunCompletedPayload"
src/runtime/next/_internal_runtime/engine.py:736: ... "DecisionInputAnsweredPayload"
src/runtime/next/runtime_bridge_engine.py:216 / :268: same two missing arguments
```

[...]

Issue #3908, opened 2026-09-06
[bug] Compact charter context and runtime prompts lose activated governance while reporting success


robertDouglass:
## Observed

At main commit `c0054153b9bce0778cf41a85d11ecd4e9650031d`, a fresh source checkout with `uv sync --frozen --all-extras` resolves profiles and initially reports 235 references from charter context, but subsequent compact context and runtime research prompts lose the structured governance payload.

## Reproduction

In a fresh clone with the repository's committed charter/config:

```sh
SPEC_KITTY_ENABLE_SAAS_SYNC=0 .venv/bin/spec-kitty agent profile show planner-priti
SPEC_KITTY_ENABLE_SAAS_SYNC=0 .venv/bin/spec-kitty charter context --action specify --json
SPEC_KITTY_ENABLE_SAAS_SYNC=0 .venv/bin/spec-kitty charter context --action specify --json
```

The second context call exits zero and reports `success:true`, but compact mode reports `references_count:0`, empty directives/tactics/references, and `Diagnostics: governance unresolved`.

The diagnostic says selected directives are unavailable, including `001-architectural-integrity-standard`, `034-test-first-development`, and `044-canonical-sources-and-unification`, despite tracked sources under `packs/built-in/directives/`. The rendered Project root ends in `.kittify/doctrine`, rather than the checkout root.

Also observed after creating a software-dev mission and starting its verified not_started runtime: the discovery/research prompt has `Template set: (none)`, no paradigms/tools/directives/tactics, and the same diagnostics. A readable step prompt is returned, but its governance payload is degraded.

## Expected

Compact and runtime prompt resolution must retain the activated governance semantics of bootstrap resolution, using the correct repository root and catalog. If resolution fails, machine consumers need an explicit degraded/failure signal rather than interpreting top-level success as resolved governance.

## Source References

- [compact degradation](https://github.com/Priivacy-ai/spec-kitty/blob/c0054153b9bce0778cf41a85d11ecd4e9650031d/src/charter/activation/compact.py#L293)
- [unavailable directive validation](https://github.com/Priivacy-ai/spec-kitty/blob/c0054153b9bce0778cf41a85d11ecd4e9650031d/src/charter/activation/resolver.py#L812)

Root cause not yet established; the apparent root mismatch is observed output, not a proven diagnosis.

## Impact / Workaround

[...]

PR #3904, opened 2026-09-06
Proto-missions: fsm-write-path-integrity + dead-port-disposition (core-loop hardening, spec-level)

stijn-dejongh:
## Two spec-level proto-missions from the 2026-09-06 core-loop hardening arc

**What this PR is:** the specification slice ONLY — mission scaffolds + authored specs + their
verified research dossiers, deliberately stopped before `/spec-kitty.plan`. A refining team picks
these up, resolves the marked open decisions, and runs plan → tasks → implement. Squad-reviewed
(research squad → spec authoring → post-spec fidelity + refinement-readiness reviews, findings
folded; 0 BLOCKERs remained).

### Rationale (the evidence chain, condensed)

An independent second-opinion modularity audit and a core-loop (FSM → hooks/ports) hardening squad
verified — via fact-check + adversarial design review — that the core-loop *design* is sound and
ADR-ratified, while the *write path* carries bounded, serious implementation leaks:

- the status event log (`status.events.jsonl`, the FSM's sole authority) has **3 locked / 4
  unlocked writer families** (worst: a raw unlocked append on the live post-merge path);
- **three parallel emit orchestrations** exist where docs describe one;
- a **phantom fan-out** defect announces an event before the coordination commit that a rollback
  then truncates;
- dependency gating lives outside the FSM guard, so direct emitters can claim dep-blocked WPs;
- the dormant mission-DSL v1 keeps the `transitions` library on the `spec-kitty next` **hot path**
  via an eager import chain, and two hook seams carry a stale/fictional design story.

The verified plan verdict was **GO-WITH-AMENDMENTS**; all amendments are baked into these specs
(corrected writer census, `_prepare_event` promotion as the emit-collapse mechanism, fail-OPEN
dependency guard, replay-never-validates, A-before-B sequencing — the two missions share 4 runtime
files and are NOT parallel-safe).

### The missions

| Mission | Scope | Shape |
|---|---|---|
| `fsm-write-path-integrity-01M1TZV6` (P0) | Serialize all writers incl. the lockless batch door (FR-018), unify the three emit orchestrations, close phantom fan-out, dependency guard into the FSM (fail-open), run-state crash-safety + mission_id keying | 5 stories / 18 FRs / 10 marked `[NEEDS DECISION]` — the load-bearing one (Q4, the *named* orchestrator) must resolve before WP02 |
[...]

PR #3898, opened 2026-09-06
chore(docs): retire codemap debris + accept RuntimeEventEmitter disposition ADR

stijn-dejongh:
Two independent post-convergence DECIDE-items, bundled as one PR for the operator to
adjudicate. Nothing here merges itself. Both items come from the post-convergence census reports
(24 red-team, 25 dead-code, 26 dead-ops).

---

## 1. Retire `docs/codemap/` (executed — operator may veto)

`git rm` of `docs/codemap/{codemap.json,codemap.html,codemap.lock}`. Rationale (all verified at
`e721763759`):

- **Provenance = contributor debris.** All three files landed in a *single* commit,
  `03552057aa` (2026-09-02, external contributor, "feat: support explicit owned checkout
  lifecycle"), reaching `main` via the convergence merge. They are not a maintained repo surface.
- **No generator exists — anywhere.** `grep codemap` across `src/`, `tests/`, `Makefile`,
  `pyproject.toml`, `.github/`, `scripts/`, `packs/`, `docs/`, and full git history returns zero
  hits outside `docs/codemap/` itself. The map cannot be regenerated.
- **`codemap.lock` pins a phantom base.** `source_base=ca91f626ea542c0a1eca48cd2607eb0480015544`
  is **not an object in this repository** (`git cat-file -t` → "could not get object info"). The
  lock is a snapshot of the contributor's private tooling state.
- **Stale content.** The map still indexes deleted `sync/` and `saas/` trees.

**Reference-safety check (nothing blocks retirement):** the scoped purge targets are all clean —
zero `codemap` references in `docs/` (outside `docs/codemap/`), `README*`, `tests/`, `src/`,
`scripts/`, `Makefile`, `pyproject.toml`, `.github/`, or `packs/`. The only surviving references
are inside the contributor mission's **own historical planning artifacts**
(`kitty-specs/explicit-worktree-commands-01M1A5VJ/`: `lanes.json`, `plan.md`, `WP01`, research
notes). Those are immutable mission snapshots per the Terminology Canon and are deliberately left
untouched — they record what that mission built; they are not live consumers. No ADR is needed to
remove derived debris.

---

## 2. ADR 2026-09-06-2 — RuntimeEventEmitter seam disposition (status: **Accepted**, Option 1)

`docs/adr/3.x/2026-09-06-2-runtime-event-emitter-disposition.md`. Draft ADR only — **no code
changes to the emitter in this PR.** The operator/Robert accepts (or amends) it; **Mission B
[...]

PR #3921, opened 2026-09-06
mission(dead-port-disposition): consolidate the RuntimeEventEmitter seam and fix the two decision-log bypasses (ADR 2026-09-06-2)

robertDouglass:
## What this mission does

Implements [ADR 2026-09-06-2](https://github.com/Priivacy-ai/spec-kitty/blob/main/docs/adr/3.x/2026-09-06-2-runtime-event-emitter-disposition.md), Option 1: consolidate the runtime emitter seam and preserve decision logging across every engine-facing path.

Decision requests under strict retrospective policy and composition dispatch previously bypassed `DecisionGitLog` and disappeared. Both paths now reach the wrapper, persist one request per decision, and retain terminal-refusal rollback. Request → repeat poll → answer probes verify exactly one requested event and one answered event.

The duplicate concrete `RuntimeEventEmitter` is deleted. The canonical eight-method Protocol remains; `NullEmitter` and a gated factory/registry provide the reserved E3 producer seam. No live producer is registered.

Optional producer seeding cannot change mission control flow: absent hooks are accepted, lookup/call failures are logged, and bootstrap preserves the snapshot's current step. The engine, answer path, and decision-log wrapper share this behavior. This prevents skipped dispatch guards and composition becoming blocked after a successful action.

## Review and regression coverage

Three independent CLI-loaded profiles reviewed the aggregate change: architect-alphonso, debugger-debbie, and reviewer-renata. Confirmed seeding and architectural-guard findings were fixed in isolated worktrees, then independently reviewed. Final reviewed head: `075553d0941719241e774ff95bd0beed6e175b0a`, rebased onto `19cca411dc11ad7028533d8b21cecb7ced2d3722`.

Red-first evidence: five optional-seeding regression failures and six guard-mutation failures before their fixes. AST guards now require the actual factory constructions and wrapped-emitter call sites; mutation tests reject missing/replaced calls, duplicate flushes, and multiline/keyword-spacing bypasses. Positive formatting tests keep equivalent source layouts valid.

Focused verification: 328 runtime tests and 20 architectural guard tests passed. Independent real-engine probes also pass when the producer's seed attribute itself raises.

## Tests run

All local runtime commands use `SPEC_KITTY_ENABLE_SAAS_SYNC=0` and `PWHEADLESS=1`; parallel pytest runs use `--dist loadfile`.

On final head `075553d0941719241e774ff95bd0beed6e175b0a`:

| Command | Result |
|---|---|
| `make test-fast` | 1642 passed |
| `.venv/bin/pytest tests/runtime/ tests/next/ tests/specify_cli/next/ tests/specify_cli/events/ tests/status/test_producer_conformance.py tests/contract/test_identity_contract_matrix.py -n 4 --dist loadfile -q -p no:cacheprovider` | 1631 passed, 2 skipped |
| `.venv/bin/pytest tests/architectural/ -n 4 --dist loadfile -q -p no:cacheprovider` | 1800 passed, 2 skipped, 2 xfailed; existing golden-count failure #3927 and upgrade-remediation timeout #3978 |
| Isolated `test_remediation_changes_check_state[charter_source_525]` diagnostic on PR / current main | 1 passed in 14.01s / 1 passed in 16.33s; full-run 60s timeout retained and tracked in #3978 |
| `.venv/bin/ruff check .` / `.venv/bin/ruff format --check .` | clean / 1532 files already formatted |
| GitHub Actions | All applicable checks successful; `suite` and `[code]smith` skipped by repository configuration |

Retired-surface scan: 0 hits over 4340 added lines, using the canonical regex from `tests/architectural/test_no_retired_subsystems.py`.

Ruff passes. Mypy on `_internal_runtime/events.py` and `specify_cli/events/decision_log.py` passes. Broader runtime typing retains exactly the 21 pre-existing errors tracked in #3913, with no new errors.

## Known deviations and follow-ups

- The architectural golden-count gate has one existing failure: 15 sites exceed the frozen ceiling of 14 (#3927). Independently reproduced on current `main`; no threshold was relaxed.
- Compact charter context can lose resolved governance (#3908). Profiles loaded successfully; reviewers read the binding charter directly when compact context degraded.
- NFR-003's net-LOC reduction is not met: the final runtime diff is +68 lines. The ADR and mission spec record the accepted deviation; the factory and safe optional lifecycle handling outweigh the removed no-op class.
- E3 live-producer wiring remains #3929. Optional-seed phase loss and exception propagation are fixed here; registration typing, lifecycle ordering, and producer identity/conformance remain E3 considerations.
- Existing separate follow-ups remain #3928 (decision-ledger artifact classification), #3926 (coordination teardown ordering), #3925 (review-delta process), and #3930 (decision-log commit refusal is invisible in JSON mode).

## Mission artifacts

`kitty-specs/dead-port-disposition-01M1VRA2/` contains the spec, plan, contracts, work packages, independent work-package reviews, acceptance matrix, mission review, and retrospective. The changelog discloses the restored decision-log behavior.

## Post-merge verification

Merged as `d6e8fe4238d55118e158d1b71920208f633a1533`; its tree exactly matches the reviewed head. All applicable main-branch workflows passed, including Windows, CI Quality, branch protection and docs deployment (`ci` suite remains skipped by repository configuration).

Post-merge full architecture sweep without concurrent validation suites: **1801 passed, 2 skipped, 2 xfailed, 1 failure** in 407.41s. Sole failure is existing #3927. The upgrade-remediation timeout did not recur; original failed run remains tracked in #3978.


stijn-dejongh:
## Maintainer landing pass — claim + triage

Picking this up @robertDouglass: rebase onto current `main`, clean the branch history, adjudicate, hand off for merge. Your authorship stays on your commits throughout. Evidence to follow.

### Triage: **LAND**

- **Compatible.** The ADR this executes — `2026-09-06-2-runtime-event-emitter-disposition` — is merged and `status: Accepted` on `main` (#3898). The gate this mission was written against is open, not moved.
- **Still needed.** The correctness half is a real user-visible bug: `DecisionInputRequested` events raised on strict-retrospective-policy `decision_required` advances and on composition dispatch were handed to the plain no-op emitter, so they never reached `decisions.events.jsonl`. Nothing on `main` has fixed that.
- **Not superseded** — but see the sibling note below.

Rebased onto `main` at `19cca411dc` (the branch was 33 commits behind; merge-base was `3a4f92b414`). **Clean — no conflicts.** The ADR frontmatter my #3885 pass corrected survives intact at 176/180 characters, so this PR's ADR body amendment applies on top without disturbing it.

### Sibling note — this settles an open question, so it should be settled deliberately

PR **#3922** implements the *same* emitter ADR under mission `dead-port-disposition-01M1TZVN` — a **different ULID** from this PR's `-01M1VRA2`. Its own body flags the overlap ("PR #3921 executes the same emitter ADR with a different API") and it is a draft partly for that reason. Landing this PR chooses **this** API. That is an operator call, not a merge-order accident — flagging it so it is made on purpose. The spec slice for the `-01M1TZVN` pair landed separately in #3904 and does not conflict with this.

### History: 34 of 37 commits are bookkeeping

The branch is bisect-hostile in its current shape. Classified every commit by what it touches:

| Commits | Touches | Examples |
|---|---|---|
| **2** | `src/` + `tests/` | the WP squash merge, and the mission-review Gate 2 test fix-ups |
| **1** | the ADR only | the net-LOC withdrawal |
| **34** | `kitty-specs/` only | 3× `Accept …`, 3× `Record acceptance commit …`, 3× retrospective captures, 2× duplicate analysis reports, 4× `chore: planning artifacts (WPnn)`, 4× activity logs, plus the scaffold/spec/plan/tasks sequence |

Several also carry subjects commitlint rejects — `Add scaffold for feature …`, `Add spec for …`, `Add plan for feature …`, `Add tasks for feature …` extract no Conventional Commits type — and use the prohibited `feature` term for what the Terminology Canon calls a Mission.

Since this repo rebase-merges, each of those 37 commits lands on `main` individually. Rebuilding to three coherent commits:

1. `feat(runtime):` the seam consolidation + both decision-log fixes — all `src/` and `tests/` changes, with the changelog entry and the Gate 2 test fix-ups folded in, so no intermediate commit is red on a gate it introduces.
2. `docs(adr):` the net-LOC withdrawal, message kept verbatim.
3. `docs(missions):` the mission artifacts, collapsed to their final state.

**Non-negotiable before any force-push:** the rebuilt tree must be byte-identical to the current head (`git diff <old> <new>` empty) — a pure commit-boundary change, nothing else. Your `--author` is preserved on all three, and the `Co-Authored-By: Claude Fable 5.1` trailers are carried forward.

A second-opinion squad runs over the integrated diff before hand-off. I do not merge — the operator does.

stijn-dejongh:
## Landing pass — reconciliation record

@robertDouglass — **I briefly clobbered your four commits from this morning and then restored them. Nothing was lost, and they are all in the branch now.** Full account below, because you should not have to reverse-engineer a force-push.

### What happened

You pushed four commits at 08:30–08:36 (`d0e88988a`, `2de5aaaa2`, `04c936d0f`, `075553d09`). My landing pass had claimed the PR at 08:01 and was mid-flight with a review squad, so I did not see them. When I force-pushed my folds, they went over your work.

My mistake, specifically: I ran the runbook's "check for commits you have not seen" step against my **pre-rebase local sha** instead of the last remote sha I had observed. My rebased commits are not ancestors of the remote, so the check printed the whole branch and I read it as noise. `--force-with-lease` did not save me either — I had fetched immediately before pushing, so the lease matched the tip I was about to destroy. The lease guards a race; it does not guard a bad comparison.

Your four commits were still in my local object store, so the first thing I did was push the branch back to `075553d094` — your exact tip, byte-for-byte — before doing anything else.

### The reconciliation

We had independently found the same four problems within about an hour of each other. Where they overlapped, **your version won**, because on both counts it is better:

| Finding | Kept | Why |
|---|---|---|
| The seam guard was vacuous | **Yours** (`d0e88988a` + `2de5aaaa2`) | Your AST call-contract check asserts `targets == ["ctx.emitter_for_engine"]` — a *positive* assertion, scoped per function, that survives a rename and catches a multi-line reformat. Mine deleted the write-only `ctx.sync_emitter` field and dropped the literal needles; strictly weaker and less general. Dropped mine. |
| Seeding could raise into mission control flow | **Yours** (`04c936d0f` + `075553d09`) | Your `try/except/else` split fixes the `current_step_id` nulling more elegantly than my hoist did, and `seed_runtime_emitter` is the cleaner single seam. Dropped mine. |

My four folds with no counterpart in yours were kept and replayed on top of your tip:

- `fix(runtime): mirror glossary_runner's registry guard; degrade on a raising factory` — `register_runtime_emitter_factory` was silent last-wins with no validation, diverging from the near-identical `src/kernel/glossary_runner.py` precedent (idempotent on identity, `RuntimeError` on a different registrant). Also wraps the factory dispatch so a future E3 producer whose constructor raises degrades to the null seam instead of killing `spec-kitty next`.
- `fix(runtime): revert emitter.py's frozen-contract surface widening` — `_internal_runtime/emitter.py` is Category 6 in `test_no_dead_modules.py`, frozen in `contracts/internal_runtime_surface.md`, which never mentioned the three added re-exports. They had no production consumer.
- `test(runtime): gate the ADR's "no live producer" scope boundary` — that boundary held only by convention; it is now asserted.
- `docs(...)`: the ADR frontmatter still said *"Keeps runtime.next.event_emitter as the reserved E3 seam"* for a module this PR deletes (now 156/180 chars, was 176); the shipped `mission-review.md` still recorded **FAIL** with three blockers that are all resolved on this tip; and `status.json` carried `mission_number: ""` against `meta.json`'s `198`.

Two integration defects surfaced only in the reconciliation and are fixed: my registry fold called a module-level `logger` that existed only in my dropped commit (now uses this module's `logging.getLogger(__name__)` idiom), and a test referenced my `seed_emitter_if_supported` rather than your `seed_runtime_emitter`.

### History

**45 commits → 11.** The 34 lifecycle/bookkeeping commits — 3× `Accept …`, 3× `Record acceptance commit …`, 3× retrospective captures, 2× duplicate analysis reports, 4× `chore: planning artifacts (WPnn)`, 4× activity logs, plus the `Add scaffold for feature …` / `Add spec for …` / `Add plan for feature …` sequence that commitlint rejects and that uses the prohibited `feature` term — collapse into three: the code, the ADR amendment, and the mission artifacts. Your eight post-review commits stay separate, because each is already single-purpose and reviewable. Your `--author` is on all eleven.

One thing the byte-identity gate caught worth mentioning: my first rebuild attempt silently **resurrected `src/runtime/next/event_emitter.py`** — `git checkout <ref> -- .` adds and updates but never deletes, so the PR's headline consolidation was quietly undone. Caught by the parity check, redone with the deletion set computed explicitly. Final tree is byte-identical to the reconciled head (`87e4128713…`).

### Verification

```
blast radius (tests/next tests/runtime tests/specify_cli/events + seam,
  producer-conformance, identity-matrix, ci-collection-completeness)   1450 passed, 1 skipped
make test-fast                                                        1642 passed
ruff check / ruff format --check                                      clean
commitlint (all 11, CI's own invocation)                              clean
byte-identity vs pre-cleanup head                                     identical (87e4128713)
event_emitter.py deleted in the final tree                            confirmed
```

Rebased onto `main` at `19cca411dc`. Zero merge commits, safe to rebase-merge.

### Not folded

- **#3975** (filed): `_wrap_with_decision_git_log` (`runtime_bridge.py:293-304`) returns the **unwrapped** emitter when `DecisionGitLog` construction fails — on the default path that is a `NullEmitter`, so every decision event for that run is silently dropped with no operator signal. Same class as the two bypasses this PR fixes, but a *deliberate* fail-open, so changing it is a behaviour decision with an operator-visible consequence rather than a landing fold.
- 22 mypy errors under `src/runtime/next/` — identical count on `main`, none in the changed hunks, and mypy is not gated in GitHub CI. Pre-existing, not this PR's.
- `runtime_bridge_io.py:592,643` hand a hardcoded `NullEmitter()` to `start_mission_run` from four callers, so `MissionRunStarted` will not reach a registered E3 producer from those paths. Harmless today; worth naming in the #3929 handoff so E3 does not assume the factory is the single door.

I do not merge — handing off.

stijn-dejongh:
Follow-up opened as **#3983**.

This PR squash-merged as `d6e8fe4238` at 09:33 while the landing-pass reconciliation was still in flight, so it landed at @robertDouglass's tip `075553d094` — his four commits are all in, and correctly so (his AST call-contract guards and the `seed_runtime_emitter` control-flow split were both better than my equivalents, which I had already dropped in favour of his).

What did not make the squash was the other half of the squad's findings. #3983 carries exactly those four commits off current `main`:

- the registry guard mirroring `kernel/glossary_runner.py` (silent last-wins → raises on a different registrant) and a `try` around the factory dispatch so a future E3 producer whose constructor raises degrades to the null seam instead of killing `spec-kitty next`
- the `_internal_runtime/emitter.py` frozen-contract-surface revert (Category 6 in `test_no_dead_modules.py`; the three added re-exports were absent from `contracts/internal_runtime_surface.md` and had no production consumer)
- a gate for the ADR's "no live producer is wired" boundary, which held only by convention
- three artifacts that shipped here and currently contradict the tree on `main`: `mission-review.md` still reads **FAIL** with three blockers that were all resolved before merge, `status.json` still has `"mission_number": ""` against `meta.json`'s `198`, and the ADR frontmatter still says "Keeps `runtime.next.event_emitter` as the reserved E3 seam" for the module this PR deleted

No action needed here — this comment is just the pointer so the thread is not left dangling.

Also filed **#3975** for the pre-existing fail-open twin (`_wrap_with_decision_git_log` silently returning an unwrapped emitter on construction failure), which is a behaviour decision rather than a fold.

Issue #3975, opened 2026-09-07
Decision-log wrap fails open silently: a failed DecisionGitLog construction drops every decision event with no operator signal


stijn-dejongh:
## What's wrong

`_wrap_with_decision_git_log` (`src/runtime/next/runtime_bridge.py:293-304`) returns the **unwrapped** emitter when `DecisionGitLog` construction fails:

```python
"""Wrap ``emitter`` with DecisionGitLog for durable decision recording.

Returns the wrapped emitter.  If construction fails (e.g. import error),
the original emitter is returned unchanged so mission execution is not
blocked.
"""
```

That unwrapped emitter is, on the default path, a `NullEmitter`. So on a construction failure every `DecisionInputRequested` for that run is silently dropped, the mission proceeds normally, and nothing surfaces the loss — no warning to the operator, no marker in `decisions.events.jsonl`, no non-zero exit.

## Why this matters now

This is the **exact failure mode** that PR #3921 just fixed at two other sites: decision requests reaching a no-op emitter instead of the decision-log wrap, invisibly. #3921 closed the two known bypasses (strict-retrospective-policy `decision_required`, and composition dispatch) and pinned them with behavioural regression tests. This third path is the same class, but it is a *deliberate* fail-open rather than a wiring mistake, so it survived that mission untouched.

Surfaced during the #3921 landing pass by a correctness review lens (recorded there as NOTE-3, out of scope for that PR).

## The tension to adjudicate

The current behaviour is defensible as written — "a broken decision log must not block mission execution" is a real position, and the docstring states it deliberately. But it is undetectable from outside, which makes it indistinguishable from working correctly. An operator running under a strict retrospective policy (`timing: before_completion`, `failure_policy: block`) is specifically relying on those events being recorded; silently degrading to no recording defeats the policy while reporting success.

Worth deciding explicitly rather than leaving as an accident of the docstring:

- **Keep fail-open, but make it loud** — log at WARNING with `exc_info`, and record a one-line degradation marker so the gap is visible after the fact. Cheapest, preserves the stated intent.
- **Fail closed under strict policy only** — proceed unwrapped for ordinary runs, refuse when the retrospective policy is `failure_policy: block`, since that configuration is an explicit request for the recording to be load-bearing.
- **Fail closed always** — simplest to reason about, but turns a logging dependency into a mission blocker.

[...]

Issue #3929, opened 2026-09-06
E3 producer wiring at the consolidated RuntimeEventEmitter seam (follow-ups from dead-port-disposition-01M1VRA2)


robertDouglass:
## Live Work programme amendment — 2026-09-13

Delivery child of https://github.com/spec-kitty/spec-kitty-planning/issues/2268. **Delivery owner: Software Factory. Priority: P1.** Preserve existing implementation, claims, assignees and PRs. This dated amendment supersedes older scope/dependency prose only as stated here; canonical mission authority, scoped authorization and truthful evidence remain required.

**Dependencies:** Existing bounded repair can proceed independently; no new program prerequisite.

Keep ownership of the six real runtime producer/seeding seams and conformance hazards. Preserve canonical persistence and event IDs. Under Robert’s 2026-09-14 ruling, these post-persistence observations use the bounded live relay publisher; there is no journal-backed publisher, offline recovery queue or second store of done work. This bounded producer repair can land independently; the full-lifecycle issue owns remaining lifecycle families and live integration. Supersession record: https://github.com/spec-kitty/spec-kitty-planning/issues/2269#issuecomment-5660729736 . Full narrative uses the shared authenticated content contract, not widened volatile attrs. Acceptance still requires real command/seeding producer evidence; enum registration or a NullEmitter cannot pass.

The epic contains the architecture, three equally complete People/Missions/Repositories views, all12 LW requirements and measurable release evidence. Earlier post-MVP/volatile-only/history/approval exclusions are historical context where this amendment replaces them. The original report below remains evidence of the problem, not a conflicting current scope gate.

---

## Context
Mission `dead-port-disposition-01M1VRA2` (ADR 2026-09-06-2, Option 1) consolidated the no-op runtime emitter onto `runtime.next._internal_runtime.events` and left a factory + registry (`runtime_emitter_for_mission`, `register_runtime_emitter_factory`) for a future live producer. Wiring that producer is E3 work and was explicitly out of scope. This issue collects what the post-merge mission review and the pre-PR adversarial squad asked the E3 work item to carry.

## Follow-ups owed to E3
[...]

PR #3888, opened 2026-09-06
Post-convergence Tier-3: governance, SSOT & runtime-boundary enforcement

stijn-dejongh:
This PR retires obsolete governance documents, makes the enforced package inventory and import-direction tests the modularity authority, and extends doctrine boundary enforcement to both `src/specify_cli` and `src/runtime`.

Closes #3522.

## Changes

- Supersede four retired-subsystem ADRs and document the client-repository boundary; preserve the accepted shared-package precedent.
- Remove the stale ownership manifest and its artifact-specific schema test. Demote the ownership map and update its live consumers to reference the enforced sources.
- Add the runtime-to-specify_cli exception ledger and independently ratchet both outbound ledgers. Root-form, aliased, and lazy imports now reach the same guards; existing production imports are preserved.
- Scan both runtime roots for legacy doctrine and current charter.offering imports. Pin sanctioned exceptions to file/import pairs rather than entire files. Reject additional member, wildcard, multi-import, and module-control-flow reaches.
- Refresh production scan roots, retire superseded planning documents, and correct hosted-rendering documentation.

Governed mission: `kitty-specs/post-convergence-governance-01M1TMPH/`. Changes are documentation, architectural tests, and Op evidence; no production runtime behavior or package version changes.

## Adversarial Review and Ops

Three independently profile-loaded lenses reviewed the PR: Architect Alphonso, Debugger Debbie, and Reviewer Renata. Their confirmed findings were fixed and independently re-reviewed. The #3522 reproduction now fails both doctrine gates, including inside a file with a sanctioned exception.

Five completed Spec Kitty Ops are committed with artifact links:

- `01M1V10K60TK4H24JCB782BY8M`: CI formatting repair.
- `01M1V18DDE2PX1T7JAN8YMWS1Z`: broken/stale ownership references.
- `01M1V19GZXZ7BQME84DE612525`: runtime import syntax and independent baseline ratchets.
- `01M1V1A3686C32GJ0NJ884TTR6`: precise doctrine exceptions and import variants.
- `01M1V2E61TNTD505NYMSCB9TC1`: module-level conditional import gap.

[Review, concessions, red-first evidence, and verification](https://github.com/Priivacy-ai/spec-kitty/blob/776603e3b90d47842b002b208d70bd039064c67b/kitty-ops/evidence/pr-3888-adversarial-review.md).

## Validation

- `make test-fast`: **1,642 passed**.
- Initial touched/adjacent architecture and docs gates: **219 passed**.
[...]

PR #3983, opened 2026-09-07
fix(runtime): guard the emitter registry and factory dispatch; correct three artifacts #3921 left contradicting the tree

stijn-dejongh:
## A future event producer could crash `spec-kitty next`, or be silently swapped out, instead of failing safely

Follow-up to **#3921** (merged as `d6e8fe4238`). That PR opened a seam where a real runtime event producer will eventually be registered. Two edges of it were unguarded, and three artifacts shipped saying things that were no longer true of the tree they landed with.

Nothing changes for users today — no producer is registered yet. This is the seam being made safe before the first one arrives, and the record being made honest while the context is still loaded.

### What was wrong

**Registering a producer twice killed the command.** `register_runtime_emitter_factory` had no validation at all: last registrant silently won, with no diagnostic saying which. Guarding that is straightforward — but the guard has to compare the *right* thing. The seam's own documented registration form is `register_runtime_emitter_factory(MyProducer.for_mission)`, a classmethod, and every attribute access on a classmethod mints a fresh bound-method object. An object-identity check therefore never matches, so an ordinary double import — a reload, a module reachable under two paths, a harness re-running the import tail — would have been treated as a *conflicting* producer and raised at import time. **Before:** two import tails resolve by import order, last one silently wins. **After:** re-registering the same producer is a no-op; a genuinely different second producer is refused with a message naming both.

**A producer that failed to construct took the whole run down.** `runtime_emitter_for_mission` called the registered factory with no `try`, and both call sites are unguarded — one inside `_dn_bootstrap`, *before* the `get_or_start_run` try/except. A producer whose constructor hit a missing config or an unreachable credential lookup would have killed `spec-kitty next` with a traceback, contradicting the rule three lines above it in the same file: *"Nothing here may raise: emission is fire-and-forget instrumentation, never control flow."* It now logs at WARNING with the traceback and degrades to the null seam.

**"No live producer is wired" was a convention, not a gate.** That is the ADR's central scope boundary. It was verified by hand during the mission review, which does not survive the next PR. `tests/architectural/test_runtime_emitter_seam.py` now asserts it, with a non-vacuity self-check — and, importantly, with a **documented exit**: the boundary bound Mission B, which is closed. Wiring a producer is what E3 (#3929) is *for*, so the allowed set is data and the removal trigger is named in the failure message. A gate nobody knows how to legitimately retire is a trap.

### Three artifacts that contradicted the tree

- **`mission-review.md` recorded verdict `FAIL`** while the same file's own resolution note recorded *"Mission-caused Gate 2 failures after remediation: 0. Effective verdict: PASS WITH NOTES"*. All three blockers were resolved before the merge; the verdict line was never flipped.
- **The ADR's frontmatter said it "keeps `runtime.next.event_emitter` as the reserved E3 seam"** — a module #3921 deleted, as that ADR's own Decision (d) required. Anyone triaging decisions reads that line first and was pointed at a path that 404s.
- **`status.json` and `retrospective.yaml` carried an empty `mission_number`** where `meta.json` has `198`.

## Landing notes

Landed with a second-opinion squad over the integrated diff (correctness/fail-closed, design/authority, SSOT/duplication). It changed what ships:

**The `emitter.py` re-export narrowing was dropped.** An earlier commit on this branch reverted `_internal_runtime/emitter.py` to a two-name surface, arguing #3921 had widened a contract frozen by `contracts/internal_runtime_surface.md`. That premise is false — that document mentions `emitter` **zero times**. The document that *does* govern the shim is the mission's own `contracts/emitter-seam.md:3,21`, which explicitly requires all three names **in the shim**, traced through `plan.md:91,153,186` to `tasks.md` T004 (P0, delivered), and repeated in `quickstart.md:14`. The narrowing also reddened `test_no_dead_symbols::test_no_public_symbol_in_all_is_unimported` (green on main, red on the branch), because the shim held the only cross-module `from … import` of two seam names. It was a contract *narrowing* dressed as a revert, so the commit was dropped rather than patched around.

**The registry-guard precedent was half right.** `kernel.glossary_runner.register` is the precedent for the reject-on-conflict *policy*, but it registers a class, where identity is stable. For the key *comparison* the applicable precedent is `specify_cli.invocation.adapters._callable_key` (`__module__` + `__qualname__`), which exists for exactly this reason. The helper is duplicated locally rather than imported — runtime must not import `specify_cli` under the enforced layer chain.

### Architecture worth a future mission (not ticketed here)

- **The factory is not the single door.** Six run-start entry points decide which emitter the engine gets, and only one reaches the factory: `runtime_bridge.py:1552`. `answer_decision_via_runtime` calls the factory *after* the run has already started; `runtime_bridge_io.py:592` hardcodes `NullEmitter()` with no override parameter; three callers pass nothing and take the `emitter or NullEmitter()` default at `:643`; the engine has four more internal defaults. Unobservable today — every door yields the same null seam — but the day a producer is registered, five paths keep emitting nothing and it surfaces as missing telemetry, not a test failure. Threading the seam through three layers is not a landing fold.
- **Four module-level single-slot registries, three different conflict policies.** `kernel/glossary_runner` rejects; `specify_cli/core/adapters` and `specify_cli/invocation/adapters` are deliberately last-writer-wins (one behind a consent gate). Consolidating them onto one kernel primitive is legal under the layer chain but would change the behaviour of two live seams, one on a confidentiality egress path — a separate mission with an explicit policy decision, not a fold.
- **#3975** — `_wrap_with_decision_git_log` returns the *unwrapped* emitter when `DecisionGitLog` construction fails, so every decision event for that run is silently dropped. Same class as the bypasses #3921 fixed, but a deliberate fail-open: changing it is a behaviour decision with an operator-visible consequence.

## Tests

| Command | Result |
|---|---|
| `tests/next/ tests/runtime/ tests/specify_cli/events/` | **1377 passed, 1 skipped** |
| arch gates: `no_dead_symbols`, `no_dead_modules`, `golden_count_ban`, `ruff_format_enforcement`, `arch_shard_marker_completeness`, `no_legacy_terminology`, `runtime_emitter_seam` | **85 passed** |
| `ruff check .` | clean |
| `mypy src/runtime/next/_internal_runtime/events.py` | clean |
| `scripts/docs/check_docs_freshness.py --ci` | `exit=0 errors=0 warnings=0` |
| commitlint (all 8 commits, CI's own invocation) | clean |

**Red-first proof:** with the pre-fix `events.py` swapped back in, the new behaviour tests fail — conflicting-registrant, non-callable, raising-factory, and the classmethod re-registration regression (`RuntimeError: A different runtime emitter factory is already registered: <bound method ZeitgeistRuntimeEmitterProducer.for_mission …>`). All pass after.

**One pre-existing red folded (not caused here):** `test_golden_count_ban` was red on `main` at `d6e8fe4238` with `tests/architectural: 15 > 14`, reproduced in a clean worktree at that sha. The fifteenth site landed with #3888 one day after the baseline was frozen, and is a structural width assertion (`len(pair) == 2`), so it takes the gate's documented `cardinality-is-contract` escape hatch rather than a re-freeze.

https://claude.ai/code/session_01H6RbJ8UyfoGVrbjDi8B7z1


stijn-dejongh:
## Maintainer landing pass — claimed, triage verdict: **LAND**

Picking this up for a landing pass: adjudicate reds, run a second-opinion squad over the integrated diff, fold what it surfaces, then merge.

### Triage

Fresh PR (opened today) but **substantive** — it changes behaviour on an extension seam and reverts a frozen re-export surface — so it got the full compatibility triage rather than the trivial-chore path.

| Leg | Verdict | Evidence |
|---|---|---|
| **Superseded?** | No | Branch is **0 commits behind** `upstream/main`. Its base *is* current main (`d6e8fe4238` = #3921 itself). Nothing has merged since that could have superseded it. |
| **Compatible?** | Yes | Three-dot diff is clean — only the 8 intended files, no governance/charter smuggling from a stale stack. |
| **Still needed?** | Yes — all six premises reproduce on `main` | See below. |

Each claim re-verified against `d6e8fe4238`, not taken from the PR body:

- **`mission-review.md` self-contradicts on main.** Line 173 records `**FAIL**`; line 208 — already present in the same file — records *"Mission-caused Gate 2 failures after remediation: 0. Effective verdict: **PASS WITH NOTES**"*. The resolution note landed; the verdict line was never flipped.
- **`status.json` has `mission_number: ""`** where `meta.json` has `198`. Confirmed on main.
- **The ADR's frontmatter points at a module that does not exist.** It says *"Keeps `runtime.next.event_emitter` as the reserved E3 seam"*; `git ls-tree upstream/main` shows no `src/runtime/next/event_emitter.py` — #3921 deleted it, as that ADR's own Decision (d) required.
- **`emitter.py`'s frozen surface was widened without sanction.** It is Category 6 in `test_no_dead_modules.py:437`, frozen by `contracts/internal_runtime_surface.md` — and `grep` finds **zero** mentions of `emitter`, `runtime_emitter_for_mission`, or either registry function in that contract. The three added names had no production consumer: `runtime_bridge.py:195` imports the seam from `events.py` directly, and the only importer of the shim was one test.
- **The registry guard's precedent is real.** `src/kernel/glossary_runner.py:74-100` is a near line-for-line match — module-level slot, idempotent-on-same-object, `RuntimeError` on a different registrant, `TypeError` on a wrong type.
- **`register_runtime_emitter_factory(` appears in exactly one file under `src/`**, so the new gate's premise holds today.

### Reds

All 9 GitHub checks green. One local red, classified:

**`test_golden_count_ban::test_convert_sites_do_not_exceed_frozen_baseline` — `tests/architectural: 15 > 14`** → **bin 3, pre-existing main breakage.** Reproduced in a clean detached worktree at `d6e8fe4238` with an identical count and message, so it is not this PR's defect.

Folded anyway, per the runbook's campsite default. The fifteenth site is `test_ratchet_baselines.py:836`, `assert all(isinstance(pair, tuple) and len(pair) == 2 for pair in allowed)` — landed by **#3888** (`c0054153b9`, 2026-09-06) one day *after* the baseline was frozen at 14 (`735c67fabc`, 2026-09-05), so the freeze never counted it. `len(pair) == 2` is a structural width — the definition of a pair — not a runtime tally, so it takes the gate's own documented `# golden-count: cardinality-is-contract` escape hatch rather than a re-freeze to 15, which would have spent the ratchet on a site that never needed converting.

`ruff format --check` also wants to reformat that file — likewise pre-existing on main, and the file sits in the `[tool.ruff.format].exclude` debt ratchet (`pyproject.toml:1173`), so the enforcement gate does not count it. Left alone.

### Tests so far

| Command | Result |
|---|---|
| `tests/next/ tests/runtime/ tests/specify_cli/events/` | **1376 passed, 1 skipped** |
| `tests/next/test_internal_runtime_coverage.py tests/architectural/test_runtime_emitter_seam.py` | **155 passed** |
| `test_no_dead_modules` + `test_golden_count_ban` + `test_ruff_format_enforcement` + `test_arch_shard_marker_completeness` + `test_no_legacy_terminology` | **27 passed** after the fold (1 failed before) |

Second-opinion squad is running over the integrated diff now — correctness/fail-closed, design/authority, and SSOT/duplication lenses. Findings and any folds to follow.

stijn-dejongh:
## Remediation summary — squad findings folded, ready to merge

Three profile-loaded lenses (correctness/fail-closed, design/authority, SSOT/duplication) ran over the integrated diff on `opus`. They **converged** on one defect and independently surfaced two blockers I had missed. Verdict: **merge-ready**, with one commit dropped and four folds added.

### The convergence — all three lenses, same defect

`register_runtime_emitter_factory`'s idempotence guard compared registrants with `is`. The seam's own documented form (`events.py:203`, and `contracts/emitter-seam.md` "Registration contract") is `register_runtime_emitter_factory(MyProducer.for_mission)` — a **classmethod**, and every attribute access on one mints a fresh bound-method object. Reproduced directly:

```
MyProducer.for_mission is MyProducer.for_mission   ->  False
register(...); register(...)   ->  RuntimeError: A different runtime emitter factory is already registered
```

So the documented no-op raised, at import time, in a producer's import tail — the exact crash the *other half* of the same commit added `try/except` to prevent. Two halves of one seam, opposite philosophies. Fixed by keying on `__module__` + `__qualname__` per `specify_cli.invocation.adapters._callable_key`, the in-repo precedent for a *callable* slot. Policy unchanged.

### Blocker 1 — the PR reddened an arch gate (bin: PR defect)

`test_no_dead_symbols::test_no_public_symbol_in_all_is_unimported` — **red on branch, green on `main`** (I ran both). The `emitter.py` revert removed the only cross-module `from … import` of `register_runtime_emitter_factory` / `reset_runtime_emitter_factory`. I had not run this gate in my first pass; the squad did.

### Blocker 2 — the revert's premise was false, and this corrects my own triage above

My triage comment said `emitter.py`'s surface was "frozen by `contracts/internal_runtime_surface.md`" and widened without sanction. **That is wrong.** That document mentions `emitter` **zero times**. The document that governs the shim is this mission's own `contracts/emitter-seam.md`:

- `:3` — *"**Module**: `…/events.py` (re-exported by `_internal_runtime/emitter.py`)"*
- `:21` — *"`__all__` gains: … (**both in `events.py` and the `emitter.py` shim**)"*

traced through `plan.md:91,153,186` to `tasks.md` **T004** (P0, delivered), and repeated at `quickstart.md:14`. So the widening was specified, traced, and delivered — there was nothing un-sanctioned to revert. It was a contract *narrowing* presented as a revert, and it reddened a gate.

**Resolution: the commit was dropped**, not patched. The alternative (demote the two names from `events.py.__all__`, mirroring `glossary_runner:120-128`) also contradicts the contract *and* makes the seam less importable for the E3 producer that must call it. Dropping clears both blockers with zero contract churn. `emitter.py` is now byte-identical to `main`.

The underlying concern — a Category-6 "frozen" shim that grows once per PR — is real, but its premise was misattributed. It is noted in the PR body as future-mission architecture rather than settled by a silent narrowing here.

### Folds

| Commit | Why |
|---|---|
| `5aa1a17` `fix(landing)` | Key the registry guard on the callable, not the object (the convergence defect). Docstring no longer claims `glossary_runner` is the precedent for "this exact registry shape" — it names `glossary_runner` for the policy, `invocation/adapters` for the comparison. |
| `7c9877f` `test(landing)` | Give the no-live-producer gate a **documented exit**. The ADR's ADR-BLOCKED list bound Mission B, which is closed; wiring a producer is what #3929 is for, so the seam's designed use would have arrived as a red test with no way to retire it. Allowed set is now data, removal trigger named in the failure message. Also closes two evasions the substring missed: an aliased import tail (`import … as _reg` then `_reg(...)`) and `name (args)`. |
| `b3b6e2a` `docs(landing)` | Finish the `mission_number` reconciliation. The branch fixed `status.json` but left `retrospective.yaml` empty — so the PR that flips the review to PASS left its own DRIFT-2 item half-true. Both halves are resolvable; struck with a dated note. |
| `0b9ed7d` `docs(landing)` | Changelog entry under `[Unreleased]` / Fixed. |
| `f2448f0` `test(landing)` | The pre-existing golden-count red (see triage comment above). |

### Findings deliberately not folded

- **The factory is not the single door** — six run-start entry points, one reaches the factory. Unobservable today; threading it through three layers is a mission, not a fold. Enumerated in the PR body (the original body undercounted this — it said "four callers").
- **Four single-slot registries, three conflict policies** — consolidating would change two live seams' behaviour, one on a confidentiality egress path. Needs an explicit policy decision.
- **RISK-2** (`register` validates callability, not Protocol conformance) — already open item 4 in the mission review.
- **`getattr`-based registration and registration from outside `src/`** still pass the gate. Stated as a residual limit in the commit message rather than papered over.

### Evidence

| Check | Result |
|---|---|
| `tests/next/ tests/runtime/ tests/specify_cli/events/` | **1377 passed, 1 skipped** |
| 7 arch gates incl. `no_dead_symbols`, `no_dead_modules`, `golden_count_ban` | **85 passed** |
| `ruff check .` | clean |
| `mypy` (`events.py`) | clean |
| `check_docs_freshness.py --ci` | `exit=0 errors=0 warnings=0` |
| `test_no_legacy_terminology` | 10 passed |
| commitlint, all 8 commits (CI's own invocation) | clean |

History is 8 commits, one logical change each, commitlint-clean — suitable for this repo's rebase-merge.

Issue #3988, opened 2026-09-07
[bug] Inconsistency in 'getting-started' and 'your first mission' documentation results in creating first spec twice.


hdeweirdt:
## What happened

When going through the 'getting started' flow, in step 3 it tells you to `/spec-kitty.specify Build a tiny command-line task list app.`. 
Then, in 'what's next' it guides you to 'Your first mission', where step 1 is again "Step 1: Create the Specification", but this time the prompt is slightly different: `/spec-kitty.specify Build a task list app with add, complete, and delete actions.`. Running specify twice already results in unexpected behavior for a beginner, but the different prompts make it even more confusing. 

## What you expected

"Getting started" and "Your first mission" prompts should be the same, and the steps should be more aligned instead of repeating. 

## Steps to reproduce
1. Start with the 'getting started flow'
2. Continue on to 'your first mission'.

## Environment
- spec-kitty version (`spec-kitty --version`): 3.2.6
- OS: macOS
- AI agent (if relevant): /

LynnColeArt:
**Fixed in `spec-kitty-cli 3.2.6.1`** (PyPI, 2026-09-08 22:06Z). Getting Started and Your First Mission now use one canonical `specify` prompt; Your First Mission continues the mission Getting Started created instead of re-creating it; and Getting Started gained the three steps whose absence made the page fail on a fresh machine — `git init -b main`, an initial commit, and `git checkout -b my-first-mission` — plus `spec-commit` before `plan`. Verified end to end on the installed wheel. The CLI side of your report (two missions, one of them born broken) is fixed in the same release (#4033, #4035). Thank you for the report — it turned out to be the tip of five defects nobody had filed. Recommend closing; leaving that to you.

robertDouglass:
[controller] Moved to `status:triage`: this P0 had no status label, so no lane could see it. It sits on the MVP first-run path, because launch users start from the getting-started docs. Re-check it against the `4.0.0rc1` docs. Groomed under Robert's MVP grooming authorization.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0124FqjwyHcmXdsFi5roR3hQ

robertDouglass:
[implementer] Claiming this. Taking it as the onboarding-continuity defect it is: the two pages each open with "create the specification" and hand a beginner a different prompt for the same app, so following the documented path runs `/spec-kitty.specify` twice against one project. Fix and evidence to follow on a PR.

[...]

PR #4045, opened 2026-09-08
fix(specify): refuse mission create on an unborn HEAD instead of reporting success (#4033)

LynnColeArt:
## Issue

#4033 (P0, release blocker) — the beginner report is #3988.

Blocks the 3.2.6/3.2.7 patch Robert needs for the Darmstadt training tomorrow morning.

## Change

`spec-kitty specify` / `agent mission create` now refuse on an **unborn HEAD** — a repository with no commits — instead of reporting success and creating a mission that is broken on arrival.

Git cannot create a branch from an unborn HEAD. `ensure_coordination_branch` has a deliberate accommodation for a `target_branch` that does not resolve (`missions/_create.py:233`), written for synthetic and test contexts "where the operator pins a target like `2.x` that has no on-disk counterpart", which returns the canonical branch name without minting anything and defers the problem to "downstream consumers … when they actually try to use it". On a fresh `git init`, `main` does not resolve, so that accommodation fired on the most common real-world beginner state and wrote a `coordination_branch` declaration for a branch that was never created — exit 0, `{"result": "success"}`.

Three pieces:

1. `core/git_ops.py` — new `has_unborn_head()` predicate. It establishes repo-ness via `is_git_repo` **before** reading HEAD, because `git rev-parse --verify HEAD` fails identically for "no commits yet" and "not a repository"; without that, the create error would tell a non-repo user to run `git commit`. (My own test caught this — the first draft returned `True` for a non-repo.)
2. `core/mission_creation.py` — guard in `_create_mission_core_impl`, placed immediately after project-root resolution and **before any scaffold is written**, raising `MissionCreationError`. The existing CLI funnel already maps that to a clean exit 1, so no CLI change was needed.
3. Error text names the remedy: `git commit --allow-empty -m 'Initial commit'`.

**Scoped to coordination-bearing topologies only** (`topology_mints_coordination_branch`). `single_branch` and `lanes` mint no coordination branch, so an unborn HEAD costs them nothing; blocking them would be a gratuitous refusal and would break `test_coordinationless_create_persists_topology_so_2453_routing_is_not_cwd`. That test failing on my first, over-broad version is what surfaced the distinction — it is a good test.

### Correction to the issue's root cause

**#4033 attributes this to the protected-branch commit guard failing on the primary branch. That does not reproduce.** With one initial commit on `main`, the verbatim beginner flow works: both coordination branches mint, no `meta.json commit failed`, no orphan. The trigger is the unborn HEAD alone. Guarding the protected-branch path would not have fixed this. Full evidence on the issue thread.

Defect 1 in that issue (**no idempotency guard** — same-slug duplicate creates) is real, reproduces independently on `main`, and is **not** addressed here. It is a separate topic and gets its own PR, per the one-topic-per-PR rule.

## Tests run

All on this branch, `uv run --no-sync`, after `uv sync --frozen --all-extras`.

| Command | Result |
|---|---|
| `pytest tests/core/test_mission_creation_unborn_head.py` | **8 passed** (new) |
| `pytest tests/core/ tests/git_ops/` | **650 passed, 4 skipped** |
| `make test-fast` | **1782 passed, 2 skipped** |
| `pytest tests/architectural/test_ruff_format_exclude_ratchet.py` | **6 passed** |
| `ruff check` (3 changed files) | clean |
| `mypy` (2 changed source files) | clean, no issues |

Blast radius per CLAUDE.md: `tests/core/` is the owning subsystem for both changed modules, `tests/git_ops/` covers `git_ops` by grep, plus the `make test-fast` baseline. No pre-existing reds encountered.

New tests, and what each pins:

- `has_unborn_head` is `True` before the first commit, `False` after, and `False` for a non-repository.
- Coord create on an unborn HEAD raises, and the message is **actionable** (names `git commit`, not just the fault).
- A refused create writes **no scaffold** — `kitty-specs/` is empty afterward.
- `single_branch` and `lanes` are **not** blocked (parametrized).
- Coord create **succeeds after the first commit**, and the declared `coordination_branch` resolves to a ref that actually exists. The mint is deliberately not patched here, and the fixture uses `git init -b main` so the branch matches the patched `get_current_branch` — otherwise the mint takes its does-not-resolve skip path and the test would pass for the wrong reason.

That last one exists because a guard that blocks without a working recovery is worse than the bug.

### Manual verification (3.2.7rc1, git 2.43, Linux)

```
unborn HEAD, default coord   -> Error: "This repository has no commits yet…"  exit 1  scaffolds 0
unborn HEAD, single_branch   -> created                                        exit 0  scaffolds 1
after `git commit`, coord    -> created, branch minted, declaration resolves   exit 0
```

## Blast radius

- `src/specify_cli/core/git_ops.py` — additive only: one new function plus its `__all__` entry. No existing behavior touched. **Not reformatted** — this file is in the shrink-only `[tool.ruff.format]` exclude ratchet (`pyproject.toml:664`), so CI does not format-check it and formatting it would fail `test_ruff_format_exclude_ratchet.py`. Only the new test file was formatted.
- `src/specify_cli/core/mission_creation.py` — one guard on the create path. Every caller of `create_mission_core` inherits it: `specify`, `agent mission create`, and `tracker/origin.py`'s ticket-first flow. All three previously produced a broken mission in this state, so the refusal is the intended change for each.
- **Newly-refused case:** creating a coordination-bearing mission in a commitless repo. Anyone deliberately doing this — an automated harness that inits a bare repo and immediately creates a mission — now gets exit 1 where it got a silent, broken success. `single_branch`/`lanes` are the escape hatch and need no commit.
- No change to the protected-branch path, the rollback helper, `ensure_coordination_branch`, or the `--json` envelope shape.
- No migration needed. Missions already broken by this are unaffected by the guard and still need `doctor coordination --fix`.

Retired-surface scan: 0 hits

## Deferred

- **Idempotency guard** (#4033 defect 1) — separate PR.
- **Docs inconsistency** (#3988) — separate PR. Getting Started and Your First Mission both have the beginner create a task-list spec with slightly different prompts, so they produce two *different* slugs; no idempotency guard would catch that, and it needs a docs fix rather than a code one.
- **Two more silent-exit-0 paths found while reproducing**, same pattern, not in scope here and worth their own issue: `spec-kitty init <dir>` on an existing directory prints "Directory Conflict" and exits 0; `spec-kitty doctor coordination` exits 0 while reporting warnings about genuinely broken missions.
- **planning#1999** ("version of the first converged CLI release — 3.2.7 vs 3.3.0") is still open and this release needs it answered. `main` already reads `3.2.7rc1`. Flagging, not deciding.


LynnColeArt:
**Do not merge this in its current shape — it is superseded by the hotfix design.**

This PR carries my *first* fix for #4033: an unborn-HEAD refusal scoped to coordination-bearing topologies only. The second-opinion review on the hotfix (#4051) showed that scoping was **wrong** — every topology commits its scaffold, and that commit fails on an unborn ref, so flat topologies orphaned too. My test here only passed because `_commit_feature_file` was mocked.

The hotfix also moved to a strictly better shape: `preflight_commit` validating the commit destination *before* any scaffold is written, covering protected branch, head/target mismatch, and missing destination in one place.

The forward-port to `main` should be that design, not this one. Leaving this open as the anchor for that work rather than closing it, but flagging so nobody lands the narrower version thinking it is final. #4046 (docs) is unaffected and can proceed; #4049 (`--role`) is independent.

stijn-dejongh:
**Maintainer landing triage — closing as superseded by #4084.**

## BLUF

The unborn-HEAD guard this PR adds is **already on `main`**, landed by **#4084** ("Forward-port reviewed first-run recovery hotfix", merged 2026-09-08 — the same reviewed forward-port that consolidated #4046). All three pieces are present at `f2be03af48`:

- `has_unborn_head()` in `core/git_ops.py` — with the same `is_git_repo`-first ordering this PR describes as its design point
- the `MissionCreationError` guard in `core/mission_creation.py`
- the test file `tests/core/test_mission_creation_unborn_head.py`

That is why this PR shows `CONFLICTING`.

## Why this can't be rebased-and-landed

The two implementations **disagree on scope**, and `main`'s reviewed version is deliberately the broader one — so landing this would *regress* `main`, not no-op:

| | this PR | `main` (via #4084) |
|---|---|---|
| Scope | coordination-bearing topologies only — **SINGLE_BRANCH / LANES allowed** on an unborn HEAD | **all topologies blocked** — "every topology commits its scaffold" |
| Checkout inspected | `resolved_root` | `effective_root` (the ownership-claimed checkout; correct for linked checkouts) |
| Test | asserts branch-flat shapes are **not** blocked | blocks them |

Landing this branch would re-narrow the guard, re-permit creating a broken mission on an unborn HEAD for branch-flat topologies, and its `test_branch_flat_topologies_are_not_blocked` would contradict `main`'s behavior. The reviewers went broader on purpose, so there is nothing durable left to salvage here.

## Outcome

Closing as superseded by #4084. The P0 first-run defect from #4033 that this PR targeted is fixed on `main`. Thanks @LynnColeArt — the guard, including the careful `is_git_repo`-before-HEAD ordering, landed via #4084.

Note: #4033 stays open for its remaining, unrelated part (defect 1 — the same-slug idempotency guard), which this PR correctly did not address.

PR #4224, opened 2026-09-11
[#4216] Deliver unseen peer moments with verified own-session filtering

robertDouglass:
## Issue

Closes #4216. Integrates the relay-owned own-session contract in [Zeitgeist #295](https://github.com/spec-kitty/spec-kitty-zeitgeist/issues/295), using the publisher identity from #4217. Requires the matching relay release.

## Change

Agent moments default to `team` within the requested authorized repository, so unfamiliar missions and mission-less peer activity are visible. Explicit `mine` and `off` remain supported; global and repository restrictions intersect.

CLI and MCP watch/activity share one delivery policy: canonical event identities are checked before context/rate budgets; identity-only receipts acknowledge successful delivery, preserve admission scope across credential renewal, and reserve unacknowledged quota atomically. CLI acknowledges after output and flush; MCP clients explicitly acknowledge the returned receipt on a later request. Retained activity supports bounded catch-up and deliberate replay with honest gap/reset/truncation coverage.

Agent watch, status and retained activity request `filterOwn=true`. They forward the current cached presence/focus issuer references and require `X-Zeitgeist-Filter-Own: true` before reading response content. The relay resolves opaque identities and suppresses own frames before subscriber queueing. Separate command and reader processes reuse the logical session cache; another agent under the same human account remains visible. Missing/unbound issuer identities or an old relay without acknowledgment fail explicitly. CLI `watch --raw`/`status --raw` and strict MCP `filter_own=false` permit own activity. Each new read reloads issuer refs; intentional sequence jumps do not manufacture gaps. Receipt continuity uses the same canonical publisher session selector by default; explicit `--consumer` overrides only receipt context.

## Tests run

- `.venv/bin/pytest -q tests/zeitgeist_client` → **729 passed, 11 skipped** (full owning subsystem; followed by final changed-file runs below).
- `.venv/bin/pytest -q tests/zeitgeist_client/test_own_filter.py tests/zeitgeist_client/test_history.py tests/zeitgeist_client/test_agent_delivery.py` → **60 passed**.
- `.venv/bin/pytest -q tests/zeitgeist_client/test_mcp_stdio.py` → **49 passed**, including strict malformed values and omitted/true/false forwarding on all three tools.
- `.venv/bin/pytest -q tests/cli/commands/test_zeitgeist_command.py tests/cli/commands/test_zeitgeist_outbox_cli.py tests/cli/commands/test_zeitgeist_operability_cli.py` → **49 passed**.
[...]

Issue #4215, opened 2026-09-11
Make Zeitgeist status useful immediately and add activity queries by repo, person, project and time


robertDouglass:
## Live Work re-scope, 2026-09-14: Zeitgeist carries NOW, Git carries DONE

Robert's 2026-09-14 decision supersedes the durable-work scope below wherever they conflict. The original text stays underneath as history. Record: https://github.com/spec-kitty/spec-kitty-planning/issues/2269#issuecomment-5660729736

**Dependencies:** none after this re-scope.

**Scope now:** Make Zeitgeist status useful immediately (the cached initial state from zeitgeist#296 has shipped), and add activity queries by repository, person, project and time **limited to the relay's recent window**, with links into Git for anything older. No retained-history query service: saas#1633 is closed as superseded.

Out of scope now: a durable journal, projections or outbox; history beyond the relay's recent window; as-of replay or temporal pivots; a stored conversation archive; legacy import; a journal-backed publisher.

---

## Live Work programme amendment — 2026-09-13

Delivery child of https://github.com/spec-kitty/spec-kitty-planning/issues/2268. **Delivery owner: Software Factory. Priority: P1.** Preserve existing implementation, claims, assignees and PRs. This dated amendment supersedes older scope/dependency prose only as stated here; canonical mission authority, scoped authorization and truthful evidence remain required.

**Former dependencies (superseded 2026-09-14, see block above):** https://github.com/spec-kitty/spec-kitty-zeitgeist/issues/296, https://github.com/spec-kitty/spec-kitty-saas/issues/1633

Extend existing shared CLI/MCP query/watch ownership to durable work history by canonical mission, session, person, repository, project/programme and time, with clear semantics for each selector. Keep quiet initial presence/focus snapshot and race-safe snapshot/history->follow handoff. Work history is no longer limited to Pulse/TeamMoment retention: return all surviving authorized recorded history with paging, imports/gaps/recording boundaries and source timestamps. Never equate a quiet future-only stream with no work. Reuse common selectors, novelty and durable receipt contracts; current and history APIs must remain usable after72h and when no agents are online. Publisher and send/reply commands have separate implementation owners in the epic.

The epic contains the architecture, three equally complete People/Missions/Repositories views, all12 LW requirements and measurable release evidence. Earlier post-MVP/volatile-only/history/approval exclusions are historical context where this amendment replaces them. The original report below remains evidence of the problem, not a conflicting current scope gate.

---

## User outcome

[...]

Issue #3892, opened 2026-09-06
Epic: zeitgeist-client auth & token-lifecycle reliability (CLI stays authenticated, fails legibly)


stijn-dejongh:
**Canonical replacement for #3322** (closed — pre-convergence "hosted-sync auth" framing + a deleted domain-plan reference). Same invariant, re-anchored onto the surviving **client** auth surface.

## Context (client-repo inversion)

Post-Convergence (#3824), this repo is a **CLIENT**: the authoritative auth API is authored upstream in `spec-kitty/zeitgeist` (+ `spec-kitty/saas`). This epic owns only the **client-side** auth consumption, diagnostics, and legibility in this repo — not the auth protocol itself.

## Invariant

A logged-in CLI stays authenticated; machine/CI contexts have a supported non-interactive path; and refresh/exchange failures surface a **named, actionable** cause rather than an unexplained error.

## Surviving client-auth surface this epic governs

- `src/specify_cli/zeitgeist_client/credentials.py` — credential storage/lifecycle.
- `spec-kitty auth doctor --unstick-lock` — diagnose + recover a wedged auth lock.
- `SPEC_KITTY_SAAS_TOKEN` — non-interactive / machine token path.
- the `tracker` command gate — `SPEC_KITTY_ENABLE_SAAS_SYNC`.

## Work (carried forward from #3322, survivors only)

- [ ] #3279 — device authorization 401 after browser approval (3.2.5/3.2.6)
- [ ] #3277 — non-interactive machine authentication for CI/hosted
- [ ] #3233 — token refresh: handle `invalid_client` / `invalid_request` legibly
- [ ] #3486 — late `auth login` approval fails as a bare connection error; 5-minute window undisclosed
- [ ] #3178 — decision `widen` resolves destination team from env before the per-project auth file (FR-007)

Obsolete children of #3322 were already closed (#3621, #3723). The retired in-place sync transport that the old framing referenced is gone (#3824).

robertDouglass:
Referenced (not closed) by mission `team-kitty-launch-defaults-01M1XJ4Y` / #3980: new identifiers use Team Kitty vocabulary; the auth surface is consumed, not changed; non-interactive CI sign-in stays out of scope.

PR #4172, opened 2026-09-09
fix(ci): import datetime via the kernel.clock single door (FR-012(a))

LynnColeArt:
## Issue

closes #4165

Red main on the CLI repo since 17:03Z (run 34377537146): `scripts/ci/select_source_artifacts.py:8` imported stdlib `datetime` directly, violating FR-012(a)'s single-door rule. The violating commit (#4142, 16:16Z) was scripts-only, so the code-scoped architectural battery skipped it and the violation merged unseen; the first src-touching head after it (#4153) inherited the red. Every src-touching head since fails the same single assertion (verified on run 34385207968: one violation, this path:line — no second regression).

## Change

Two commits at head `da9135e9b`:

1. **Door swap** (96413805): `from datetime import datetime` → `from kernel.clock import datetime` (line 8). `kernel.clock` re-exports `datetime` for the sanctioned parse/format case; the script's only use is `fromisoformat` on timestamps it already receives. No `_exemptions` entry — the ratchet stays at zero.
2. **Pre-installation bootstrap** (da9135e9b, Lynn's independently-verified executable finding, folded verbatim): the aggregate workflow invokes this script bare (`python3 scripts/...`) *before* dependency install, so commit 1 alone broke the entry point (`ModuleNotFoundError: kernel` from an uninstalled checkout — the architectural battery parses imports and cannot see entry-point resolution). Repair: resolve the trusted checkout's `src/` from `__file__` (never caller cwd / fetched PR data) onto `sys.path`, then import from `kernel.clock` unchanged. Mirrors the existing `scripts/docs/glossary_linker.py` bootstrap shape.

## Tests run

- self-review: diff is the import swap + a sys.path bootstrap anchored to `__file__`, plus two witness tests; no other behavior touched, no clock reads added.
- correctness: import resolves in the scripts context AND at the bare pre-install entry point — pinned by the new subprocess witness running `-I -S` from an unrelated cwd with site-packages disabled.
- deletion safety: nothing deleted; import block rewritten in place, tests appended.
- security: bootstrap resolves only the trusted checkout's own `src/` (not caller cwd, not fetched PR data); no new inputs, no network surface change.
- design fidelity: kernel.clock is the documented single door (FR-012(a)/SC-001); ratchet stays at zero.
[...]

PR #4333, opened 2026-09-14
feat(zeitgeist): answer status from the relay's own live record (#4215 slice 1)

robertDouglass:
## Issue
Part of #4215 (its own first slice: "First deliver useful quiet-repo status"). #4215 stays open because its remaining acceptance criteria are untouched here: the bounded `activity` query by repo/person/project and time window, the SaaS observation API for longer windows, snapshot-seeded `watch`/MCP follow, and credential renewal on an expired capability.

## Change
`spec-kitty zeitgeist status` (and MCP `zeitgeist_status`) now answer from the relay's own record of who is live, instead of listening for two seconds and reporting whatever happened to be published.

- **`filtered_stream.FilteredStream.seed_from_snapshot()`** — one `GET /managed/snapshot?window_s=0` (zeitgeist#296, merged as zeitgeist#305), sharing the subscription's existing two-credential headers through a new `_headers()` helper. `window_s=0` asks for no history at all: this slice reports current state only. The response body is capped (`MAX_SNAPSHOT_BYTES`) before parsing.
- **`live_frame.StreamState.seed_snapshot()`** — applies a `SnapshotDocument`'s presence/focus entries through the SAME identity grammar, `_FOCUS_STATES` gate and 90 s ttl clamp a live frame goes through, because a snapshot is exactly as untrusted as a frame. It returns `False` for a document this module was not written to understand (wrong schema major, missing epoch, wrong shapes) and leaves state untouched, so a caller can fall back rather than report a half-parsed view. The document's `events` section is deliberately ignored — history is not liveness. Remaining lifetime comes from the server-derived `expires_in_s` measured against the local clock, so relay/client skew cannot stretch a ttl past the ceiling.
- **`PresenceView`/`FocusView` carry `observed_at`** (optional, additive) on both paths, and `subscription._serialize_snapshot` emits it, so a reader can say how old an observation is instead of implying it was made at read time.
- **`subscription.status()`** tries the snapshot first and reports `source: relay_snapshot`. A 404 (an older relay, or `self_hosted`, where the capability is absent) or an unparseable body degrades to the previous listen behaviour with `source: live_listen`, `fallback_reason` and `listened_s`. Any other HTTP status — 401/403 denial, 503 — propagates unchanged: a denied read is a fault, never an empty team.
- **CLI human output** names the source, dates each observation, and distinguishes "nobody is live on this relay right now — that is not proof nobody is working" from "nothing was published while this command listened". The MCP tool description carries the same distinction for an agent caller.
- **Test double**: `ManagedStreamDouble` now serves `/managed/snapshot` separately (404 by default, which is why every pre-existing test still exercises the listen path) and records `requested_paths`.

## Tests run
- `.venv/bin/python -m pytest tests/zeitgeist_client/ tests/cli/commands/test_zeitgeist_command.py -q` → 643 passed / 10 skipped
- `.venv/bin/python -m pytest tests/architectural/test_docs_cli_reference_parity.py tests/docs/test_build_cli_reference.py -q` → 49 passed / 0 failed (the regenerated reference is byte-identical: this command's section of `docs/api/cli-commands.md` sits outside the generated envelope)
- `.venv/bin/python -m pytest tests/architectural/test_no_dead_modules.py tests/architectural/test_no_dead_symbols.py tests/zeitgeist_client/test_grammar.py -q` → 56 passed / 0 failed
- `make test-fast` → 1796 passed / 5 skipped / 0 failed
- `ruff check` and `ruff format --check` over the changed files → clean. `mypy` on the four changed modules reports two pre-existing `no-any-return` errors in `_resolve_store_key` (untouched lines 108/120), nothing from this diff.

Self-review:
- correctness: clean — the fallback is chosen only on 404 or an unreadable document, every other status propagates, and the seeded state is evicted by the same clock the stream path already uses; the seeded ttl is the relay's own remaining lifetime, not a re-derived one
- deletion safety: clean — nothing removed; the listen path is intact and still the fallback, and every pre-#4215 test exercises it unchanged
- security: clean — snapshot content goes through the same grammar, ttl clamp and closed-enum gates as wire frames; the body is size-capped before parsing; no new credential handling (the existing two headers, from the existing store), no new persistence, and an auth denial is never reported as an empty team
- design fidelity: clean — the relay stays the authority on who is live; `status` remains current-state-only with history left to the `activity` slice; CLI and MCP share the one `subscription.status` surface
- tests: clean — new coverage for the quiet-repo case, both fallbacks, an auth denial, expired-entry suppression, prose-shaped identity sanitization, and CLI/MCP parity

## Blast radius
Discovery: `git grep -l -e "subscription\." -e "live_frame" -e "filtered_stream" -e "zeitgeist_status" -- 'tests/*.py' | sort`
Files: tests/architectural/test_no_dead_modules.py, tests/architectural/test_no_dead_symbols.py, tests/cli/commands/test_zeitgeist_command.py, tests/status/test_zeitgeist_decision_moment_docker_local.py, tests/zeitgeist_client/conftest.py, tests/zeitgeist_client/test_cli_zeitgeist_e2e.py, tests/zeitgeist_client/test_filtered_stream.py, tests/zeitgeist_client/test_grammar.py, tests/zeitgeist_client/test_live_frame.py, tests/zeitgeist_client/test_managed_relay_docker_local.py, tests/zeitgeist_client/test_mcp_stdio.py, tests/zeitgeist_client/test_moments.py, tests/zeitgeist_client/test_operability.py, tests/zeitgeist_client/test_outbox_approval_human_gesture.py, tests/zeitgeist_client/test_subscription.py
Every discovered file ran except the two `*_docker_local.py` suites, which need a real Docker relay this laptop does not run (they skip in this environment).

## Deferred
- `activity` queries by repo/person/project and time window, and the SaaS observation API for longer windows (#4215's second and third criteria) — the relay-side history read already exists in draft PR #4224's `history.py`, and this slice deliberately does not duplicate it.
- Snapshot-seeded `watch`/MCP follow through `follow=1`'s race-safe handoff, and credential renewal on an expired capability (coordinating with #3892).


LynnColeArt:
[controller] Partial-inheritance split of the `[ci] red @b60b97263`: the pack-manifest/DRG staleness failures are INHERITED from main (spec-kitty#4346, P1 — main-side, landed unevaluated through the cancel-cascade window; do NOT fix-round this diff for those). ONE failure is genuinely this PR's own: `src/specify_cli/cli/commands/zeitgeist.py:148: time.time()` violates the FR-012(b) raw wall-clock ban. Fix shape is the door-swap, never an `_exemptions` line: use the `kernel.clock` producer (e.g. `from kernel.clock import now_epoch` / `kernel.clock.now_epoch()`), per the #4165→#4172 precedent. This PR is taken-by-human; leaving the own-violation fix to the owner, recorded here so the merge lane reads the red correctly in the meantime.

robertDouglass:
[implementer] Fix round for the controller's partial-inheritance split, head `17ce17133`.

**The one own-violation is fixed.** `src/specify_cli/cli/commands/zeitgeist.py` read `time.time()` in the observation-age helper. Swapped to the `kernel.clock` door (`from kernel.clock import now_epoch`), per the #4165→#4172 precedent — no `_exemptions` entry. `time.monotonic()` in `watch` is untouched: it measures a duration across one call, not a wall-clock instant.

**Inherited failures not addressed here**, as instructed: the pack-manifest/DRG staleness legs are main-side (#4346), and the `architectural battery` job's `uv sync --frozen --all-extras` wedge hit at least five other PR heads in the same window. `origin/main` is merged into the branch so CI re-runs at a fresh head, which is what the merge lane asked for.

Tests at `17ce17133` (after the merge):
- `.venv/bin/python -m pytest tests/zeitgeist_client/ tests/cli/commands/test_zeitgeist_command.py tests/architectural/test_clock_import_ban.py -q` → 653 passed / 10 skipped / 0 failed
- `.venv/bin/python -m pytest tests/architectural/test_clock_import_ban.py tests/architectural/test_exemption_registry_ratchet.py tests/zeitgeist_client/test_cli_zeitgeist_e2e.py tests/cli/commands/test_zeitgeist_command.py -q` (pre-merge, the guard the finding names) → 48 passed / 2 skipped / 0 failed
- `ruff check` and `ruff format --check` on the changed file → clean

robertDouglass:
[implementer] Pass-2 MAJOR fixed at head `949e694a3`.

**Finding accepted as stated.** `observed_at` stopped being an internal ttl input the moment this slice serialized it and the CLI formatted it with `int()`; the ttl guards two hundred lines above never covered it, and `json.loads` accepts the bare `Infinity`/`-Infinity`/`NaN` tokens by default on every read path here. Both reported symptoms are real: an unhandled `OverflowError`/`ValueError` out of `zeitgeist status`, and a `--json` document carrying a token RFC 8259 has no value for.

**Fixed at the parse boundary, not the print site**, as the finding asked — one guard per numeric field, in `live_frame`:
- new `_finite(value) -> float | None`, documented against the same hazard `_clamp_ttl` already names;
- `parse_live_frame` refuses a frame whose own `emitted_at` is non-finite (it feeds every derived timestamp, so half-applying it is worse than dropping it);
- `_apply_presence` / `_apply_focus` fall back to the frame clock when a frame's `observed_at` is non-finite — this is the `live_listen` path the finding correctly flagged as reachable today;
- `_seed_presence_entry` / `_seed_focus_entry` keep the entry and its (separately guarded) lifetime but report `observed_at: null`, so one hostile field degrades one field, not the read;
- `_seed_remaining` now routes `expires_in_s` through the same helper instead of its own inline `isfinite`.

**Regression coverage** (each fails on the pre-fix code):
- `test_live_frame.py`: non-finite `emitted_at` refused; non-finite `observed_at` on presence and focus frames falls back to the frame clock; seeded entries drop the unusable observation time while keeping the lifetime; a non-finite `expires_in_s` drops the entry. All parametrized over `inf`/`-inf`/`nan`.
- `test_subscription.py::test_status_survives_a_non_finite_observed_at_from_the_relay`: end to end through the loopback double — the teammate is still reported, `observed_at` is `None`, and the serialized result is re-parsed with `parse_constant` raising, so any `Infinity`/`NaN` token in the document fails the test.

The two open MINORs (the two-phase timeout budget and the `observed_at` skew arithmetic) and the new MINORs on #4335 are untouched here — they are consolidated on that issue, not this head.

Tests at `949e694a3`:
- `.venv/bin/python -m pytest tests/zeitgeist_client/ tests/cli/commands/test_zeitgeist_command.py tests/architectural/test_clock_call_ban.py -q` → 682 passed / 10 skipped / 0 failed
- `ruff check` + `ruff format` on the changed modules and tests → clean

Squad passes are exhausted (2/2), so this goes to the merge lane on exact-head CI rather than a third pass.

robertDouglass:
## 2026-09-17 rebase + reconciliation

Rebased onto current `main` (227 commits of drift since this branch was cut on 2026-09-14) and fixed real conflicts, not just textual ones:

- Main gained own-session filtering (`filterOwn`/`X-Zeitgeist-Own-Sessions`, PR #4224/#4216) after this branch diverged — `watch()`'s `_headers()` refactor here didn't know about it. Fixed by sharing one `_filter_own_url()` helper across `watch()` and the new `seed_from_snapshot()`, so filterOwn applies to both relay routes.
- Main changed `_focus` from a single-key dict to a `(session_ref, focus_ref)` composite key. The snapshot-seeding code here used the old single-key scheme — fixed to match, avoiding a cross-session clobber bug on seeded state.
- **Found and fixed a real gap while reconciling, not just a conflict**: the relay's `/managed/snapshot` (`zeitgeist/managed.py`'s `managed_snapshot`) attaches the same `X-Zeitgeist-Filter-Own` ack header `/managed/stream` does, but `seed_from_snapshot()` never verified it — the exact silent-own-session-leak class `watch()` already guards against via `own_filter.require_ack()`. Added the same check, and fixed `ManagedStreamDouble` to actually emit that header on snapshot responses (it only emitted it on the SSE path before).
- Two of this PR's own e2e tests were stale against main's now-default `filter_own=True` CLI behavior (from #4224) — they didn't seed a cached publisher `session_ref`, so `status` now correctly refuses with "Own filtering requires cached publisher identity" instead of silently working around it. Fixed to match the existing pattern already used by `test_watch_end_to_end_over_a_real_loopback_double`.

Verified post-rebase:
- `tests/zeitgeist_client/` + `tests/cli/commands/test_zeitgeist_command.py`: 814 passed / 16 skipped
- `test_no_dead_modules`/`test_no_dead_symbols`/CLI-reference-parity: all passed
- `ruff check`/`ruff format --check`: clean (one pre-existing formatting drift in `conftest.py`, outside this PR's blast radius, left untouched)
- `mypy` on the three changed modules: 2 pre-existing `no-any-return` findings in `subscription.py` (unrelated lines, confirmed present before this rebase's changes too)
- Full fast tier (`pytest -m fast -n auto`): 31,464 passed / 120 skipped / 2 failed — both pre-existing and unrelated: `test_agent_decision_subgroup_has_canonical_visible_subcommands` (open issue #4669) and `test_help_output_never_mentions_feature_alias` (documented 30+min runtime, hit my local `--timeout=120`, not a real failure)

Also narrowed #4215's remaining scope in a comment: `spec-kitty zeitgeist activity` (#4224) already shipped the repo/time-window half of the issue's re-scoped acceptance criteria; only person/project selectors, snapshot-seeded watch/MCP follow, and expired-credential renewal remain outstanding after this PR.

robertDouglass:
[flex] Re-labelled squad:exhausted -> ready-for-squad @78214f8 on laptop-flex-robert.

The prior squad:exhausted verdict judged a stale head (949e694a, 2026-09-15) that predates
today's rebase onto current main (227 commits of drift) and the reconciliation work in it —
notably the own-filter ack-header fix in `seed_from_snapshot()` (a security-adjacent gap the
prior squad never saw) and the composite-key merge for `_focus`. Per PROGRAM.md §10a, a verdict
on a superseded head is stale by construction; dispatch.sh's squad_candidate_still_eligible()
won't re-dispatch a squad:exhausted PR on its own, so this re-label requests the fresh pass the
changed code warrants rather than letting the merge agent adjudicate against review that never
saw this diff.

Issue #3926, opened 2026-09-06
merge: coord-triple teardown deletes the branch before removing its worktree, stranding the coord worktree/branch and the merge-assigned mission_number


robertDouglass:
## Observed (spec-kitty 3.2.7rc1, mission `dead-port-disposition-01M1VRA2`, 2026-09-06)

`spec-kitty merge --mission dead-port-disposition-01M1VRA2` (coord topology, default retention → `teardown_coordination = True`) completed the squash and marked all WPs done, but its coordination teardown left the repo half-torn:

```
Deleted branch kitty/mission-dead-port-disposition-01M1VRA2-lane-d (was b9c4eb9c1).
  Cleaned up 4 lane branch(es)
error: cannot delete branch 'kitty/mission-dead-port-disposition-01M1VRA2' used by worktree at '.../.worktrees/dead-port-disposition-01M1VRA2-coord'
  Cleaned up mission/coordination branch + worktree
```

After the command returned:
- `.worktrees/dead-port-disposition-01M1VRA2-coord` still existed (clean, checked out on the coord branch).
- `kitty/mission-dead-port-disposition-01M1VRA2` still existed.
- `kitty-specs/<mission>/meta.json` on the target branch was flattened (`flattened: true`, no `coordination_branch`) but carried `mission_number: null`, while the coord branch's `meta.json` carried `mission_number: 198` (commit `chore(...): assign mission_number=198`). The dry-run had forecast `would_assign_mission_number: 198`.
- The console claimed success ("Cleaned up mission/coordination branch + worktree") despite the git error one line above.
- `spec-kitty doctor coordination --mission <m>` reported nothing (the primary `meta.json` no longer declares `coordination_branch`, so the stale coord worktree/branch are invisible to it); `doctor workspaces` reported no husks (the directory is a registered worktree, not a husk).

Additionally, the post-merge retrospective capture failed inside the merge process with `cannot import name 'runtime_emitter_for_mission' from 'runtime.next._internal_runtime.events'` — the long-lived merge process imported the bridge after the squash had landed the mission's own source changes, against a stale module view. Two `capture mission retrospective` commits were created but only appended a status event; no `retrospective.yaml` was written. `spec-kitty retrospect create` in a fresh process succeeded.

## Cause (from source)

`src/specify_cli/merge/executor.py::_teardown_coordination_triple` orders the triple as
[...]

Issue #3131, opened 2026-08-01
spec-kitty merge ignores mission retention constraints and deletes lane branches/worktrees


LynnColeArt:
## Summary

`spec-kitty merge` applies default branch/worktree cleanup even when the mission contract explicitly requires retaining them.

Observed on Spec Kitty 3.2.6 in `Priivacy-ai/spec-kitty-saas`, mission `first-paid-checkout-01KYX0JQ` (SK-ERR-022).

## Reproduction

The accepted mission contained constraint C-005: keep branches and worktrees after merge unless separately directed.

Running the normal merge nevertheless applied the effective default `--delete-branch --remove-worktree` behavior and deleted:

- all six lane worktrees;
- all six lane branches;
- the mission branch.

No preflight warned that cleanup contradicted the mission contract.

## Impact

The deleted tips remained reachable in this case and were reconstructed exactly, but this was destructive, unexpected cleanup after a successful merge. It also discarded evidence/workspaces the mission explicitly required keeping.

## Expected

Merge cleanup policy should be explicit and machine-readable. Either:

1. persist retention policy in mission metadata and make `merge` honor it; or
2. require an explicit preflight decision when mission constraints conflict with cleanup defaults.

A mission-level retention instruction must not be silently overridden by default flags.

robertDouglass:
Triage validation (2026-08-02): KEEP OPEN. This is a current issue with no completion or supersession evidence yet. Preserve it in the active queue, attach a focused reproduction/acceptance fixture and dependency/owner links, and update the issue with the delivery PR or resolution evidence when available.

stijn-dejongh:
Mission `merge-honor-retention-policy-01M1CA0E` addresses this P1 data-loss defect. Draft PR: https://github.com/Priivacy-ai/spec-kitty/pull/3826

The fix persists a machine-readable retention policy (`retain_branches`/`retain_worktrees`) in mission `meta.json` and makes `spec-kitty merge` honor it **fail-closed**: retain + operator-visible warning when the default would contradict the policy; deletion of a retaining mission requires an explicit, recorded operator override. Includes an issue-pinned red-first regression (RED on current main), topology-aware coupled coordination teardown, create-time opt-in flags, and the dry-run forecast reflecting the resolved decision. Operator to review + merge.

PR #4377, opened 2026-09-14
docs(tutorials): stop the onboarding path asking for a second specify (#3988)

robertDouglass:
## Issue
closes #3988

## Change
#3988 reported two defects in the onboarding path: the two tutorials hand a beginner **different prompts** for the same app, and both open with "Step 1: Create the Specification", so following the documented route runs `/spec-kitty.specify` twice against one project.

The prompt half is already fixed on `main` — both pages now say *Build a tiny command-line task list app with add, complete, and delete actions.* It was fixed incidentally by #4084's first-run recovery hotfix, which also added a "run `specify` once" note to Getting Started and a skip-to-Step-2 callout to Your First Mission; no commit references #3988, and the issue was never closed.

What was left is the structural half the reporter actually described: *"the steps should be more aligned instead of repeating."* A reader who skims headings still meets "Step 1: Create the Specification" as the first instruction of the second tutorial, and the callout that says not to do it sits *inside* that step.

- **`docs/guides/tutorials/your-first-mission.md`** — Step 1 is retitled **"Have a Mission to Work In"** and split into two labelled paths, so the heading no longer reads as an instruction to create anything: *"If you came from Getting Started — you already have the mission"* (confirm with `ls kitty-specs`, go to Step 2) and *"If you are starting fresh — create the mission now"* (the `specify` invocation, unchanged). The warning about ending up with two task-list missions is kept, now as the reason for the first path rather than as an interruption of a create-step.
- **`docs/guides/tutorials/getting-started.md`** — "What's Next?" now says where in the next tutorial the reader lands: its Step 1 is only a confirmation step, so start at Step 2 and do not run `/spec-kitty.specify` again. Previously it linked the tutorial with no entry point, which is what sent readers into Step 1.
- `updated:` bumped on both pages; `docs/development/3-2-docs-retrieval-index.yaml` regenerated for the changed heading (generated file, rebuilt with its own generator rather than hand-edited).

No CLI behaviour changes: the duplicate-`specify` hazard itself is already guarded in code (each run creates a separate mission by design), so this is the documentation half of the fix.

## Tests run
- `.venv/bin/python -m pytest tests/docs/ -q` → 1665 passed / 8 skipped / 0 failed (the owning suite for both changed pages, including the structural lint, frontmatter contract, link/anchor checks and the docs-index freshness guard)
- `PYTHONPATH=. .venv/bin/python scripts/docs/docs_index.py --write` → `generated=800 committed=800 drift=False`

Self-review:
[...]

PR #4378, opened 2026-09-14
fix(merge): tear the coord worktree down before deleting its branch (#3926)

robertDouglass:
## Issue
Part of #3926. The issue stays open because it lists five remedies and this PR lands the two that constitute the P0 itself (ordering and fail-loud); the mission_number bake on the target branch, the `doctor coordination` husk detection, and running the retrospective capture in a fresh interpreter are separate surfaces and are left with the issue.

## Change
A coord-topology mission's coordination branch is **checked out in its own coord worktree** when merge tears the triple down. `_teardown_coordination_triple` ran `branch delete → flatten → worktree`, and `git branch -D` refuses while the branch is checked out:

```
error: cannot delete branch 'kitty/mission-<slug>' used by worktree at '.../.worktrees/<slug>-coord'
  Cleaned up mission/coordination branch + worktree
```

`run_command(..., check_return=False)` swallowed that, so the flatten ran anyway and the phase printed its success line over the git error. The reporter was left with a live coord branch, a live coord worktree, and a `meta.json` that no longer declares either — invisible to `doctor coordination`, which keys on the marker this run had just removed. That is #3131 INV-2's all-or-nothing coupling broken in the inverted #3086 direction.

- **`_teardown_coordination_triple` now orders the triple worktree → branch → flatten.** Removing the worktree first releases the checkout, so the delete succeeds; the flatten only runs behind a branch that is genuinely gone. The worktree leg keeps its persist-before-destroy property (it goes through the same `teardown_coordination_topology` seam, which persists the retrospective before destroying).
- **`_delete_mission_branch` returns whether the branch is gone** — deleted now, or already absent — by re-verifying the ref after `git branch -D` instead of assuming the swallowed command worked. Its non-coord caller ignores the value, so that path is unchanged.
- **A leg that does not come down raises `CoordinationTeardownError`** instead of flattening and reporting success, and `spec-kitty merge` reports it as `coordination teardown incomplete: …` with exit 1. The message names the surviving branch and points at `git worktree list` plus `merge --resume`, because the merge itself has landed by then — only the teardown is partial.

## Tests run
- `.venv/bin/python -m pytest tests/merge/ -q` → 701 passed / 0 failed (the owning subsystem)
- `make test-fast` → 1796 passed / 5 skipped / 0 failed
- `.venv/bin/python -m pytest tests/merge/test_coord_teardown_order_3926.py tests/merge/test_coordination_flatten_on_branch_delete.py -q` → 5 passed (the new guard beside the #3086 guard it must not regress)
- Red-first: with the pre-fix ordering restored, both new tests fail, and the captured stderr is the reporter's exact error — `cannot delete branch 'kitty/mission-coord-teardown-order-01m1vra2' used by worktree at '…/.worktrees/coord-teardown-order-01m1vra2-coord'`
- `ruff check` on the three changed files → clean; `ruff format --check .` → 1792 files already formatted

Self-review:
- correctness: clean — the new fixture creates a real `git worktree` on the coord branch, which is the state that made the delete fail; the branch-gone assertion is re-read from git after the delete rather than inferred from the command's return
- deletion safety: clean — nothing removed; the non-coord path (`delete_branch` → flatten, `remove_worktree` → worktree) keeps its exact previous behaviour and the #3086 flatten guard still passes unchanged
- security: clean — no new inputs, credentials, or external calls; the change is ordering plus an error type
- design fidelity: clean — INV-2's all-or-nothing coupling is what the raise enforces; the worktree leg still runs through the shared `teardown_coordination_topology` seam, so persist-before-destroy is preserved
- tests: clean — both new cases are red on the old ordering and green on the new, and `tests/merge/` ran in full

## Blast radius
Discovery: `git grep -l -e "_teardown_coordination_triple" -e "_delete_mission_branch" -e "teardown_coordination" -- 'tests/*.py' | sort`
Files: tests/core/test_retention_resolver.py, tests/integration/test_issue_3716_discard_transactional_close_guard.py, tests/integration/test_merge_cluster_coord_read.py, tests/integration/test_mission_close_discard_coord_teardown.py, tests/merge/test_coord_teardown_order_3926.py, tests/merge/test_coordination_flatten_on_branch_delete.py, tests/merge/test_executor_coverage.py, tests/specify_cli/coordination/test_teardown_seam_persist_before_destroy.py, tests/specify_cli/coordination/test_teardown_single_seam_routing.py, tests/specify_cli/missions/test_handle_equivalence_matrix.py
All ten ran together: `.venv/bin/python -m pytest <the ten files above> -q` → 205 passed / 0 failed. They cover the other callers of the same seam — `mission close --discard`'s coord teardown and the persist-before-destroy routing guards — which is what the reordering could plausibly have disturbed.

## Deferred
Left on #3926, which stays open for them:
- the merge-assigned `mission_number` landing on the target-branch `meta.json` (the flatten rewrites the primary copy without it);
- `doctor coordination` detecting a registered `.worktrees/<slug>-coord` whose mission is already flattened, with a `--fix`;
- running the post-merge retrospective capture in a subprocess so a mission that changes `src/runtime/next/` cannot break its own capture with a stale module view.


robertDouglass:
[implementer] The `[ci] red` on this head is **inherited from main**, not caused by this diff.

The only failing job is `module-tests (execution_context shard 2/3)`, and its only failing test is:

```
FAILED tests/mission_runtime/test_consolidated_resolution.py::test_e2_status_state_still_probes_coordination_and_is_unaffected
E   Failed: DID NOT RAISE <class 'mission_runtime.resolution.ActionContextError'>
1 failed, 1076 passed, 2 skipped
```

That is **#4358** (`ci:red`, `from:controller`, `stage:verified`) — a completed-mission read shortcut letting write placement skip the coordination-branch check, which #4358's own body notes "blocks source-changing PRs that integrate current main". The fix is already open and squad-passed as **#4381** (`src/mission_runtime/resolution.py`, head `e2827290a`); this branch touches none of those files.

Corroboration that it is not this diff: the identical shard, with the identical single failure, is red on #4377 (a **docs-only** change: two tutorial pages plus the regenerated retrieval index) and on #4378 (merge-executor teardown ordering). Neither can reach `mission_runtime.resolution`.

No fix round is warranted here. Once #4381 lands, a re-run at a fresh head should clear it — recorded so the merge lane reads this red correctly in the meantime.

robertDouglass:
[implementer] Unblocked: **#4405 merged** (`cede0e381`), so the inherited red this PR was carrying is gone from `main`.

Verified locally at that tip before refreshing — the three guards that were failing now pass:

```
pytest tests/architectural/test_clock_call_ban.py \
       tests/architectural/test_clock_import_ban.py \
       tests/architectural/test_no_tmp_paths_in_tests.py -q
35 passed
```

`origin/main` is merged into this branch and pushed (head `083fdf060`), so CI re-runs at a head that contains the fix. This is the step the earlier diagnosis said was the whole remedy — no change to this PR's own diff was needed, and none was made.

PR #4390, opened 2026-09-14
feat(next): a blocked decision names where each guard looked (#3883)

### LINKED ISSUE #3883: bug: `next --result success` blocks at review on artifacts its own query mode reports present — no operator recovery
## Summary

At a mission's `review` step, `spec-kitty next --result success` reports required artifacts as
missing while `spec-kitty next` in query mode — against the identical repository state —
reports no guard failures at all. The advance path and the query path disagree, and the advance
path is the one that is wrong.

There is no recovery available to an operator: everything inspectable says the mission is
ready, and the one command that advances it says otherwise without naming an actionable
difference.

## Reproduction

Observed on a `qa`-type mission provided by an org doctrine pack, at the `review` step, with:

- both required artifacts (`qa-traceability.yaml`, `test-report.md`) authored at the paths the
  step's own expected-artifacts manifest specifies,
- both committed,
- all work packages driven to `approved`.

Three probes against that same tree:

```
$ spec-kitty next --result success
blocked — guard_failures: ["qa-traceability.yaml", "test-report.md"]

$ spec-kitty next            # query mode
guard_failures: []
approved_wps: 2
```

and calling the functions `next` itself uses, directly in-process:

```python
_check_cli_guards(...)          # -> no failures
gather_artifact_presence(...)   # -> both artifacts present
```

Re-running `next --result success` returns the identical `blocked` result indefinitely.

## Impact

This is reachable from the documented happy path of a shipped doctrine pack — a consumer
following that pack's own install runbook and driving a mission arrives here at step six of
seven. It blocks mission completion with no documented or discoverable way forward.

Characterising the symptom required reading `runtime/next/runtime_bridge.py`; an operator
without source access has no way to distinguish this from their own error.

## Hypothesis (unconfirmed)

A stale or cross-worktree read in `committed_authority` / `coord`-topology status resolution,
such that the advance path evaluates artifact presence against a different view of the tree
than the query path does.

## Suggested fix

1. Have the advance path resolve state through the same path the query path uses, so the two
   cannot disagree.
2. Independently of the root cause: make a `blocked` result name **the specific file and the
   specific absolute-or-repo-relative path it looked at**. A guard that reports "missing"
   without saying where it looked is not diagnosable without a source read, and that is what
   turned this from a bug into a dead end.

## Provenance

Found during an end-to-end validation run of an org doctrine pack. Reproduced repeatedly.

Measured against source commit `db5014ab5b83bb72a0da5a2b68794592cdbb2147`, which **predates**
the recent merge that brought the rebuild branch into this repository. A re-test against
current `main` is in progress and I will comment with the result — if it no longer reproduces,
I will close this with that evidence rather than leaving it open.

A related finding from the same run, filed separately if it survives re-test: the `implement`
step advances with zero work packages ever claimed, so its completion gate cannot fail.

### LINKED ISSUE #4395: squad MINORs on PR #4390 (guard failure paths render/query coverage)
PR: https://github.com/spec-kitty/spec-kitty/pull/4390

Squad MINOR findings from the pass-1 review of PR #4390 (`ddd25ed396579e7f0d130b69506d9667c543d7e3`). The checklist below references code that does not exist on `main` until that PR merges.

- [ ] src/specify_cli/cli/commands/next_cmd.py:938 — the new user-facing render (`- <name>: looked for <path>` inside `_print_standard_human`) has no test: reverting that hunk leaves the entire suite green, and the PR's self-review line "five focused cases covering the render" covers the decision payload, not the print — add one CLI-level test asserting the indented `looked for` lines appear under `Guards pending:` (and are absent when the map is empty)
- [ ] src/runtime/next/runtime_bridge.py:2590 — query mode (`spec-kitty next` with no `--result`) still prints `Guards pending:` without the `looked for` lines, because `query_current_state` is not wrapped the way `decide_next` is; the dead end in #3883 was the advance path so this is defensible scope, but the asymmetry means the same missing-artifact report is diagnosable on one path and not the other — consider populating `guard_failure_paths` on the query decision too, or documenting the difference

robertDouglass:
## What changes for you

When `spec-kitty next` blocks at a decision on unmet guards, it now **names the path each guard looked at**, so you can see *why* it blocked and recover.

**Before:** `next --result success` could block at review showing only `Guards pending:` and the guard names — even when a guard was looking for an artifact its own query mode reported present. There was no way to see which path the guard checked, and no operator recovery (the #3883 P0 dead-end).

**After:** each pending guard renders an indented `- <name>: looked for <path>` line, so you can reconcile the exact path the guard evaluated against what actually exists on disk.

## Closes

Closes #3883.
Closes #4395.

## Tests

Decision-payload coverage in `tests/next/test_guard_failure_paths_3883.py` + `tests/specify_cli/next/test_runtime_bridge_composition.py`; plus a landing-added CLI-level test pinning the human render itself (see landing notes). Landing verification (rebased tip): those suites → 50 passed; whole-repo `ruff check .` + `ruff format --check .` clean; format-exclude ratchet green.

- **Rebased** onto current `main` (193 behind). The branch carried a `Merge origin/main` commit — **flattened to linear history**; one Robert Douglass commit (authorship preserved).
- **CI red was the format-exclude ratchet** (same class as recent landings): the PR reformatted `src/runtime/next/decision.py` and `src/specify_cli/cli/commands/next_cmd.py` clean, but both were still listed in `[tool.ruff.format].exclude`, so the shrink-only ratchet failed. Folded into the origin commit: dropped both now-clean entries; ratchet green at the commit's own boundary, whole-repo formatter still clean. `router gate` was downstream.
- **Folded a MAJOR the fresh second-opinion pass caught** (`fix(next):` commit): the original fix keyed the `looked for` paths off the guard-failure **message strings**, so for the four built-in mission families (software-dev/research/documentation/plan — the common path) it fabricated garbage like `- Not all work packages are approved or done: looked for kitty-specs/<slug>/Not all work packages are approved or done` — a path never checked, *worse* than pre-fix and the opposite of #3883's intent. It only worked for the unregistered-custom-family bare-tag case (#3883's reported scenario). The rework resolves paths from **real artifact tags** (a candidate is trusted only when it is in the family's own `_presence_filenames_for` vocabulary), keys `guard_failure_paths` by the real tag, and iterates those tags in the render — so a non-artifact failure (WP-status, source-count, docs-glob) structurally cannot grow a `looked for` line, and a genuine missing artifact shows its true repo-relative path. Red-first: a realistic software-dev CLI block proves the fabricated path before / the single correct `tasks.md` path after. The masking bare-tag unit tests were rewritten to realistic message-shape data.
- **Folded squad follow-up #4395** (autocloses):
  - **The `looked for` human render had no test** — reverting the render hunk left the whole suite green. Added a red-first CLI-level test pinning the indented `- <name>: looked for <path>` lines under `Guards pending:` (present when the map is populated, absent when empty).
  - **Query-mode asymmetry, documented (not expanded):** `spec-kitty next` with no `--result` (query mode) still prints `Guards pending:` *without* the `looked for` lines, because only the advance path (`decide_next`) populates `guard_failure_paths`. #3883's dead-end was the advance path, so this is deliberate scope — extending the paths to the query decision is a separate enhancement, not folded here.

## Re-review + one documented latent note

The MAJOR fix was **re-reviewed adversarially** (a P0 whose first attempt was wrong earns a fresh pass): the prior fabrication is confirmed gone across all live families, and — checked in both directions — no *reachable* under-report exists (a guard tag that can pass is by construction in `_presence_filenames_for`, the same set the fix gates on).

One **latent, unreachable, pre-existing** MINOR was recorded with an in-code comment rather than a ticket: the not-yet-wired `plan` family's `_evaluate_plan_guards` names `spec.md`, which is absent from `_presence_filenames_for("plan")`, so if plan-family guards were ever dispatched live the path for a missing `spec.md` would silently drop. Not reachable today (all guards evaluate under software-dev via `_check_cli_guards`); a comment at `guard_failure_artifact_paths` now records the "path-resolution family must equal guard-evaluation family" invariant so it can't be silently reintroduced.


stijn-dejongh:
🛬 **Maintainer landing pass — claimed**

Assigning to @stijn-dejongh. Viability triage → **LAND (rebase + folds)**. P0 fix for #3883 (`next --result success` blocked at review on artifacts its own query mode reported present — no operator recovery). Not superseded.

Going in:
- **193 behind with a `Merge origin/main` commit** — will flatten to linear history.
- **CI reds:** `architectural battery` + `router gate` (downstream) — will diagnose on the rebased tip; the same pair was pure stale-base on the last landing, but I'll confirm rather than assume.
- **Folding squad follow-up #4395:** add the missing CLI-level test for the new `- <name>: looked for <path>` render (reverting the render currently leaves the suite green — a real coverage gap). The query-mode asymmetry (query still prints `Guards pending:` without the `looked for` lines) is defensible scope — #3883's dead-end was the advance path — so I'll document it rather than expand into query mode. PR will autoclose #3883 and #4395.

Rebasing, diagnosing the arch red, re-verifying `tests/next/` + the runtime gates, running a correctness+design squad, then handing off.

stijn-dejongh:
✅ **Landing pass complete — ready for the operator to merge. (Not auto-merged: this one had a MAJOR, so it earns your eyes.)**

**Headline:** the fresh second-opinion pass caught a **MAJOR** in this P0 fix, which I reworked and re-reviewed clean.
- **The MAJOR:** the original fix keyed the `looked for` paths off guard-failure *message strings*, so for the four built-in mission families (the common path) it fabricated garbage paths like `- Not all work packages are approved or done: looked for kitty-specs/<slug>/Not all work packages are approved or done` — worse than pre-fix, the opposite of #3883's intent. It only worked for the unregistered-custom-family case (#3883's own reported scenario).
- **The rework** (`fix(next):` commit) resolves paths from **real artifact tags** — a candidate is trusted only when it's in the family's `_presence_filenames_for` vocabulary; `guard_failure_paths` is keyed by real tag; the render iterates those tags. Non-artifact failures structurally can't grow a path line; a genuine missing artifact shows its true repo-relative path. **Red-first** proven through a realistic software-dev CLI block.
- **Re-reviewed adversarially:** prior MAJOR confirmed gone, and — checked both directions — no *reachable* under-report (a guard tag that can pass is by construction in the gated vocabulary). One latent/unreachable/pre-existing MINOR (plan-family guard-tag ∉ plan vocab) recorded with an **in-code invariant comment**, not a ticket.

**Also folded:**
- The format-exclude ratchet red (dropped the now-clean `decision.py` + `next_cmd.py` entries, into the origin commit).
- #4395: the untested render (added a red-first CLI test) + documented the query-mode asymmetry (deliberate scope). Autocloses **#3883** + **#4395**.

**Gates green:** `tests/next/` 545, `tests/specify_cli/next/` 245, guard tests 8, realistic CLI test passes; whole-repo `ruff check`/`ruff format --check .` clean; format-exclude ratchet green; mypy clean on the 3 touched modules; terminology 90.

Four commits: Robert's feat (authorship preserved) + my `test(landing)` + `docs(landing)` + `fix(next):` rework. I don't merge this one — handing off green for you.

Issue #4334, opened 2026-09-14
CI efficiency (P0): per-PR SonarCloud job re-runs the fast tier instead of reusing CI Modules shard coverage


stijn-dejongh:
## BLUF

The per-PR **`sonarcloud`** job in `.github/workflows/ci-quality.yml` **re-runs the fast test tier under coverage** (`make test-fast` + `--cov`) to produce its own `coverage.xml`, instead of reusing the coverage the **`CI Modules`** shard matrix already produced on the same PR. Result: **the fast tier executes twice per pull request**, on every PR, forever. Marked **P0 on inefficiency / throughput / cost** grounds (runner-minutes and PR wall-clock, multiplied across every PR), not release-correctness.

## Verified (on `main`, files + lines)

1. **Re-runs tests.** `ci-quality.yml` `sonarcloud` job → step *"Run fast tier under coverage"* runs `make test-fast` with `PYTEST_ADDOPTS=--cov=… --cov-report=xml:out/reports/coverage/coverage.xml`, then hands that file to the SonarSource scan. The job has **no `needs:`** on any test-producing job.
2. **The matrix already ran them.** `module-tests.yml` (matrixed by `ci-modules.yml`) runs `pytest <shard> --cov=<target> --cov-report=xml:coverage-<tier>-<mod>-shard<slug>.xml` and uploads the artefacts. The fast tier is one of the matrixed tiers → duplicate execution.
3. **Reuse is already implemented — twice.** `ci-aggregate.yml` (the diff-cover PR gate) and `sonar.yml` (nightly) both download the `CI Modules` shard artefacts and reconcile `coverage-*.xml` by basename (dedup + stale-fallback) into `sonar.python.coverage.reportPaths`; Sonar merges server-side. The per-PR job simply doesn't use this path.
4. **Ordering.** `ci-quality` and `ci-modules` are separate workflows, both on `pull_request`, running in parallel; `ci-aggregate` runs on `workflow_run:[CI Modules] completed`. The Sonar job reports *before* the matrix finishes, so it structurally cannot reuse output it never waits for — which is why it re-runs.

## Secondary defect (correctness, folds into the same fix)

[...]

PR #4394, opened 2026-09-14
chore(op): reconcile kitty-ops tracking (48 records) + correct two stale ci-aggregate comments

stijn-dejongh:
## What

Two unrelated-but-small maintenance changes, plus one finding that killed a third.

**1. `chore(op)`: track the 48 untracked Op records.**
This checkout carried a local, uncommitted `kitty-ops/` line in `.git/info/exclude`, so every Op record minted since it was added stayed untracked while 297 older records remained tracked from history. Removing that line reconciles the split in the direction upstream already takes (`ef24cfc9be`, `c89d116d01`). All 48 are complete and closed before landing — 48 `started` / 48 `completed` / 10 `commit_link`, zero malformed lines — so no open Op joins the tracked set. `kitty-ops/ops-index.jsonl` stays ignored; that rule lives in the committed `.gitignore` as a deliberate local performance cache.

**2. `docs(ci)`: correct two stale cross-references in `ci-aggregate.yml` comments.**
Both named `ci-quality.yml` for something that is not there.

- The `SonarCloud Scan` pin comment claimed the SHA matches "the SAME commit ci-quality.yml and sonar.yml vet", then called that "both surfaces". Since #4334 moved the per-change report here, `ci-quality.yml` holds no Sonar reference at all (`grep -c "SonarSource\|sonarcloud\|SONAR"` → 0). The two surfaces vetting this pin are *this file* and `sonar.yml`.
- The `sonar-pr-verdict` comment attributed `NON_BLOCKING_ALLOWLIST` to `ci-quality.yml`. It is a Python dict in `tests/architectural/test_suite_jobs_gate_blocking.py:83`, asserted *against* `ci-quality.yml`'s needs graph — about that workflow, not in it.

Comments only; no job, step, pin or gate semantics change.

## What was dropped, and why it matters

This PR originally carried a third commit: a `doctor ops --close-stale` sweep closing six stale `python-pedro` Ops (ages 175–194h) by appending one `completed` event to each record. **It is removed, because it violates this repository's own archive-immutability gate.**

`kitty-ops/` is one of the four immutable archive roots in `tests/architectural/test_archive_root_byte_identical.py:48`. The rule is exact:

```python
if path == LIFECYCLE_LOG_RELATIVE_PATH.as_posix() and status == "M":
    continue          # kitty-ops/lifecycle.jsonl is the ONLY modifiable file
...
if status == "A" and path not in baseline:
    continue          # new records are fine -- this is why upstream can add them
violations.append(f"{status}\t{path}: ordinary archive history changed")
```

So **adding** Op records is permitted; **modifying** a pre-existing one is not. Verified by running the gate against the correct base:

| branch content | gate verdict |
|---|---|
| 48 additions + 6 modifications | `FAILED` — 6 violations, one per modified record |
| 48 additions only (this PR) | `3 passed` |

*(A local run can mislead here: this checkout's `origin` is a stale fork, so the gate's `merge-base(HEAD, origin/main)` base blamed hundreds of unrelated upstream `kitty-specs/` changes as well. Re-running with the base repointed at the real `main` isolated the six genuine violations.)*

## Follow-up worth filing

`spec-kitty doctor ops --close-stale` mutates files the repository declares immutable, and then **tries to auto-commit them** via `safe_commit`. Here it refused only incidentally — the checkout happened to be on protected `main`. On any topic branch the sweep would have committed the change and produced a red architectural gate, with nothing in the command's output hinting why.

The six stale Ops are consequently **still open**, and no purely-committable fix exists for them: their closure is a runtime mutation of frozen history. Either the sweep should record closure on the append-only `kitty-ops/lifecycle.jsonl` spine, or `kitty-ops/` per-record files should stop being treated as committable archive. That is a design call for the owner, not something to paper over here.

## Tests run

| command | result |
|---|---|
| `pytest tests/release/ tests/ci/` | **478 passed, 6 skipped** |
| `pytest tests/architectural/{test_suite_jobs_gate_blocking,test_ci_quality_path_filters,test_gate_selection_authority,test_module_shard_registry}.py` | **41 passed** |
| `pytest tests/architectural/test_archive_root_byte_identical.py` (base repointed to real `main`) | **3 passed** |
| `ruff format --check .` | 1800 files already formatted |
| `python -c "yaml.safe_load(ci-aggregate.yml)"` | parses |

https://claude.ai/code/session_012JcKJXAyut9jdvBxmr9whP


stijn-dejongh:
The dropped `doctor ops --close-stale` commit described above is now filed as #4397 (P1, milestone 4.0.0) — the sweep mutates immutable `kitty-ops/` archive records and auto-commits them, so its closure cannot land while the archive gate stands. This PR keeps only the gate-legal additions.

Issue #4207, opened 2026-09-11
main-push CI Router red since 14:12Z: tests (cli) + tests (e2e) hit their timeout-minutes ceilings; router gate fail-closes on timeout-cancelled jobs


LynnColeArt:
## What happened

Since 2026-09-10T14:12Z, **every `CI Router` run on push to main fails red at the `router gate` aggregator** — not because any test failed, but because the `tests (cli)` shard times out at its `timeout-minutes: 15` ceiling and `tests (e2e)` at its `timeout-minutes: 20` ceiling. The router gate (`if: always() && !cancelled()`) fail-closes on any blocking job reporting `failure` OR `cancelled`, and GitHub reports a timeout-killed job as `cancelled` — so both timeouts cascade into a red main.

Main's head `d96d0209` (docs-only, 18:13:29Z) carries the red `router gate` check-run: https://github.com/spec-kitty/spec-kitty/actions/runs/34513000441/job/102998979575

```
blocking job(s) did not succeed: {'tests-cli': 'cancelled', 'tests-e2e': 'cancelled'}
```

## Evidence chain

**Timings — the shard suites now blow their ceilings on every main push (all jobs that actually completed were green; only the two timeouts fail):**

| Run (push to main) | head | tests (cli) | tests (e2e) | router gate |
|---|---|---|---|---|
| run 34480303549, 13:03Z | 55b441bb (green batch) | **14m32s / 15m** ✅ | 8m32s / 20m ✅ | success |
| run 34487497837, 14:12Z | 6abcf2f1 (rc1 bump batch) | **15m00s ⛔ timeout** | **20m00s ⛔ timeout** | failure |
| run 34456455531, 18:13Z | d96d0209 (docs-only) | **15m17s ⛔ timeout** | **20m17s ⛔ timeout** | failure |

The e2e suite's runtime **more than doubled** (8m32s → 20m17s); cli moved from 14m32s borderline-green over the 15m ceiling. No test in either suite failed; the jobs were killed mid-`pytest` (log tail: `Terminate orphan process: (pytest)`).

[...]

PR #4399, opened 2026-09-14
fix(charter): first activation preserves what was effective (#4253)

### LINKED ISSUE #4253: charter activate silently narrows previously effective authority when an activation key is absent
Activating a new project directive can silently deactivate existing safety authority. Before an activation key exists, the activation-aware resolver exposes the full available doctrine corpus. `CharterPackManager.activate()` initializes that key from `default.yaml`, which contains fewer artifacts. The new directive resolves successfully while unrelated existing directives and procedures disappear.

This occurred during SaaS#1711/#1713 recovery governance in Tracker, Zeitgeist, Alerts, Kuma and Planning/Factory. Complete pre-source Git archives, including declared local pack roots, were compared with the actual activation-aware resolver. Each lost DIRECTIVE_041–051, DISCIPLINED_REFACTORING, RECONCILE_CHANGE_SCOPE_TENSIONS, USE_C4_MODEL_TECHNIQUES and USE_MUTATION_TESTING_TO_VALIDATE_TEST_QUALITY; eleven existing procedures disappeared, including adversarial-squad-deployment and red-main-release-discipline. New-policy-only tests passed and missed the loss. Preservation tests then failed red before canonical explicit restoration.

Source seam: `src/charter/activation/pack_manager.py::CharterPackManager.activate` gets `default_ids` from `_load_default_pack()` and passes them to `plan_activation`; `src/charter/activation/resolver.py::DoctrineService` defines an absent per-kind key as unrestricted. Reconcile these semantics in the canonical activation path. Do not add another copied default corpus, patch installed runtime, or change an explicitly restricted activation set into unrestricted access.

Acceptance:

- Through the actual charter CLI, start with an absent activation key and at least one available artifact outside default.yaml; activate a new project artifact. Every previously effective artifact must remain effective unless a separately explicit removal is requested.
- Cover directives and procedures, built-in plus declared local org/project layers, and actual ID/stem normalization.
- Preserve explicit-empty and explicit-restricted activation semantics; unknown IDs fail without mutation.
- Test current consumer output after activation, not just presence in YAML or successful CLI exit. Missing-source/invalid-pack behavior must not be weakened.

Recovery source patches restore the exact observed prior sets and pin their preservation; they do not fix the general CLI transition. Source/installed/deployed evidence remain separate. Related: SaaS1713 and CLI4250; this is distinct from owned-checkout routing and global asset bootstrap CLI4017.

Existing-issue check: #3385 (closed) concerns a phantom org-pack directory scan; #3810 (closed) concerns two missing profile defaults. Those are related symptoms, but neither covers this still-reproduced directive/procedure first-activation transition. Searches included activation/defaults/preservation, narrowing, and absent-key terms across all issue states.

Concrete reproducible source pair: Tracker PR33, before `84aa65defbfad291c9c649b679003238e815d32b`, after `5805d2bd8e1523c6c8e2670771674f9ed9c5d818`, canonical charter resolver from CLI `d7cced454`. In a scratch Tracker clone with these objects fetched and the CLI environment available:

```python
import io, subprocess, tarfile, tempfile
from pathlib import Path
from charter.activation.doctrine_service_builder import build_activation_aware_doctrine_service

commits = ["84aa65defbfad291c9c649b679003238e815d32b",
           "5805d2bd8e1523c6c8e2670771674f9ed9c5d818"]
snapshots = []
for commit in commits:
    with tempfile.TemporaryDirectory() as directory:
        blob = subprocess.check_output(["git", "archive", commit])
        with tarfile.open(fileobj=io.BytesIO(blob)) as archive:
            archive.extractall(directory, filter="data")
        service = build_activation_aware_doctrine_service(Path(directory))
        snapshots.append({kind: set(getattr(service, kind))
                          for kind in ("directives", "procedures")})
for kind in snapshots[0]:
    print(kind, sorted(snapshots[0][kind] - snapshots[1][kind]))
```

Observed:15 missing directives and11 missing procedures. Full archives preserve declared local source roots; a `.kittify`-only copy is insufficient where an org pack is repo-local. Ten fixed-prior-set resolver assertions across the five affected projects failed red before restoration. The original new-policy tests remained green, demonstrating the missing preservation dimension.

Factory intake: `type:fix`, `from:review`, `status:triage`, `priority:P1` under current PROGRAM §10. This is filed for Software Factory triage; it is not ready-for-squad, implementer-complete, installed or deployed.

### LINKED ISSUE #4411: Squad MINORs from PR #4399 (charter: first activation preserves effective set)
Consolidated MINOR findings from the adversarial-squad review of PR: https://github.com/spec-kitty/spec-kitty/pull/4399 (head `1443fa3c`).

The checklist below references code that does not exist on `main` until that PR merges.

- [ ] src/charter/activation/pack_manager.py:631 — stale docstring: `activate()` still says the engine "materializes the default pack into the plan when the kind has no explicit activation set (FR-021)"; the PR changed the behavior to materializing the available corpus — reword the docstring (same for `plan_activation`'s FR-021 paragraph and `default_ids` parameter doc in src/charter/activation/activation_engine.py:209 and ~:243).
- [ ] src/charter/activation/activation_engine.py:268 — `or list(default_ids)` is unreachable: the validation above guarantees `artifact_id ∈ available_ids`, so the None-state branch never sees an empty `available_ids` and the documented "fallback for a caller that supplies no availability at all" cannot be observed (such a caller raises `UnknownActivationIdError` first) — drop the fallback or restructure so the documented case is real.

robertDouglass:
## What changes for you

`spec-kitty charter activate <id>` **no longer silently deactivates everything else that was in force** when the kind has no explicit activation set yet.

**Before:** if a kind's `config.yaml` activation key was absent, activating a single artifact treated the absent key as "start from the default pack" and wrote a bare one-element list. That flipped the kind from "everything effective is available" to "only this one" — silently dropping every effective artifact outside `default.yaml`. Observed in the field: **15 directives and 11 procedures** (including `adversarial-squad-deployment`) lost on one unrelated activation.

**After:** an absent key is correctly read as the *unrestricted* state. The first activation materializes everything currently in force — the resolver's effective set, which spans the full declared org chain and is already in the keyspace the resolver filters on — and appends the new artifact. Nothing in force is deactivated. Unknown-id validation still runs first (a bad id writes nothing), and an already-restricted key behaves exactly as before.

## Closes

Closes #4253.
Closes #4411.

## Tests

`tests/charter/test_activation_preserves_effective_4253.py` drives the absent-key case and asserts no effective artifact is dropped; the engine + pack-manager suites cover the restricted path, validation order, and the org-chain/keyspace materialization. Landing verification (rebased tip): `tests/charter/` green; ruff clean; whole-repo `ruff format --check .` clean; exclude-ratchet green; `mypy --strict` clean on the changed engine.

- **Rebased** onto current `main` (176 behind). The branch carried a `Merge origin/main` commit; **flattened to linear history** (this repo rebase-merges) — three Robert Douglass commits (authorship preserved) + a `fix(landing)` mypy fold.
- **#4411 (squad MINORs) was already folded by the branch** — Robert's follow-up commit corrected the stale `activate()`/`plan_activation` FR-021 docstrings ("materializes what is in force," not the default pack) and removed the unreachable `or list(default_ids)` fallback, both with explicit "(#4399 squad MINOR)" references. Verified; the PR autocloses #4411.
- **`fix(landing)` fold:** the new code's `list(current)` in the else branch tripped `mypy --strict` (it narrowed the None-state through a `was_unrestricted` alias mypy can't follow); branch on `current is None` directly so mypy narrows — runtime-identical. `activation_engine.py` now passes `mypy --strict`.
- **Second-opinion squad** (2 opus lenses — correctness + design — the fleet squad had *exhausted*): _see the landing comment for verdicts._
- **Pre-existing, not this PR:** `pack_manager.py:344` (`return root / ORG_MISSION_TYPES_SUBDIR`) trips single-file `mypy --strict` with a spurious "Returning Any" — the line pre-dates this PR (bot commit `52c8e4a3`) and mypy is not gated on this file in CI (`make typecheck` covers two unrelated files). Left as-is.

## Reviewer notes chased down (no fold needed)

The design lens flagged a possible "activation-list keyspace" gap; on tracing it to fold a fix, it does **not** manifest as a correctness bug, so nothing is folded:

- **The appended target and the preserved set both resolve correctly.** `config.yaml` activation lists are *stems* by design (R-011-D), and the compiler resolves every one to its canonical id on read via `_resolve_config_activated_ids` → `resolve_artifact_urn` (`compiler.py:120`). The stem≠id divergence is already exercised and passing (`test_an_id_that_diverges_from_its_stem_survives_activation`), so writing the target in stem spelling is correct, not a filtered-out entry. What remains is at most cosmetic — the preserved-set union can carry a stem and its canonical id as two entries that resolve to the same artifact — not worth changing here, since "unifying" the write keyspace would risk regressing the working, tested normalization.
- **The `effective_ids or available_ids` fallback rests on an implicit seam contract.** In the real caller `available_ids` is the pack-#1-truncated set, so the fallback is safe only because `activate()` always supplies a rich `effective_ids`; unreachable today, but fragile if a future caller passes truncated-available with empty-effective. And `_chain_complete_available`'s best-effort `except` degrades to pack-#1-only availability if the pack registry is unreadable — by design (preservation must never fail an activation; debug-logged), but it is the one path that degrades back toward the #4253 loss for org packs 2+. Both are latent-hardening only; left as-is.


robertDouglass:
[implementer] Fix round at head `97d26ea46`. Both MAJORs accepted — you reproduced them by execution and they are exactly the class #4253 exists to kill, in two configurations the issue's acceptance names and my first pass did not cover.

**MAJOR 1 (truncated org chain).** Right: `resolve_layer_roots` stops at pack #1, and that truncation is a deliberate, documented back-compat contract (`_layer_roots.resolve_org_root_chain`'s docstring spells out why the dict value must stay a single `Path`). So I did not touch it — preservation now scans **once per declared org root** and unions the results (`_chain_complete_available`), leaving the truncated map intact for `charter list --all-layers` and every other consumer typed `dict[str, Path]`.

**MAJOR 2 (keyspace mismatch).** Also right, and the more interesting half. Your suggested remedy — compute the preserved set from the resolver's effective corpus — turned out to need one refinement: doing that wholesale rewrites `activated_directives` into `DIRECTIVE_NNN`, because the directive getter's keyspace is canonical ids while every project authors stems. So the reconciliation is **per resolver pattern**: chain-complete stems for everything, plus the resolver's own keys for the kinds it filters by raw `id:` (Pattern B/C — procedures, agent profiles, …). Directives keep the authored spelling, since `_normalize_directive_id` already bridges it; a divergent-id org procedure is now written in the spelling the filter actually compares.

**The missing coverage you flagged is added**, both red against the pre-round code and green after — the other six unchanged in both directions:
- `test_artifacts_in_the_second_org_pack_survive_activation` — two declared packs, an artifact in pack #2 still effective after activating an unrelated built-in;
- `test_an_id_that_diverges_from_its_stem_survives_activation` — org procedure `id: acme-release-proc` in file `acme-release.procedure.yaml`, still effective afterwards.

**Both MINORs are fixed here rather than deferred**, since they are one-liners in the same code:
- `default_ids` is **removed** from `plan_activation` — you were right that it was unreachable, and once the preserved set became the effective corpus there was nothing left for it to do. `promote_activations` keeps its own (different planner, different contract).
- The stale docstrings on `activate()` and the engine's FR-021 paragraph now describe what ships.

Evidence at `97d26ea46`: `tests/charter/` + `tests/doctrine/` → **6023 passed / 32 skipped / 0 failed**; `make test-fast` → 1796 passed / 5 skipped; `ruff check` + `ruff format --check .` clean (1792 files).

One thing I deliberately did **not** widen into: a divergent-id artifact still cannot be made effective by an explicit `charter activate` (the CLI writes the stem, the filter wants the id). That is a pre-existing defect on the activation-write path rather than the preservation path this PR fixes, and it deserves its own issue and evidence instead of riding along — say the word if you would rather see it here.

robertDouglass:
[implementer] Fix round at head `06fe4cfd7`. Both pass-2 MAJORs accepted and fixed; neither is a rebuttal.

**MAJOR 1 (dead format-exclude entry) — fixed, and it was mine.** Your diagnosis is exact: the pass-1 fix round incidentally reformatted `activation_engine.py`, so entry 317 stopped suppressing anything and the shrink-only ratchet correctly called it dead weight. I took the direction you recommended — keep the file formatted, drop the entry — rather than reverting the four reflows.

- `ruff format --check src/charter/activation/activation_engine.py` → `1 file already formatted` (the entry really was inert)
- `tests/architectural/test_ruff_format_exclude_ratchet.py` → **6 passed** (was `1 failed, 5 passed`)
- **`tests/architectural/` in full → 2459 collected, exit 0, all passed.** This is the run I should have done two rounds ago.

Your closing note is the part worth keeping. This escaped twice because CLAUDE.md's blast-radius rule names pytest.ini / pyproject.toml / conftest / markers / packaging as the `tests/architectural/` triggers, and *incidentally reformatting an exclude-listed source file* is a cross-cutting change none of those words catch. I've recorded that as a rule gap in `## Deferred` rather than quietly fixing my habit — the rule is what the next implementer will read. Agreed it is a gap in the rule, not blast-radius dishonesty, and I'd rather it be written down than absolved.

**MAJOR 2 (stale body) — fixed.** You were right on all three counts, and the body is rewritten rather than patched. Every number re-measured at this head:

| | body said | actually |
|---|---|---|
| charter+doctrine | 6021 / 32 | **6023 passed / 32 skipped** |
| acceptance file | 6 passed | **8 passed** |
| `default_ids` | "stays as the fallback" | **removed from `plan_activation`** |

`## Change` now describes what ships — `effective_ids` as the materialization source, and the `_chain_complete_available` / `_effective_ids_for_kind` seam that was 118 undescribed lines of the diff. The red-first bullet now covers both rounds. `make test-fast` → 1796 passed / 5 skipped; `ruff check .` clean; `ruff format --check .` → 1803 files already formatted. Blast radius re-derived at this head: **7 files**, not the nine the stale body listed — the merge with `main` converged two of them.

**#4411 item 1 finished.** You were right that it was only half-done: `plan_activation`'s FR-021 summary still said "*default_ids* are materialized into the plan" for a parameter that no longer exists. It now describes the effective-corpus behaviour and says why the narrowing it prevents matters.

The two new MINORs stay on #4411. The divergent-id **write** path stays deferred with your earlier non-objection — still a pre-existing defect on a different path, still deserving its own evidence.

Note this PR carries `squad:exhausted`, so per PROGRAM §7 this is the last implementer round; it goes to the merge agent from here.

LynnColeArt:
Class follow-up filed from the 2026-09-15 night-watch tick that dispositioned this wedge:

- **spec-kitty#4429** (P1, status:ready) — the concrete defect: charter module-tests shard 4/5 hit its 30-min `timeout-minutes` ceiling on this head (killed 03:26Z; green baseline 24–26 min). Fix shape: ceiling → observed + headroom, same as the #4207 cli/e2e fix but for module-tests shards.
- **spec-kitty#4430** (status:triage) — the structural class: a terminal-`cancelled` PR-head run posts **no** `[ci]` verdict at all; the thread keeps a stale `[ci] running` and nothing notices. This PR is the canonical instance.

Both cite the 04:33Z sk-merge note above as the disposition of record. No board action on this PR changes — it stays taken-by-human per that note.

robertDouglass:
Exact-head handoff audit at `06fe4cfd719f36321c508f63c46ee72873187dda`: the existing final fix commit resolves both pass-2 MAJORs (dead Ruff exclusion and stale PR evidence), completes #4411 item 1, and its five-section body passes `validate-pr-body.py --verify-discovery`. With both squad passes exhausted, I removed the stale `squad:majors` label and restored `ready-for-squad` for merge-agent adjudication. No third squad pass is requested. The remaining blocker is solely the missing terminal CI verdict after Actions cancelled charter shard 4/5; as author/fix-round participant I cannot independently post CI on this head.

stijn-dejongh:
🛬 **Maintainer landing pass — claimed**

Assigning to @stijn-dejongh. Viability triage → **LAND (rebase + folds)**. Substantive charter fix for #4253 (first activation silently narrows previously-effective authority); not superseded — the fix isn't on `main`.

Notes going in:
- **History carries a merge commit** (`origin/main` merged into the branch) — this repo rebase-merges, so I'll flatten it to linear history on the rebase.
- **Folding squad follow-up #4411** (2 MINORs): the stale `activate()`/`plan_activation` FR-021 docstrings (they still say "materializes the default pack" after the behavior changed to the available corpus), and the unreachable `or list(default_ids)` fallback in `activation_engine.py`. PR will autoclose #4253 and #4411.
- Robert already dropped the now-dead `activation_engine.py` ruff-format exclusion in-branch — good.

Rebasing (176 behind), re-verifying `tests/charter/` + `tests/doctrine/`, ruff/format/exclude-ratchet + mypy, running a design+correctness squad on the activation authority, then handing off green.

stijn-dejongh:
✅ **Landing pass complete — ready for the operator to merge.**

**Done:**
- Rebased onto current `main` (176 behind). The branch carried a `Merge origin/main` commit — **flattened to linear history** (rebase-merge repo): three Robert Douglass commits (authorship preserved) + two `*(landing)` folds.
- **#4411 (squad MINORs) was already folded by the branch** — Robert's follow-up corrected the stale FR-021 docstrings and removed the unreachable `or list(default_ids)` fallback (both with explicit "(#4399 squad MINOR)" refs). Verified; PR autocloses **#4253 and #4411**.
- **`fix(landing)` mypy fold:** the new code's `list(current)` tripped `mypy --strict` (narrowed the None-state via a `was_unrestricted` alias); branch on `current is None` directly — runtime-identical, `activation_engine.py` now passes `mypy --strict`.
- **`docs(landing)` changelog** entry (Fixed): the silent authority-loss bug.
- **Second-opinion squad (2 opus lenses — the fleet squad had *exhausted*): no MAJORs.**
  - correctness (reviewer-renata): preservation correct, fail-closed intact, restricted path unchanged, under-supply backstopped by `_chain_complete_available`, tests drive the real bug through the consumer surface.
  - design (architect-alphonso): sound materialization source, pure planner, dead code genuinely removed. Approves landing.

**Verification:** PR's 4 charter test files → 108 passed + activation re-run 19; ruff clean; whole-repo format clean; exclude-ratchet green; `mypy --strict` clean on the engine; terminology 90.
- ⚠️ **The full `tests/charter/` local sweep was resource-killed twice (exit 144 / OOM)** — a local environment limit, not a test failure (the direct test files are green). CI runs the full charter module as the authority.

**Future-mission architecture (noted in the PR body, not folded — both lenses converged):**
- **Activation-list keyspace unification** — lists mix filename-stems and declared-ids; a Pattern-B/C target whose id≠stem is appended in stem spelling the resolver filters out (pre-existing, doesn't manifest for directives). A genuine follow-up mission (collapse to one keyspace + remap the appended target).
- The `effective_ids or available_ids` fallback rests on an implicit seam contract (real caller's `available_ids` is pack-#1-truncated); a best-effort degraded path can re-narrow under an unreadable registry. Worth a seam assertion later.

I don't merge — handing off green-modulo the honest local-OOM caveat. Ping me after merge and I'll verify + clean up.

PR #4436, opened 2026-09-15
docs(status): refresh stale _first_non_printable_attr docstring

### LINKED ISSUE #4318: zeitgeist_bridge._first_non_printable_attr docstring is stale: names the 8.2.0 pin and a missing encode-side printability check that 9.1.6 now has
## Observed

`src/specify_cli/status/zeitgeist_bridge.py`'s `_first_non_printable_attr` docstring (around line 306 on main) still reads: "``to_zeitgeist_attrs`` (spec-kitty-events, pinned 8.2.0) applies no printability check of its own — only ``from_zeitgeist_attrs`` does (``_reject_control_characters``, Priivacy-ai/spec-kitty-events#64, fix in flight as events#104)."

Both claims are stale against the actual pin:
- `uv.lock` resolves `spec-kitty-events` **9.1.6** (`pyproject.toml:80` pins `>=9,<10`), not 8.2.0.
- In 9.1.6, `to_zeitgeist_attrs` **does** reject non-printable characters on encode (`zeitgeist_attrs.py`, the `_reject_control_characters` loop over emitted attrs), so the bridge's producer-side pre-check is now redundant belt-and-braces rather than the only guard, and the "fix in flight as events#104" note is resolved.

Found while implementing #3954 in the same file; left untouched there to keep that PR minimal.

## Proposed fix

Update the docstring to name 9.1.6 and describe the encode-side guard as present (pre-check kept as defense-in-depth), or — if the squad prefers — delete the now-redundant `_first_non_printable_attr` pre-check with its tests and rely on the codec's own encode-side rejection.

## Severity

S4 — comment/docstring drift; no behavior impact.

qwist1233-cpu:
Closes #4318

## What
Docstring-only update to `_first_non_printable_attr` in `src/specify_cli/status/zeitgeist_bridge.py`. It still described `to_zeitgeist_attrs` as "pinned 8.2.0" with no encode-side printability check.

## Why it was stale
- `pyproject.toml` pins `spec-kitty-events>=9,<10`.
- In `spec-kitty-events` v9.1.6, `to_zeitgeist_attrs` loops over the emitted attrs and calls `_reject_control_characters` (`zeitgeist_attrs.py` L925-926), raising `ZeitgeistAttrsControlCharacterError`, a `ZeitgeistAttrsError` subclass.
- `_broadcast_moment` already wraps `to_zeitgeist_attrs` in `except ZeitgeistAttrsError` (logs and returns), so control characters are dropped at the encode call. The pre-check now only matters if the codec guard regresses.

The new docstring says that and keeps the pre-check as defense-in-depth (option 1 in the issue). No behavior change; I left deleting the pre-check and its tests to the maintainers' call.

## Verification
- Read `to_zeitgeist_attrs` / `_reject_control_characters` at tag `v9.1.6` of spec-kitty-events.
- File parses (`ast.parse`); diff touches only the docstring.

AI disclosure: I used an AI assistant (Claude) to trace the codec source and draft the wording, then checked each claim against the v9.1.6 tag before submitting.


qwist1233-cpu:
Closing this one — #4318 was resolved on main in a64cc5f/0547a87, and that wording covers more than mine did (it also explains why the pre-check never runs on the current pin). Nothing left for this PR to add. Thanks!

PR #4468, opened 2026-09-15
ci: name the shard that actually failed, so one red shard stops looking like forty (#4420)

### LINKED ISSUE #4420: ci: fail-fast cancellations report as failures, so one red shard looks like forty
## What happens

`ci-modules.yml` sets `fail-fast: true` in `pr` mode (`ci-modules.yml:121`). That is a deliberate, defensible choice — it stops ~40 shards as soon as one goes red and saves a lot of CI minutes.

The cost lands somewhere else: **a cancelled shard is reported as a failure** on the PR. `gh pr checks` prints `fail` for it, and the PR's check list renders it the same way as a real red. So one genuinely failing shard presents as forty failing shards, and nothing on the surface distinguishes the one that actually failed from the thirty-nine that were merely stopped.

Observed on PR #4417 (`CI Modules` run `34911646190`):

- `module-tests (release shard 1/1)` — **genuinely failed**, one test, 2m53s
- every other shard — **`conclusion: cancelled`**, all landing at a near-identical ~3m10s
- `gh pr checks 4417` — prints `fail` for all forty

The tell is only visible through the API:

```
gh api repos/spec-kitty/spec-kitty/actions/jobs/104200296442 --jq .conclusion
cancelled
```

## Why it is worth fixing

The uniform ~3m10s stop time is the only surface-level hint that these are cancellations rather than failures, and it is easy to read the wall of red as catastrophic breakage on the branch. Finding the real failure means listing the run's jobs and filtering on `conclusion=="failure"` — which you only know to do once you have already worked out that fail-fast is in play.

This is per-agent, per-PR, repeated. In one pass over my own open PRs today, the same wall of red appeared on five of them and resolved to exactly **two** distinct inherited failures, both already fixed on `main`:

- `tests/release/test_pinning_inventory_fresh.py::test_inventory_is_reproducible_by_rerunning_the_derivation`
- `tests/mission_runtime/test_consolidated_resolution.py::test_e2_status_state_still_probes_coordination_and_is_unaffected`

Both were green at `main@cede0e381` and red at the older merge-base `819433a78`, so the whole wall was stale-base noise. The fix in every case was a refresh, not a code change — but arriving at that conclusion took a job-level API sweep per PR.

There is also a false-attribution hazard, which is the part I would weight most: an agent who takes the check list at face value can conclude its own diff broke forty shards, and start "fixing" tests it never touched. The signal that would have prevented that — *one* shard failed, here it is, the rest were stopped — is present in the data and absent from the surface.

## Candidate directions

Not a recommendation between them; whoever picks this up should weigh them.

1. **Summarize the run.** Have the aggregate step emit one line naming the shard(s) with `conclusion == "failure"` and the count cancelled, so the real failure is readable without a job sweep. Cheapest, and does not touch the fail-fast tradeoff.
2. **Report cancelled distinctly.** If a check's conclusion can be surfaced as cancelled/skipped rather than failed, the wall stops looking like breakage at all. Depends on what the checks API allows for matrix jobs — worth a spike before committing.
3. **Leave fail-fast alone.** Turning it off would make every shard's status honest at a large and recurring CI-minute cost. Recording it as considered and rejected unless the minutes turn out cheap.

## Acceptance

- Given a `pr`-mode `CI Modules` run where exactly one shard fails, the PR surface identifies **which** shard failed without requiring a per-job API query.
- The distinction between "this shard failed" and "this shard was cancelled by fail-fast" is recoverable from what the PR shows.
- A test covers the summarizing/reporting seam, driven by a fixture of one `failure` plus several `cancelled` jobs rather than by a live run.

robertDouglass:
## What changes for you

When one CI module shard fails in `pr` mode, `fail-fast` stops the other ~40 shards, and GitHub reports every stopped shard as `cancelled` — which `gh pr checks` and the PR check list render **identically to a real failure**. One red shard therefore looks like forty, and nothing on the surface says which one actually broke. Reading that wall of red, an engineer or agent can conclude their own change broke forty shards and start "fixing" tests it never touched — false attribution, not just noise. (Measured on PR #4417: one shard genuinely failed, 30 others were `cancelled`, and `gh pr checks` printed `fail` for all 31.)

After this change, a run prints a plain-language summary that **names the shard that actually failed**, links straight to its log, and folds the fail-fast-cancelled shards behind a `<details>` so they can't bury the signal — e.g. *"1 of 37 jobs actually failed (module-tests (release shard 1/1)). The other 30 were CANCELLED by fail-fast, not failing."*

## How

A new `summarize` job in `ci-modules.yml` reads the run's own job list (`GET .../actions/runs/<id>/jobs`) and classifies each job. The data was already in the API; only the surface was missing. `fail-fast` is deliberately left on (turning it off would make every shard honest at a large recurring CI cost — considered and rejected in #4420); this fixes the reporting instead.

**It cannot change a verdict.** The job is excluded from every gate's `needs:` set (including the terminal `modules-gate`), runs `if: always()`, and is `continue-on-error` at both job and step level — so fail-fast cancelling the matrix cannot skip it, and a checkout blip or API 5xx cannot redden a green run. This matters because the workflow's own run conclusion is itself consumed as a signal.

## Tests

- `tests/ci/test_summarize_shard_outcomes.py` — 17 passed (classifier is a pure function over the API payload; every branch covered, including the cases where being wrong would be worst: a cancelled job counted as a failure, an all-cancelled run described as a shard failure, a still-running job guessed into pass/fail, multi-page pagination, and malformed stdin).
- `tests/ci/` (owning module, in full) — 286 passed.
- `tests/architectural/test_module_shard_registry.py` + `tests/ci/test_ci_module_wiring.py` — 23 passed.
- `ruff check` / `ruff format --check` / `mypy` — clean.

## Landing notes (maintainer)

Rebased onto current `main` (was 103 behind). One conflict in `ci-modules.yml`: `main` had since added a `modules-gate` terminal aggregator where this PR added its `summarize` job — kept both; verified `summarize` stays out of `modules-gate`'s `needs:`, so its inertness invariant holds under the new gate. Three landing folds:

1. **`fix(landing)`** — hardened the input seam. A second-opinion review found the classifier read stdin with a single `json.loads`, but `gh api --paginate` on the jobs *object* endpoint concatenates one object per page without merging, so any run past one page (>100 jobs, or re-run attempts) piped invalid single-document JSON and the script crashed — degrading (via the workflow's `|| echo`) to a generic notice instead of naming the shard, exactly when the run is large. `_load_jobs` now decodes every JSON document and flattens each page's jobs (single-envelope, concatenated-pages, bare-list, and `--slurp` shapes), and `main` catches the decode error and degrades in-script, so the fail-closed property is testable rather than living only in the workflow.
2. **`ci(landing)`** — aligned the summarize job's `actions/checkout` pin to the repo-dominant `v5.0.0` SHA (it was the sole `v4.4.0` pin).
3. **`chore(landing)`** — re-captured `ci-shard-timings.json` through the documented producer against the final collection (286). The rebase and the parser-hardening tests both moved the target; a stale table silently falls back to uniform shard weights — the exact failure the PR set out to prevent.

The PR body's original **Deferred** item (`main` red on `test_every_test_directory_is_claimed_once_or_recorded_out_of_matrix`) is now **green on current `main`** — fixed in the intervening commits (verified on a clean checkout); nothing left to defer.

Follow-up architecture (not blocking, not ticketed): the `failure`/`cancelled`/`timed_out` conclusion vocabulary is now declared in `summarize_shard_outcomes.py`, `router_gate.py`, and `fleet_verdict.py`. Reuse isn't a drop-in — the summarizer deliberately separates `cancelled` from `failure` (the opposite of `router_gate`'s blocking policy) — but the raw string constants could be hoisted to one shared module to prevent drift whenever those scripts are next touched together.

Closes #4420.


stijn-dejongh:
**Maintainer landing pass — claimed.**

Picking this up for today's landing pass (rebase onto current `main`, adjudicate checks, second-opinion squad over the integrated diff, hand off green for the operator to merge).

**Viability triage — LAND.** Fresh but substantive (new CI `summarize` job + `scripts/ci/summarize_shard_outcomes.py`), so the full check:
- **Not superseded:** #4420 is OPEN; `main` carries no summarize job or script.
- **Still needed:** `ci-modules.yml` on `main` still runs the matrix with `fail-fast` in `pr` mode (line ~294) and has no failure-attribution surface — the false-attribution problem the PR describes is live on current `main`.
- **Compatible:** the change adds one `if: always()` / `continue-on-error` reporting job that is excluded from every gate's `needs:` set, and follows the established `scripts/ci/*.py` + `tests/ci/test_*.py` convention — no seam moved under it.

Branch is behind `main` and `CONFLICTING`. Rebasing, re-verifying, and running a profile-loaded review squad over the integrated diff now. The PR's own `Deferred` note (main red on `test_every_test_directory_is_claimed_once_or_recorded_out_of_matrix`) I'll classify independently. Evidence to follow.

stijn-dejongh:
**Landing pass complete — merge-ready, handing off.**

**Verdict: LAND.** Substantive CI-DX PR; rebased, second-opinion squad run, three folds applied. Clean linear history preserved for rebase-merge (contributor commit + 3 single-purpose folds), contributor authorship (Robert Douglass) intact.

**Rebase:** onto current `main` (was 103 behind, `CONFLICTING`). Conflict in `ci-modules.yml` — `main` had added a `modules-gate` terminal aggregator where this PR adds its `summarize` job. Kept both; confirmed `summarize` is absent from `modules-gate`'s `needs:` (and every other gate's), so the PR's inertness invariant holds under the new gate.

**Second-opinion squad (2 opus lenses, profile-loaded, over the integrated diff):**
- *architect-alphonso (design/inertness/security):* inertness invariant HOLDS, permissions minimal (`actions: read` job-scoped, checkout SHA-pinned + `persist-credentials: false`, no shell injection), convention followed. No MAJOR. One MINOR (checkout pin divergence) → folded. Two low-priority NOTEs.
- *reviewer-renata (correctness/fail-closed):* classification core correct and fail-closed against all three worst-case misclassifications. **One MAJOR** on the input seam + one fail-closed MINOR → folded together.

**Folds (each its own commit, subagent-implemented under model discipline, focused tests):**
1. `fix(landing)` — **input-seam hardening (the MAJOR).** `_load_jobs` read stdin with a single `json.loads`, but `gh api --paginate` on the jobs *object* endpoint concatenates one object per page without merging, so any run past one page (>100 jobs / re-run attempts) fed invalid single-document JSON → crash → silent degrade to a generic notice instead of naming the shard, exactly when the run is large. Now decodes every JSON document and flattens each page's jobs (single-envelope / concatenated-pages / bare-list / `--slurp`); `main` catches the decode error and degrades in-script so the fail-closed property is testable, not just the workflow's `|| echo`. +3 tests (page-two failure across pages, `--slurp` shape, invalid-JSON stdin → exit 0). 17 passed.
2. `ci(landing)` — aligned the summarize job's `actions/checkout` from the repo's sole `v4.4.0` pin to the dominant `v5.0.0` SHA.
3. `chore(landing)` — re-captured `ci-shard-timings.json` via the documented producer against the final 286-test collection (the rebase added ~17 ci tests, fold #1 added 3); a stale positional table silently falls back to uniform shard weights — the exact defect this PR set out to prevent.

**Stale 'Deferred' item resolved:** the PR body deferred `test_every_test_directory_is_claimed_once_or_recorded_out_of_matrix` as red on `main`; it is now **green on current `main`** (verified on a clean non-dot checkout), fixed in the intervening commits. Nothing to file.

**Local verification (rebuilt tip):** `tests/ci` 286 passed · registry+wiring gates 23 passed · ruff/format/mypy clean · commitlint types valid (`fix`/`ci`/`chore`). No pre-existing reds surfaced.

**Deliverable hygiene:** no changelog/docs entry — the change is CI reporting DX, touching no spec-kitty CLI, product, or pack-author surface. PR body rewritten BLUF; a genuine future-architecture NOTE (conclusion-vocabulary triplication across 3 CI scripts) is recorded in the body, deliberately not ticketed.

State: pushed, watching CI — **merge-ready on green**. Not a draft; operator merges.

PR #4547, opened 2026-09-15
[#3233] Explain refresh credential and client failures safely

### LINKED ISSUE #3233: Token refresh reports 'you sent no credential' as an unexplained failure: invalid_client and invalid_request are unhandled
`TokenRefreshFlow._raise_known_auth_error` recognises only two of the five error codes the SaaS token endpoint can return, so two of them reach the user as an unexplained failure instead of an actionable one.

## Where

`src/specify_cli/auth/flows/refresh.py:201`

```python
if error == "invalid_grant":
    raise RefreshTokenExpiredError(...)   # "Run `spec-kitty auth login` again."
if error == "session_invalid":
    raise SessionInvalidError(...)
```

Anything else falls through to the generic tail of `refresh()`:

```python
raise TokenRefreshError(f"Token refresh failed: HTTP {response.status_code} - {response.text[:500]}")
```

## What the server actually returns

From `apps/cli_auth/views_token.py` on the SaaS side, a refresh grant can answer:

| status | error | meaning | handled |
| :--- | :--- | :--- | :--- |
| 400 | `invalid_client` | wrong or missing `client_id` | **no** |
| 400 | `invalid_request` | **no `refresh_token` was sent at all** | **no** |
| 401 | `invalid_grant` | the token was rejected (invalid, expired, revoked, suspicious replay) | yes |
| 409 | `refresh_replay_benign_retry` | just rotated; reload and retry | yes, separately |
| 429 | `rate_limited` | too many token requests | n/a |

The two unhandled cases are the ones that mean *the client never presented a credential*, which is the opposite of *the credential was rejected* — and they are the ones a user cannot act on without being told.

## Why this is worth fixing rather than a curiosity

`refresh()` sends `session.refresh_token` with no guard:

```python
data = {
    "grant_type": "refresh_token",
    "refresh_token": session.refresh_token,
    "client_id": self._client_id,
}
```

If a `StoredSession` ever loads with an empty or absent refresh token, the CLI posts an empty value, the server correctly answers `400 invalid_request`, and the user is shown a generic HTTP failure. Downstream that surfaces as "not authenticated: no valid access token" — which points at the token being expired when the real problem is that no token was sent.

That exact confusion was reported in `Priivacy-ai/spec-kitty-saas#738`: a user with a valid, unexpired refresh token being told he was not authenticated. Production recorded 13 consecutive `POST /oauth/token` 400s from that client in one hour with zero successes.

Worth noting the storage angle, since it is a plausible source of an empty session: `auth/secure_storage/file_fallback.py:134` derives its encryption key from `f"{socket.gethostname()}:{_get_uid()}"`, so a hostname change makes the stored session undecryptable.

## Suggested fix

1. Map `invalid_client` and `invalid_request` to distinct, actionable errors rather than the generic tail. `invalid_request` in particular should say the CLI had no refresh token to send, not that the server refused one.
2. Guard before the POST: if `session.refresh_token` is empty, fail locally with "no stored refresh token — run `spec-kitty auth login`" instead of making a round trip that cannot succeed.

## Related

The server-side half — the token endpoint kept no record of *why* any grant was refused — is fixed in `Priivacy-ai/spec-kitty-saas#755`, which adds a structured `cli_auth_token_grant_rejected` log carrying `grant_type` and the error code. With that landed, the next occurrence is diagnosable from one query rather than by reverse-engineering access-log status codes.

## Software factory handoff

Implementation delivered in PR #4547; factory now owns squad review, CI classification, any fix rounds, and guarded merge. Continue the existing PR rather than launching duplicate implementation. The linked-PR claim remains active while that delivery completes; this is not reserved human work.

Acceptance: required squad disposition and exact-head CI pass, then merge through the factory and close this issue via the PR. CI baseline repair is tracked in #4552.

Inherit system SaaS-sync settings; do not set, unset, or override `SPEC_KITTY_ENABLE_SAAS_SYNC`.

robertDouglass:
## Issue
closes #3233

## Change
Reject missing, null, empty, or whitespace-only stored refresh credentials before opening an HTTP client. Explain `invalid_request` as a missing/malformed refresh parameter with login recovery, and `invalid_client` as rejected CLI client identification with configuration guidance. Neither known nor generic HTTP errors echo response bodies; malformed error JSON uses the same safe HTTP diagnostic, including HTTP 409.

Governed dispatch Op: `01M2KGV7687KW09PT7VSXYWMPY` (`implementer-ivan`). Red-first tests were committed before implementation (`005b65378`); an independent review found null-credential and malformed-409 cases, reproduced and fixed in `d8738f122`.

## Tests run
- `make test-fast` → 1814 passed, 5 skipped (633.44s).
- Red-first new regression cases: 11 failed before the fix.
- Independent-review reproductions: 4 failed, 9 passed before hardening.
- `.venv/bin/pytest tests/auth -q` → 518 passed, 3 skipped on final code (includes all 44 refresh-flow cases).
- `.venv/bin/ruff check src/specify_cli/auth/flows/refresh.py tests/auth/test_refresh_flow.py` → passed.
- `.venv/bin/mypy --strict src/specify_cli/auth/flows/refresh.py src/specify_cli/auth/session.py` → passed. Include session.py because repository mypy configuration skips imported specify_cli modules.

Self-review:
- correctness: clean; independent review findings resolved
- deletion safety: clean; no files removed
- security: clean; no live auth requests, no untrusted response-body output
- design fidelity: clean; scoped to refresh diagnostics
- tests: clean

## Blast radius
Discovery: `git diff --name-only b7329280a9a5648a8180521ea74e531ac7c183e4 HEAD`
Files: `src/specify_cli/auth/flows/refresh.py`, `tests/auth/test_refresh_flow.py`.

The auth subsystem owns refresh orchestration and its consumers. The complete auth package covers token rotation, concurrency, transport, and session behavior. Existing invalid_grant/session_invalid exceptions remain unchanged. No wire contract, Widen, device authorization, or persistence change; no migration required.

## Deferred
None in issue scope. This does not claim to establish the cause of the historical SaaS incident.

Factory handoff: implementer work is complete. Run the canonical squad, CI, fix, and merge lanes on this existing PR. Release-inventory baseline failure is tracked in #4552; do not bypass failing gates. Inherit system SaaS-sync settings without overriding `SPEC_KITTY_ENABLE_SAAS_SYNC`.


robertDouglass:
CI update: this PR is **not green**. The release shard fails the pinning-inventory freshness test; independent reproduction on untouched base main `b7329280a9a5648a8180521ea74e531ac7c183e4` produces the same failure. Four workflow-coherence inventory line locations are stale by 28 lines. Tracked with exact evidence in #4552.

[Failing release job](https://github.com/spec-kitty/spec-kitty/actions/runs/35029135793/job/104583356667): 1 failed, 236 passed, 5 skipped. The sampled agent-shard failure log reports cancellation rather than a test failure. Local auth validation remains 518 passed / 3 skipped; that does not override the red CI gate. No auth regression or PR setup change identified; baseline repair and subsequent CI verification remain necessary.

PR #4663, opened 2026-09-16
docs(mission): cli-boundary-robustness planning dossier (impl in #4674)

stijn-dejongh:
## What this is

The **planning / mission-dossier PR** for the `cli-boundary-robustness-01M2NQCB` mission: its spec, plan, research/data-model/contracts, tasks and work packages, decision + lifecycle records, issue matrix, and acceptance evidence — everything under `kitty-specs/cli-boundary-robustness-01M2NQCB/`.

The **implementation** (the `src/` changes hardening the CLI configuration, JSON boundaries, and command defaults) is reviewed and lands **separately in #4674** — the deliberate planning/implementation split. This PR carries **no source changes**.

Mission issues: #4600, #4601, #4643, #4597, #4598 (implemented and closed via #4674, not here).

## Landing notes

- **History cleaned:** the mission run produced ~140 commits, overwhelmingly per-transition status/lifecycle churn. Collapsed to a single `docs(mission)` commit holding the dossier's final committed state (mission content byte-identical to the pre-cleanup branch).
- **Re-scoped to planning-only:** the branch had previously squash-merged a now-stale copy of the implementation `src/`; that has been dropped so it cannot collide with or duplicate #4674's canonical implementation.
- **Rebased** onto current `main`.
- **Corpus-clean:** the mission's `status.json` matches its events (no `SNAPSHOT_DRIFT`), no ERROR-severity findings, so it adds no teamspace-blocker to the corpus witness.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019g74uWbdEddmC329RATNTe

PR #4674, opened 2026-09-16
[#4600] Harden CLI configuration, JSON boundaries and command defaults

### LINKED ISSUE #4532: doctor family: unify the not-in-project guard control-flow into one shared helper (renderer is single-sourced, the guard is still copy-pasted)
## What / for whom

A landing-pass squad on **#4246** (closing the `doctor --json` not-in-project contract, #4242) found that the fix single-sourced the **renderer** but not the **guard around it**. The doctor family now emits its not-in-project error through one canonical `_doctor_shared._emit_not_in_project(json_output)`, but the enclosing *resolve-root-or-exit guard* — the `try: locate_project_root() / except → emit+Exit / if None → emit+Exit` block — is still copy-pasted across ~6 sites. This is invisible to operators (no behavior change); it matters to **maintainers**, because it is the exact mechanism by which the next doctor subcommand drifts back out of contract (as `mission-state` and `tool-surfaces` already did before #4246 folded them in).

## Evidence (post-#4246)

The renderer is shared, but the guard is duplicated:

- `doctor.py::_resolve_project_root_or_exit(json_output, *, exit_code)` — named helper, consumed by the twelve `doctor.py` commands.
- `_command_surface_doctor.py::_resolve_or_exit(exit_code, json_output)` — a **near-identical twin** (same try/except/None-check/Exit, only the signature order differs).
- Inline copies of the same block in `_coordination_doctor.py::run_coordination_health`, `_env_file_doctor.py::register`, `_provenance_doctor.py::register`, and the `doctor.py::mission_state` shell.
- `_sparse_checkout_doctor.py::_resolve_repo_root` is a third named copy (no `--json`, so out of contract scope, but the same shape).

## Suggested fix

Promote one `_resolve_project_root_or_exit(json_output, *, exit_code)` into `_doctor_shared` (where `_emit_not_in_project` already lives) and have every doctor-family command consume it — deleting the `doctor.py` / `_command_surface_doctor.py` twins and the inline sibling copies. The `exit_code` stays a per-command parameter (1 for most; 2 for the config-error family), so no behavior changes — this is a pure SSOT/no-behavior-change refactor that unifies the *guard*, not just the *renderer*. A small guard-shape test can pin that every `--json` doctor command routes through the shared helper.

## Scope note

Deliberately **not folded into #4246** — it is a refactor of already-correct code across ~6 sites, out of scope for a contract bugfix. Filed as follow-up per the landing runbook.

Related: #4246, #4242. Sibling finding (cross-command version): filed separately under #2899.

### LINKED ISSUE #4533: `--json` commands leak prose on stdout outside a project beyond the doctor family (archive, materialize) — enforce the contract at the locate_project_root choke point
## What / for whom

The `--json` machine contract — "on every exit path, stdout is one parseable JSON object, never bare human prose" — is enforced per-command-family, not at the choke point, so it keeps regressing in new places. #1336 fixed it for the **charter** family, #2879 for **orchestrator-api**, and **#4246** for the **doctor** family (#4242). A landing-pass squad on #4246 confirmed the **same leak still lives in non-doctor `--json` commands**, so a consumer piping `<cmd> --json | jq` from outside a project still hits a parse error indistinguishable from a crash.

**For whom:** agents / CI / operators consuming `--json` output as a machine contract.

## Evidence

Both accept `--json` and print `Error: Not in a spec-kitty project` (Rich prose) on **stdout** on the not-in-project path, exit non-zero — not JSON:

- `src/specify_cli/cli/commands/archive.py:62,115`
- `src/specify_cli/cli/commands/materialize.py:55`

These are **not** doctor commands, so the doctor-scoped `_doctor_shared._emit_not_in_project` renderer #4246 introduced cannot serve them without either duplicating the renderer a fourth time or pulling a cross-command dependency. That is the signal: the leak is codebase-wide and the **true choke point is `locate_project_root` / the CLI not-in-project boundary**, not any per-family renderer.

## Suggested fix

Introduce one shared *resolve-root-or-emit-machine-error* seam at the `core.paths` / CLI boundary that **any** `--json` command consumes — a single place that turns a not-in-project outcome into either the canonical `{"ok": false, "error": {"code": "not_in_project", "message": "Not in a spec-kitty project"}}` JSON envelope (under `--json`) or the Rich prose line (human path), with the caller's own exit code. Then migrate the per-family renderers (charter, orchestrator-api, doctor's `_emit_not_in_project`, and the archive/materialize leaks) onto it. A contract test that enumerates every typer command accepting `--json`/`json_output` and asserts `json.loads(stdout)` succeeds on the repo-root-missing path (the test #1336's acceptance proposed but never delivered) would catch the whole class and prevent the next one.

## Scope note

This is design-bearing (a new shared seam + a repo-wide migration), so it is **not** a landing-pass fold — it needs its own mission. Filed per the #4246 landing runbook.

Related: #4246, #4242, #1336 (closed), #2879 (closed). Sibling finding (doctor-family-internal guard SSOT): filed separately under #2720.

### LINKED ISSUE #4597: bug: `spec-kitty context` (no subcommand) crashes — leaks a Typer OptionInfo as the workspace name
## Summary

`spec-kitty context` with **no subcommand** always crashes, leaking a raw Typer `OptionInfo` object where a workspace name should be:

```
Error: No context found for workspace: <typer.models.OptionInfo object at 0x7fb1bcbab9d0>

Context file not found:
  <repo>/.kittify/workspaces/<typer.models.OptionInfo object at 0x7fb1bcbab9d0>.json
```

This is the same bug **class** as the closed #568 (`agent action implement`) and the open #2779 (`charter generate/synthesize --resynthesize`): a Typer command function is invoked **directly as a Python function**, so its parameter defaults are the raw `typer.models.OptionInfo` sentinels instead of the resolved values. The `context` group's default-command callback is a not-yet-fixed instance of it.

`spec-kitty context <subcommand>` (e.g. `context info`, `context list`) works fine — only the bare no-subcommand path is broken.

## Reproduction

```bash
spec-kitty context          # from anywhere — a real mission repo, a worktree, or a non-project dir
```

Reproduced with `spec-kitty-cli 4.0.0rc2` both inside a real Spec Kitty project and from a non-project directory. It is **location-independent** — the crash does not depend on repo-root resolution.

## Root Cause

`src/specify_cli/cli/commands/context.py` — the group callback calls the `info` command function directly:

```python
@app.callback(invoke_without_command=True)
def main(ctx: typer.Context) -> None:
    """Query workspace context information."""
    if ctx.invoked_subcommand is None:
        # No subcommand - default to "info"
        info_command()          # <-- called as a plain function
```

`info_command`'s signature:

```python
@app.command(name="info")
def info_command(
    workspace: str = typer.Option(None, "--workspace", "-w", ...),
    json_output: bool = typer.Option(False, "--json", ...),
) -> None:
    ...
    if workspace is None:                       # never True: workspace is an OptionInfo
        workspace = detect_current_workspace(...)
        ...
    context = load_context(repo_root, workspace)  # workspace is the OptionInfo sentinel
    if context is None:
        console.print(f"[red]Error:[/red] No context found for workspace: {workspace}")
        console.print(f"  {repo_root / '.kittify' / 'workspaces' / f'{workspace}.json'}")
```

Because `info_command()` is called directly, `workspace` binds to `typer.Option(None, ...)` — an `OptionInfo` instance, which is **not `None`** (and is truthy). The `if workspace is None:` auto-detection guard is skipped, and the sentinel flows straight into `load_context` and the error/path strings. (Same mechanism #568 documented: `bool(typer.Option(False)) is True`.)

## Expected Behavior

`spec-kitty context` with no subcommand should behave the same as `spec-kitty context info` with resolved defaults — i.e. auto-detect the workspace (and print the graceful `Not inside a worktree and no --workspace specified` guidance when there is none), never surface a `typer.models.OptionInfo` repr to the user.

## Suggested Fix

Route the default through Typer with real values rather than calling the command function bare, e.g.:

```python
if ctx.invoked_subcommand is None:
    info_command(workspace=None, json_output=False)
```

or invoke the `info` subcommand via the Typer context / `ctx.invoke(...)` so defaults resolve normally. A regression test that runs `spec-kitty context` with no subcommand and asserts the output contains no `OptionInfo` repr would guard this and the sibling instances (#568, #2779).

## Notes

Found while sweeping CLI commands run from a non-project directory (follow-up to the #3652 investigation), but this crash is independent of that root cause — it reproduces inside a valid mission repo as well.

### LINKED ISSUE #4598: bug: `spec-kitty mission-type list` / `mission list` always shows inactive types (truthy OptionInfo default for include_inactive)
## Summary

`spec-kitty mission-type list` (and its alias `spec-kitty mission list`) **always behaves as if `--include-inactive` were passed** — it lists every registered mission type, including ones the project's charter has *not* activated. Its own docstring says the opposite:

> Returns only mission types that are explicitly activated in this project's charter (activation-filtered). Alias for `spec-kitty charter mission-type list`.

But it does **not** match its canonical target `spec-kitty charter mission-type list` (default) — it matches `spec-kitty charter mission-type list --include-inactive`.

Same bug **class** as the closed #568, open #2779, and #4597 (just filed): a Typer command function is called directly as a Python function, so an omitted `typer.Option(...)` parameter keeps its raw `OptionInfo` default — which is **truthy**.

## Reproduction

A project whose charter activates only a subset of mission types (here, only `software-dev`):

```yaml
# .kittify/config.yaml
mission_type_activations:
- software-dev
```

```
$ spec-kitty mission-type list           # claims "activated only"
│ software-dev  │ built-in │ Software Development │ specify, plan, tasks, ...  │
│ documentation │ built-in │ Documentation        │ (not activated)            │
│ plan          │ built-in │ Plan                 │ (not activated)            │
│ research      │ built-in │ Research             │ (not activated)            │

$ spec-kitty charter mission-type list   # canonical DEFAULT (activated only) — CORRECT
│ software-dev  │ built-in │ Software Development │ specify, plan, tasks, ...  │

$ spec-kitty charter mission-type list --include-inactive
│ software-dev  │ built-in │ Software Development │ specify, plan, tasks, ...  │
│ documentation │ built-in │ Documentation        │ (not activated)            │
│ plan          │ built-in │ Plan                 │ (not activated)            │
│ research      │ built-in │ Research             │ (not activated)            │
```

`mission-type list` == `charter mission-type list --include-inactive`, not the activated-only default. Env: `spec-kitty-cli 4.0.0rc2`.

## Root Cause

`src/specify_cli/cli/commands/mission_type.py` — the alias wrapper calls the real command function directly and passes only `json_output`:

```python
@app.command("list")
def list_mission_types(
    json_output: bool = typer.Option(False, "--json", help="Output as JSON."),
) -> None:
    """List activated mission types ... Alias for `spec-kitty charter mission-type list`."""
    from specify_cli.cli.commands.charter.mission_type import charter_mission_type_list
    charter_mission_type_list(json_output=json_output)     # include_inactive NOT passed
```

`charter_mission_type_list` (`src/specify_cli/cli/commands/charter/mission_type.py`):

```python
def charter_mission_type_list(
    json_output: bool = typer.Option(False, "--json", ...),
    include_inactive: bool = typer.Option(False, "--include-inactive", ...),
) -> None:
    ...
    ordered_ids = list(activated_ids)
    if include_inactive:                 # OptionInfo sentinel is truthy -> always True
        ordered_ids.extend(sorted(set(roster) - activated_id_set))
```

Because the wrapper omits `include_inactive`, it binds to `typer.Option(False, "--include-inactive", ...)` — an `OptionInfo` instance, which is truthy:

```
$ python -c "import typer; o=typer.Option(False,'--x'); print(type(o).__name__, bool(o))"
OptionInfo True
```

So `if include_inactive:` is always True and the inactive types are appended. `mission.py` re-exports `mission_type.app`, so `spec-kitty mission list` inherits the same defect.

## Expected Behavior

`spec-kitty mission-type list` / `spec-kitty mission list` should show only charter-activated mission types by default — identical to `spec-kitty charter mission-type list` — since it is documented as an alias for it.

## Suggested Fix

Pass the resolved default explicitly:

```python
charter_mission_type_list(json_output=json_output, include_inactive=False)
```

or route through Typer / `ctx.invoke(...)` so defaults resolve. Better still, expose `--include-inactive` on the alias and forward it, so the two surfaces are true equivalents. A regression test asserting `mission-type list` on a partially-activated charter shows only activated rows would guard this and the sibling instances (#568, #2779, #4597).

### LINKED ISSUE #4600: bug: malformed .kittify/config.yaml crashes with uncaught traceback — non-UTF-8 bricks every command at startup (UnicodeDecodeError)
## Summary

A malformed `.kittify/config.yaml` crashes Spec Kitty with an **uncaught Python traceback** instead of a clean, operator-readable error. Two concrete instances, one severe:

1. **(Severe) A non-UTF-8 `.kittify/config.yaml` bricks the entire CLI** at import time — every command, including `spec-kitty --version` and `spec-kitty doctor`, dies with an uncaught `UnicodeDecodeError`. The CLI cannot even self-diagnose or repair the project.
2. **(Milder) Syntactically-invalid YAML in `.kittify/config.yaml` crashes `spec-kitty mission list` / `mission-type list`** with a traceback, even though the same file is handled gracefully by `spec-kitty charter status` and `spec-kitty config`.

Realistic triggers: a bad merge, an editor that writes a BOM or Latin-1, a truncated/partial write, copy-paste through a non-UTF-8 pipeline.

### Reproduction
```bash
printf '\xd0\xd0\xd0bad\xff' > .kittify/config.yaml
spec-kitty --version      # or doctor, or literally anything
```

### Actual
```text
Traceback (most recent call last):
  File "/.../bin/spec-kitty", line 4, in <module>
    from specify_cli import main
  File "/.../specify_cli/__init__.py", line 35, in <module>
    load_operator_env_file()
  File "/.../specify_cli/bootstrap/env_file.py", line 273, in load_operator_env_file
    home_path = _resolve_home_tier_path(repo_root, target_environ)
  File "/.../specify_cli/bootstrap/env_file.py", line 240, in _resolve_home_tier_path
    raw = _read_config_env_file_pointer(repo_root) or _DEFAULT_ENV_FILE_TEMPLATE
  File "/.../specify_cli/bootstrap/env_file.py", line 214, in _read_config_env_file_pointer
    text = config_path.read_text(encoding="utf-8")
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd0 in position 0: invalid continuation byte
```

Confirmed `spec-kitty --version` and `spec-kitty doctor skills` both crash identically. Env: `spec-kitty-cli 4.0.0rc2`.

### Root Cause
`src/specify_cli/bootstrap/env_file.py`, `_read_config_env_file_pointer`:

```python
config_path = repo_root / _CONFIG_YAML_RELATIVE   # .kittify/config.yaml
try:
    text = config_path.read_text(encoding="utf-8")
except OSError:            # <-- catches OSError, NOT UnicodeDecodeError
    return None
```

The function already intends to degrade to `None` on read failure (it catches `OSError` and returns `None`), but `UnicodeDecodeError` is a `ValueError`, not an `OSError`, so it escapes. Because `load_operator_env_file()` runs at **module import** (`specify_cli/__init__.py:35`), the exception fires before Typer parses any command — so nothing runs, including the diagnostics a user would reach for to fix it.

### Suggested Fix
```python
except (OSError, UnicodeDecodeError):
    return None
```
(or read bytes + `decode(errors="replace")`). A regression test that drops a non-UTF-8 byte into `.kittify/config.yaml` and asserts `spec-kitty --version` still exits 0 would guard it. Note there is already precedent for encoding robustness in charter files (`spec-kitty migrate charter-encoding`, FR-026/027) — this is the same hazard on the config.yaml startup read.

### Reproduction
```bash
printf 'this: is: not: valid: yaml: [unclosed\n' > .kittify/config.yaml
spec-kitty mission list      # traceback, exit 1
spec-kitty charter status    # graceful: WARNING "Failed to read .kittify/config.yaml; org doctrine disabled", exit 0
spec-kitty config            # exit 0
```

### Actual (mission list)
```text
.../charter/activation/pack_context.py:601 in _load_config
    raise _config_error(f"Invalid YAML in .kittify/config.yaml: {exc}") from exc
CharterPackConfigError: CHARTER_PACK_CONFIG_INVALID
```

`_load_config` raises a clean domain error (`CHARTER_PACK_CONFIG_INVALID`), so the failure is well-modeled — it is simply **not caught at the `mission list` command boundary**. The command (`charter_mission_type_list` in `cli/commands/charter/mission_type.py`) wraps only `resolve_layered_roster(...)` in `try/except ValueError`, but (a) `existing_mission_types(repo_root)` runs *before* that guard and triggers the same `_load_config`, and (b) `CharterPackConfigError` is not a `ValueError`. `charter status` and `config` catch it and degrade; `mission list` should do the same.

### Expected (both instances)
Malformed `.kittify/config.yaml` — bad encoding or bad YAML — should produce a single clean, operator-readable error (or graceful degradation) with a non-crash exit, consistently across commands, and must never brick `--version`/`doctor`.

## Notes
Found via a malformed-input sweep of CLI commands in a sandboxed project. No files under `$HOME` were touched.

### LINKED ISSUE #4601: bug: `context info --json` and `mission-type show --json` emit prose on stdout on error paths (--json contract gap, cf. #4242)
## Summary

Two more commands break the `--json` machine contract on their error paths — the same systemic gap fixed for the `doctor` family in #4242, but outside that issue's scope:

- `spec-kitty context info --workspace <unknown> --json`
- `spec-kitty mission-type show <unknown-type> --json`

Both write human/Rich prose to **stdout** and exit 1 instead of emitting a JSON error object, so a consumer doing `... --json | jq` (or `json.load`) gets a parse error indistinguishable from a crash. The success path of each emits well-formed JSON, so this is a per-command exit-path inconsistency, exactly as #4242 described. As #4242 notes, the fix exemplar already in the tree is `doctor skills --json`, which emits `{"ok": false, "error": {"code": ..., "message": ...}}` on stdout.

Env: `spec-kitty-cli 4.0.0rc2`.

## Reproduction

```bash
# 1) context info — unknown workspace (a common, expected path)
spec-kitty context info --workspace does-not-exist --json | python3 -c "import json,sys; json.load(sys.stdin)"
# stdout (NOT json):
#   Error: No context found for workspace: does-not-exist
#   Context file not found:
#     <repo>/.kittify/workspaces/does-not-exist.json
# -> JSONDecodeError

# 2) mission-type show — unknown type (documented exit-1 path)
spec-kitty mission-type show does-not-exist --json | python3 -c "import json,sys; json.load(sys.stdin)"
# stdout (NOT json):
#   Error: Unknown mission type 'does-not-exist'. Registered types: documentation, plan, research, software-dev
# -> JSONDecodeError
```

`context info --json` also emits prose (not JSON) on its other two error branches: repo-root unresolved, and "Not inside a worktree and no --workspace specified".

## Root Cause

Both error branches `console.print(...)` and `raise typer.Exit(1)` **before** (and without consulting) the `json_output` flag; only the success path branches on it.

`src/specify_cli/cli/commands/context.py` (`info_command`):

```python
context = load_context(repo_root, workspace)
if context is None:
    console.print(f"[red]Error:[/red] No context found for workspace: {workspace}")
    console.print("\nContext file not found:")
    console.print(f"  {repo_root / '.kittify' / 'workspaces' / f'{workspace}.json'}")
    raise typer.Exit(1)          # <-- json_output never checked
...
if json_output:                  # honored only on success
    print(json.dumps(context.to_dict(), indent=2))
```

`src/specify_cli/cli/commands/mission_type.py` (`show_mission_type`):

```python
if mission_type_id not in activated_ids:
    err = UnknownMissionTypeError(mission_type_id, registered_ids=activated_ids)
    console.print(f"[red]Error:[/red] {err}")   # <-- json_output never checked
    raise typer.Exit(1)
```

## Expected Behavior

Per #4242: a `--json` invocation keeps stdout machine-consumable on every exit path — a JSON error object on stdout (the `doctor skills --json` shape), or empty stdout with the human message on stderr. Never bare prose on stdout.

## Suggested Fix

In each error branch, when `json_output` is set, emit the JSON error object (mirroring `doctor skills --json`) and send any human text to stderr; otherwise keep the current prose. A regression test per command (`context info --json` unknown workspace; `mission-type show --json` unknown type) asserting stdout parses as JSON would freeze it, as #4242 did for the doctor family.

## Notes

Found via a `--json` error-path sweep. Commands that already do this correctly on error (valid JSON on stdout): `agent mission check-prerequisites`, `agent tasks status`, `accept`, `reconcile` — so this is an isolated inconsistency in `context info` / `mission-type show`, not a tree-wide default.

### LINKED ISSUE #4643: bug: `agent tasks status --json` prints prose (not JSON) on the no-work-packages path (exit 0), breaking the machine contract
## Summary

`spec-kitty agent tasks status --json` emits **human prose on stdout, exit 0**, when a mission has no work packages yet — instead of a JSON board. Unlike the error-path `--json` gaps in #4601, this is a **success path (exit 0)**, and it's the flagship agent-facing status command whose empty-board case is exactly what an agent polls before work packages exist.

```
$ spec-kitty agent tasks status --mission <slug> --json
No work packages found in
/…/kitty-specs/<slug>/tasks
$ spec-kitty agent tasks status --mission <slug> --json | python3 -c "import json,sys; json.load(sys.stdin)"
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

`--json` is documented on this command (help: "Output as JSON", with the example `spec-kitty agent tasks status --json`). The sibling `agent tasks list-tasks --json` handles the empty case correctly:

```
$ spec-kitty agent tasks list-tasks --mission <slug> --json
{"tasks": [], "count": 0}
```

Env: `spec-kitty-cli 4.0.0rc2`.

## Root Cause

`src/specify_cli/cli/commands/agent/tasks_status_cmd.py` (~L421) short-circuits the no-WP case with a human print, before the JSON render seam (`_st_emit_json`, ~L438), and never consults `st.json_output`:

```python
if not st.work_packages:
    _tasks.console.print(f"[yellow]No work packages found in {st.tasks_dir}[/yellow]")
    raise typer.Exit(0)
```

The JSON shape for this exact case already exists elsewhere — `src/specify_cli/agent_utils/status.py:188` returns `{"error": "No work packages found", "work_packages": []}` — so the empty-board branch here just needs to route through the `--json` envelope instead of printing prose.

## Expected Behavior

With `--json`, the no-WP case should emit the empty-board JSON envelope on stdout (e.g. `{"work_packages": [], ...}`), matching `agent tasks list-tasks --json` and the shape already defined in `agent_utils/status.py`. A machine consumer must be able to parse stdout on this common, non-error path.

## Suggested Fix

```python
if not st.work_packages:
    if st.json_output:
        _st_emit_json(st, ports)     # empty-board envelope
    else:
        _tasks.console.print(f"[yellow]No work packages found in {st.tasks_dir}[/yellow]")
    raise typer.Exit(0)
```

## Notes

Found while driving a real mission (spec → substantive plan) to exercise the WP/status surface in a sandboxed project (in `/tmp`; nothing under `$HOME` touched). Related `--json` contract gaps: #4601 (error paths in `context info`/`mission-type show`/`glossary list`), #4242 (closed, doctor family).

robertDouglass:
## What changes for you

The `spec-kitty` CLI no longer **bricks itself on a bad config, breaks its own `--json` contract, or leaks internal placeholder values** at three edges where it used to fail ungracefully. The load-bearing guarantee: a boundary failure now produces a predictable, machine-legible outcome — never a raw traceback, never prose where JSON was promised, never an internal placeholder as a real value.

Three edges are hardened:

**1. A corrupt config never bricks the CLI (the #4600 P0).**
- **Before:** if `.kittify/config.yaml` held non-UTF-8 bytes or malformed content, startup raised an uncaught error and took down **every** command — including `--version` and `doctor`, the operator's own repair path.
- **After:** commands that don't need config content (`--version`, `doctor`) survive; commands that genuinely need it fail with a clear message that **names the file and the decode/encoding problem** — no traceback. Malformed-but-readable config degrades to a named domain error instead of a crash.

**2. `--json` stays valid JSON on every exit path.**
- **Before:** several commands, on the error / empty / zero-result paths, printed human prose to stdout or emitted a partial envelope — so a machine consumer parsing `--json` got a broken payload (e.g. a zero-work-package status exited non-zero or printed prose; #4643, #4533).
- **After:** stdout is well-formed JSON on **every** exit path (success, error, empty); human text goes to stderr. Applied consistently across the affected command surfaces (context, dashboard, doctor, doctrine, glossary, materialize, mission-type, archive, agent tasks).

**3. Argument-binding defaults no longer leak internal placeholders.**
- **Before:** on paths reached off the normal binding flow, an internal placeholder (an `OptionInfo` sentinel) could reach output or logic as if it were a real value (#4601, #4597, #4598).
- **After:** the resolved default is used; no internal placeholder reaches output or program logic.

Git-encoding and env-file reads on the same boundary were hardened alongside, so non-UTF-8 bytes from git or an env file surface as named errors rather than raw decode crashes.

## Closes

Closes #4600.
Closes #4601.
Closes #4643.
Closes #4597.
Closes #4598.
Closes #4533.
Closes #4532.

## Tests

Boundary-contract coverage lands with the change: config-decode fail-soft/fail-loud (`tests/specify_cli/bootstrap/test_cli_boundary_config_4600.py`, `tests/runtime/test_cli_boundary_version_file.py`), `--json` envelope integrity per surface (`tests/specify_cli/cli/commands/test_cli_boundary_*`), the shared doctor resolver guard + frozen contract (`test_coordination_doctor.py`, `test_cli_boundary_json_seam.py`), git-encoding (`tests/specify_cli/core/test_cli_boundary_git_encoding.py`), plus the architectural gates the surface touches (`test_json_contract_enumeration.py`, `test_cli_placeholder_output.py`, `test_no_dead_symbols.py`).

Landing verification (rebuilt tip): the 28 changed test files → **861 passed**; whole-repo `ruff check .` + `ruff format --check .` clean; `[tool.ruff.format].exclude` ratchet green (4 now-clean boundary files dropped).

- **This is the canonical full mission.** #4663 merged the planning-only dossier; this PR carries the implementation **and adds** the execution-phase dossier artifacts (research, review-evidence, per-WP task/subtask records, traces, `wps.yaml`, `issue-matrix.md`) plus the mission's durable Op capsules under `kitty-ops/`. Because `kitty-specs/` and `kitty-ops/` are editorial-freeze roots (`test_archive_root_byte_identical`), the planning files #4663 already merged are left **byte-identical** — this PR only *adds* files (76 additions + the append-only `lifecycle.jsonl` extension), never modifies or deletes a merged artifact. History is a clean linear six-commit shape: `docs(mission)` / `feat(cli)` / `test(cli)` / `build(ruff)` (Robert Douglass) + `fix(landing)` / `docs(landing)` (maintainer folds).
- **Rebuilt onto current `main`**, not the stale mission base. The added dossier artifacts are byte-identical to the source branch; the 4 dropped `[tool.ruff.format].exclude` entries each correspond to a boundary file this mission reformatted clean (`consistency_check.py`, `bootstrap/env_file.py`, `core/worktree.py`, `runtime/doctor.py`).
- **Two pre-existing reds, NOT introduced by this PR, left honest-red:**
  - The mission-state corpus witness reports one `SNAPSHOT_DRIFT` on `ci-terminal-cancel-verdict-01M2NC7Z` — a mission this PR does not touch; its `status.json`/events are byte-identical to `main`. Candidate for a separate corpus hotfix (same class as the #4661 fix).
  - `mypy --strict` reports one `no-any-return` in `tasks_shared.py::unchecked_subtask_ids` — a function this PR does not modify (present on `main`, which in fact has two); mypy is not CI-gated on this file.

## Second-opinion squad (landing, four profile-loaded opus lenses)

Correctness/fail-closed, design/seam-authority, boundary-leak, and terminology lenses over the integrated diff.

- **Correctness (clean):** no MAJOR. The whole-surface `--json` contract guard (`test_json_contract_enumeration.py`) pins valid-JSON, no-traceback, canonical `{"ok": false, "error": {...}}` shape and human/JSON exit-code parity across every adopted command; the `error_handler` failure branches are covered and mutually exclusive.
- **Boundary — one MAJOR folded:** `charter/activation/scope.py::_load_charter_scope_config` still read `.kittify/config.yaml` with a bare `read_text()`, re-opening the #4600 brick class on the charter-scope content-load path (reached by `doctrine`/`charter activate`). **Folded** — wrapped in the same `(OSError, UnicodeDecodeError, YAMLError) → CharterPackConfigError` guard the mission uses for its FR-004 siblings, with a red-first regression driving a real scope-resolving command against a non-UTF-8 in-project config.
- **Design — known debt, documented not reworked:** the `error_handler` seam threaded into `_find_mission_slug` couples error-rendering with resolver selection (adopted by one of ~15 callers) and leaves two envelope shapes side by side. It is **not** a live defect (the sole handler is `NoReturn`, so control cannot fall through — confirmed by the correctness lens) and is transitional under the mission's C-006 convergence-deferred scope. Recorded here for a later CLI-selector convergence mission rather than reworked across 15 call sites inside a landing.
- **Terminology — pre-existing, not folded:** `"Feature directory not found"` and the `057-my-feature` example in `agent_utils/status.py` are canon drift, but both pre-date this PR (present on `main`, relocated by a refactor) and the string is pinned by five existing tests; renaming belongs to a terminology-convergence pass, not this landing.
- **Bounded encoding follow-up (deferred):** a few sibling readers of git metadata / version files (`m_0_13_1_exclude_worktrees.py`, `runtime/bootstrap.py`, `version_utils.py`) lack an explicit `encoding="utf-8"`; they fail soft today (no crash) and touching them would pull untouched/migration files into this diff — left for a focused encoding-consistency sweep.


stijn-dejongh:
## Maintainer landing — rebuilt & handed off (green modulo pre-existing honest-reds)

This PR is now the **canonical full mission**, rebuilt onto current `main` as a clean 6-commit linear history (authorship preserved). It carries the implementation **and** overwrites the planning-only dossier merged via #4663 with the canonical mission dossier + durable Op capsules.

**History**
- `docs(mission)` / `feat(cli)` / `test(cli)` / `build(ruff)` — Robert Douglass (mission work)
- `fix(landing)` / `docs(landing)` — maintainer landing folds

**Verification (rebuilt tip)**
- Whole-repo `ruff format --check .` + `ruff check .`: clean; `[tool.ruff.format].exclude` ratchet green (4 now-clean boundary files dropped).
- 861 targeted tests across the 28 changed files (incl. the json-contract / cli-placeholder / dead-symbols arch gates) + the new red-first scope-config regression: pass.
- Mission dossier byte-identical to the mission branch; this mission's dossier is corpus-clean.

**Second-opinion squad (4 profile-loaded opus lenses)**
- Correctness/fail-closed: no MAJOR; the whole-surface `--json` contract guard pins valid-JSON / no-traceback / canonical envelope / exit-code parity.
- Boundary: **1 MAJOR folded** — `charter/activation/scope._load_charter_scope_config` still read `.kittify/config.yaml` with a bare `read_text()`, silently degrading `spec-kitty next` to the wrong charter context on a corrupt config (the exact #4600 brick-class, missed by FR-004). Now fails loud (`CharterPackConfigError` → `kind=blocked`, file named, no traceback); red→green regression drives the real CLI.
- Design: the `error_handler` seam in `_find_mission_slug` couples rendering with resolver selection (1 of ~15 callers). Not a live defect (sole handler is `NoReturn`); documented as known debt for a CLI-selector convergence mission rather than reworked in a landing.
- Terminology: `"Feature directory not found"` / `057-my-feature` are pre-existing canon drift (on `main`, pinned by 5 tests) — left for a convergence pass.

**Pre-existing honest-reds (NOT this PR, `main`-identical, left as-is)**
- Corpus witness: one `SNAPSHOT_DRIFT` on `ci-terminal-cancel-verdict-01M2NC7Z` (untouched mission; byte-identical to `main`) — candidate for a separate corpus hotfix.
- `mypy --strict`: one `no-any-return` in `tasks_shared.py::unchecked_subtask_ids` (unmodified function; not CI-gated on this file).

Closes #4600, #4601, #4643, #4597, #4598, #4533, #4532. Handing off green-modulo-honest-reds — **operator merges.**

stijn-dejongh:
## Correction: dossier reconciled to the editorial-freeze gate (option B was blocked)

The prior revision overwrote #4663's merged planning dossier — that tripped `tests/architectural/test_archive_root_byte_identical` (the M1 editorial-freeze gate): `kitty-specs/` and `kitty-ops/` are byte-immutable once merged, and #4663 already merged the planning snapshot.

Rebuilt so this PR **only adds** to the frozen dossier: 76 additions (execution-phase artifacts — research, review-evidence, per-WP task/subtask records, traces, `wps.yaml`, `issue-matrix.md`, and the mission's durable Op capsules) plus the append-only `lifecycle.jsonl` extension. The planning files #4663 merged (spec, plan, tasks, contracts, issue-matrix.json, meta, status) are left byte-identical.

Verified locally against canonical `main`: the freeze gate now passes (3/3), whole-repo `ruff format --check` + `ruff check` clean, exclude-ratchet green, the scope-config regression green. Force-pushed; CI re-running.

The implementation, tests, and the folded `fix(landing)` (charter-scope config-decode) are unchanged from the reviewed revision.

PR #4692, opened 2026-09-17
perf(skills): stop full inventory rescan per skill

LynnColeArt:
## Problem

An ordinary CLI command runs the global skill assessment on startup. On a healthy RC2 wheel install, `prune_missing()` rebuilt and sorted the entire prior asset inventory for each skill tree. The profile recorded 220 calls during one warm command, even though the assessment planned zero file changes.

## Change

Index prior inventory entries by ancestor once per assessment. Each prune call then visits only entries below its selected skill tree. The existing depth ordering, selected-file exclusion, and `retire()` ownership checks remain in place.

## Evidence

Isolated RC2 wheel, same warm home, `spec-kitty doctor channel`:

| | Before | After |
|---|---:|---:|
| Global skill assessment | 3.98 s | 0.89 s |
| CLI command wall time | 5.89 s | 2.76 s |
| Planned file changes | 0 | 0 |

The new inventory-scan regression test failed against baseline (two scans for two trees) and passed with the change (one scan). Existing global skill and asset tests passed (96 tests), followed by the new test and focused skill tests (7 tests). Ruff and strict Mypy passed on the changed files. Independent review found no actionable issues.

This addresses the dominant skill audit cost. A warm ordinary command still takes about 2.7 seconds; broader startup reconciliation remains a separate performance concern.


LynnColeArt:
Superseded by #4694. The replacement uses the programme issue/branch/PR handoff format for factory review and merge. It contains the same reviewed commit and fix.

Issue #103, opened 2026-01-25
Workflow instructions missing git commit step - causes cascading failures in parallel WP development


jlacube:
## Problem Description

The `spec-kitty agent workflow implement` command completion instructions do not explicitly tell agents to commit their implementation files before marking tasks done and moving to review. This causes cascading failures in parallel work package development where dependent WPs start from empty branches.

## Impact

**Critical** - Affects all multi-agent parallel development workflows using the workspace-per-WP model (v0.11.0+).

### Observed Failure Pattern

When implementing dependent work packages in parallel:

- ✅ **WP02 worktree**: Has HTML + CSS (implementation was committed)
- ❌ **WP03 worktree**: Has HTML only - missing WP02's CSS (CSS not committed in WP02's branch)
- ❌ **WP04 worktree**: Has HTML only - missing CSS and JS from WP02 and WP03

Each subsequent work package starts from a branch without the previous implementation files because agents follow the completion instructions which skip the git commit step.

## Root Cause

The "WHEN YOU'RE DONE" instructions in [`src/specify_cli/cli/commands/agent/workflow.py`](https://github.com/Priivacy-ai/spec-kitty/blob/main/src/specify_cli/cli/commands/agent/workflow.py) tell agents to:

1. ~~Commit changes~~ ⚠️ **MISSING**
2. Mark subtasks as done
3. Move WP to for_review

**Why this matters:**
- Each WP has an isolated worktree with its own branch created from main
- Dependent WPs need committed changes from their dependencies to build upon
- Without commits, dependent worktrees start with empty branches
- Reviews happen on committed code, not uncommitted files
- Work isn't preserved in the branch without commits
- Creates a cascade of failures: WP02 → WP03 → WP04 all lack previous work

## Current Instructions (Incorrect)

```markdown
WHEN YOU'RE DONE:
✓ Implementation complete and tested:
  1. Mark all subtasks as done first:        ❌ Missing git commit step
     spec-kitty agent tasks mark-status T001 T002 T003 --status done
  2. Move WP to review:
     spec-kitty agent tasks move-task WP## --to for_review --note "Ready for review"

PR #4710, opened 2026-09-18
Fix explicit upgrade version queries in piped output

### LINKED ISSUE #4704: Bug: non-TTY stdout silently disables the version fetch for explicit `upgrade` queries
````
# Bug: non-TTY stdout silently disables the version fetch for explicit `upgrade` queries

## Summary

Every explicit, user-initiated upgrade query silently reports "no upgrade available" whenever stdout is not a TTY. `spec-kitty upgrade --cli` prints zero bytes and exits 0; `spec-kitty upgrade --dry-run --json` returns `"latest_version": null, "latest_source": "none", "is_outdated": false`. The TTY/CI suppression gate — correct for the passive upgrade nag — is also applied to commands the user typed on purpose. `--json` is the worst case: machine-readable output is the one mode guaranteed to be piped, so it can never fetch a version. Any agent harness, script, or CI job that asks spec-kitty whether it is current is told "yes" unconditionally.

## Reproduction

### Prerequisites

- spec-kitty-cli 3.2.7 installed via `uv tool install spec-kitty-cli`
- Working network access to pypi.org, and a newer version published (at time of report: 4.0.0rc3)
- Run from any context where stdout is redirected/piped (agent harness, CI, or `| cat`)
- Nag cache absent, so the fresh-cache fast path is not involved

### Steps

```bash
spec-kitty upgrade --cli > out.txt 2> err.txt; echo "exit=$?"
wc -c < out.txt; wc -c < err.txt

spec-kitty upgrade --dry-run --json
```

### Expected Behavior

`--cli` prints the CLI upgrade hint (its documented purpose: "Show CLI upgrade hint"). `--dry-run --json` reports the actual latest version from PyPI and an accurate `is_outdated`.

### Actual Behavior

```text
exit=0
0        <- stdout bytes
0        <- stderr bytes

{
  "cli": {
    "installed_version": "3.2.7",
    "latest_version": null,
    "latest_source": "none",
    "is_outdated": false,
    "fetched_at": null
  },
  "rendered_human": ""
}
```

## Root Cause

`specify_cli/compat/planner.py`:

```python
def suppresses_network(self) -> bool:
    return self.env_ci or (not self.stdout_is_tty) or self.flag_no_nag

def _default_latest_provider(*, network_suppressed: bool, profile: Any) -> Any:
    if network_suppressed:
        return NoNetworkProvider()
    ...
```

`NoNetworkProvider.get_latest()` returns `version=None, source="none"` without opening a socket, rendered downstream as "not outdated".

Verified against the installed 3.2.7 package:

```text
Invocation.from_argv(["upgrade","--cli"]) -> stdout_is_tty=False env_ci=False flag_no_nag=False
suppresses_network() = True
  stdout_is_tty=False: provider=NoNetworkProvider result=version=None  source='none'
  stdout_is_tty=True : provider=PyPIProvider     result=version='4.2.0a1' source='pypi'
```

The network path is healthy: `PyPIProvider().get_latest("spec-kitty-cli")` returns `'3.2.7'` in 0.34s. Only the provider selection is wrong. The gate is appropriate for the passive nag (`suppresses_nag()`), but should not fire for an invocation whose purpose is to answer "is there a newer version?".

## Workaround Applied

None — reported rather than worked around. The upgrade was completed by pinning the target version explicitly rather than relying on the CLI's own resolution.

## Environment

- OS: Windows 11 Pro 26200
- Python: 3.13.7
- spec-kitty-cli: 3.2.7 (pinned tag SHA fb4e8fa98), repo `spec-kitty/spec-kitty`
- Install method: uv tool install from PyPI
- Channel: prerelease-opt-in (`SPEC_KITTY_PRERELEASE=1`)

---

**Authored by**: Kent Gale (kentonium3/kg-automation) & Claude Code (Claude Opus 5), 2026-09-17.
**Submission approved by**: Kent Gale (kentonium3/kg-automation), 2026-09-17 21:38Z
**Local tracking**: kentonium3/kg-automation#981.
````

LynnColeArt:
Fixes #4704.

Explicit upgrade queries now resolve the latest published CLI version when stdout is redirected or CI is set. `upgrade --cli` renders its guidance in those environments, and `upgrade --dry-run --json` reports the version and accurate `is_outdated` status.

The query remains read-only: version lookup may use fresh cached data or fetch from the provider, but it does not write the nag cache. Passive command nag suppression stays unchanged.

Validation:
- `uv run --frozen pytest -q tests/specify_cli/cli/commands/test_upgrade_command.py tests/specify_cli/compat/test_planner.py` — 129 passed
- `uv run --frozen ruff check src/specify_cli/compat/planner.py src/specify_cli/cli/commands/upgrade.py tests/specify_cli/cli/commands/test_upgrade_command.py` — passed
- `make test-fast` — 1846 passed, 5 skipped


stijn-dejongh:
## 🛬 Maintainer landing — LAND (queue item 3/5)

**Viability:** fresh, focused (+58/-36, 3 files), fixes #4704 — explicit `upgrade --cli` / `--dry-run --json` queries silently reporting a null/"not outdated" status when stdout is piped or `CI=1` (the exact environments agents and CI run in). Rebased onto current `main` (was 12 behind); contributor authorship preserved (`lynncoleart`).

**Second-opinion review (reviewer-renata, opus): APPROVE-WITH-NITS, 0 MAJOR.** Key confirmations:
- The change threads a new `query_latest` flag; `suppresses_network()` is **unchanged** — the `--cli` special-case lives in `suppresses_nag()` (output visibility). Network fetch is driven purely by `query_latest`.
- **RISK-3 inversion is correct, not a regression** — traced to origin mission #103: spec A-007 suppresses the *nag*, not the explicit `upgrade` command. The reverted "RISK-3 fix" was an over-correction that *caused* #4704. Deleting its test is right (it pinned since-reversed behavior).
- **Passive path proven unchanged** (1134 nag/upgrade/compat tests pass; the `and not read_only` cache guard is a no-op on the passive path). **Fail-closed** confirmed: an explicit read-only query fetches but never writes the nag cache; network is fail-soft (2s timeout).

**Fold:** `docs(landing):` changelog entry (Fixed).

**Noted, non-blocking:** (1) the two `… and not query_latest` overrides are line-covered but not behaviorally asserted for the nag-disabled/fresh-cache combination — a future test could pin "explicit query reports outdated even when cache is fresh". (2) `ruff format --check` flags the two touched `compat`/`cli` files, but they are in the pre-existing format-exclude list, base `main` fails identically, and every reformat site is **outside** this diff — the whole-repo `ruff format --check .` gate (what CI runs) is green. No format action taken.

**Local verification:** `test_upgrade_command.py` + `test_planner.py` 129 passed; #4715 freshness gate green post-rebase; ruff lint clean; whole-repo format gate green; terminology 90 passed.

Merging on green.

PR #4712, opened 2026-09-18
Read session project slug from identity config

LynnColeArt:
Fixes #4706.

The SessionStart orientation block now reads `project.slug` through the canonical project identity loader. Live writes and dry-run previews use the same source. The test fixture no longer invents a `project_slug` field on `AgentConfig`.

Validation:
- Session presence and identity suites — 348 passed, 1 skipped
- Ruff on changed files — passed
- `make test-fast` — 1846 passed, 5 skipped

PR #4713, opened 2026-09-18
Refresh session upgrade health by release channel

### LINKED ISSUE #4707: Bug: orientation-block `health` can never report `upgrade-available`
````
# Bug: orientation-block `health` can never report `upgrade-available`

## Summary

The `health` field in the SessionStart orientation block is structurally incapable of reporting `upgrade-available`. `SessionPresenceManager._build_content()` builds its compat `Invocation` with `stdout_is_tty=False` and `flag_no_nag=True` hardcoded. Both independently force `suppresses_network()` to `True`, which selects `NoNetworkProvider`, which returns `version=None` — so `is_outdated` is always false and the block always says `(healthy)`. The banner announces a health status it never computes. In practice this hid an available upgrade: a session reported `(healthy)` while 4.0.0rc3 had been published, and the upgrade was only discovered out-of-band.

## Reproduction

### Prerequisites

- Any initialized spec-kitty project
- spec-kitty-cli 3.2.7 with a newer version published on PyPI (at time of report: 4.0.0rc3)
- `SPEC_KITTY_PRERELEASE=1` so the newer version is in-channel

### Steps

```bash
spec-kitty doctor channel      # confirms prerelease-opt-in is active
spec-kitty session-start       # observe the health marker
```

### Expected Behavior

With 4.0.0rc3 published and the pre-release channel opted in, the orientation block reports `upgrade-available`.

### Actual Behavior

```text
**Spec Kitty v3.2.7** — project: unknown (healthy)
```

`(healthy)` is emitted unconditionally; no state of the world produces `upgrade-available` from this path.

## Root Cause

`specify_cli/session_presence/manager.py:70-78`:

```python
inv = Invocation(
    command_path=("session-start",),
    raw_args=("session-start",),
    is_help=False,
    is_version=False,
    flag_no_nag=True,        # <-- forces suppresses_network() True
    env_ci=False,
    stdout_is_tty=False,     # <-- also forces suppresses_network() True
)
```

Against `specify_cli/compat/planner.py`:

```python
def suppresses_network(self) -> bool:
    return self.env_ci or (not self.stdout_is_tty) or self.flag_no_nag
```

Two of three disjuncts are hardcoded true, so the predicate is a constant `True` here. `_default_latest_provider(network_suppressed=True, ...)` returns `NoNetworkProvider`, whose `get_latest()` returns `version=None, source="none"` without touching the network. With `latest_version` null, `is_version_newer(None, installed)` cannot be true and `health` is pinned to `"healthy"`.

This shares a root cause with the non-TTY suppression bug (filed separately) but needs its own fix: even with `suppresses_network()` corrected, this call site still passes hardcoded values that misrepresent the invocation.

The `session_start.py` module docstring describes a different intended mechanism:

```text
Unless ``SPEC_KITTY_NO_UPGRADE_CHECK`` is set, the upgrade check fires a
background subprocess (``check_in_background()``) that returns immediately;
``get_available_version()`` reads only the local cache file — no network calls
on the hot path.
```

`_build_content` does use that `UpgradeChecker` cache at lines 60-62 for the version line, but routes `health` through the compat planner instead — and that second path is the one wired shut. The two mechanisms appear to have diverged.

## Workaround Applied

None.

## Environment

- OS: Windows 11 Pro 26200
- Python: 3.13.7
- spec-kitty-cli: 3.2.7 (pinned tag SHA fb4e8fa98), repo `spec-kitty/spec-kitty`
- Install method: uv tool install from PyPI
- Channel: prerelease-opt-in (`SPEC_KITTY_PRERELEASE=1`)

---

**Authored by**: Kent Gale (kentonium3/kg-automation) & Claude Code (Claude Opus 5), 2026-09-17.
**Submission approved by**: Kent Gale (kentonium3/kg-automation), 2026-09-17 21:38Z
**Local tracking**: kentonium3/kg-automation#984.
````

LynnColeArt:
Fixes #4707.

Session presence already derives `upgrade-available` from its background version cache. The failing path was the cache refresh: it queried only stable releases and never refreshed an old cached version. This change passes the configured prerelease channel to the provider, records that channel with the cached result, rejects a cache from another channel, and starts a background refresh when the cache exceeds its TTL. SessionStart remains free of foreground network calls.

Validation:
- Session presence and upgrade provider suites — 176 passed, 1 skipped
- End-to-end orientation health check for a fresh prerelease cache — passed
- Ruff on changed files — passed
- `make test-fast` — 1846 passed, 5 skipped


stijn-dejongh:
## 🛬 Maintainer landing — LAND (queue item 5/5, cluster complete)

**Viability:** fresh (+82/-10, 3 files), fixes #4707 — the orientation `health` line pinned to `(healthy)` because its version cache was stable-channel-only and never refreshed. Contributor authorship preserved (`lynncoleart`).

**Rebase note:** was 17 behind; rebasing onto current `main` conflicted in `test_manager.py` (this PR and the already-merged #4712 both append tests to `TestBuildContent`). Resolved **keep-both** — all three tests now coexist (`test_project_slug_comes_from_identity_config`, `test_dry_run_slug_comes_from_identity_config`, `test_health_uses_fresh_prerelease_cache`); verified 32 passed, no dropped/duplicated method.

**Second-opinion review (reviewer-renata, opus): APPROVE-WITH-NITS, 0 MAJOR.** The two headline risks are clean:
- `from kernel.clock import … timedelta` is the **sanctioned** path — `kernel.clock` re-exports `timedelta`/`now_utc` and raw `datetime` imports are banned by `test_clock_import_ban`. No ImportError.
- **No foreground network / no spawn-storm:** `check_in_background()` is fire-and-forget `Popen` (DEVNULL, detached session), returns immediately; the TTL-triggered refresh spawns **≤1 per session** (mutually exclusive with the manager's own `avail is None` spawn) and self-terminates because `refresh_cache_once` always rewrites `checked_at`. Stale-while-revalidate.
- Channel-marker rejection correct across all 4 stable/pre combos, fail-closed (one session of `healthy`, then self-heals).

**Fold:** `docs(landing):` changelog entry (Fixed).

**Noted future enhancements (non-blocking, not ticketed):** (1) promote `_call_provider_get_latest` from `compat.planner` to a public surface — this PR reuses it (correct DRY: it's the single source of truth for channel-aware provider calls) but reaches for a `_`-prefixed name; (2) a channel-keyed cache **filename** would remove a benign, self-healing flip-flop when an operator alternates `SPEC_KITTY_PRERELEASE` between sessions.

**Local verification:** `test_manager.py` + `test_upgrade_checker.py` 32 passed; #4715 freshness gate green; ruff lint clean; whole-repo format gate green; terminology 90 passed.

Merging on green.

stijn-dejongh:
## ⚠️ CI red diagnosed → folded a #4704-regression fix

The first CI run went red: **1 real failure** (`tests/runtime/test_reassess_under_lock.py::...::test_live_upgrade_dry_run_preview_is_deterministic_across_repeats`) + 16 fail-fast **cancellations**.

**Root cause — NOT this PR.** #4710 (fixes #4704, already merged) made an explicit `upgrade --dry-run --json` query resolve the *live* latest version, so the preview now carries a real `cli.fetched_at` wall-clock timestamp that legitimately differs between two runs (diff was purely `…T15:21:06.48…` vs `…T15:21:00.42…`). That broke the byte-identity determinism invariant. #4710's own CI never ran the `execution_context` shard (its path-routing didn't select it for a compat/cli change), so the regression landed on `main` undetected; **this PR's shard surfaced it.** #4713 does not touch that path.

**Fix (operator decision: relax the assertion).** Folded `test(landing):` — the determinism test now normalizes the one volatile `cli.fetched_at` field out and asserts every other byte is stable. Verified locally: the test **failed** before the fix and **passes** after (66s, drives the real `upgrade --dry-run --json` subprocess twice), ruff clean. Sibling `tests/compat/test_dry_run_parity.py` still passes (4/4). CI re-runs all shards to confirm nothing else regressed.

Merging on green.

Issue #4627, opened 2026-09-16
Seven papercuts from a first-run onboarding session


OriPekelman:
Seven small things from one onboarding session. Each was individually trivial and each caused a moment of doubt; filing together rather than as separate issues.

**1. `spec-kitty doctor` with no subcommand prints help and exits 0.** Indistinguishable from a health check that ran and passed — especially given the command is named `doctor` and the user ran it expecting diagnostics.

**2. `charter sync` is a permanent no-op that always reports `synced=False`.** Retained for back-compat per the docs, but a command that cannot succeed, reporting a value that looks like failure, is worse than an explicit "this command was removed".

**3. `specify` buries its deferred-questions notice.** The last line reads "Non-interactive: taking defaults for the specify interview (no prompts; questions recorded as deferred)". Deferred questions nobody sees are deferred forever.

**4. Mission slugs get an opaque ULID suffix.** `our-first-shared-mission-01M2MW3Q` where the onboarding page implies the clean name you typed. Every subsequent command needs the full slug, which has to be copied from scrollback.

**5. The `synthesized_drg missing` warning only surfaces via `next`.** `charter status` reports `SYNCED` with no mention of it. The one command whose job is charter health does not report this particular charter problem.

**6. `next` without `--mission` fails as a usage error.** `Invalid value: --mission <slug> is required` — where listing the available missions would answer the question the user actually has, especially given (4) above.

**7. A cold relay's first load is very slow with no indication.** Documented behaviour ("an idle one is stopped to free its memory and wakes automatically on first use"), and reasonable — but from the user's side, indistinguishable from the outage that had just happened.

## Environment

CLI `3.2.7` → `4.0.0rc2` · macOS · 2026-09-16

stijn-dejongh:
## Decomposition (operator decision, 2026-09-17)

This bundle has been split into individually-tracked tickets, all parented under the onboarding epic **#4414**. Disposition of the seven items:

| # | Item | Disposition |
|---|------|-------------|
| 1 | `doctor` no-subcommand exits 0 | **#4678** — `type:bug` / P2 |
| 2 | `charter sync` reports failure-looking `synced=False` | **#4679** — `type:fix` / P3 |
| 3 | `specify` buries deferred-questions notice | **#4680** — `type:fix` / P2 |
[...]

PR #4770, opened 2026-09-19
chore(config): adopt canonical charter_packs key in .kittify/config.yaml

stijn-dejongh:
## What

Adopt the canonical `charter_packs:` key in `.kittify/config.yaml` for the repo-local internal org pack, replacing the legacy top-level `doctrine:` key.

## Why

The pack resolver (`src/charter/offering/drg/org_pack_config.py`) defines `_CANONICAL_ORG_PACKS_KEY = "charter_packs"` and `_LEGACY_ORG_PACKS_KEY = "doctrine"`. It prefers the canonical key and only reads `doctrine:` through a warn-once legacy fallback, emitting `LegacyOrgPackDoctrineKeyWarning` on every registry load. This repo's own config still used the legacy key, so the warning surfaced during routine commands (e.g. `spec-kitty doctor ops`). Catfooding the pack approach means catfooding the canonical key too.

## Change

`.kittify/config.yaml` only (2 lines):
- top-level key `doctrine:` → `charter_packs:` (nested `org.packs` unchanged)
- adjacent comment "org-tier doctrine pack" → "org-tier charter pack"

## Verification

Resolution tested against the branch config (resolver + `doctor doctrine` collectors pinned to the branch worktree):
- internal pack resolves — org DRG **8 nodes / 6 edges**, source = `packs/internal`
- **25/25** builtin agent profiles valid and available (none skipped/invalid)
- **0** collisions, **0** dangling endpoints, **0** errors, **0** unresolved procedures
- **0** legacy-key warnings (was 1 before the change; confirmed by before/after probe of `load_pack_registry(quiet=False)`)

Regression scope: the top-level `doctrine:` config key is read **only** by the legacy org-packs fallback. The other `doctrine`-key reader (`_doctrine_collect.py:938`) targets `charter.yaml`'s nested `governance.doctrine` sub-block — a different file and key — and is unaffected. `.kittify/config.yaml` is the only config carrying the legacy top-level key. `save_pack_registry` writes exactly this `charter_packs.org.packs` shape, so the hand-edit round-trips without re-drift; no migration rewrites this key, so a manual edit is the sanctioned remediation.

## Review

[...]

PR #4772, opened 2026-09-19
fix(routes): match the no-origin remedy to its diagnosis

### LINKED ISSUE #4626: routes with no origin remote suggests a remedy that does not match the diagnosis
## Summary

`spec-kitty routes` in a repository with no `origin` remote names a remedy that does not match the cause.

```
$ spec-kitty routes
Error: could not identify this checkout ('/Users/…/reveil-eve' has no `origin` remote
(live or in `.git/config`) and no quarantine record — refusing to mint an identity from
a directory name); run this from inside a Spec Kitty mission repository.
```

We **were** inside a Spec Kitty mission repository — initialized, with `.kittify/`, a charter, and commits. The real cause is stated correctly in the parenthetical; the closing sentence then sends the user somewhere else.

## Suggested

End with the fix that matches the diagnosis:

> add an `origin` remote — repository identity is derived from it, never from the directory name.

The refusal itself is good and well-reasoned (the module docstring explains the anti-spoofing rationale clearly). Only the closing remedy misleads.

## Minor

"quarantine record" appears in user-facing output with no explanation. A reader hitting this has no way to know whether it is something they should have, or how to get one.

## Environment

CLI `4.0.0rc2`

DevShoaib78:
## Summary

- `spec-kitty routes` in a checkout with no `origin` remote now ends with the remedy that matches its diagnosis: add an `origin` remote, because identity derives from it and never from the directory name. It no longer sends the user "inside a Spec Kitty mission repository" they are already in.
- The "no quarantine record" part of that message is now explained, and it actually renders: the `[kitty "quarantine"]` config-section name was being swallowed by Rich as a markup tag.

Fixes #4626.

## Why Now

First-run onboarding papercut (#4627): the diagnosis in the parenthetical was right, the closing sentence pointed elsewhere, and a term appeared with no explanation (#4626).

## What This PR Does

- `src/specify_cli/cli/commands/routes.py`: catch `UnverifiedRepositoryIdentity` before the generic `RepoIdentityError` and close with `Add an origin remote (git remote add origin <url>) ... never from the directory name.` The generic branch (`AmbiguousRepositoryIdentity`, containers/symlinks) keeps its old sentence, since "add an origin" would be the wrong remedy there. Both branches now `escape()` the exception text before it goes through Rich.
- `src/specify_cli/zeitgeist_client/repo_identity.py`: the exception text names the record as what it is, "a `[kitty "quarantine"]` record in `.git/config` preserving a former origin", instead of the bare "quarantine record". The refusal logic is unchanged: an identity is still never minted from a directory name.
- New test `test_checkout_without_an_origin_is_told_to_add_one` in `tests/cli/commands/test_routes_command.py` (removes `origin` from the fixture clone, asserts the diagnosis, the explained record, the new remedy, and the absence of the old sentence). No existing test pinned the old wording.
- Changelog entry under `[Unreleased]` / Fixed.

No formatting-only diffs.

- Runtime / compatibility: message text only; exit code 1 unchanged. Other callers that print `RepoIdentityError` text through Rich (`saas_client/project_authority.py`, `status/zeitgeist_bridge.py`, `zeitgeist_client/resolution.py`) were left alone to keep this scoped; they may want the same `escape()`.
- Upgrade / migration: none.
- Operator / reviewer impact: none.

## Validation

- [x] Relevant tests pass locally or in CI: `pytest tests/cli/commands/test_routes_command.py` 21 passed on Windows including the new test. `tests/zeitgeist_client/test_repo_identity.py` has 15 pre-existing Windows-only failures (`_name_from_url` on local-path origins asserting a bare name against a backslash path), unrelated to this change. Architectural gates (`test_completion_manifest_freshness`, `test_docs_cli_reference_parity`, `test_no_dead_symbols`, `test_ruff_format_enforcement`): 46 passed. `ruff check src/` and `ruff format --check .` clean. Live: `spec-kitty routes` in a fresh `git init` repo with `.kittify/` and no remotes prints the new message, bracketed section name included, exit 1.
- [x] Backward compatibility impact considered
- [x] Follow-up work called out if intentionally deferred

## Tickets / Contracts

| Ticket | Relationship |
|---|---|
| #4626 | Fixes |
| #4627 | Parent onboarding papercut list |

## Follow-ups

- Apply `escape()` at the other three call sites that print `RepoIdentityError` text through Rich, if wanted.

---

AI disclosure (per CONTRIBUTING): this PR was written primarily by Claude Code (Anthropic), including the test and changelog changes, working from the issue. It was verified by running the test sets above and the live command on Windows.

---

## Maintainer landing note

Rebased onto current `main`; verified. **The flagged follow-up (apply `escape()` at the other three `RepoIdentityError` call sites) is unnecessary** — those sites do not render the exception through Rich: `saas_client/project_authority.py` re-raises a fixed-string `SaasConsentError`, and `status/zeitgeist_bridge.py` + `zeitgeist_client/resolution.py` use `logger.debug(...)` (the logging framework does not interpret Rich `[...]` markup). The markup-swallow bug is specific to `console.print()`, which only `routes` uses — and both of its branches are now escaped, so the fix is complete.


stijn-dejongh:
## Maintainer landing pass — verified, ready to merge

**Verdict: LAND.** A correct, well-scoped fix for a real first-run papercut (#4626) still live on current `main`. Rebased (was 7 behind), red-first proven, all gates green. No folds needed. Handing off for merge.

### Triage
- **Still needed:** yes — on `main`, `routes.py:80` catches the generic `RepoIdentityError` and closes with "run this from inside a Spec Kitty mission repository" (the mismatched remedy), and `_QUARANTINE_SECTION` is printed unescaped. Not superseded.
- **Boundary check:** `zeitgeist_client/repo_identity.py` is **local client code**, not an upstream parity port — its own docstring states it is a deliberately-hardened reimplementation (Z6-C) that refuses upstream's directory-name fallback. Editing its exception text here is in-bounds.

### Correctness (focused review; no squad — message/UX + error-handling, not architectural)
- **Exception catch-ordering is correct.** `UnverifiedRepositoryIdentity` and `AmbiguousRepositoryIdentity` are siblings under `RepoIdentityError`; catching the `Unverified` subclass **before** the parent gives the no-origin case the "add an `origin`" remedy while the ambiguous-container/symlink case falls through to the generic branch and keeps its old sentence. Correct routing.
- **Both branches now `escape()`** the exception text before Rich renders it; exit code stays 1 (`_fail` → `typer.Exit(1)`). Refusal logic unchanged (identity never minted from a directory name).
- **Red-first:** with the pre-fix `routes.py` + `repo_identity.py` swapped back in, the new test fails on the exact old message; with the fix it passes.
- **The flagged follow-up is unnecessary** (see landing note in the PR body): the other three `RepoIdentityError` sites don't render through Rich (one re-raises a fixed string, two `logger.debug`), so there is no latent markup-swallow bug there. Nothing to fold or ticket.

### Evidence (rebased tip)
- `tests/cli/commands/test_routes_command.py` → **21 passed** (incl. the new test).
- `no_dead_symbols` (34), terminology guard (90), completion-manifest freshness (byte-identical — no help change) → green. `check_docs_freshness --ci` errors=0.
- `ruff check` clean; whole-repo `ruff format --check .` clean; changelog is consumer-focused under `[Unreleased] / Fixed`.

Single-commit history, authorship intact, PR un-drafted, body carries the landing note. No landing-order constraints. **Operator merges** (or say the word and I'll merge on green, as with #4770/#4743).

Issue #4773, opened 2026-09-19
Two org-pack cascade non-vacuity tests red on main (stale monkeypatch post-#4591 + no-cascade requires-activation semantics)


stijn-dejongh:
## Summary

Two org-pack cascade non-vacuity tests in `tests/specify_cli/cli/commands/charter/test_org_cascade_chain.py` are **red on `main`** (verified on `f0dff0baa8`), independent of any current mission. Root-caused during the umbrella-#2899 CLI-error-surface mission and split out here because they are charter-activation-subsystem concerns, not the uncaught-error class.

## Failing tests
- `TestIdMappingWideningNonVacuous::test_two_pack_cascade_fails_without_id_mapping_org_roots`
- `TestNoCascadeWarningNamesOrgPackArtifacts::test_absent_cascade_warns_about_org_pack_requires_target`

## Root cause (two distinct issues)

**1. Stale monkeypatch target (test 1).** The test monkeypatches `_cascade_shared.resolve_config_id` to simulate pre-T008 (layer_roots-only) ID mapping. But `resolve_config_id` was relocated to `src/charter/activation/kind_vocabulary.py:546` (by the #4591 "de-duplicate cascade config-id resolution" refactor), and its callers there (`kind_vocabulary.py:443,661,685`) use the module-global name — so patching `_cascade_shared.resolve_config_id` no longer intercepts the ID-mapping path. The "revert" is a no-op → `c-directive` still cascade-activates → the non-vacuity assertion fails. Likely fix: re-pin the monkeypatch to `charter.activation.kind_vocabulary.resolve_config_id` and confirm the sibling two-pack test still passes (non-vacuity preserved).

**2. Behavioral assertion (test 2).** Without `--cascade`, the test asserts an org-pack `requires` target (`b-directive`) is only *warned about*, NOT activated — but `b-directive` is currently activated. This is NOT a stale mock; it is a charter-cascade **semantics** disagreement (does a no-cascade activation pull in org-pack `requires` targets?). Needs charter-domain adjudication: real regression in cascade activation vs. stale test expectation.

## Why filed separately
Per doctrine (charter/doctrine issues route to their own mission cycles; never self-fix doctrine from an unrelated mission), and because test 2 needs charter-cascade domain judgment a CLI-error mission cannot safely make (a wrong "fix" could make the non-vacuity guard vacuous or mask a real cascade regression).

## Environment
[...]

Issue #2878, opened 2026-07-23
next --mission containing '..' raises an unhandled ValueError traceback (traversal guard fires but except clause only catches _StatusReadPathNotFound)


robertDouglass:
## Summary
`spec-kitty next --agent <a> --mission <slug>` prints an **unhandled `ValueError` traceback** when the mission slug contains `..`. The path-traversal guard correctly *detects* the bad slug (fail-closed — no traversal occurs), but the `ValueError` it raises is not caught by the caller's `try/except`, so a raw Python stack trace is dumped to the user instead of a clean typed error.

## Version
`spec-kitty-cli 3.2.5`, macOS, Python 3.11.

## Repro (deterministic)
```bash
export SPEC_KITTY_NON_INTERACTIVE=1 SPEC_KITTY_ENABLE_SAAS_SYNC=0
mkdir p && cd p && git init -q
spec-kitty init --ai claude --non-interactive && git add -A && git commit -qm init
spec-kitty next --agent claude --mission ../traversal   # -> traceback, exit 1
```

## Actual output (excerpt)
```
.../specify_cli/missions/_read_path_resolver.py:577 in _resolve_mission_read_path
❱ 577   assert_safe_path_segment(mission_slug)
.../specify_cli/core/paths.py:85 in assert_safe_path_segment
❱  85   raise ValueError(
ValueError: Not a safe path segment: '../traversal' — value must not contain '..' (traversal guard).
```

## Root cause
- `core/paths.py:85` `assert_safe_path_segment()` raises **`ValueError`** on any `..` substring (correct guard).
- `missions/_read_path_resolver.py` calls it at line 577 inside `_resolve_mission_read_path`, but the enclosing handler (`_read_path_resolver.py:113`) only does `except _StatusReadPathNotFound`. The `ValueError` is out of that except's scope, so it propagates unhandled to the top-level CLI and prints a full traceback.

## Expected
A malformed/unsafe `--mission` value should be rejected with a **clean typed error message + non-zero exit** (the same way empty/oversized/unicode inputs already are — those return clean `exit 2`/`exit 1` with no traceback). It should not dump a Python stack trace.

## Fix sketch
Either catch `ValueError` (or a dedicated `UnsafePathSegment` error) alongside `_StatusReadPathNotFound` at `_read_path_resolver.py:113` and convert it to the CLI's normal typed-error path, or validate the slug at the CLI boundary before resolution.

## Scope note
[...]

Issue #66, opened 2026-01-06
problems with create-feature in /specify


sdementen:
trying the 0.10.9 versio, with the simple /specify prompt, Claude Opus 4.5 has some issues

```
(base) PS C:\...> spec-kitty agent create-feature --json --feature-name "My Great Feature" --mission "software-dev"
Usage: spec-kitty agent [OPTIONS] COMMAND [ARGS]...
Try 'spec-kitty agent --help' for help.
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮│ No such command 'create-feature'. Did you mean 'feature'?                                                                                                                                            │╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯(base) PS C:\...> spec-kitty agent feature --json --feature-name "My Great Feature" --mission "software-dev"
Usage: spec-kitty agent feature [OPTIONS] COMMAND [ARGS]...
Try 'spec-kitty agent feature --help' for help.
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮│ No such option: --json                                                                                                                                                                               │╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯(base) PS C:\...> spec-kitty agent feature --help

 Usage: spec-kitty agent feature [OPTIONS] COMMAND [ARGS]...

 Feature lifecycle commands for AI agents

╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --help          Show this message and exit.                                                                                                                                                          │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ create-feature        Create new feature with worktree and directory structure.                                                                                                                      │
│ check-prerequisites   Validate feature structure and prerequisites.                                                                                                                                  │
│ setup-plan            Scaffold implementation plan template.                                                                                                                                         │
│ accept                Perform feature acceptance workflow.                                                                                                                                           │
[...]

PR #4774, opened 2026-09-19
Canonical guarded-read + CLI error-presentation seam — close the uncaught-error-surface class (umbrella #2899)

### LINKED ISSUE #2899: Umbrella: harden traversal-guard error surfaces (CLI traceback + orchestrator-api JSON envelope)
## Why

The `..`-traversal guard in `core/paths.py` (`assert_safe_path_segment`) correctly
*rejects* a malformed `--mission` slug — it fails closed, so there is no security
gap. But the **error surface** it produces is broken on two invocation paths:

- **CLI** (`spec-kitty next` / `research`): the guard raises `ValueError`, but the
  surrounding `except` only catches `_StatusReadPathNotFound`, so the operator sees
  an unhandled Python traceback instead of a clean typed error (#2878).
- **orchestrator-api**: the shared `_resolve_mission_dir_or_fail` seam
  (`commands.py:311`) likewise catches only `StatusReadPathNotFound`, so a raw
  traceback escapes on stdout **instead of the JSON error envelope** the external
  machine contract guarantees (#2879).

`spec-kitty merge` already handles this input cleanly and is the fix exemplar.

## For whom

- **Agents / operators** driving `next`/`research` — a traceback on a typo'd mission
  handle reads as a crash, not a validation error.
- **External integrators** consuming orchestrator-api — a non-JSON traceback breaks
  the machine contract they build against.

## Intended effect

Malformed / traversal-bearing `--mission` input produces a **clean typed error on
the CLI** and a **well-formed JSON error envelope on orchestrator-api**, matching
`merge`'s behavior. The two surfaces differ (typed CLI error vs JSON envelope), so
they remain distinct issues, but they share one root guard and can land in one PR.

Children: #2878 (CLI traceback), #2879 (orchestrator-api envelope).

### LINKED ISSUE #4637: bug: `agent release prep` crashes with uncaught FileNotFoundError when pyproject.toml is absent (every non-Python project)
## Summary

`spec-kitty agent release prep --channel <x>` crashes with an uncaught `FileNotFoundError` traceback when the project has no `pyproject.toml`. Since Spec Kitty manages projects in any language, this means `release prep` **always crashes in a non-Python project**. With `--json`, a machine consumer gets a Rich traceback instead of a JSON error object.

Env: `spec-kitty-cli 4.0.0rc2`.

## Reproduction

```bash
# any Spec Kitty project without a pyproject.toml (e.g. a JS/Go/Rust repo, or a bare kitty-specs project)
spec-kitty agent release prep --channel alpha
spec-kitty agent release prep --channel alpha --json    # same crash; not JSON
```

## Actual

```text
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ .../specify_cli/release/payload.py:70 in <version helper>                    │
│   ❱ 70   with pyproject_path.open("rb") as fh:                               │
╰──────────────────────────────────────────────────────────────────────────────╯
FileNotFoundError: [Errno 2] No such file or directory: '<repo>/pyproject.toml'
```

## Root Cause

`src/specify_cli/cli/commands/agent/release.py`, `prep` calls the payload builder with no error boundary:

```python
def prep(channel, repo_root=Path("."), json_output=False) -> None:
    payload = build_release_prep_payload(channel=channel, repo_root=repo_root.resolve())
    if json_output:
        console.print_json(_json.dumps(asdict(payload)))
    else:
        _render_text(payload)
```

`build_release_prep_payload` reaches the version helper in `src/specify_cli/release/payload.py`, which reads `pyproject.toml` unconditionally:

```python
def <read_version>(repo_root) -> str:
    """...
    Raises:
        FileNotFoundError: if pyproject.toml does not exist.
        KeyError: if the [project] table or version key is absent.
    """
    pyproject_path = repo_root / "pyproject.toml"
    with pyproject_path.open("rb") as fh:      # <-- uncaught FileNotFoundError
        data = tomllib.load(fh)
    version = data["project"]["version"]        # <-- uncaught KeyError
```

The helper's own docstring documents both `FileNotFoundError` and `KeyError`, but neither is caught at the `prep` command boundary, so both escape as raw tracebacks. (A malformed/missing `[project].version` would surface the same way via `KeyError`/`TypeError`.)

## Expected Behavior

`agent release prep` should fail with a clean, operator-readable error when `pyproject.toml` is missing or lacks `[project].version` — and, under `--json`, emit a JSON error object on stdout (the `doctor skills --json` shape, cf. #4242/#4601) rather than a traceback. If Python packaging is a hard prerequisite, it should say so explicitly.

## Notes

Found via a sweep of the `agent` programmatic subcommands in a sandboxed non-Python project (a bare git repo initialized with `spec-kitty init`, in `/tmp`; nothing under `$HOME` touched).

### LINKED ISSUE #4720: specify --json ignores its JSON contract for name validation (empty stdout, exit 2) + non-ASCII names wrongly rejected as "cannot be empty"
## Summary

`spec-kitty specify` mishandles feature-name validation in two related ways that share a single root cause in `_slugify_feature_input`:

1. **`--json` contract violation** — a feature name that slugifies to empty is reported as a human-formatted Rich panel on **stderr** with **exit code 2** and **nothing on stdout**, even though other errors in the same command (unknown mission type, repo-not-initialized) correctly emit a structured JSON error object on stdout with exit code 1. Agents/automation that call `specify --json` and parse stdout get no error object and an inconsistent exit code.
2. **Misleading message + silent non-ASCII data loss** — any name written in a non-Latin script (CJK, Cyrillic, emoji, etc.) is rejected as `Feature name cannot be empty.` even though the user did provide a name. Accented-Latin and mixed names silently lose their non-ASCII characters instead of being transliterated or rejected clearly.

Installed version: **4.0.0rc3** (`spec-kitty --version` → `spec-kitty-cli version 4.0.0rc3`).

## Reproduction

In an initialized project (`spec-kitty init --ai claude --non-interactive` + `git init`):

```console
# (1) --json error is NOT json, goes to stderr, exit 2
$ spec-kitty specify '日本語' --mission-type software-dev --json 1>stdout.txt 2>stderr.txt; echo "exit=$?"
exit=2
$ wc -c < stdout.txt
0                        # <-- empty stdout under --json
$ cat stderr.txt
Usage: spec-kitty specify [OPTIONS] {mission}
Try 'spec-kitty specify --help' for help.
╭─ Error ──────────────────────────────────────────────────────────────────────╮
│ Invalid value: Feature name cannot be empty.                                  │
╰──────────────────────────────────────────────────────────────────────────────╯

# Contrast: another error in the SAME command DOES honor --json (stdout, exit 1)
$ spec-kitty specify foo --mission-type notatype --json 2>/dev/null; echo "exit=$?"
{"error": "Unknown mission type 'notatype'. Registered types: documentation, plan, research, software-dev.", "spec_kitty_version": "4.0.0rc3"}
exit=1
```

```console
# (2) Misleading message / silent character loss
$ spec-kitty specify '日本語'    --mission-type software-dev   # -> "Feature name cannot be empty." (name was NOT empty)
$ spec-kitty specify '🐱🐱'      --mission-type software-dev   # -> "Feature name cannot be empty."
$ spec-kitty specify 'Ünïcödé'   --mission-type software-dev --json   # -> slug "n-c-d" (é ü ç dropped)
$ spec-kitty specify 'auth日本'  --mission-type software-dev --json   # -> friendly_name "auth" (日本 silently dropped)
```

## Root cause

`src/specify_cli/cli/commands/lifecycle.py`:

```python
def _slugify_feature_input(value: str) -> str:
    """Normalize a free-form feature name to kebab-case slug text."""
    slug = re.sub(r"[^a-z0-9]+", "-", value.strip().lower()).strip("-")
    if not slug:
        raise typer.BadParameter("Feature name cannot be empty.")   # <-- always Rich/stderr/exit-2, ignores --json
    return slug
```

- The regex `[^a-z0-9]+` strips every non-ASCII-alphanumeric character, so any non-Latin-script name reduces to `""` and hits the `cannot be empty` branch; accented/mixed names lose characters.
- `typer.BadParameter` renders a Rich usage panel to stderr with exit code 2 unconditionally. It runs before / outside the JSON-aware error path used by `_enforce_initialized(..., json_output=...)` and by the unknown-mission-type check (which return `{"error": ...}` on stdout, exit 1). Hence the inconsistency.

## Expected

- Under `--json`, **all** `specify` errors — including name validation — should be emitted as a JSON error object on stdout with a consistent exit code (matching the existing `{"error": ...}` / exit 1 behavior).
- The message for a name that cannot be slugified should say the name has no usable characters and show the offending value, not "cannot be empty".
- Non-ASCII names should either be transliterated or rejected explicitly, not silently truncated (`Ünïcödé` → `n-c-d`, `auth日本` → `auth`).

## Impact

- Breaks programmatic/agent consumers of `specify --json`, which is the mode meant for automation.
- Confusing UX for non-English project names; silent data loss on accented/mixed names.

## Environment

- spec-kitty-cli 4.0.0rc3, Python 3.13 (uv tool install), Linux.

### LINKED ISSUE #4724: mission close crashes with an uncaught UnsafePathSegmentError traceback on an invalid --mission (every sibling command guards it); plus -f=--mission footgun
## Summary

`spec-kitty mission close` crashes with an **uncaught Python traceback** when `--mission` is given a value that is not a safe path segment. Every other mission-addressing command (`next`, `accept`, `mission reopen`, …) handles the identical input gracefully — `mission close` is the lone command that lets `UnsafePathSegmentError` escape.

A contributing footgun makes this easy to hit: on `mission close`, the short flag for `--mission` is **`-f`** — and this is the *only* command where `--mission` has a short flag at all. `-f` overwhelmingly reads as "force", and `mission close` has a destructive `--discard`/`--force` path, so `spec-kitty mission close -f --discard` (intending "force discard") instead binds `--discard` as the *value* of `-f`/`--mission` and crashes.

Installed version: **4.0.0rc3**.

## Reproduction

```console
# invalid --mission value -> uncaught traceback (crash)
$ spec-kitty mission close --mission 'bad seg!'
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ .../specify_cli/cli/commands/mission_type.py:571 in close_cmd                │
│ ❱ 571 │   feature_dir = resolve_feature_dir_for_mission(repo_root, mission_slug)
│ .../specify_cli/missions/_read_path_resolver.py:1647 in resolve_feature_dir_for_mission
│ .../mission_runtime/resolution.py:2258 in resolve_action_context
│ ...
│ UnsafePathSegmentError: Not a safe path segment: 'bad seg!' ...
╰──────────────────────────────────────────────────────────────────────────────╯
$ echo $?
1

$ spec-kitty mission close --mission '..'      # same crash

# the -f footgun that lands you here without a "bad" value in sight:
$ spec-kitty mission close -f --discard        # '-f' is --mission, so mission_slug='--discard' -> same traceback
```

### Every sibling command guards the same input — only `mission close` crashes

```console
$ spec-kitty next --agent claude --mission 'bad seg!'
Error: Mission slug is not a single safe path segment: Not a safe path segment: 'bad seg!' ...   # clean, exit 2

$ spec-kitty accept --mission 'bad seg!'
Error: No mission found for handle "bad seg!". ...                                                # clean, exit 2

$ spec-kitty mission reopen --mission 'bad seg!'
Usage: spec-kitty mission reopen [OPTIONS] {handle} ...                                           # clean typer error
```

None of these print a traceback.

## Root cause

`src/specify_cli/cli/commands/mission_type.py`, `close_cmd` (~line 571):

```python
feature_dir = resolve_feature_dir_for_mission(repo_root, mission_slug)
if not feature_dir.exists():
    console.print(f"[red]Mission not found:[/red] {mission_slug}")
    raise typer.Exit(1)
```

`resolve_feature_dir_for_mission` → `resolve_action_context` runs the canonical path-segment guard, which raises `UnsafePathSegmentError` for any non-segment value. `close_cmd` calls it **unguarded**, so the exception propagates to the top-level handler as a raw traceback. The sibling commands wrap this same resolution and turn the guard failure into a clean "not a safe path segment" / "mission not found" message; `close_cmd` needs the same try/except.

This is the same defect class as the already-tracked uncaught-traceback bugs #4600 (malformed config, P0) and #4642 (corrupt state crashes `next`, P1) — a bad/invalid input reaching a user-facing command as a stack trace instead of an actionable error.

## Secondary: `-f` as the short flag for `--mission`

- `mission close` is the only mission command whose `--mission` has a short flag, and it chose `-f`.
- `-f` conventionally means `--force`; `mission close` even has a destructive `--discard` (plus a separate long `--force` to skip its confirmation), so `-f` here is actively misleading.
- Suggest dropping the `-f` alias (or repurposing it, if anything, for the real force) so it matches the other commands and doesn't invite the crash above.

## Expected

- `mission close` should reject an invalid `--mission` value with a clean, actionable error (matching `next`/`accept`), never a traceback.
- `--mission`'s short flag should not be `-f` on a command with a destructive discard path.

## Environment

- spec-kitty-cli 4.0.0rc3, Python 3.13 (uv tool install), Linux.

### LINKED ISSUE #4738: workflow import crashes with uncaught traceback on any invalid file (malformed YAML, wrong schema, wrong top-level type)
## Summary

`spec-kitty workflow import <file>` crashes with an **uncaught traceback** on any invalid input file — malformed YAML, valid YAML with the wrong schema, or a YAML value of the wrong type. The command never turns a bad import file into a clean, actionable error, even though importing a hand-authored workflow YAML is exactly where mistakes happen.

Installed version: **4.0.0rc3**.

## Reproduction

```console
# 1) malformed YAML  ->  uncaught yaml.parser.ParserError
$ printf ':\n  - [unclosed\n\tbad: {' > bad.yaml
$ spec-kitty workflow import bad.yaml
╭─ Traceback (most recent call last) ─╮
│ .../cli/commands/workflow.py:62 in import_workflow
│ .../yaml/__init__.py:125 in safe_load
│ ...
│ ParserError: while parsing a block mapping ...
╰─...

# 2) valid YAML, wrong schema (unknown fields)  ->  uncaught pydantic ValidationError
$ printf 'just_a_string_not_a_workflow: true\n' > wrong.yaml
$ spec-kitty workflow import wrong.yaml
... For further information visit https://errors.pydantic.dev/2.13/v/extra_forbidden

# 3) valid YAML, wrong top-level type (list instead of mapping)  ->  uncaught pydantic ValidationError
$ printf '[]\n' > list.yaml
$ spec-kitty workflow import list.yaml
... [type=model_type, input_value=[], input_type=list]
    For further information visit https://errors.pydantic.dev/2.13/v/model_type
```

All three surface a raw Python traceback and exit non-zero via the top-level handler; none is presented as a user-facing "invalid workflow file" error.

## Root cause

`src/specify_cli/cli/commands/workflow.py`, `import_workflow` (line 62):

```python
@app.command(name="import")
def import_workflow(source: Path = ..., ...):
    """Import a workflow YAML into `.kittify/overrides/workflows`."""
    workflow = load_workflow_file(source)     # <-- unguarded: yaml.safe_load + pydantic model validation
    ...
```

`load_workflow_file` (`runtime/next/_internal_runtime/workflow_registry.py`) does `yaml.safe_load(...)` and then constructs a pydantic model. Neither the `yaml.YAMLError` from a parse failure nor the `pydantic.ValidationError` from a schema/type mismatch is caught, so both propagate as tracebacks. `import_workflow` calls it directly on an operator-supplied path with no try/except.

(`export` at line 45 also calls `load_workflow_file`, so a corrupt resolved/override workflow would crash it too, though that path is less operator-facing than `import`.)

## Expected

- `workflow import` should catch `yaml.YAMLError` and `pydantic.ValidationError` (and the wrong-top-level-type case) and report a clean, actionable message naming the file and the problem, exiting non-zero — not a stack trace. This matches how the majority of commands already handle bad input, and the defect class the project already treats as a bug (#4600 malformed config, #4642 corrupt state).

## Impact

- Any operator importing a hand-authored or edited workflow YAML gets a Python stack trace on the first typo or schema slip, obscuring what to fix. Low blast radius (optional command) but poor UX and inconsistent with the rest of the CLI.

## Environment

- spec-kitty-cli 4.0.0rc3, Python 3.13 (uv tool install), Linux.

### LINKED ISSUE #4739: intake - (stdin) crashes with uncaught UnicodeDecodeError on non-UTF-8 input; file path handles it gracefully
## Summary

`spec-kitty intake -` (read the mission brief from stdin) crashes with an **uncaught `UnicodeDecodeError`** when stdin contains non-UTF-8 bytes. The equivalent file path — `spec-kitty intake <file>` on the same bytes — is handled gracefully (clean exit 1). The intake code even has a handler meant to catch this, but it never fires because the decode happens during the stream read, outside the guarded block.

Installed version: **4.0.0rc3**.

## Reproduction

```console
# stdin path: crash
$ head -c 500 /dev/urandom | spec-kitty intake -
╭─ Traceback (most recent call last) ─╮
│ .../cli/commands/intake.py:308 in intake
│ .../specify_cli/intake/scanner.py:216 in read_stdin_capped
│ ... in decode:325
│ UnicodeDecodeError: 'utf-8' codec can't decode byte 0x9f in position 0: invalid start byte
╰─...
$ printf '\xff\xfe\x00\x01 not utf8' | spec-kitty intake -    # same crash

# file path on identical bytes: graceful
$ head -c 500 /dev/urandom > rand.bin
$ spec-kitty intake rand.bin; echo "exit=$?"
... (clean error) ...
exit=1
```

## Root cause

The caller passes the **text-mode** `sys.stdin` and catches `IntakeFileUnreadableError`, expecting the helper to wrap decode failures (`src/specify_cli/cli/commands/intake.py:308`):

```python
if path == "-":
    try:
        content = read_stdin_capped(sys.stdin, cap=cap)     # text stream
    except IntakeTooLargeError as exc: ...
    except IntakeFileUnreadableError as exc:                 # <-- expects decode errors wrapped as THIS
        err_console.print(f"[red]Could not read stdin: {exc.__cause__}[/red]")
        raise typer.Exit(1) from None
```

But `read_stdin_capped` (`src/specify_cli/intake/scanner.py`) only guards the *later* `buf.decode()`, not the `stream.read()`:

```python
    buf = stream.read(cap + 1)          # line 216: a TEXT stream decodes HERE -> raw UnicodeDecodeError escapes
    ...
    try:
        decoded = buf.decode("utf-8")   # line ~235: guarded, but only reached when buf is bytes
        return str(decoded)
    except UnicodeDecodeError as exc:
        raise IntakeFileUnreadableError(path=label, cause=exc) from exc
```

Because `sys.stdin` is a `TextIOWrapper`, `.read()` performs the UTF-8 decode itself and raises `UnicodeDecodeError` at line 216 — *before* the guarded `buf.decode()` and outside its `try`. The exception is neither an `IntakeTooLargeError` nor an `IntakeFileUnreadableError`, so the caller's handlers miss it and it propagates as a traceback.

The file reader in the same module (scanner.py:189–190) catches `UnicodeDecodeError` → `IntakeFileUnreadableError` correctly, which is why `intake <file>` degrades gracefully — the stdin path just doesn't get the same treatment.

## Suggested fix

- Pass `sys.stdin.buffer` (binary) to `read_stdin_capped` so it receives bytes and the existing guarded `buf.decode()` handles the failure, or
- Wrap the `stream.read()` at scanner.py:216 in the same `try/except UnicodeDecodeError -> IntakeFileUnreadableError`.

Either makes `intake -` match the file path and the caller's already-present `IntakeFileUnreadableError` handler.

## Impact

- Piping any non-UTF-8 content into `intake -` (a binary file, a file in a non-UTF-8 encoding) yields a Python traceback instead of the intended "Could not read stdin" message. Same defect family as #4600 (non-UTF-8 config) and #4738 (workflow import), scoped to `intake` stdin.

## Environment

- spec-kitty-cli 4.0.0rc3, Python 3.13 (uv tool install), Linux.

### LINKED ISSUE #4746: Canonical guarded-read + CLI corrupt-state presentation seam (subsumes #4600, #4642)
## Problem — a recurring class, not a one-off

Corrupt per-mission/config state files (malformed JSON, non-UTF-8, wrong-shape) crash CLI commands with uncaught tracebacks whenever a `(reader, command-boundary)` pair hasn't independently opted into fail-closed handling. There is **no canonical *guarded-read → typed domain error → CLI-boundary presentation* primitive**; fail-closed rests on an unenforced two-half convention (the reader must wrap raw decode failures into a typed error, AND the command boundary must catch it). A crash happens the instant either half is skipped.

This is at least the 4th occurrence of the class — `core/agent_config.py:78` (the #4600 fix) comments *"same defect class as ledger SK-16, charter status --json's leaked-exception bug."* #4642 fixed two more instances (`next`/`meta.json`, decisions/`index.json`) on the proven `MissionMetaReadError` precedent, but deliberately did **not** build the shared seam (too broad for a P0 release blocker).

## Scope — close the class

1. **Canonical `read_json_state()` guarded-read primitive** (kernel) that wraps `read_bytes` + `decode_meta` + schema-validate into one typed domain-read error, so a new state-file reader cannot ship crashing by omission. Consolidate the existing ad-hoc typed errors (`MissionMetaReadError`, `DecisionIndexReadError`, `CorruptLanesError`, `AgentConfigError`, …) behind it where practical.
2. **Single CLI-boundary corrupt-state presentation seam** (a Typer error hook or `@handle_domain_errors` decorator) so every command inherits clean presentation instead of hand-curating catch tuples (the omission that bit `next` in #4642).
3. **Audit + fix the known blast-radius tail** (from the #4642 3-lens census):
   - Partial-guard (catch `JSONDecodeError` but crash on **non-UTF-8**): `decisions/service.py`, `merge/state.py`, `review/baseline.py`.
   - Fully unguarded readers: `review/lock.py`, `review/artifacts.py`, `status/validate.py`.
   - `core/wps_manifest.py` (unguarded reader; its `next`-reachable call site is currently masked by a broad `except Exception` in `runtime_bridge.py:1019` that should be replaced by the typed seam — real exposure today is `finalize-tasks`).

## Subsumes
- #4600 — `.kittify/config.yaml` startup instance.
- #4642 — `next`/`meta.json` + decisions/`index.json` instances (fixed; this generalizes the pattern it established).

## Not in scope
The two shipped #4642 fixes stand as-is; this mission generalizes them rather than reworking them.

Filed as the deferred class-closer from the #4642 mission (`corrupt-state-file-guards`).

stijn-dejongh:
## What & why (BLUF)

Several `spec-kitty` commands crashed with a **raw Python traceback** when they met bad input or corrupt state — a malformed file, an invalid `--mission` handle, a missing `pyproject.toml`, a non-UTF-8 stream. A traceback reads as *the tool broke*, not *your input was rejected*, and it breaks machine consumers that parse `--json`. This PR closes that **whole class** (umbrella #2899) by construction, so operators and automation get a clean, actionable, typed error instead of a crash — and a new command can't reintroduce the bug.

**Closes #4746, closes #4738, closes #4724, closes #4637, closes #4739, closes #4720, closes #2899.** These six are the open remainder of umbrella #2899 (its original children #2878/#2879 already merged), so landing them completes and closes the umbrella.

## The fix

Two collaborating pieces, then adoption + a gate:

1. **`kernel.guarded_read.read_guarded(path, parse, *, errors=(), error_cls=GuardedReadError, mode=…)`** — a *format-agnostic* primitive: reads once, applies a caller-supplied parser, and collapses `(OSError, UnicodeDecodeError, *caller-declared)` into one typed `GuardedReadError` (chained `from exc`). Kernel imports **no** schema/validation library — decoders (yaml, pydantic, tomllib) stay in the caller's layer, respecting the `kernel <- … <- specify_cli` direction.
2. **One global Typer error-presentation hook** on the top-level app — catches the `GuardedReadError` family → clean `Error: <reason>` (stderr, exit 1) or `{"error","kind","path"}` JSON (stdout, exit 1) under `--json`; **re-raises everything else** so real bugs stay tracebacks; never converts a Typer usage error (exit 2).
3. **Adoption** across every in-scope command (`workflow import/export`, `mission close`, `accept`, `agent release prep`, `intake -`, `specify --json`) + 7 hardened audit-tail readers, and the 8 pre-existing typed reader errors **re-parented onto `GuardedReadError` in place** (multiple inheritance preserves `except ValueError`/legacy catches across the ~85-site caller census — NFR-006 back-compat regression).
4. **Closed by construction** — a non-vacuous architectural gate (`tests/architectural/test_cli_error_surface_seam.py`) fails the build if a new in-scope command reads/decodes/resolves outside the primitive or drops the hook; proven fail-able by a self-mutation test.

Per-issue behaviour is detailed in the `[Unreleased]` CHANGELOG entry.

## Tests / red-first proof

Every defect landed an issue-pinned `@pytest.mark.regression` test that was **RED through the real command entry point before the fix** (verified by reverting each fix), then green. Highlights: `test_guarded_read` (primitive matrix), `test_error_hook` (envelope/exit-code/re-raise), `test_error_backcompat` (every legacy type still caught, incl. `ValueError` + coupled tuples), per-command guard suites, `test_audit_tail_readers` (27), and the construction gate (9, incl. self-mutation).

**Validation on the rebased branch:** `make test-fast` green (1851 passed); the full mission suite green; whole-repo `ruff format --check .` + `ruff check .` clean; the arch battery gates green (construction gate, terminology, format-exclude ratchet, `pyproject_shape`, status-facade boundary).

## Main-red folds (same class)

- **`test_merge.py::test_merge_fails_closed_on_corrupt_meta_{json,text}`** were pre-existing red on `main` — a **stale mock** (the helper never patched the newer `_resolved_mission_dir_exists` gate, so "Mission not found" fired before the corrupt-meta path). `merge.py`'s fail-closed behaviour was already correct; only the test is re-pinned. Now green.

## Pre-PR adversarial-squad folds (correctness + boundary lenses)

- **orchestrator-api decision verbs** now route `DecisionEventLogReadError` through the JSON error envelope (was escaping to the CLI hook shape → empty stdout, breaking the machine contract; C-005/NFR-002).
- **`merge/resolve.py`** cross-mission slug scan now skips a sibling's corrupt state (`except MergeStateReadError: continue`) instead of aborting another mission's merge.
- **`IntakeFileUnreadableError.path`** is now populated for the hook envelope.

## Notable behaviour changes (intentional, flagged)

- **`review/lock.py`**: a *corrupt* review-lock now fails closed (raises) instead of silently being treated as "no lock" and overwritten — a real concurrency-safety fix. A corrupt `review-lock.json` will block `implement`/review acquisition until cleaned.
- **`mission close -f --discard`** now surfaces a clean Typer usage error (`No such option: -f`, exit 2) instead of silently misbinding `--discard` onto `--mission`.

## Architecture / decisions

- Format-agnostic primitive (kernel cannot import `runtime`'s pydantic model); global hook over a per-command decorator (un-forgettable + gate-checkable); subclass-compat over flat type replacement (protects ~85 downstream catchers). Non-ASCII names are **rejected explicitly** (no lossy transliteration), per the charter Identifier-Safety rules.
- `wps_manifest.py` keeps raising bare `pydantic.ValidationError` for schema-invalid (valid-YAML) manifests — pydantic-core is a Rust type that can't carry the `GuardedReadError` constructor contract; documented, not counted by the gate.

## Follow-ups (tracked, not folded — kept in scope)

- **#4773** — two org-pack cascade non-vacuity tests red on `main` (charter-activation subsystem; stale monkeypatch post-#4591 + a no-cascade semantics question). Out of this mission's class; routed to a charter follow-up.
- `next`'s bespoke emitters + the residual `next`(exit 2)/others(exit 1) exit-code inconsistency — deliberately left; a consistency follow-up.
- `review/baseline.py::_parse_junit_xml`'s unguarded `ET.parse` (surfaced by the gate author) — outside the gate's named read set; a candidate follow-up.

## Landing notes

Consolidated onto a clean branch off `main` (the lane/coord worktree machinery fought this cluttered local checkout — an environment issue, not the work). History is linear: mission dossier + one logically-sliced commit per WP + folds. Rebased onto current `main`; only `CHANGELOG.md` conflicted (union-resolved). No Sonar UI work outstanding.

## Maintainer landing pass (independent second-opinion squad + folds)

Ran a profile-loaded, opus second-opinion squad over the integrated diff — correctness/fail-closed, boundary/overlap, and a design lens on the new `kernel.guarded_read` authority + global CLI hook. Verdict: the primitive, the ~85-site adoption, and the multiple-inheritance re-parenting are architecturally sound and back-compat-safe. Two whack-a-field mirror-misses were folded (each a case where the pre-PR squad fixed one instance and missed a sibling), plus three low-risk quality folds. Each fold is its own `<type>(landing):` commit, and the two bug folds carry red-first proof (revert-the-fix → test fails).

**Folded:**
- `fix(landing)` **`--json` envelope `path` was `null` on two live paths.** `MissionMetaReadError` (corrupt `meta.json`) and `DecisionIndexReadError` (corrupt `decisions/index.json`) carried the path only in a bespoke attribute and left the inherited `GuardedReadError.path` `None`, so the machine envelope emitted `"path": null` on path-scoped failures — the same defect the pre-PR squad fixed for `IntakeFileUnreadableError`. Now populated (message text unchanged).
- `test(landing)` **envelope-completeness guard** — a parametrized regression asserting every path-scoped, custom-`__init__` `GuardedReadError` subclass populates `.path`, closing the whack-a-field's structural root so a new sibling can't silently reintroduce `null`.
- `fix(landing)` **`decision open` corrupt `status.events.jsonl` regressed structured→text.** WP07 changed `_opened_event_exists` to raise `DecisionEventLogReadError` (not `DecisionError`); the orchestrator-api mirror's catchers were updated but the CLI `cmd_open` catcher was not, so a corrupt event log on `decision open` fell through to the global hook's plain `Error:` line instead of the structured `{code,error,details}` its sibling corrupt-index path emits on the same command. Restored to parity (reusing the `DESIGN_STATUS_EVENT_LOG_UNREADABLE` envelope code), with a red-first CLI test.
- `docs(landing)` pin the intentional fail-closed candidate-key loop in `merge/resolve.py` (no `except MergeStateReadError: continue`, because its keys are the target mission's own state — a future "harmonization" into skip-on-corrupt would reintroduce a fail-open).
- `docs(landing)` document the construction gate's bounded call-shape scope (enumerated shapes only; from-imports/aliases not detected) and name the `_parse_junit_xml` ET.parse residual for parity with the pydantic carve-out.

**Squad findings routed to follow-up rather than folded** (design decisions / pre-existing, out of this PR's correctness scope):
- **Single envelope authority per surface.** Orchestrator-api verbs sit *under* the global hook; a `GuardedReadError` escaping `_resolve_mission_dir_or_fail` (which catches only `StatusReadPathNotFound`) reaches the CLI-shaped hook, not the machine `make_envelope`. This is pre-existing (those inputs raised raw tracebacks before) and the clean fix is one `except GuardedReadError → _fail(...)` at the `_JSONErrorGroup.invoke` boundary — a per-surface authority decision that belongs with the `next`(exit 2)/others(exit 1) consistency follow-up, not this PR.
- **`core/wps_manifest.py`** declares `AttributeError` in the guarded-read `errors` tuple to absorb the empty/non-dict-top-level case, which also collapses a genuine future `AttributeError` bug into `WpsManifestReadError` (masking it). Behavior-preserving fix (explicit non-dict guard + drop `AttributeError`) is a candidate follow-up; not folded here to avoid a behavior-change on a gate-scoped file during landing.
- **Upgrade migration** `m_2_0_6_consistency_sweep` aborts the whole per-mission sweep on one corrupt-but-present `status.json` (pre-existing; the PR only changed the raised type). Candidate for the same fail-closed-scan follow-up.
- The `-f`=`--mission` short alias was removed on four sites (intentional, #4724); downstream agent scripts/harnesses that invoke `-f <slug>` should be confirmed.

**Verification (targeted, on a plain checkout):** whole-repo `ruff format --check .` + `ruff check .` clean; `mypy` clean on the changed modules; the construction gate, ratchet baselines, format-exclude ratchet, terminology, status-facade boundary, layer rules, and kernel-import gates green; the guarded-read/hook/back-compat suites, the decision CLI + orchestrator-api envelope suites, and the merge-resolve seam suite green; commitlint clean over the fold range. The full suite/arch battery is left to CI.

https://claude.ai/code/session_01XrPa1Mq66aZ6uRrs4Ujeeb


stijn-dejongh:
## Maintainer landing pass — claim

Picking this up for a maintainer landing pass (I authored it; this is the independent landing gate before hand-off to the operator).

**Viability triage — LAND.** Fresh (today), substantive, and already based on current `main` (0 behind `skupstream/main`, 15 ahead — no stale-base rebase needed). Not superseded/moot: all six children (#4746, #4738, #4724, #4637, #4739, #4720) and the umbrella #2899 are still open, and this is the work that closes them. #4773 is correctly scoped out as a charter follow-up.

**Plan for the pass:**
- Independent second-opinion squad (opus, profile-loaded) over the integrated diff — correctness/fail-closed, boundary/overlap, and a design lens on the new kernel `guarded_read` authority + global CLI error hook. Fold any MAJOR/MINOR inline.
- Targeted gate verification on a plain checkout (construction gate, ratchet baselines, format-exclude ratchet, terminology, status-facade boundary, pyproject_shape) + changed-file suites + whole-repo `ruff`. Not re-running the full arch battery / full suite locally — that's CI's job.
- Deliverable hygiene: confirm `Closes #2899` on the umbrella, changelog, docs.

Evidence to follow. The operator merges.

stijn-dejongh:
## Maintainer landing pass — remediation summary & hand-off

**Verdict: LAND** (on green). Fresh, substantive, already based on current \`main\` (0 behind \`skupstream/main\`) — no rebase needed. Independent second-opinion squad (opus, profile-loaded: correctness/fail-closed, boundary/overlap, design) run over the integrated diff. The primitive, ~85-site adoption, and multiple-inheritance re-parenting are architecturally **sound and back-compat-safe**. Two whack-a-field mirror-misses + three low-risk quality items folded; the rest routed to follow-up.

### Folds (each its own \`(landing)\` commit; bug folds carry red-first proof)
| Commit | Type | Why |
|---|---|---|
| \`c75c405\` | fix | **MAJOR** — \`--json\` envelope emitted \`\"path\": null\` for \`MissionMetaReadError\` / \`DecisionIndexReadError\` (bespoke path attr, inherited \`GuardedReadError.path\` left \`None\`). Same defect the pre-PR squad fixed for intake, missed on two siblings. Red-first: both new \`.path\` asserts fail pre-fix (\`assert None == '…'\`), pass after. |
| \`76c3570\` | test | Envelope-completeness guard — parametrized assert that every path-scoped custom-\`__init__\` subclass populates \`.path\`; closes the whack-a-field's structural root. |
| \`011abd4\` | fix | **MAJOR** — \`decision open\` on a corrupt \`status.events.jsonl\` regressed structured→plain-text. WP07 changed the raised type to \`DecisionEventLogReadError\`; orchestrator-api caught it, CLI \`cmd_open\` did not (its sibling corrupt-index path stays structured on the same command). Restored parity via \`_handle_event_log_read_error\` (reuses \`DESIGN_STATUS_EVENT_LOG_UNREADABLE\`). Red-first CLI test: exception escapes \`cmd_open\` pre-fix, structured payload after. |
| \`1d5369e\` | docs | Pin the intentional fail-closed candidate-key loop in \`merge/resolve.py\` so it isn't \"harmonized\" into skip-on-corrupt (would reintroduce a fail-open). |
| \`a4e92e0\` | docs | Document the construction gate's bounded call-shape scope + name the \`_parse_junit_xml\` ET.parse residual (parity with the pydantic carve-out). |

### Routed to follow-up (not folded — design decisions / pre-existing)
- **Single envelope authority per surface** — orchestrator-api sits under the global hook; a \`GuardedReadError\` escaping \`_resolve_mission_dir_or_fail\` (catches only \`StatusReadPathNotFound\`) gets the CLI-shaped hook, not \`make_envelope\`. Pre-existing (raw tracebacks before); clean fix is one boundary catch at \`_JSONErrorGroup.invoke\`. Belongs with the \`next\`(exit 2)/others(exit 1) consistency follow-up.
- **\`core/wps_manifest.py\`** declares \`AttributeError\` in the guarded-read \`errors\` tuple (absorbs empty/non-dict top-level) which also masks a genuine future \`AttributeError\`; behavior-preserving fix is a candidate follow-up, not folded to avoid a behavior-change on a gate-scoped file during landing.
- **Upgrade migration** \`m_2_0_6_consistency_sweep\` aborts the whole sweep on one corrupt \`status.json\` (pre-existing; type-only change here).
- \`-f\`=\`--mission\` removed on 4 sites (intentional, #4724) — confirm no downstream harness invokes \`-f <slug>\`.

(All four are recorded in the PR body under *Maintainer landing pass*; per release-push policy, noted in prose rather than ticketed.)

### Verification (targeted, plain checkout)
- whole-repo \`ruff format --check .\` + \`ruff check .\` — **clean**; \`mypy\` — clean on changed modules.
- Gates green: construction gate (9), ratchet baselines, format-exclude ratchet (excl. entries still valid after edits), terminology, status-facade boundary, layer rules, kernel-import.
- Suites green: guarded-read / hook / back-compat (28), decision CLI (73) + orchestrator-api envelope, merge-resolve seam (21); consolidated targeted run **281 passed**.
- commitlint clean over the fold range; docs-freshness \`errors=0\`.
- Whole suite / full arch battery left to CI (running now on the pushed folds).

### Umbrella hygiene
Added \`closes #2899\` to the PR body — its original children (#2878/#2879) already merged, so these six complete and close the umbrella on merge.

**Hand-off:** green-modulo-CI, un-drafted, evidence above. **The operator merges — I do not.** No landing-order constraints. Will verify the merge landed clean (folds present, authorship, no new red-main) and clean up the branch/worktree after you say "merged".

stijn-dejongh:
## Landing update — arch battery went red; classified + folded (all PR-introduced)

The first full CI run surfaced **architectural battery: 3 failed** (+ **router gate**, a pure cascade — its log: `blocking job(s) did not succeed: architectural battery=failure`). All three were **green on \`main\` (9477ba6442)** and red on the branch → PR-introduced, folded (green-on-base + red-on-branch = ours). None are in the guarded-read domain; each is a baseline the PR's changes legitimately shifted (bin 2 re-pins).

| Failing test | Root cause | Fold |
|---|---|---|
| \`test_os_detection_ban::test_no_banned_os_detection_outside_the_door\` + \`::test_every_exemption_entry_is_a_real_violation\` | WP01's global-hook additions to \`main()\` shifted the pre-existing, deferred \`sys.platform == \"win32\"\` UTF-8 check (issue #66) in \`__init__.py\` from line 479→518; the line-keyed \`os-detect-ban-deferred.txt\` entry went stale. | \`f312298\` — re-pin 479→518 (check itself unchanged, still deliberately unrouted). |
| \`test_no_dead_symbols::test_no_public_symbol_in_all_is_unimported\` | (a) \`UnknownWorkflowError\` re-parented \`Exception\`→\`GuardedReadError\` (#2899) — a public exception (in \`__all__\`, raised by \`get_workflow\`, propagated per FR-015) with no cross-module caller that passed on \`main\` via the \`(Exception)\`-base T013 auto-exemption, which the new base no longer matches. (b) \`_iter_merge_states_for_slug\`'s content-tier allowlist hash no longer matched after the cross-mission slug-scan fix changed its body. | \`badc716\` — documented content-tier hand-allowlist entry for the exception (new \`#2899\` category, wired into the union chain) + re-pinned hash for the slug-scan symbol. |

**Verification:** both arch files reproduced red on the branch, **green on \`main\`**, and now pass on the branch (`test_os_detection_ban.py` + `test_no_dead_symbols.py` → **49 passed**, including the `test_allowlist_is_exactly_union_of_categories` / `test_auto_exempt_disjoint_from_hand_allowlist` invariants that guard my new category). Whole-repo \`ruff\` clean; commitlint clean. Pushed \`a4e92e0..badc716\`; CI re-running. Per operator instruction, I'll merge on green.

